PR #78, opened 2020-09-03
MPE: Sync production

sergiogcx:
We need to sync the new router settings in staging and make sure they work merging to production.


sergiogcx:
Not a good PR, closing.

PR #81, opened 2020-09-21
3941 lz new project

lazavazav:
This branch has been created to add the new-project user form to the moped-editor app.
It is using React-Hook-Form to create the form with styling from material-ui and matx. It currently logs the data to the console. The next commit will add the functionality to send user data to the  moped_project database. It is currently being rendered from the navigation.js file and can be found in src/views/forms.


lazavazav:
Latest commits have added the functionality to save user data to database and a Get Data page has been added to UI to start adding database queries. An initial query of displaying data from the moped_project table is working.

lazavazav:
new project commit
newProject feature can query the database to populate the select fields and all input fields are saved in moped_project table.

sergiogcx:
Thank you @lazavazav 👍  I will be reviewing your code first thing in the morning.

lazavazav:
last commit to delete team project files etc.....created a new branch for team project.
Closes cityofaustin/atd-data-tech#3941

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "git.ignoreLimitWarning": true
+}
```
sergiogcx, line 3: VS Code rules should not be committed to the repo, try adding a rule to the .gitignore file so that the .vscode folder is ignored.

moped-editor/src/app/App.jsx
```diff
@@ -7,15 +7,30 @@ import { Router } from "react-router-dom";
...
+  return new ApolloClient({
+    link: new HttpLink({
+      uri: 'https://coa-moped.herokuapp.com/v1/graphql',
+    }),
+    cache: new InMemoryCache(),
+  });
+ };
+
+const client = createApolloClient();
+
```
sergiogcx, line 29: This is very good, I like how you created a client and how you provided parameters for a default uri and cache policy.  Right now there isn't a lot of clarity, but we will definitely need to stay away from hard-coded URIs soon, also there will be situations where we will need to have different cache policies for different pages.

It's not really necessary, but if you have the time, take a look at how we have done this in the past:
https://github.com/cityofaustin/atd-vz-data/blob/9c54201924df3d893a95a184efb390648eae0ed5/atd-vze/src/App.js#L31

moped-editor/package.json
```diff
@@ -20,6 +21,7 @@
     "aws-amplify-react": "^4.1.22",
     "axios": "^0.19.0",
     "babel-polyfill": "^6.26.0",
+    "client": "0.0.1",
```
sergiogcx, line 24: Can't tell what this library is for:
https://www.npmjs.com/package/client

The code hasn't been updated in 6 years 😸 
https://github.com/ContinuumBridge/client-js

moped-editor/package.json
```diff
@@ -64,7 +67,7 @@
   },
   "scripts": {
     "start": "env-cmd -e local npm run start:local",
-    "start:local": "export HTTPS=true && react-app-rewired start",
+    "start:local": "react-app-rewired start",
```
sergiogcx, line 70: We really need to keep the HTTPS, this is because any communications between the Hasura server and the local client in HTTP will be unencrypted (plain text), and we would be passing important admin keys or JWT tokens.

I suppose it is ok when running agains the Heroku Hasura instance for testing, but let's put it back in there for now.

moped-editor/src/app/views/forms/NewProject.jsx
```diff
@@ -0,0 +1,280 @@
...
+        current_phase
+        current_status
+        eCapris_id
+        fiscal_year
+        capitally_funded
+        start_date
+      }
+    }
+  }    
+ `;
```
sergiogcx, line 60: These are really great! I am excited! 😃 

I'm not too opinionated as to where to put queries, but as per Mateo's documentation, we are going to try to follow this pattern: https://app.gitbook.com/@atd-dts/s/moped-technical-docs/dev-guides/code-organization

You could save all those queries in a file like that, and then import them.

moped-editor/src/app/views/forms/NewProject.jsx
```diff
@@ -0,0 +1,280 @@
...
+    let project_name=data.newProject;
+    let project_description=data.ProjDesc;
+    let eCapris_id=data.eCaprisId;
+    let capitally_funded=data.capitalFunded;
+    let start_date=data.date;
+    let current_phase=data.Phase;
+    let project_priority=data.Priority;
+    let current_status=data.Status;
+    let fiscal_year=data.FiscalYear;
+    addProject({variables: {project_name, project_description, eCapris_id, current_phase, current_status, project_priority, fiscal_year, capitally_funded, start_date}});
```
sergiogcx, line 75: Why not just pass the data.* variables directly to the addProject function? ie:

```
addProject({variables: {data.newProject, data.ProjDesc, data.eCaprisId, ...}});
```

moped-editor/src/index.jsx
```diff
@@ -5,40 +5,18 @@ import Amplify from "aws-amplify";
...
-                name: "testApi",
-                endpoint: config.apiGateway.URL,
-                region: config.apiGateway.REGION
-            }
-        ]
-    }
-});
-
-
-ReactDOM.render(<App />, document.getElementById("root"));
```
sergiogcx, line 41: We really need the Amplify library here, sorry things are confusing right now, but let's put this block of code back in your branch whenever you get a chance.

moped-editor/src/app/App.jsx
```diff
@@ -29,6 +44,7 @@ const App = () => {
         </MatxTheme>
       </Provider>
     </AppContext.Provider>
+     </ApolloProvider>
```
sergiogcx, line 47: These are fine, but it would be nice to have some stylistic tabulation to make it easier to read.

sergiogcx (commented):
The code looks good, feel free to make changes as suggested. I don't think we are going to merge the code right now, at least until next week. We are in the middle of finishing the hasura migration to AWS, whenever that is done we are going to provide credentials, the API endpoints, and proper environment configuration.

sergiogcx:
Also, when running the branch I noticed this bug when I entered the eCapris ID: "98765"

moped-editor/src/app/views/forms/NewProject.jsx
mddilley: One additional option that might be worth exploring here or down the road at some point: https://hasura.io/docs/1.0/graphql/core/mutations/insert.html#insert-a-single-object
(the second example called "Using variables:")

If the names of the form fields are updated to the format that Hasura is expecting, it might be possible just to pass `data` in this mutation. I know it breaks JS convention with the snake case keys in `data`, but it might be cool to have them map directly. Just a thought and not sure how possible that is. Eager to take a closer look at all of this! Cool stuff!

moped-editor/package.json
lazavazav: I took it out to run my dev server as it kept giving me an error. I will put it back before committing next time.

moped-editor/package.json
lazavazav: Honestly, I don't know what it is either. I installed a couple of packages before deciding on apollo client so maybe it came along for the ride on something. I will delete it on my local copy to confirm its uselessness.

moped-editor/src/app/views/forms/NewProject.jsx
lazavazav: I checked out the link and it is definitely worth looking into. I'll try both suggestions and see what happens.

lazavazav:
I went ahead and changed eCapris id to String.  It originally was bpChar in hasura but I think JD changed it since.

PR #89, opened 2020-10-12
Moped API | Implement API functionality

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4014

This PR adds a CRUD API for managing Moped users. I welcome any feedback on error handling or anything at all - especially in routes that makes requests to different AWS resources one after another. I'll update the user role restriction on the write routes once we have a solid idea of who should/should not create/modify users. 

@sergiogcx - I also modified the `normalize_claims` decorator to make it work on multiple routes in the user blueprint. If you think that will break things, let me know, and I'll create a separate helper. I plan on sharing the reasoning in a dev sync one day.

[Documentation](https://team-1600951431491.atlassian.net/l/c/moBtXtwU)


moped-api/claims.py
```diff
@@ -1,5 +1,6 @@
 import json
 from flask_cognito import _request_ctx_stack, current_cognito_jwt
+from functools import wraps
```
sergiogcx, line 3: This is cool! I learned something new today! :)

moped-api/users/helpers.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    fernet_key = get_secret(AWS_COGNITO_DYNAMO_SECRET_NAME)
+    claims_str = json.dumps(claims)
+    encrypted_claims = encrypt(fernet_key=fernet_key, content=claims_str)
+    dynamodb = boto3.client("dynamodb", region_name="us-east-1")
+    dynamodb.put_item(
+        TableName=AWS_COGNITO_DYNAMO_TABLE_NAME,
+        Item={"user_id": {"S": user_id}, "claims": {"S": encrypted_claims}},
+    )
+
```
sergiogcx, line 203: This entire file is great stuff!

moped-api/users/users.py
```diff
@@ -0,0 +1,163 @@
...
+@users_blueprint.route("/", methods=["POST"])
+@cognito_auth_required
+@normalize_claims
+def user_create_user(claims: list) -> str:
+    """
+    Returns created user details
+    :return str:
+    """
+    if is_valid_user(current_cognito_jwt) and has_user_role("user", claims):
+        cognito_client = boto3.client("cognito-idp")
```
sergiogcx, line 72: This line repeats itself a lot in this file, would it not be better to declare it once globally and let function access the cognito_client?

moped-api/claims.py
mddilley: Same here! Planning on sharing how I tripped over this in dev sync this week. 👍

moped-api/users/users.py
```diff
@@ -0,0 +1,163 @@
...
+@users_blueprint.route("/<id>", methods=["DELETE"])
+@cognito_auth_required
+@normalize_claims
+def user_delete_user(id: str, claims: list) -> str:
+    """
+    Returns created user details
+    :return str:
+    """
+    if is_valid_user(current_cognito_jwt) and has_user_role("user", claims):
+        response = cognito_client.admin_delete_user(UserPoolId=USER_POOL, Username=id)
```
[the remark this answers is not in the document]
mddilley: Great catch! 🙏  This is now fixed in the new branch.

mddilley:
Starting a new branch and PR.

PR #93, opened 2020-10-19
Moped API | Implement API functionality

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4014

This PR adds a CRUD API for managing Moped users. I welcome any feedback on error handling or anything at all - especially in routes that makes requests to different AWS resources one after another. I'll update the user role restriction on the write routes once we have a solid idea of who should/should not create/modify users. 

@sergiogcx - I also modified the `normalize_claims` decorator to make it work on multiple routes in the user blueprint. If you think that will break things, let me know, and I'll create a separate helper. I plan on sharing the reasoning in a dev sync one day.

[Documentation](https://team-1600951431491.atlassian.net/l/c/moBtXtwU)


sergiogcx (approved):
A ton of real cool stuff in this branch!

PR #99, opened 2020-10-27
Moped API | Add tests to user management API

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4143

[Documentation](https://app.gitbook.com/@atd-dts/s/moped-technical-docs/infrastructure/api/testing)

This PR adds some test coverage to the helpers and routes in the User Management API. The tests mostly focus on the route protections and Cognito actions. The write routes also have a lot of repetitive patching that I hope to abstract away in the future, but they work for now. There is also room to extend test coverage around the requests to DynamoDB.


sergiogcx (approved):
I can't believe it works so well! Thank you for showing us the moto library!

PR #102, opened 2020-10-28
Reimplement auth

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4241


moped-editor/src/views/auth/LoginView.js
```diff
@@ -43,16 +60,19 @@ const LoginView = () => {
         <Container maxWidth="sm">
           <Formik
```
mddilley, line 61: Here's Formik! Maybe we could swap out these login and register forms to be powered by React Hook Form, but they would perfectly fine as is. 👍

moped-editor/src/views/auth/LoginView.js
```diff
@@ -29,11 +30,27 @@ const LoginView = () => {
...
+      // wait to see if login was successful (we don't care about the return
+      // value here)
+      await login(values.email, values.password);
+
+      // mark the form as non-subbmitting
+      setSubmitting(false);
+    } catch (err) {
+      // If an error occured, showcase the proper message (we customised the
+      // message ourselves in `UserProvider`'s code)
+      setErrors({ password: err.message });
```
mddilley, line 47: Nice! I tried out an incorrect password, and I received the error message.

moped-editor/src/auth/user.js
```diff
@@ -0,0 +1,100 @@
...
+      return data;
+    });
+
+  // Make sure to not force a re-render on the components that are reading these values,
+  // unless the `user` value has changed. This is an optimisation that is mostly needed in cases
+  // where the parent of the current component re-renders and thus the current component is forced
+  // to re-render as well. If it does, we want to make sure to give the `UserContext.Provider` the
+  // same value as long as the user data is the same. If you have multiple other "controller"
+  // components or Providers above this component, then this will be a performance booster.
+  const values = React.useMemo(() => ({ user, login, logout }), [user]);
```
mddilley, line 82: This is cool! Glad to see a solution to the problem that we all discussed on Monday about potential performance cost. Might be cool add to VZE one day too.

mddilley (approved):
This looks amazing! I'm glad to see similar patterns to the auth context in VZE and also that it can be done with Cognito all the same. The memoization in the context taught me a new trick, and the user custom hook sets up a great base to role based permissions as well! I logged in with my staging username with no issues at all.

I ran into one problem when first running locally, but it was an easy fix to get up and running. 👍 
Screen Shot 2020-10-28 at 12 30 36 PM

sergiogcx (changes requested):
It won't run for me:

mateoclarke:
> It won't run for me:
> 
> 

Thanks @sergiogcx, Mike also pointed this out but I just pushed the code that should fix this import error.

PR #103, opened 2020-10-28
Moped | Roles and Permissions - Synchronize Cognito with Postgres users table #4054

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4054

[Documentation](https://app.gitbook.com/@atd-dts/s/moped-technical-docs/~/drafts/-MLEtw02D5W-Lq9plUlG/infrastructure/api/user-management-api)


moped-api/users/users.py
```diff
@@ -109,39 +119,92 @@ def user_create_user(claims: list) -> str:
...
 
-        return jsonify(response)
+        # Generate the user profile for the database
+        user_profile = generate_user_profile(
+            cognito_id=cognito_username, json_data=request.json
+        )
+        # Persist the profile in the database
+        db_response = db_create_user(user_profile=user_profile)
+
+        if "errors" in db_response:
```
mddilley, line 129: Nice, this is great so that the DB and Cognito remain synchronized. 👍

moped-api/users/users.py
```diff
@@ -1,31 +1,34 @@
-import os, boto3
+import boto3
+
 from botocore.exceptions import ClientError
-from flask import Blueprint, jsonify, abort
+from flask import Blueprint, jsonify, abort, Response
```
mddilley, line 4: Cool, TIL. This is a much better option for the return type hints in this blueprint.

mddilley (approved):
Looks great, Sergio! I really like your solutions to keeping Cognito and the DB in sync and also the responses that give feedback about both of them as well. Very cool since I was unsure how this would all work out, and it gives me a lot to think about when it comes to writing APIs.

It will be great to test this out from the frontend after the user management UI is finished out! ⚛️ 😎

PR #105, opened 2020-10-28
Editor | Implement User table

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4242

This PR adds the table list view of Moped users and a custom hook to hit the user management API. I also added a function to the auth context to get the user token. This was developed against the Heroku Hasura endpoint so that endpoint will need to be pasted [here](https://github.com/cityofaustin/atd-moped/blob/2d19fd0e5ffb0a64182b31c3f0b98e95a7ab82b7/moped-editor/src/views/staff/StaffListView/StaffList.js#L47) to fully test it. I also hardcoded the local endpoint for the user API for now. I'll update in further branches as we finalize the DevOps stuff in the editor.

#### Screenshot
Screen Shot 2020-11-06 at 11 38 52 AM


mateoclarke (approved):
🚢

moped-editor/src/views/staff/StaffListView/helpers.js
```diff
@@ -0,0 +1,37 @@
...
+export function useUserApi(method, path, payload = null) {
+  const [result, setResult] = useState(null);
+  const [loading, setLoading] = useState(false);
+  const { user, getToken } = useUser();
+
+  useEffect(() => {
+    // Use local API dev server for now
+    const url = `http://127.0.0.1:5000` + path;
+
+    const token = getToken(user);
```
sergiogcx, line 15: Ah, cool. I had been wondering. Thanks! 🙏

moped-editor/src/auth/user.js
```diff
@@ -73,13 +73,18 @@ export const UserProvider = ({ children }) => {
       return data;
     });
 
+  // Get JWT token with roles
+  const getToken = user => user && user.signInUserSession.idToken.jwtToken;
+
```
sergiogcx, line 78: I love this line! 👍

moped-editor/package.json
```diff
@@ -9,11 +9,10 @@
     "start:local": "export HTTPS=true && react-scripts start",
     "start:staging": "env-cmd -e staging npm run start:local",
     "start:production": "env-cmd -e production npm run start:local",
-    "build": "env-cmd -e local npm run build:local",
+    "build": "react-scripts build",
```
sergiogcx, line 12: Is the command not working?

sergiogcx (approved):
Amazing work as usual!

moped-editor/package.json
mddilley: Nice catch, Sergio! I didn't see that update. Not sure when this happened but maybe when I added `axios`. Looks like there were two build commands before. I'll try to run it and make sure everything looks okay.

moped-editor/src/auth/user.js
mddilley: Thanks! I just realized that this function doesn't really need a parameter since `user` is in scope in the context. I'll update and test.

PR #107, opened 2020-11-06
Rename files for component instead of "index.js"

mateoclarke:


sergiogcx (approved):
Amazing work! Thanks for looking into this :)

mddilley (commented):
🙏 I updated the User route to match this pattern.

PR #108, opened 2020-11-09
Apollo Setup

mateoclarke:
This PR completes the first part of this issue around configuring Apollo in the app. https://github.com/cityofaustin/atd-data-tech/issues/4220

It uses the Heroku deployed Hasura endpoint. I'm going to continue with local, staging, and prod instances in the next PR.

In this PR we're now using apollo to populate the date on this the Staff List view and that file has the basic pattern for accessing gql queries as hooks in a component with Apollo.
http://localhost:3000/app/staff


moped-editor/.env-cmdrc
```diff
@@ -2,16 +2,19 @@
...
+    "REACT_APP_HASURA_ENDPOINT": "https://coa-moped.herokuapp.com/v1/graphql"
   },
   "production": {
     "REACT_APP_AWS_COGNITO_REGION": "us-east-1",
     "REACT_APP_AWS_COGNITO_USER_POOL_ID": "us-east-1_Zc3pNWX51",
-    "REACT_APP_AWS_COGNITO_APP_CLIENT_ID": "ins01e2a8d3vd8apvnd0jv10c"
+    "REACT_APP_AWS_COGNITO_APP_CLIENT_ID": "ins01e2a8d3vd8apvnd0jv10c",
+    "REACT_APP_HASURA_ENDPOINT": "https://coa-moped.herokuapp.com/v1/graphql"
   }
 }
```
sergiogcx, line 20: This is great, thank you. However, these are not likely to work once deployed. The start scripts package.json control which environment variables are loaded at runtime, and it seems a little different from the originals currently in production:

https://github.com/cityofaustin/atd-moped/blob/4fb212ceaee74ec2c4c6dcc2f6a0984fb2cb4c76/moped-editor/package.json#L66

sergiogcx (approved):
Great work!

mddilley (approved):
Awesome, thanks for setting this up! 🙌  I was going to update this branch's base but I will leave it as is to avoid conflicts when you merge. I'll update in a future branch. 👍

PR #111, opened 2020-11-11
Moped | Single Sign On | Refactor Claims Structure in DynamoDB #4422

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4422

Keeps the functionality of the API the same, but it switches the claims to use email identifiers as opposed to Cognito UUIDs, that way we can have two or more cognito accounts (Cognito and AzureAD) both pointing to the same record in DynamoDB where we get our claims.

### DIFF
--- a/moped-api/users/users.py
+++ b/moped-api/users/users.py
@@ -196,7 +196,7 @@ def user_update_user(id: str, claims: list) -> (Response, int):
 
         if roles:
             user_claims = format_claims(id, roles)
-            put_claims(id, user_claims)
+            put_claims(user_email=user_profile["email"], user_claims=user_claims)
 
         response = {
             "success": {
--- a/moped-api/users/users.py
+++ b/moped-api/users/users.py
@@ -168,6 +168,10 @@ def user_update_user(id: str, claims: list) -> (Response, int):
         if not profile_valid:
             return jsonify({"error": profile_error_feedback}), 400
 
+        # Retrieve current profile (to fetch old email)
+        user_info = cognito_client.admin_get_user(UserPoolId=USER_POOL, Username=id)
+        user_email_before_update = get_user_email_from_attr(user_attr=user_info)
+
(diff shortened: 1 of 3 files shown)


moped-api/users/users.py
```diff
@@ -193,9 +198,13 @@ def user_update_user(id: str, claims: list) -> (Response, int):
             UserAttributes=updated_attributes,
         )
 
+        # Delete the email if it is different
+        if user_email_before_update != json_data["email"]:
```
mddilley, line 202: I'm probably overlooking something, but what would happen if a user's email updated but their roles did not? It looks like their claims would be deleted from DynamoDB and not replaced.

mddilley (approved):
Awesome stuff, Sergio! I left a couple questions, and I'm glad to see your SSO work paying off and that the required updates to the API weren't extreme. 🙌

moped-api/users/users.py
```diff
@@ -34,7 +35,7 @@ def user_list_users() -> (Response, int):
         cognito_client = boto3.client("cognito-idp")
 
         user_response = cognito_client.list_users(UserPoolId=USER_POOL)
-        user_list = user_response["Users"]
+        user_list = list(filter(lambda user: "azuread_" not in user["Username"], user_response["Users"]))
```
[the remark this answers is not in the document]
sergiogcx: Correct, whenever someone uses SSO it creates another Cognito user that always contains the "azuread_" prefix, and both accounts need access to the same claims.

moped-api/users/users.py
sergiogcx: It's a great point... admittedly I've only tested when roles are provided and it worked well deleting the claims with the old email and creating the new roles with the new email (line 207). The validator requires roles to be provided, so I would expect it to enforce the update to include the new information.

PR #112, opened 2020-11-11
Moped | API | Lint API Code #4451

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4451

Linted using Black


mddilley (approved):
Awesome, thanks for keeping the API looking ✨😎!

PR #114, opened 2020-11-12
Moped | Single Sign On | Make API Populate Cognito UUID in DynamoDB #4455

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4455

This PR adds the cognito_uuid column in the users table in DynamoDB, it lives right along with the claims and the user email. We need the cognito_uuid to be provided when searching with an email, as opposed to browsing through the entire Cognito list of users.

The Lambda that generates tokens for AzureAD needs that Cognito UUID, and it passes it to Hasura. Normal cognito accounts are not affected at all.

### DIFF
--- a/moped-api/claims.py
+++ b/moped-api/claims.py
@@ -186,18 +186,23 @@ def format_claims(user_id: str, roles: list) -> dict:
     }
 
 
-def put_claims(user_email: str, user_claims: dict):
+def put_claims(user_email: str, user_claims: dict, cognito_uuid: str = None):
     """
     Sets claims in DynamoDB
     :param str user_email: The user email to set the claims for
     :param dict user_claims: The claims object to be persisted in DynamoDB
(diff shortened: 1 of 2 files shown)


moped-api/claims.py
```diff
@@ -186,18 +186,23 @@ def format_claims(user_id: str, roles: list) -> dict:
     }
 
 
-def put_claims(user_email: str, user_claims: dict):
+def put_claims(user_email: str, user_claims: dict, cognito_uuid: str = None):
```
mddilley, line 189: This is nice, Sergio. I wondered if the data structure in the Cognito user payload would present a problem in the future when trying to get user attributes, and this is a great solution since we are already storing things in DynamoDB. 🙌

mddilley (approved):
Good stuff here! 👍

PR #115, opened 2020-11-12
Moped | Single Sign On | Pre-token generator needs to know Cognito ID #4454

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4454

This updates the token generator to load the Cognito UUID from DunamoDB and makes it part of the token for Hasura.


mddilley (commented):
Makes sense to me. Thanks for updating the DynamoDB values to store the Cognito UUID! 👍

PR #116, opened 2020-11-12
Get hasura endpoints working for staging & prod envs

mateoclarke:
In this PR we've added some user auth helper functions that enable us to get tokens and roles based on different environment settings.


sergiogcx (approved):
Thank you for taking care of this one! I did make a little contribution for the API endpoints, those will be needed in the future as well, snuck them into your branch if you don't mind, they are harmless.

mddilley (approved):
🙌 🙏

PR #117, opened 2020-11-14
Lz 4220  new project

lazavazav:
added newProjectView feature to code from mc_4220_v2_local_staging-prod branch.

Noticed that apollo-boost was being used to create apollo client. Added @apollo/client as a dependency to use the useMutation and useQuery hooks. Also added react-hook-form and lodash. Currently, getting a failed to fetch network error upon loading the new project page as queries to the new endpoint are not working and are dependent for the feature to load.


mateoclarke:
Hey @lazavazav,

I've added a few commits to your pull request:

- First, I merged main into your branch to isolate the differences in your code.
- A lot of the changes that I made were simply autoformats from [Prettier](https://prettier.io/). It would be good to add Prettier to your VSCode setup. We already have a `.prettierrc` file at the root so it should just start autoformatting for you. You may need to configure your editor to autoformat on save. 
- I removed the `<Page>` component from the `MapProjectGeometry`, `ProjectTeamTable`, and `DefineProjectForm`. Since they are all just steps within the same page, that seemed to make more sense and improve the `title` attribute for the page.
- I updated `moped_coa_staff` table reference to `moped_users`. This change to the table name was something we updated after migrating away from the Heroku test instance. Also removed `staff_id` and `full_name` fields from the query that weren't used and caused GraphQL error.
- I updated some layout and spacing issues but there's probably a few more things to address, for example the padding around the Add Staff icon.
- I updated the "Loading..." text to use the Material UI Spinner

Here is the before/after:

### Before
Screen Shot 2020-11-18 at 5 13 33 PM

### After
Screen Shot 2020-11-19 at 1 15 20 AM

## Next steps
I didn't dig too much into the causes of any of this yet but, I encountered a few buggy things about the form UI that I think we'll still need to address in the next pull request. I can make a new issue for these items too.

- The text input (ex: project name) seemed a little laggy. Wondering if there's a more performant way to handle state changes.
- The Add Staff icon needs some styling and padding.
- When I select a staff, and then remove them, the form hangs and gives me `TypeError: Cannot read property 'workgroup' of null` related to the `handleGroupChange` function in the.
- Submit form returns GraphQL error.

mateoclarke (approved):
@lazavazav, once you've had a chance to review the comments I made and changes I added, I think we're good to merge this as is. 🚢 

I did note some follow-up work required to get the form submitting and fully functional. If you want to talk about any of this let me know and I can also make an issue for the follow-up work and we can assign it to you, or one of our other devs.

lazavazav:
Thanks so much Mateo for working on it. I pushed changes to my copy and it
looks good ...at least what I can see so far. I have Docker installed but
still need to get hasura cli and get hasura cluster going to be able to see
the data coming in and see the rest of the forms I believe. Thanks for the
docs on all of that. I'm good with going ahead with the merge and having
the follow up work assigned to me.

PR #119, opened 2020-11-20
Editor and API | Implement Update User password form

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4410

This PR adds a password update form to the `ProfileDetails` component and extends the user blueprint in the Flask API to handle the password change request with Cognito. I also patched up a regression that broke the add user form (by changing a hidden type field to being hidden with CSS) and updated the `StaffForm` submission to navigate back to the user table on success. The `Results` component is also renamed as `StaffTable` to be clearer.

#### Password update with field validations
passwordForm


sergiogcx (approved):
Amazing work as always! 🥇

mateoclarke (commented):
Great stuff in here Mike, great to see feature implementation on both the front and backend in the same PR.

jim_carrey_typing

PR #120, opened 2020-11-24
update sub folder routing /moped

mateoclarke:
This PR fixes (🤞 ) the bug I introduced where hitting the staging URL (https://moped.austinmobility.io/moped/) shows as a blank white page.

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -1,8 +1,9 @@
 {
-  "name": "react-material-dashboard",
-  "author": "DeviasIO",
+  "name": "atd-moped-editor",
+  "author": "ATD Data & Technology Services",
   "licence": "MIT",
   "version": "1.0.0",
+  "homepage": "/moped",
   "private": false,
   "scripts": {
     "start": "env-cmd -e local react-scripts start",
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -1,7 +1,7 @@
 {
   "name": "atd-moped-editor",
   "author": "ATD Data & Technology Services",
-  "licence": "MIT",
+  "licence": "CC0 1.0 Universal",
   "version": "1.0.0",
   "homepage": "/moped",
   "private": false,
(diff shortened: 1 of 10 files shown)


moped-editor/package.json
```diff
@@ -1,8 +1,9 @@
 {
-  "name": "react-material-dashboard",
-  "author": "DeviasIO",
+  "name": "atd-moped-editor",
+  "author": "ATD Data & Technology Services",
   "licence": "MIT",
```
sergiogcx, line 4: Not opinionated, but should we change the license type to GPL or something like that? Just out of curiosity

sergiogcx:
My only observation is that there is an image that does not load, not important but I wonder if there are any other resources elsewhere in the code that do not load (CSS, other images, etc):




The `/static` string does not occur too often in the app, but it might be worth refactoring now rather than later:

sergiogcx (approved):
It works! Thank you so much for working on this! 🙌

mateoclarke:
@sergiogcx good catch on the assets that didn't load. I'll fix and refactor that before merging. 🙏

mddilley (approved):
Checked out the deploy preview and also locally. I didn't encounter the same image bug that Sergio mentioned, but it definitely worth looking into. Everything looks great to me! 🎉

moped-editor/package.json
mateoclarke: yeah, I'm also not opinionated but I think the standard me and @johnclary had discussed a while back was just public domain CC0 1.0 Universal. 

ex: https://github.com/cityofaustin/atd-dockless-dataviz/blob/master/LICENSE

Probably something we should get back into the habit of updating across our projects.

johnclary: yep, that's generally what i use but also often forget to add it. i didn't realize the creative commons license [was available in github's pre-configured list](https://choosealicense.com/licenses/cc0-1.0/), but it is now. nice!

PR #122, opened 2020-11-28
API | Add update password route test and update existing tests

mddilley:
This PR updates the User API test coverage to work with the latest improvements and also adds a test for the new password update route. I also fixed the order of nested `@patch` decorators in the test methods to pass in arguments in the right order. https://docs.python.org/3/library/unittest.mock.html#quick-guide

Screen Shot 2020-11-27 at 4 36 41 PM

### DIFF
--- a/moped-api/tests/test_users.py
+++ b/moped-api/tests/test_users.py
@@ -19,17 +19,6 @@ def create_user_claims():
     }
 
 
-# "cognito_user_id": cognito_id,
-#         "date_added": generate_iso_timestamp(),
-#         "email": json_data.get("email", None),
-#         "first_name": json_data.get("first_name", None),
-#         "last_name": json_data.get("last_name", None),
-#         "is_coa_staff": is_coa_staff(json_data["email"]),
-#         "status_id": json_data.get("status_id", 1),
-#         "title": json_data.get("title", None),
-#         "workgroup": json_data.get("workgroup", None),
-#         "workgroup_id": json_data.get("workgroup_id", None),
-
 mock_users = [
     {"email": "neo@test.test", "password": "test123", "roles": ["moped-viewer"]},
     {"email": "morpheus@test.test", "password": "test123", "roles": ["moped-viewer"]},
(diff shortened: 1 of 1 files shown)


moped-api/tests/test_users.py
```diff
@@ -220,19 +220,29 @@ def test_creates_user_no_auth(self):
     @patch("flask_cognito._cognito_auth_required")
     @patch("users.users.is_valid_user")
     @patch("users.users.has_user_role")
+    @patch("users.users.is_valid_user_profile")
+    @patch("users.users.format_claims")
     @patch("users.users.put_claims")
+    @patch("users.users.generate_user_profile")
+    @patch("users.users.db_create_user")
```
sergiogcx, line 227: This is tremendous! 👍

moped-api/tests/test_users.py
```diff
@@ -424,7 +460,8 @@ def test_deletes_user(
         )
         _get_current_object = Mock(return_value=claims)
         patch(
-            "claims.current_cognito_jwt", Mock(_get_current_object=_get_current_object),
+            "claims.current_cognito_jwt",
+            Mock(_get_current_object=_get_current_object),
         ).start()
```
sergiogcx, line 465: Reading into Mock and Patch documentation :)

moped-api/tests/test_users.py
```diff
@@ -438,14 +475,70 @@ def test_deletes_user(
...
+            content_type="application/json",
+        )
+        response_dict = self.parse_response(response.data)
+
+        # Get user list to compare length
+        user_list_response = self.client.get("/users/")
+        user_list = self.parse_response(user_list_response.data)
+
+        assert isinstance(response_dict, dict)
+        assert isinstance(response_dict.get("success", None), dict)
```
sergiogcx, line 544: Thanks for your cleverness Mike! :) I enjoy learning from you

sergiogcx (approved):
🚢

moped-api/tests/test_users.py
mddilley: Thanks! There is a lot of patching in this branch to get through the routes so if we ever decide we need more coverage for the DynamoDB actions then hopefully it will provide a good base for that. 👍

PR #123, opened 2020-12-01
4507 lz proj notes

lazavazav:
added notes column to step 2 of new project. added new column "notes" to moped_proj_personnel table. While testing, encountered graphql errors such as no workgroup column exists or workgroup_id can't be null. I made small changes to the table to allow the mutation to go through and be tested. With these changes, was successful adding data to table on final step.


moped-database/migrations/1606833580979_alter_table_public_moped_project_alter_column_project_priority/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."moped_project" ALTER COLUMN "project_priority" TYPE text;
```
sergiogcx, line 1: Is there a reason we are using text and not integers? integers are easier to index and more performant than text, but it is not a big problem whatsoever.

moped-editor/src/views/projects/DefineProjectForm.js
```diff
@@ -99,10 +99,8 @@ const DefineProjectForm = props => {
   if (fiscalLoading) return <CircularProgress />;
   if (fiscalError) return `Error! ${fiscalError.message}`;
 
-  console.log(phases);
-
```
sergiogcx, line 103: Thank you for getting rid of the console log 👍

moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -172,13 +173,15 @@ const NewProjectView = () => {
       $role_name: String! = ""
       $first_name: String! = ""
       $last_name: String! = ""
+      $notes: String! = ""
```
sergiogcx, line 176: I did not know you could have defaults in a hasura graphql query 🙌  this is excellent!

moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -222,15 +226,16 @@ const NewProjectView = () => {
...
       let name_split = name_array.split(" ");
       let first_name = name_split[0];
       let last_name = name_split[1];
       let workgroup = value.workgroup;
       let role_name = value.role;
-      addStaff({ variables: { workgroup, role_name, first_name, last_name } });
-    });
+      let notes = value.notes.userInput;
+      addStaff({ variables: { workgroup, role_name, first_name, last_name, notes } });
+     });
```
sergiogcx, line 238: The other devs seem ok with using lodash, I do have a recommendation to import specific libraries from lodash as opposed to the entire library. Not a big deal, if it is hard to refactor it is not important.

moped-database/migrations/1606833580979_alter_table_public_moped_project_alter_column_project_priority/up.sql
lazavazav: JD wanted to switch to text for the user and in our code we have both integer and text so we could do the mutation with int. It was originally working fine in the old endpoint as int but with the new endpoint, I got a graphql error that it was expecting text so I changed it just to get everything working as a quick fix and so I could see if the other parts of the mutations were working.

sergiogcx (approved):
Thank you @lazavazav! 🙌  Merge away! 🚀  I have learned a lot from you while reviewing your code! :)

moped-database/migrations/1606833580979_alter_table_public_moped_project_alter_column_project_priority/up.sql
sergiogcx: I think it will be fine... as long as we are not expecting a lot of different priorities :)

PR #124, opened 2020-12-01
Moped API | Handle CORS in the MOPED User Management API

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4583

This PR adds CORS setup to the users blueprint in the Flask app. CORS was already set up app-wide by Sergio 🙏 but it seems that blueprints require their own initialization.


sergiogcx (approved):
Sweet! Thanks for looking into this!

PR #128, opened 2020-12-03
Kickstand | gqlAbtract.js updates and import into Moped Editor via Kickstand #4616

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4616

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -41,7 +41,7 @@
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@material-ui/styles": "^4.10.0",
     "apollo-boost": "^0.4.9",
-    "atd-kickstand": "github:cityofaustin/atd-kickstand#4616-sg-abastract",
+    "atd-kickstand": "github:cityofaustin/atd-kickstand#main",
     "aws-amplify": "^3.3.4",
     "axios": "^0.21.0",
     "chart.js": "^2.9.3",
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -41,6 +41,7 @@
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@material-ui/styles": "^4.10.0",
     "apollo-boost": "^0.4.9",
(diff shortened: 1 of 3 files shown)


mddilley:
@sergiogcx I was about to review this and the other PR but noticed that the builds failed with `Error: Could not resolve './graphql/GQLAbstract' from src/index.js`. Not sure what is going on since everything looks good in the code exports/imports.

sergiogcx:
@mddilley Ah, my bad. It's still not very clear how we are going to do this. The way I made this work is by using `npm link`, I am happy to correct things as needed but I need direction.

sergiogcx:
The definitions are in the branch build though:
https://github.com/cityofaustin/atd-kickstand/blob/4db9103b11f5347cecc408c08c01caa577293788/dist/index.js

mddilley:
Ah gotcha, that makes total sense! 👍

moped-editor/package.json
```diff
@@ -41,6 +41,7 @@
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@material-ui/styles": "^4.10.0",
     "apollo-boost": "^0.4.9",
+    "atd-kickstand": "github:cityofaustin/atd-kickstand#4616-sg-abastract",
```
mateoclarke, line 44: This works for now, but we'll want to figure out some semver pattern with kickstand I think and use a non-feature branch. I bet you've already thought about this too.

mateoclarke: Also just realized, if we delete this branch from Github (like we normally do after merging a feature branch) it will break.

mateoclarke (approved):
If we can figure out that whats causing that no breaking build error, this is good to merge. Lmk if we need to work through that together.

moped-editor/package.json
sergiogcx: Let me make it the main branch

sergiogcx: After updating it to main branch, it looks like it works on my end:

PR #131, opened 2020-12-04
Moped Editor | Fix HTTPS on Start-Local #4631

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4631

To test, simply run `npm run start` and check if the HTTPS is there (even if it says "not secure")

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -6,7 +6,7 @@
   "homepage": "/moped",
   "private": false,
   "scripts": {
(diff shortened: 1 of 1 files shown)


mateoclarke (approved):
🚢

mateoclarke:
thanks for fixing this Sergio.

mddilley (commented):
👍

PR #132, opened 2020-12-04
Editor | Use error responses from User Management API for Add User form feedback

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4494

This PR updates the user add/edit forms to display errors return from the user API. I also fixed a bug that prevented these forms from submitting successfully in the current staging/production apps. Open to any feedback on whether we should make the API errors human-readable on the front-end or not. I definitely like the way the API returns the regex in the errors the way it does currently.

#### API errors in Add User form
Screen Shot 2020-12-04 at 1 32 46 PM

### DIFF
--- a/moped-editor/src/views/staff/helpers.js
+++ b/moped-editor/src/views/staff/helpers.js
@@ -32,6 +32,7 @@ export function useUserApi() {
     axios(config)
       .then(res => {
         setResult(res.data);
+        setError(null); // Clear errors from previous attempts
         setLoading(false);
         !!callback && callback();
       })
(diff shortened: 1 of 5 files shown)


moped-editor/src/views/staff/helpers.js
```diff
@@ -28,13 +29,29 @@ export function useUserApi() {
...
+      .catch(err => {
+        setError(err.response.data.error);
+      });
   };
 
-  return [result, loading, requestApi];
+  return { result, error, loading, requestApi };
 }
+
+const errorsToTranslate = {
```
mateoclarke, line 47: 💅

mateoclarke (approved):
🚢

PR #134, opened 2020-12-08
4130 lz nested mutation

lazavazav:
Issue 4130 asked for a nested mutation to send the project_id and other data from the UI to the database at the same time upon user clicking on the finish button at the end of step 3 of the newProject feature. It is working after this last commit fairly well.

For some of the array relationships, I added extra data coming from the UI instead of creating dummy data. For example, in moped_proj_financials, eCapris_id is needing to be added so I went ahead and had the data come from the UI rather than defining it as dummy data in the object array. Did the same for the project_date in moped_proj_dates. It's coming from the UI. Please advise if that is what we want or should we just add it as dummy data like the rest of the non-nullable columns. If the column was set as nullable, I didn't add the dummy data at all.

One problem I have encountered when testing is that if I don't choose true for capitally funded and add an eCapris_id value in the UI, I get a graphql error that says 
(foreign key violation, insert or update on table "moped_proj_phases" violates foreign key constraint "moped_phase_history_phase_name_fkey" ) 

I went in to database and changed eCapris_id to nullable but still got that error.

Finally, just for my own knowledge, I would like to know why it is necessary to have certain rows defined under affected rows such as line 229, 244, 254 and even 249 and 230. If I removed any or some of those lines which I tried to do since they seemed unnecessary, graph ql errors happened all over the place and the mutation was not successful. 

Thanks

Closes https://github.com/cityofaustin/atd-data-tech/issues/4130


zappfyllc (dismissed):
eCapris as nullable is 100% correct: my apologies, that constraint existed because I thought project_financials would require an eCapris ID but I think there are plenty of cases where there either won't be an eCapris ID or it will get added AFTER a project has already been entered.

For the date_added columns in these tables, I just want to clarify that the database should handle those timestamps, you don't need to push any data (project start date, for example) into those columns during the mutation. As long as that's the case, this commit is good to go!

zappfyllc (commented):
You can remove moped_proj_entities and moped_proj_timeline for now: let's address those later when we have more in-depth requirements for what data we need saved in terms of who/when, which are still details being sorted out by stakeholders in some of our work sessions/alignment.

For moped_proj_phases, the phase name is going to have to match the name in the moped_phases table exactly to be inserted correctly and not get a foreign key constraint warning. moped_proj_phases is a join table in the many-to-many relationship between moped_project and moped_phases (i.e. moped_project > moped proj_phases < moped_phases table, if that makes sense)

moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -138,17 +138,80 @@ const NewProjectView = () => {
...
+          capitally_funded: $capitally_funded,
+          project_priority: $project_priority,
+          moped_proj_entities: {
+            data:{
+              date_added: "01-21-2021",
+              project_sponsors: "{Sponsor A, Sponsor B}",
+              project_groups: "A",
+              partners: "{A, B}",
+              workgroups: 1,
+              project_personnel: "{Employee A, Employee B}"
```
sergiogcx, line 158: I think these values can be hard-coded here for now, what would be the strategy to pass variables to the project entities? :)

sergiogcx:
I am seeing this error whenever I try to create a project:

sergiogcx:
I decided to start over, deleted the local database and restarted it. Went over again at it and saw another bug:



This is the query it attempted to execute:

```
mutation NewProject(
	$project_name: String! = "", $project_description: String! = "", $current_phase: String! = "", $current_status: String! = "", $eCapris_id: String! = "", $fiscal_year: String! = "", $start_date: date = "", $capitally_funded: Boolean! = false, $project_priority: String! = ""
	) {
  		insert_moped_project(
  			objects: [
  				{
  					project_name: $project_name, project_description: $project_description, current_phase: $current_phase, current_status: $current_status, eCapris_id: $eCapris_id, fiscal_year: $fiscal_year, start_date: $start_date, capitally_funded: $capitally_funded, project_priority: $project_priority, moped_proj_entities: {
  						data: {
  							date_added: "01-21-2021", project_sponsors: "{Sponsor A, Sponsor B}", project_groups: "A", partners: "{A, B}", workgroups: 1, project_personnel: "{Employee A, Employee B}"
						}
					},
					moped_proj_timeline: {
						data: {
							active_phase: "bidding", active_phase_start: "2020-06-10", active_phase_end: "2020-08-10", active_phase_length: 2, active_milestone_start: "2020-05-10", active_milestone_end: "2020-09-10", active_milestone_length: 10, current_status: $current_status, capital_projects_explorer_id: "f"
						}
					},
					moped_proj_financials: {
						data: {
							subproject_name: "Project A", subproject_status: "active", budget_total: 3000.00, expenses_total: 10000.00, expenses_ytd: 40000.00, expenses_previous: 40000.00, budget_previous: 3000.00, budget_available: 50000.00, last_updated: "2020-09-10", primary_funding_source: "capital funds", eCapris_id: $eCapris_id
						}
					},
					moped_proj_phases: {
						data: {
							completion_percentage: 10, completed: false, phase_name: $current_phase
						}
					},
					moped_proj_status_history: {
						data: {
							date_added: "2020-01-01", milestone_privacy: false, status_name: $current_status
						}
					},
					moped_proj_dates: {
						data: {
							active_date: false, date_day: 10, date_month: 6, date_year: 2020, date_type: "2020-06-10", project_date: $start_date
						}
					}
				}
			]
		) {
    affected_rows
    returning {
      project_name
      project_description
      project_priority
      current_phase
      current_status
      eCapris_id
      fiscal_year
      capitally_funded
      start_date
      moped_proj_entities {
        project_id
        project_sponsors
        date_added
        __typename
      }
      moped_proj_timeline {
        project_id
        current_status
        active_phase
        __typename
      }
      moped_proj_financials {
        project_id
        eCapris_id
        __typename
      }
      moped_proj_phases {
        project_id
        phase_name
        completed
        __typename
      }
      moped_proj_status_history {
        project_id
        status_name
        date_added
        __typename
      }
      moped_proj_dates {
        project_id
        project_date
        active_date
        __typename
      }
      __typename
    }
    __typename
  }
}
```

These are the variables it passed to the queries:

```
{
	capitally_funded: "The PRAAAAHJECT"
	current_phase: "active"
	current_status: "active"
	eCapris_id: "1245678911111"
	fiscal_year: "2021"
	project_description: "2021-01-01"
	project_name: "The project description"
	project_priority: "High"
	start_date: true
}
```

sergiogcx:
Ok, it managed to insert once. I restarted it and attempted, again. I attempted to create another project, and I tried to add two staff members, the note field is copied across all note fields whenever I type notes:

lazavazav:
Well, I'm glad you were able to insert at least one time successfully. I think the notes issue will be fixed soon with the work we are doing getting it ready for the demo.

lazavazav:
I will revisit this PR once the notes issue is fixed.

moped-database/metadata/tables.yaml
```diff
@@ -2807,6 +2807,51 @@
...
+          schema: public
+          name: moped_proj_status_history
+  - name: moped_proj_timeline
+    using:
+      manual_configuration:
+        remote_table:
+          schema: public
+          name: moped_proj_timeline
+        column_mapping:
+          project_id: project_id
```
sergiogcx, line 2854: This is great! 💟  I can see the relationships working in the local database:

zappfyllc:
@sergiogcx @lazavazav @amenity I think in the interest of closing this out, I'm going to put out the suggestion that we **remove moped_proj_entities, moped_proj_timeline, moped_proj_financials, and moped_proj_dates from the mutation**. I'll then create an issue with the corresponding features that will actually touch on these entities as a placeholder for updating this mutation when we're shipping new features. The issue here is that the exhaustive work I did with the data model was waterfall-ish, and it really makes no sense to block iterative development by inserting in data that isn't actually used yet anywhere in the application. In fact, by inserting that data, we're just creating more cleanup we'll have to do later when we inevitably implement subsequent features and start seeing really bad test data.

**moped_proj_phases and moped_status_history do need to be inserted into through this workflow** in order to maintain a running history of phase and status changes, however, because subsequent edits of these fields/columns will happen when the user edits a project.

So, if this makes sense, **the goal here is to get a working mutation for the moped_project table, with nested inserts happening in the moped_proj_phases and moped_status_history tables.**

Additionally, @sergiogcx can correct me if I'm wrong with his database expertise, but I am under the impression that timestamp fields are generated by the database when a new row is added, meaning we should not need to deal with any date_added fields in the mutation. The only date field that will need to be appropriately handled is the project start date.

lazavazav:
Removed the above tables in mutation. Need to test again which I will do early this upcoming week.

sergiogcx:
@zappfyllc sounds good, I think this is not needed for now. @lazavazav let's wait on it.

PR #135, opened 2020-12-08
Adds support for SSO tokens for API Gateway #4421

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4421


mddilley (approved):
Should be easy peasy to manage users after signing in with SSO now. 🙏

PR #138, opened 2020-12-09
Adds aggregation permissions to all roles

sergiogcx:
To test, pull this branch and run hasura-cluster start and hasura-cluster console

for every table you should be able to see the permissions applied


mateoclarke (approved):
🚢

PR #139, opened 2020-12-10
Editor | Project extent placeholder

mddilley:
This PR adds a placeholder map for the project extent. It can be found in the 3rd step of the New Project stepper. Happy to update with any feedback. There is no interaction with the DB for now. Selected items are temporarily held in state.

#### Project Map
Screen Shot 2020-12-09 at 5 21 39 PM


moped-editor/src/views/projects/mapHelpers.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+
+export const MAPBOX_TOKEN = `pk.eyJ1Ijoiam9obmNsYXJ5IiwiYSI6ImNraWV4dHR0ZjAwNnYyd3FwYjFoNHduaDcifQ.--3vRm2KHq1gh5K_L0pqtA`;
```
sergiogcx, line 3: This should be part of the environment variables I think

sergiogcx:
Woah! Impressive work!

sergiogcx (approved):
Amazing work as always!

mateoclarke (approved):
Looks great Mike! 

teigen_clapping

moped-editor/src/views/projects/mapHelpers.js
mddilley: Cool, it is restricted to staging, production, and dev for now but I think that y'all are right that we should hide it away. I'll create an issue to delete this token, create a fresh one, and update where this comes from. 👍

PR #141, opened 2020-12-10
DB | Update workgroup migration

mddilley:
This PR updates the previous workgroup migration that failed its GH Action run. The problem was a single workgroup already present in the `moped_workgroup` table conflicting with the 1st insert statement.


mddilley:
Paired with Sergio to update staging and production with these insertions so this migration is no longer needed. Closing.

PR #143, opened 2020-12-11
Project Definition | Implement progress/spinner animation in "Finish" button #4668

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4668


lazavazav:
Hi Sergio, I merged your branch with a new branch I created (4666_lz_projectSummary2) that is similar to my original projectSummary branch to test out your spinner and redirects with the project summary page upon user clicking on finish. Everything looks great except when I click on finish, the data does get saved in database and shows up in the Projectstable successfully but the redirect to the summary page happens a split second later gets redirected again to the dashboard with not enough time to look at the summary page. 

In my original branch 4666_lz_projectSummary, I didn't use the fancy spinner, used a simpler setTimeout function with no success or error messages if user didn't fill everything out etc.....so this is way better. 

I also saw this error below in the console after clicking on the finish button.
index.js:1 Warning: Cannot update a component (`BrowserRouter`) while rendering a different component (`NewProjectView`). To locate the bad setState() call inside `NewProjectView`, follow the stack trace as described in https://fb.me/setstate-in-render
    in NewProjectView (at routes.js:40)

Should we try to fix my new branch 4666_lz_projectSummary2 to use for the demo on Wednesday which uses your spinner and redirects with my summary page?

mateoclarke (approved):
This looks good but I encountered a few things I want to make sure are addressed if not in this PR, a future follow-up. I know a lot is in flux so maybe these fixes are already addressed in another PR but listing here to make sure nothing slips through the cracks.

1. The first time I tried to submit the project, My mutation hit an error around the data value. I know we talked about this in sync yesterday but I just to make sure its clear who is working to fix that and where that work is happening.

Screen Shot 2020-12-15 at 3 25 53 PM

2. Why are there two buttons?  Can we remove the one we don't need?

Screen Shot 2020-12-15 at 3 25 40 PM

3. After the form submits redirect breaks and redirects to a 404. This may be due to the updated routes I mentioned in Slack and in my recent PR.

PR #144, opened 2020-12-11
Moped Editor | Direct path access | Routing issue #4680

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4680


mateoclarke (approved):
Wasn't sure if setting global state with localStorage via Context was a suggested pattern or not but I found a bunch of examples of people doing that in React articles so these seems pretty standard and simple. Thanks for bug hunting @sergiogcx!

judge_judy_fly_swatter

PR #146, opened 2020-12-13
4666 lz project summary

lazavazav:
This branch currently has a working summary page upon user submitting data in the UI and clicking on finish. The spinner spins for 2 seconds and then the summary page with the project details shows up. There is a back button on the summary page that takes user back to dashboard.  This branch does not yet have any required field validation or error message if user doesn't fill everything out. Currently, user will get a graphql error if all fields are not filled in if there is a null violation associated with the non filled out input.

I changed the start script only to get it to work on my machine. Before merging with main if you want this code, I will change it back.

Finally, there is still a bug in notes on the 2nd step where the input field doesn't clear out properly but Mike is working on that.


moped-editor/package.json
```diff
@@ -7,7 +7,7 @@
   "private": false,
   "scripts": {
     "start": "env-cmd -e local npm run start:local",
-    "start:local": "export HTTPS=true && react-scripts start",
+    "start:local": "react-scripts start",
```
sergiogcx, line 10: We need HTTPS, otherwise our connections to other parts of our stack could be compromised (Hasura), or straight-up fail (Single Sign-On).

It's only matter of putting `export HTTPS=true &&` back into the start:local script.

moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -227,16 +241,27 @@ const NewProjectView = () => {
...
+    });
+    if (!loading) {
+      setSuccess(false);
+      setLoading(true);
+      timer.current = window.setTimeout(() => {
+        setSuccess(true);
+        setLoading(false);
+        navigate("/moped/projects/new/success");
+      }, 2000);
+    }
```
sergiogcx, line 264: This is very good! thanks for showing me how to use navigate 👍

moped-editor/src/routes.js
```diff
@@ -37,6 +38,7 @@ const routes = [
       { path: "staff/edit/:userId", element: <EditStaffView /> },
       { path: "projects", element: <ProjectsListView /> },
       { path: "projects/new", element: <NewProjectView /> },
+      { path: "projects/new/success", element: <ProjectSummary /> },
```
sergiogcx, line 41: I like the success route, but it would be more efficient if we changed it tom something like `projects/:projectId`, that way we can use that same page for any project and not just the one that has been created. This may be a good question for @mateoclarke and @mddilley

moped-editor/src/views/projects/ProjectSummary.js
```diff
@@ -0,0 +1,256 @@
...
+      marginLeft: theme.spacing(1),
+    },
+  }));
+  const classes = useStyles();
+
+  const navigate = useNavigate();
+
+  const handleClick = () => {
+    navigate("/moped/dashboard");
+  };
```
sergiogcx, line 48: Maybe this should redirect back to the projects list

moped-editor/src/views/projects/ProjectSummary.js
```diff
@@ -0,0 +1,256 @@
...
+                ))}
+              </Grid>
+            </Grid>
+          </CardContent>
+          <div>
+            <Divider />
+            <Box pt={2} pl={2} className={classes.buttons}>
+              <Button className={classes.button} onClick={handleClick}>
+                Back
+              </Button>
```
sergiogcx, line 247: Maybe refactor to redirect to the projects list

sergiogcx (changes requested):
We need to have HTTPS, other than that I think the changes you have will work.

moped-editor/src/views/projects/ProjectSummary.js
lazavazav: sounds like a plan, I'll change it tonight.

moped-editor/package.json
lazavazav: added it to my latest commit. I always have to remove it when I try to start the app on my local machine after syncing again with main. No worries. It's on my radar as something to fix before final commit and PR so hopefully it won't happen again.

moped-editor/src/routes.js
lazavazav: sounds like a plan.

amenity:
Miscommunication here. 🤪  This does not close #4666, that's a backlog enhancement.

moped-editor/src/routes.js
mateoclarke: Yeah, I agree the `projects/:projectId` should be the pattern instead of a dedicated success page. Going to work on this update tonight.

mateoclarke (approved):
After including my [add-on Pull Request](https://github.com/cityofaustin/atd-moped/pull/149), I think we're good to go ahead and merge this and call the Project Summary ready for demo. We'll need to fill out features within the other tabs.

PR #149, opened 2020-12-15
Project summary updates

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4525

Screen Shot 2020-12-15 at 3 43 14 AM

### DIFF
--- a/moped-editor/src/views/projects/NewProjectView.js
+++ b/moped-editor/src/views/projects/NewProjectView.js
@@ -207,6 +207,8 @@ const NewProjectView = () => {
   const [success, setSuccess] = React.useState(false);
   const timer = React.useRef();
 
+  console.log(success);
+
   React.useEffect(() => {
     return () => {
       clearTimeout(timer.current);
--- a/moped-editor/src/views/projects/NewProjectView.js
+++ b/moped-editor/src/views/projects/NewProjectView.js
@@ -204,10 +204,7 @@ const NewProjectView = () => {
   const navigate = useNavigate();
 
   const [loading, setLoading] = React.useState(false);
-  const [success, setSuccess] = React.useState(false);
   const timer = React.useRef();
-
-  console.log(success);
 
   React.useEffect(() => {
     return () => {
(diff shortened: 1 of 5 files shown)


amenity:
So beautiful, @mateoclarke. 🤗 Seeing those tabs in action is super helpful as I think about more granular information architecture and navigation. 

Would it be possible to display `eCapris ID` — with a space and a capital "D"?

moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -207,6 +207,8 @@ const NewProjectView = () => {
   const [success, setSuccess] = React.useState(false);
   const timer = React.useRef();
 
+  console.log(success);
+
```
sergiogcx, line 211: ...Rid?

moped-editor/src/views/projects/ProjectSummary.js
```diff
@@ -19,234 +19,187 @@ import {
...
-    },
-  }));
-  const classes = useStyles();
-
-  const navigate = useNavigate();
-
-  const handleClick = () => {
-    navigate("/moped/projects");
-  };
+function TabPanel(props) {
```
sergiogcx, line 73: It never occurred to me to have two components within a file, this is clever, but maybe split into two different files?

sergiogcx (approved):
Brilliant work as always! 🥇 🚀

moped-editor/src/views/projects/ProjectSummary.js
mateoclarke: yeah, I definitely agree that this Project Summary component needs to be broken down into smaller components. I started doing a bit of that, but for the sake of time I just keep pushing ahead. This could definitely use more refactoring.

moped-editor/src/views/projects/NewProjectView.js
mateoclarke: yeah good call. the `success` var wasn't being used so I just added a console.log as a temp fix to fix the breaking build. But I think the better solution is just to remove it entirely.

mateoclarke:
> So beautiful, @mateoclarke. 🤗 Seeing those tabs in action is super helpful as I think about more granular information architecture and navigation.
> 
> Would it be possible to display `eCapris ID` — with a space and a capital "D"?

@amenity will do! thanks for the review!

PR #154, opened 2020-12-15
Project Definition | Follow-up details

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4544
Closes https://github.com/cityofaustin/atd-data-tech/issues/4719

This PR updates the new project forms to address some bugs with input state updates. The form will now preserve input updates when navigating back and forth between steps. I also ended up refactoring `DefineProjectForm` to use controlled components instead of using React Hook Form because it was the path of least resistance.

- Updated styling of Add Staff form icons and fields 
- Fixed `TypeError: Cannot read property 'workgroup' of null` related to the `handleGroupChange` function
- Submit form no longer returns GraphQL error
- Workgroup updates when staff member is selected
- Notes no longer mirror across staff member rows
- Notes state repopulates on change
- Workgroup moved to helper text below staff member name (thanks @sergiogcx and @mateoclarke for this idea in dev sync)
- Can no longer reproduce GraphQL error from https://github.com/cityofaustin/atd-data-tech/issues/4701
- Fixed a bug where all staff rows disappeared if the first row was removed
- Form capitalizes `phase_name` and `status_name` options in the UI
- Only active staff populate in the staff member field
- Capitally funded switch updates boolean state properly (switched to setting/getting `checked` attribute instead of `value`)

#### Updated team table styles
Screen Shot 2020-12-15 at 12 20 46 PM


moped-editor/src/views/projects/NewProjectView.js
```diff
@@ -257,45 +254,26 @@ const NewProjectView = () => {
   const handleSubmit = () => {
     // Change the initial state...
     setLoading(true);
-    // data from Define Project going to database
```
mateoclarke, line 260: glad to see this refactored! 👍

moped-editor/src/views/projects/ProjectTeamTable.js
```diff
@@ -28,193 +30,206 @@ const useStyles = makeStyles(theme => ({
...
+    props.setStaffRows(updatedStaffRows);
   };
 
-  const handleGroupChange = (value, item, index) => {
-    item.workgroup = value.workgroup;
-    props.setStaffRows(props.StaffRows);
-  };
+  const handleNameAndWorkgroupChange = (value, index) => {
+    const updatedName = value?.name || null;
+    const updatedWorkgroup = value?.workgroup || null;
```
mateoclarke, line 86: ohhh! nice use of the optional chaining operator!! didn't realize we were already able do this with our current webpack/babel setup

moped-editor/src/views/projects/ProjectTeamTable.js
```diff
@@ -28,193 +30,206 @@ const useStyles = makeStyles(theme => ({
...
-  let nameOption = [];
-  members.moped_users.forEach(name =>
-    nameOption.push({
-      name: name.first_name + " " + name.last_name,
-      workgroup: name.workgroup,
+  let userOptions = [];
+  members.moped_users.forEach(user =>
+    userOptions.push({
+      name: user.first_name + " " + user.last_name,
+      workgroup: user.workgroup,
```
mateoclarke, line 123: i appreciate this renaming to better match our tables 👍

mateoclarke (approved):
In this preview deploy app, this worked great! Feels like its all coming together now!!! ✨

amenity (approved):
LSTM (Looks Splendid To Me) 🙌

sergiogcx (approved):

moped-editor/src/views/projects/ProjectTeamTable.js
mddilley: Same, I saw it somewhere in Devias so I just went with it. 👍

PR #155, opened 2020-12-15
Fix some console warnings

mddilley:
This fixes a couple of console warnings that broke the `main` deploy.

#### The warnings for reference
Screen Shot 2020-12-15 at 5 31 50 PM

Screen Shot 2020-12-15 at 5 32 55 PM


sergiogcx (approved):
Thanks for this quick fix, in fact I see the same errors whenever I pulled them locally!

PR #157, opened 2020-12-16
Handle API data validation for Edit User route #4722

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4722

In the Flask API, the data validation schema being used for create and edit user routes is the same. The password field is not always present in the edit user payload coming from the UI. When it is not present, the request fails and returns an error response with validation failure details.

- Need to handle validation in the case when a user data is updated but not their password
- Revise validation for status_id (value is either 0 or 1)

### DIFF
--- a/moped-api/users/helpers.py
+++ b/moped-api/users/helpers.py
@@ -1,7 +1,7 @@
 """
 Helper methods to update the database via GraphQL
 """
-import re
+import re, copy
 from cerberus import Validator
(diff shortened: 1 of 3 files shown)


mddilley (approved):
Thanks for fixing this, @sergiogcx! 🙏

mateoclarke (commented):
👍

PR #159, opened 2020-12-16
Update Projects, Phase, and Status tables with better dummy data #4669

sergiogcx:
This is done at the local stage only. Staging and production have been updated manually.

Closes https://github.com/cityofaustin/atd-data-tech/issues/4669


mddilley (approved):
Great to have some projects in the local environment now. 🙌 

I tested locally and the migrations and data seed went smoothly. 👍

PR #163, opened 2020-12-18
Moped User Login | Give user feedback after a user submits their login

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4741

This PR adds a spinner that appears when either the email login or SSO login is used to access Moped. The [Backdrop MUI](https://material-ui.com/components/backdrop/) spinner component is used for now just make it super clear when the spinner is present. We can update to any other loading component.

### SSO
ssoSpinner

### Email login
loginSpinner

### DIFF
--- a/moped-editor/src/auth/useAuthentication.js
+++ b/moped-editor/src/auth/useAuthentication.js
@@ -85,6 +85,7 @@ const useAuthentication = () => {
   }, [refreshState]);
 
   const signIn = useCallback(() => {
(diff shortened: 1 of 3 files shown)


sergiogcx (approved):
An elegant solution!

amenity (approved):
LSTM! 👌

mateoclarke (approved):
Looks great! 

thumbs_up_thomas

PR #167, opened 2020-12-24
Project Extent | Store map feature data in Moped

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4748

This PR updates the project extent map in the New Project workflow to persist GeoJSON and also IDs for the features from the vector tile layers in the DB. These two columns could be combined in the future, but I wanted to leave it open since the number of vector tile layers available in this map is undetermined at the moment.

The two columns added to the database are named `project_extent_geojson` and `project_extent_ids`. I chose GeoJSON for now since it will be easy to work with in the next branch to visualize the project extent in the Project Summary Details view.


mateoclarke (approved):
This looks great! As we discussed on the video chat, my only feedback is that maybe we should include the Hasura metadata updates that give `moped-admin` users the ability to query and mutate these fields. We can also punt that for a larger role-based refactoring that sounds like is coming down the pipeline very soon. 

🚢 

animaniacs_map

moped-database/migrations/1608323427908_alter_table_public_moped_project_add_column_project_extent/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."moped_project" ADD COLUMN "project_extent_ids" json NULL;
```
sergiogcx, line 1: Use jsonb instead of regular json, performance is better and we can index certain fields. Also, we get to use really cool database operators.

https://hasura.io/blog/postgres-json-and-jsonb-type-support-on-graphql-41f586e47536/

moped-database/migrations/1608670107519_alter_table_public_moped_project_add_column_project_geometries/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."moped_project" ADD COLUMN "project_extent_geojson" json NULL;
```
sergiogcx, line 1: Shouldn't this be a geometry type instead of json?

https://hasura.io/blog/native-support-for-postgis-topology-operators-now-in-graphql-engine/

sergiogcx: IIRC, you can insert geometry objects as GeoJSON, isn't that right? But if we are to stay with json I would prefer we stick with jsonb.

moped-database/migrations/1608323427908_alter_table_public_moped_project_add_column_project_extent/up.sql
sergiogcx: Notes on differences:
https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/#:~:text=What's%20jsonb,string%2C%20but%20as%20binary%20code.

mddilley: Thanks for sharing these articles. They fits nicely in with those DB videos that I watched last week. 📚

moped-database/migrations/1608670107519_alter_table_public_moped_project_add_column_project_geometries/up.sql
mddilley: Swapping this out for jsonb was straightforward - looking forward to see the next phase of all this!

johnclary:
thanks @mddilley! good convo in the review. i'm getting a graphql error when I try to save a new project:

`Error: GraphQL error: no such type exists in the schema: 'jsonb'`

Looks like we're still waiting on a DB migration for this? Sorry if I jumped the gun on testing it.

mddilley:
@johnclary No worries, I'm glad that you brought this up! I'm thinking that I need to update some permissions. Mateo and I talked briefly about this at the end of Tuesday. I'll create an issue to patch this up. 🙏

mateoclarke:
@johnclary were you testing on the deploy branch or locally? I wouldn't expect you to be able to save in the deploy branch since the backend migrations wouldn't have been updated. Locally is where I was seeing a different GraphQL issue, not on trying to save though.

johnclary:
@mateoclarke i was testing in staging.

PR #169, opened 2020-12-28
Moped Tables | Implement basic table #4784

sergiogcx:
Implements initial GridTable component
- Makes full use of GQLAbstract library
- Displays columns according to configuration
- Implements pagination
- Implements sorting

Closes https://github.com/cityofaustin/atd-data-tech/issues/4784

** IT REQUIRES THE LATEST VERSION OF THE GQLAbstract LIBRARY, BEFORE YOU RUN, BE SURE TO UPDATE YOUR ATD-KICKSTAND LIBRARY IN YOUR LOCAL ENVIRONMENT **

During development, I came across an underlying issue with Apollo caching and I still could not pinpoint the issue after a couple hours trying to figure it out. This issue has been created to troubleshoot caching in the future.
https://github.com/cityofaustin/atd-data-tech/issues/4788


mateoclarke (approved):
This code looks excellent. Really appreciate all the organization and comments. The only feedback I have is visual and I'm fine doing that as a separate styling focused pass later on if you prefer for me too add these comments as a backlog issue to revisit later.

1. Padding around the loading spinner. I think the spinner could use a little room to breathe here. Could even see it being centered.

Screen Shot 2020-12-28 at 2 40 08 PM

2. The alignment of the sort arrow and the way the text wraps looks a little off visually.

Screen Shot 2020-12-28 at 2 40 15 PM

Maybe the example from the [Material UI `<DataGrid>` component](https://material-ui.com/components/tables/#data-table) could be a good example to look at.

Screen Shot 2020-12-28 at 2 46 11 PM

[Material-Table](https://material-table.com/ ) also has some good examples for visual inspiration.

3. I couldn't get a screengrab of this but I would only expect the column headers that are sortable to show the `cursor: pointer` on hover. The rest that aren't sortable, I would expect a more default cursor. The pointer implies there is interactivity.

PR #174, opened 2021-01-05
User Management | Restrict adding and editing users to Admins

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4782

This PR adds role-based restrictions to UI components and routes in Moped. It also uses the patterns for restricting access to only allow admin users to add and edit users. `GridTable` has been extended to allow custom UI components in place of the new item button in the toolbar and the inline edit buttons in the table rows. The `Can` component is passed in as the custom component for access purposes.

I threw in one small edit to the bounding box for the project map geocoder [as suggested by John](https://github.com/cityofaustin/atd-data-tech/issues/4715#issuecomment-756990098).

I also added some docs about the access control to the UI section of the Moped tech docs. https://app.gitbook.com/@atd-dts/s/moped-technical-docs/infrastructure/user-interface/access-control

### Staff Table with`moped-admin` role
Screen Shot 2021-01-12 at 2 33 33 PM

### Staff Table with `moped-editor` role
Screen Shot 2021-01-12 at 2 34 43 PM

### Staff Table with `moped-viewer` role
Screen Shot 2021-01-12 at 2 35 08 PM


moped-editor/src/auth/Can.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const Can = ({ perform, yes = null, no = null }) => {
+  const { user } = useUser();
+  const role = getHighestRole(user);
+
+  return check(rules, role, perform) ? yes : no;
+};
+
+export default Can;
```
sergiogcx, line 25: This whole file is brilliant! 💯

moped-editor/src/views/staff/StaffListViewQueryConf.js
```diff
@@ -16,14 +21,37 @@ export const StaffListViewQueryConf = {
...
+  customSingleItemButton: id => (
+    <Can
+      perform="user:edit"
+      yes={
+        <RouterLink to={`/moped/staff/edit/${id}`}>
+          <EditIcon color="primary" />
+        </RouterLink>
+      }
+    />
+  ),
```
sergiogcx, line 33: This is again very brilliant! Frankly, I dislike to see components in a configuration.

This should stay here for now, but we need to figure out a way to control this pattern.

Maybe we could have a section for custom components, maybe something like this:

```

import { StaffListViewCustomComponents } from "./StaffListViewCustomComponents";

{
...
customComponents: StaffListViewCustomComponents,
...
}

```

Custom Components file:

```
export const StaffListViewCustomComponents =  {
singleItemButton: id => (
  <Can
      perform="user:edit"
      yes={
        <RouterLink to={`/moped/staff/edit/${id}`}>
          <EditIcon color="primary" />
        </RouterLink>
      }
    />
  ),
  newItemButton: ewItemButton: (
    <Can
      perform="user:create"
      yes={
        <Button
          color="primary"
          variant="contained"
          component={RouterLink}
          to={"/moped/staff/new"}
          startIcon={<Icon>add_circle</Icon>}
        >
          Add Staff
        </Button>
      }
    />
  )
}
```

sergiogcx (approved):
It works flawlessly! 🙌 

There is more security in your code than there is at the capitol building 😹  jkd!

moped-editor/src/views/staff/StaffListViewQueryConf.js
mddilley: I like this a lot, Sergio. We might even need to recycle this component somewhere down the line. I'm going to update and pull that code into a separate file before merging. 🙏

moped-editor/src/auth/Can.js
mddilley: Thanks - it all goes back to [this Auth0 blog post](https://auth0.com/blog/role-based-access-control-rbac-and-react-apps/). Very helpful for this and VZE.

mateoclarke (approved):
Really great stuff! Love seeing this pattern set in a very reusable way with solid docs. You're a React rockstar! 🤩 

green_power_ranger_guitar

PR #176, opened 2021-01-06
Fixes to Stackable Filters & Search Bar

sergiogcx:
Build Preview:
https://deploy-preview-176--atd-moped-main.netlify.app/moped/session/signin

Implements:

- [x] If no filter is active, hide the “Apply Filters” button, “Reset Filters” button, and “You don’t have any search filters...” alert. Only display the “Add Filters” button.
- [x] When the “Reset filters” button is pressed, do not prompt the user to confirm—just reset the filters.
- [x] Set the default "Field" option to “Project Name”. Remove the “None” option.
- [x] Rename both of filter operators "Equals" and “Equals to” to "is".
- [x] Replace "is null" and "is not null" with "is blank" and "is not blank"
- [x] Make all of the filter operator names lower case
- [x] Set a default operator for each filter type. Text fields should default to “contains”. Numeric and date fields should default to "is". Remove the “None” operator option.
- [ ] When adding a date filter, clicking into the date input field should automatically open the calendar picker. Currently you must click on the calendar icon to open the date picker.
- [x] On initial load, a generic “search” input field should be visible, with placeholder text "Search by project name or ID...". This field should search project name and project id. The “Add Filters” should be visible, and, when clicked, the generic search should be hidden/replaced with the stackable filters component.
- [x] We found the dynamic placeholder text in the filter value input field to be a bit confusing. Suggest that we (1) remove the placeholder text and icons, (2) hide the input entirely until the user selects an operator, and (3) do not show the input if the user selects an operator that doesn't require an input ("is blank" or "is not blank")
- [x] Instead of showing the "Invalid Filters" modal, would be be possible disable the "Apply Filters" button until all filters are valid?


sergiogcx:
Working on showing the calendar automatically, approve if it is not a deal-breaker.

moped-editor/package.json
```diff
@@ -11,7 +11,7 @@
     "start:staging": "env-cmd -e staging npm run start:local",
     "start:production": "env-cmd -e production npm run start:local",
     "build": "react-scripts build",
-    "build:local": "GENERATE_SOURCEMAP=false react-scripts build && cp _redirects build && cp -r build moped && mv moped build",
+    "build:local": "grep 'kickstand' package-lock.json && GENERATE_SOURCEMAP=false react-scripts build && cp _redirects build && cp -r build moped && mv moped build",
```
mateoclarke, line 14: I think you mentioned why you added this `grep` in dev sync but can you explain it again for me? I thought using the commit hash instead of branch name turned out to be the solution?

mateoclarke (approved):
🚢

moped-editor/package.json
sergiogcx: Ah, yes this was to display the exact version of kickstand the build downloaded. It seems that Netlify ignores the package-lock.json file and overwrites it with whatever npm finds in `npm install`.

So I found out that Netlify kept downloading the wrong version using `kickstand@main` and switched to `kickstand@<commit id>`

PR #177, opened 2021-01-07
Changes First Name and Last Name field type from Name to Text

sergiogcx:
I noticed the First Name and Last Name fields in the database are of type Name, which was unknown to me, not to mention it lacks important string comparison operators we need. As it turns out "the name type exists only for the storage of identifiers in the internal system catalogs and **_is not intended for use by the general user_**".

https://www.postgresql.org/docs/current/datatype-character.html

This branch was created only for the migration, changes were cherry-picked from commit id: `3498946e118c995c8f82ff0ce19fe4e2ca5b9447`


mddilley (approved):
Good catch and thanks for including the docs about `name` type - TIL! 🙏

mateoclarke (commented):
👀

PR #180, opened 2021-01-13
Project Activity Log | Implements SQS Backend #4895

sergiogcx:
This implements the pipeline for event handling from Hasura

Mostly DevOps, no real way to test at the moment. The python handler is not implemented, but this PR focuses on providing the platform, not the functionality yet. New PR for that.


moped-data-events/activity_log/app.py
```diff
@@ -0,0 +1,72 @@
+#
+# Resolves the location for a crash.
```
mateoclarke, line 2: seems like some VZ specific copy/pasta here.

mateoclarke (approved):
🚢  LSTM 😄

moped-data-events/activity_log/app.py
sergiogcx: Will fix that @mateoclarke

PR #183, opened 2021-01-14
State management | Migrate to Apollo Client 3.0

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4758

This PR updates Apollo Client and all of the imports throughout the app. I left `apollo-boost` as a dependency since it is required by the GraphQL abstract in `atd-kickstand`. I also updated the way that the request headers and user role are appended to requests by following [this Apollo snippet](https://www.apollographql.com/docs/react/networking/authentication/#header). I tested multiple roles and checked the request headers for the JWT and role for Hasura, and it works well.

I also am unable to replicate the caching bug from https://github.com/cityofaustin/atd-data-tech/issues/4788 in this branch so I removed the notes about the bug and set `cache-first` fetch policies in both Project and Staff GridTable configs.

### Request headers from the React app
Screen Shot 2021-01-19 at 11 17 31 AM


sergiogcx (approved):
Thanks for completing the Apollo mission!

moped-editor/src/App.js
```diff
@@ -1,53 +1,54 @@
...
+        "x-hasura-role": role ? role : "",
+      },
+    };
+  });
+
+  const client = new ApolloClient({
+    // Join authLink and httpLink to handle auth in each request
+    link: authLink.concat(httpLink),
+    cache: new InMemoryCache(),
+  });
```
sergiogcx, line 48: Thanks for explaining how Apollo Links work! :+1:

mateoclarke (approved):
🚢

moped-editor/src/App.js
mateoclarke: +1

PR #186, opened 2021-01-20
Project Extent | Implement extent definition using CTN segments

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4892

This PR updates the vector tile layer used to define the project extent with the [CTN layer](https://tiles.arcgis.com/tiles/0L95CJ0VTaxqcmED/arcgis/rest/services/CTN_Project_Extent_Vector_Tiles/VectorTileServer). I also updated the way styles are defined with a map config object to make switching out or adding layers more straightforward in the future. I also consolidated some common logic between the `NewProjectMap` and `ProjectSummaryDetailsMap` components into a custom hook called `useHoverLayer`.

Found out about the [`get` syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) which I used to expose `this` in the map layer configs so that I could introspect the object outside of the `layerStyleSpec` and get the values for color and field ID.

#### Updated New Project map
Screen Shot 2021-01-22 at 2 22 59 PM

#### Selecting from the layer
Screen Shot 2021-01-22 at 2 23 17 PM

#### Project Summary view
Screen Shot 2021-01-25 at 5 11 36 PM

#### A longer feature
Screen Shot 2021-01-25 at 5 15 54 PM

### DIFF
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -24,16 +24,16 @@ export const mapConfig = {
     zoom: 12,
   },
   geocoderBbox: austinFullPurposeJurisdictionFeatureCollection.bbox,
-  layerConfigs: [
-    {
-      layerId: "ctn-lines",
+  layerConfigs: {
+    CTN: {
+      layerIdName: "ctn-lines",
+      layerIdField: "PROJECT_EXTENT_ID",
       type: "line",
-      layerSourceName: "CTN",
       layerColor: theme.palette.primary.main,
       layerUrl:
         "https://tiles.arcgis.com/tiles/0L95CJ0VTaxqcmED/arcgis/rest/services/CTN_Project_Extent_Vector_Tiles/VectorTileServer/tile/{z}/{y}/{x}.pbf",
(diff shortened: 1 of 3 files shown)


mateoclarke (approved):
🚢  Really awesome work here Mike! It's clear you're encountering and overcoming some pretty challenging technical problems. Thanks for sharing what you've been learning along the way! This CTN network works great and is really making sense after I played with the updated map a bit.

kermit_map

amenity:
Whooa — it's so beautiful, @mddilley! Playing with this prompted an almost instantaneous lifting of my stubborn attachment to polygons. 😍 

Works really smoothly and looks exquisite. I'm a bit concerned that we may not have enough contrast between the available and selected polygons, but I think it's actually a great opportunity to let users in to try it out and get feedback directly from them.

amenity (approved):
LSTM 👍✨

sergiogcx (approved):

sergiogcx:
Nothing short of brilliant! Thanks for letting me learn from you 🙏

PR #187, opened 2021-01-20
Project Phases Table

mateoclarke:
addressed part of https://github.com/cityofaustin/atd-data-tech/issues/4332

This PR add as "Project Phases" table to the Timeline Tab of the Project Details Page. The feature uses [Material Table](https://material-table.com/#/docs/features/editable) as a dependency because of the out of the box UI available for creating, editing, and deleting table items.

Note: The delete feature will not work in the staging instance  until the permissions/Hasura config settings are applied.

### Add
Screen Shot 2021-01-29 at 2 19 07 AM

### Edit
Screen Shot 2021-01-29 at 2 18 46 AM

### Delete
Screen Shot 2021-01-29 at 2 17 01 AM


amenity:
So slick, @mateoclarke! ✨ 😎

- What do you think about moving the "+" from the upper right to below the table as a [FAB](https://material.io/components/buttons-floating-action-button)? 

~I'm not able to delete any rows a~

amenity (approved):

moped-editor/src/views/projects/projectView/ProjectTimelineViewQueryConf.js
```diff
@@ -0,0 +1,89 @@
...
+      label_table: "Capital Funding",
+      width: "5%",
+      type: "boolean",
+    },
+  },
+  order_by: {},
+  where: {},
+  limit: 25,
+  offset: 0,
+};
```
sergiogcx, line 89: What is the purpose of this file? It looks like it was meant for a timeline view, but at the same time I don't see it being used or imported in any other files?

sergiogcx, line 89: What is the purpose of this file? It looks like it was meant for a timeline view, but at the same time I don't see it being used or imported in any other files? It also seems to reference the moped_project table and its attributes...

moped-database/seeds/1602292389297_initial_seed.sql
```diff
@@ -45,12 +45,9 @@ INSERT INTO public.moped_milestones (milestone_name, milestone_description, mile
...
-INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('preliminary design', NULL, NULL, NULL, NULL, 4);
 INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('scoping', NULL, NULL, NULL, NULL, 5);
 INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('planned', NULL, NULL, NULL, NULL, 6);
 INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('potential', NULL, NULL, NULL, NULL, 7);
-INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('hold', NULL, NULL, NULL, NULL, 8);
-INSERT INTO public.moped_phases (phase_name, phase_description, phase_rank, phase_average_length, required_phase, phase_id) VALUES ('canceled', NULL, NULL, NULL, NULL, 9);
```
sergiogcx, line 53: Thanks for sorting those out! 🙏

sergiogcx (changes requested):
It works flawlessly! 💯 🥇 

Just a question about the queryconf file and happy to approve once resolved

zappfyllc (approved):
@mateoclarke will echo everyone else and say this looks slick! I like this table better than the default MUI table that Leslie and I placed in the initial Project Teams page. Very cool!

Requests:
- Agree with @amenity regarding the initial placement of the add icon.
- Can you transform the UI values for "Active?" to Yes/No instead of True/False? Fine with true/false in the database but I think generally yes/no is preferable for user-facing copy.
- Default sorting: can you make this sort chronologically from newest to oldest phase instead of by phase name?

Additional Requirements (may be best handled by a separate issue, will defer to you): 
- User cannot add a phase name that's already been added.
- If a user adds an active phase, set any other active phases to inactive.
- Current phase stored in moped_projects should automatically reflect the newest phase chronologically where "active is true"

Comments:
- When I first clicked the add button to add in a phase, I stupidly forgot to click the checkmark to save and lost the row. Totally user error but I'm just flagging it in case we run into that error during user testing.
- The slight lag between entering a new row and saving it, and then the row appearing is something I'm noting. Is this because we're waiting to see if the mutation is successful (otherwise it would display an error)? Really not a big deal but just thought I'd point out if anyone else thought anything of it.

@mateoclarke we can add any of these that are more complex to a subsequent improvements issue so we can go ahead and merge: nothing major sticks out to me that would stop us from getting it in front of users. Just let me know what is left to do and I can add in the issue (or go ahead and do it and we'll discuss in grooming). Awesome job, and my deep thanks for wading through my issues and separating out the phases table so you could ship it without the milestones requirements!

johnclary:
awesome work, @mateoclarke! To raise one more detail—it would be nice if the date-picker calendar opened automatically when a date field is active. if y'all agree, we might move this into a separate issue to address it globally since we need this enhancement in Sergio's advanced table filters and will need it elsewhere.

moped-editor/src/views/projects/projectView/ProjectTimelineViewQueryConf.js
```diff
@@ -0,0 +1,89 @@
+export const ProjectTimelineViewQueryConf = {
+  options: {
+    useQuery: {
+      fetchPolicy: "no-cache", // Default ise "cache-first", or use "no-cache"
+      // Use no-cache as per issue https://github.com/cityofaustin/atd-data-tech/issues/4788
```
mddilley, line 5: The update to Apollo fixed this issue for the staff and project queries so I took out this comment in those configs. I know that no cache is the behavior that we will want in some places so just giving an update.

mddilley (approved):
Material Table works great for this feature, and it works great! Awesome to see another project summary tab filled in. 🙌 

I noticed a small regression in Hasura permissions for a couple fields that I mentioned below and need to be patched before merge. They were hidden in the formatting updates to the tables yaml.

moped-editor/src/views/projects/projectView/ProjectTimelineViewQueryConf.js
mateoclarke: hmm, you're right. I think I copy/pasted this file before changing directions on which react table library to use. Thanks for catching this.

mateoclarke:
Ok, I've addressed the code feedback from Sergio and Mike and taken inventory of the enhancements/changes requested by Amenity, John and JD. Going to start tackling them in a new quick PR and make a "Project Phases Tables Enhancements" for the ones that are more in depth.

PR #191, opened 2021-01-24
Changes to CSV Export

sergiogcx:
I had written something like this before, so I decided to do this manually for the time being and removed the "react-csv" library.

Tested to work in Firefox and Chrome:
https://deploy-preview-191--atd-moped-main.netlify.app/moped/projects

Initially I attempted to use another component within the react-csv library called CSVDownload, it' used as follows`<CSVDownload data={csvData} target="_blank" />;` This however presented a problem because it uses experimental javascript, which requires additional babel libraries and additional configuration settings I did not have the time to play with.

Alternatively, we could have hidden the CSVLink component, but that also required hiding the component with styles and advanced DOM logic and event emulation. This is still a possibility if this presented manual solution proves unstable.

Closes https://github.com/cityofaustin/atd-data-tech/issues/4785

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableExport.js
+++ b/moped-editor/src/components/GridTable/GridTableExport.js
@@ -243,7 +243,7 @@ const GridTableExport = ({ query, showFilterState }) => {
         onClose={handleDialogClose}
         aria-labelledby="form-dialog-title"
       >
-        <DialogTitle id="form-dialog-title">Downloading CSV</DialogTitle>
+        <DialogTitle id="form-dialog-title">{" "}</DialogTitle>
         <DialogContent>
           <Grid container spacing={3}>
             <Grid item xs={2} lg={2}>
@@ -251,7 +251,7 @@ const GridTableExport = ({ query, showFilterState }) => {
             </Grid>
             <Grid item xs={10} lg={10}>
               <DialogContentText>
-                Generating download file, please wait.
+                Preparing download, please wait.
               </DialogContentText>
             </Grid>
           </Grid>
--- a/moped-editor/src/components/GridTable/GridTableExport.js
+++ b/moped-editor/src/components/GridTable/GridTableExport.js
@@ -11,15 +11,11 @@ import {
   DialogTitle,
   DialogActions,
   DialogContentText,
-  Input,
-  FormControlLabel,
-  Switch,
   Grid,
   makeStyles,
   CircularProgress,
 } from "@material-ui/core";
 import { NavLink as RouterLink } from "react-router-dom";
-import { CSVLink } from "react-csv";
 import { useLazyQuery } from "@apollo/client";
 import moment from "moment";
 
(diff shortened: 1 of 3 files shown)


amenity (changes requested):
Super! Two modifications would be great: 
- [x] Hide "Downloading CSV" title in modal
- [ ] Change "Generating..." text to "Preparing download"

moped-editor/src/components/GridTable/GridTableExport.js
```diff
@@ -170,26 +178,28 @@ const GridTableExport = ({ query, showFilterState }) => {
...
-    }
-  }, [limit, selectAll, dialogOpen, query.limit, getExport]);
+  if (dialogOpen && !loading && data) {
+    const formattedData = formatExportData(data[query.table]);
+    const headers = dataGetHeaders(formattedData);
+    const csvString = dataToCSV(headers, formattedData);
+    setTimeout(() => {
+      // Update the state
+      setDialogOpen(false);
+      downloadFile(csvString);
```
mddilley, line 200: To track how many times this code fired, I just added a console.log here on line 203. I'm not sure exactly how it is happening, but maybe wrapping this top level code in a useEffect could help control when this fires?

Screen Shot 2021-01-25 at 11 22 48 AM

Screen Shot 2021-01-25 at 11 23 43 AM

mddilley (changes requested):
Awesome stuff, Sergio! Glad that you got this working without a library. Very cool. 💾 🙌  The CSV export file generates perfectly. 

I did noticed a bug where, after updating filters or the rows per page, multiple copies of the CSV download (using Chrome locally and in the deploy preview). I left some comments below with some screenshots. Let me know if I can help reproduce or dive in more.

moped-editor/src/components/GridTable/GridTableExport.js
sergiogcx: Thanks @mddilley I've refactored the code so that it lives in a useEffect function. I've checked by adding a console.log to the download function and I only see one run, let me know if this fixes it.

mddilley (approved):
💾  ✅  👏

moped-editor/src/components/GridTable/GridTableExport.js
mddilley: Awesome, I tested it out and can't replicate the multiple downloads anymore. 👍

mateoclarke (commented):
I think this way of exporting .csv data without using a 3rd party is really clever and simple. 

One thing I'm thinking about is whether a user would want to export data not displayed in the table. I think we have this feature in VZE, the export config isn't the same as what is displayed. Do we have the flexibility with this Export feature? It looks like the `dataToCSV` method expects the raw Hasura response so maybe so? If not, do you think it would be hard to extend? 

I think @amenity might have opinions on the necessity of that aspect of the export feature. Either way, seems like this is a great MVP implementation to me.

Thanks for the work on this! 💪

PR #193, opened 2021-01-27
Moped Database | Permissions for Activity Log

sergiogcx:
Creates access permissions for the activity log for all users:

image


mddilley (approved):
🔑  🙏

PR #197, opened 2021-01-29
Project Extent | Zoom to project extent in Project Summary view

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4819

This PR adds a custom hook that handles fit the project summary map to a bbox created from its existing feature collection. 

@amenity I think that this feature would be great for the edit map too, but, with the current code, it will zoom to the current features each time a new line is selected. I could refactor to add the option to only zoom on initial load but could use some feedback before investing time on that route.

### Edit and then update fit
editAndZoom

### Create a project
Screen Shot 2021-02-02 at 1 56 17 PM

### and then redirect to summary view with map zoomed to selection
Screen Shot 2021-02-02 at 1 56 22 PM

### DIFF
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -1,4 +1,5 @@
 import React, { useState } from "react";
+import bbox from "@turf/bbox";
 import theme from "../theme/index";
 import { Typography } from "@material-ui/core";
 import { isEqual } from "lodash";
@@ -89,6 +90,8 @@ export const mapConfig = {
     },
   },
 };
+
+export const createZoomBbox = featureCollection => bbox(featureCollection);
 
 /**
  * Get the IDs from the layerConfigs object to set as interactive in the map components
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -91,7 +91,7 @@ export const mapConfig = {
 };
 
 /**
- * Create a Mapbox LngLatBounds object from a bbox feature collection
+ * Create a Mapbox LngLatBounds object from a bbox generated from a feature collection
  * @param {Object} featureCollection - A GeoJSON feature collection
  * @return {Array} A nested array that fits the LngLatBounds Mapbox object format
  */
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -315,7 +315,7 @@ export function useHoverLayer() {
 }
 
 /**
- * Custom hook that returns a vector tile layer hover event handler and the details to place and populate a tooltip
+ * Custom hook that initializes a map viewport and fits it to a provided feature collection
  * @param {Object} mapRef - Ref object whose current property exposes the map instance
  * @param {Object} featureCollection - A GeoJSON feature collection to fit the map bounds around
  * @return {ViewportStateArray} Array that exposes the setter and getters for map viewport using useState hook
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -314,6 +314,17 @@ export function useHoverLayer() {
   return { handleLayerHover, featureId, hoveredCoords };
 }
 
+/**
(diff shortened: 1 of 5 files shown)


moped-editor/src/utils/mapHelpers.js
```diff
@@ -1,7 +1,8 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
+import { WebMercatorViewport } from "react-map-gl";
+import bbox from "@turf/bbox";
```
sergiogcx, line 3: Interesting little library :)

moped-editor/src/utils/mapHelpers.js
```diff
@@ -294,3 +313,54 @@ export function useHoverLayer() {
...
+        latitude,
+        zoom,
+      };
+    };
+
+    setViewport(prevViewport => fitViewportToBounds(prevViewport));
+  }, [featureCollection, mapRef]);
+
+  return [viewport, setViewport];
+}
```
sergiogcx, line 366: This whole function is brilliant! The typedef comment is kind of confusing at first, but then it makes prefect sense when you see the code.

moped-editor/src/utils/mapHelpers.js
```diff
@@ -294,3 +313,54 @@ export function useHoverLayer() {
...
+      if (featureCollection.features.length === 0) return viewport;
+
+      const featureViewport = new WebMercatorViewport({
+        viewport,
+        width: currentMap._width,
+        height: currentMap._height,
+      });
+      const newViewport = featureViewport.fitBounds(mapBounds, {
+        padding: 100,
+      });
```
sergiogcx, line 350: Coool! This is where the viewport gets its width, size. Zoom is automatically calculated.

moped-editor/src/views/projects/projectView/ProjectSummaryMap.js
```diff
@@ -44,8 +44,11 @@ const ProjectSummaryMap = ({
   const mapRef = useRef();
   const featureCount = sumFeaturesSelected(selectedLayerIds);
 
-  const [viewport, setViewport] = useState(mapConfig.mapInit);
   const { handleLayerHover, featureId, hoveredCoords } = useHoverLayer();
+  const [viewport, setViewport] = useFeatureCollectionToFitBounds(
+    mapRef,
+    projectExtentGeoJSON
+  );
```
sergiogcx, line 51: Elegant 💯 🥇

sergiogcx (approved):
It works very well! Thanks @mddilley I keep learning from you every day! 👍

moped-editor/src/utils/mapHelpers.js
mddilley: 🙏   I'm also curious if there is a better way to document a return that is meant to be destructured. Had the same problem with the other custom hook in this file so I'll be on the lookout. This came from a Stack Overflow discussion so I'm not super invested in it. Glad it works for now, though

amenity:
@amenity I think that this feature would be great for the edit map too, but, with the current code, it will zoom to the current features each time a new line is selected. I could refactor to add the option to only zoom on initial load but could use some feedback before investing time on that route.

moped-editor/src/utils/mapHelpers.js
```diff
@@ -294,3 +313,54 @@ export function useHoverLayer() {
...
+ * @param {Object} mapRef - Ref object whose current property exposes the map instance
+ * @param {Object} featureCollection - A GeoJSON feature collection to fit the map bounds around
+ * @return {ViewportStateArray} Array that exposes the setter and getters for map viewport using useState hook
+ */
+/**
+ * @typedef {Array} ViewportStateArray
+ * @property {Object} StateArray[0] - A Mapbox viewport object
+ * @property {Function} StateArray[1] - Setter for viewport state
+ */
+export function useFeatureCollectionToFitBounds(mapRef, featureCollection) {
```
mateoclarke, line 328: so reusable! ♻️  🌱

mateoclarke (approved):
enhance

moped-editor/src/utils/mapHelpers.js
johnclary: really cool stuff @mddilley!

PR #200, opened 2021-02-04
Moped Database | Add column to easily retrieve the operation type #5205

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5205
Closes https://github.com/cityofaustin/atd-data-tech/issues/5206

This PR basically creates a new column in the activity_log table, allows us to quickly retrieve the operation type: INSERT, UPDATE, DELETE. It will be a nullable varchar(8) and it will be indexed (since we are not expecting more than those 3 entries)

Also, it will be null by default, this field will be populated by the Lambda in SQS as it processes new events from Hasura


mateoclarke (approved):
🚢

PR #202, opened 2021-02-08
User Management | Implement default avatars #4798

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4798

PR URL:
https://deploy-preview-202--atd-moped-main.netlify.app/

### DIFF
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -95,11 +95,25 @@ export const UserProvider = ({ children }) => {
   return <UserContext.Provider value={values}>{children}</UserContext.Provider>;
 };
 
+export const getRandomColor = () => {
+  const key = "atd_moped_user_color";
+  if (localStorage.getItem(key) === null) {
+    const randomInt = Math.floor(Math.random() * Math.floor(3));
+    const colors = ["#43a047", "#3949ab", "#fb8c00", "#e53935"];
+    localStorage.setItem(key, colors[randomInt]);
+  }
+
+  return localStorage.getItem(key);
+};
+
 // We also create a simple custom hook to read these values from. We want our React components
 // to know as little as possible on how everything is handled, so we are not only abtracting them from
 // the fact that we are using React's context, but we also skip some imports.
 export const useUser = () => {
   const context = React.useContext(UserContext);
+  if (context && context.user) {
+    context.user.userColor = getRandomColor();
+  }
 
   if (context === undefined) {
     throw new Error(
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -1,6 +1,8 @@
 import React from "react";
 import Amplify, { Auth } from "aws-amplify";
 
+import { colors } from "@material-ui/core";
+
 // Create a context that will hold the values that we are going to expose to our components.
 // Don't worry about the `null` value. It's gonna be *instantly* overriden by the component below
 export const UserContext = React.createContext(null);
@@ -80,6 +82,9 @@ export const UserProvider = ({ children }) => {
       return data;
     });
 
+  // Remove the current color
+  destroyProfileColor();
+
   // Make sure to not force a re-render on the components that are reading these values,
   // unless the `user` value has changed. This is an optimisation that is mostly needed in cases
   // where the parent of the current component re-renders and thus the current component is forced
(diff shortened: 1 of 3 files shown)


moped-editor/src/auth/user.js
```diff
@@ -95,11 +95,25 @@ export const UserProvider = ({ children }) => {
   return <UserContext.Provider value={values}>{children}</UserContext.Provider>;
 };
 
+export const getRandomColor = () => {
+  const key = "atd_moped_user_color";
+  if (localStorage.getItem(key) === null) {
+    const randomInt = Math.floor(Math.random() * Math.floor(3));
+    const colors = ["#43a047", "#3949ab", "#fb8c00", "#e53935"];
```
mateoclarke, line 102: should we use colors based on our Material Theme? Or at least store these hex codes as color variables in a constants file for re-use?

moped-editor/src/layouts/DashboardLayout/TopBar.js
```diff
@@ -26,7 +25,21 @@ const useStyles = makeStyles(() => ({
...
+  const { user, logout } = useUser();
+
+  const emailToInitials = email => {
+    try {
+      const subdomain = "austintexas.gov";
+      if (!email.endsWith(subdomain)) return null;
+      const [first, last] = email
+        .replace("azure_ad", "")
+        .replace(subdomain, "")
+        .split(".");
```
mateoclarke, line 37: Should we also capitalize the string to match the example in the issue?

mateoclarke (changes requested):
All this looks nice. A few comments about formatting and where color hexes should live.

Also I think we missed another place where the Avatar icon lives: `/moped/account`.

Screen Shot 2021-02-08 at 3 59 10 PM

moped-editor/src/auth/user.js
sergiogcx: I took those from our material UI theme, I can re-work these as variables in a config file

moped-editor/src/layouts/DashboardLayout/TopBar.js
sergiogcx: Ah, good call! Mine was running capitalized, but I don't know why

moped-editor/src/auth/user.js
mateoclarke: Yeah, I think we should either use hexcode from or add to the `theme.palette` object or even make a new palette.js or color.js file in the `/theme` directory.

moped-editor/src/layouts/DashboardLayout/TopBar.js
amenity: Yes, capitalize plzzz! 🧑‍🏫

amenity:
@sergiogcx - This feature was such a sweet surprise! Thanks for taking a sec to globalize what you implemented in the Activity Feed. 😊  🙏

amenity (approved):
Looks great! Mateo covered my notes — capitalization and the _Account_ page instance.

moped-editor/src/layouts/DashboardLayout/TopBar.js
```diff
@@ -26,7 +25,21 @@ const useStyles = makeStyles(() => ({
...
+      return first.charAt(0) + last.charAt(0);
+    } catch {
```
johnclary, line 39: cool to see try/catch here. i don't consider it enough w/ javascript

moped-editor/src/layouts/DashboardLayout/TopBar.js
johnclary: @sergiogcx I take it that a user with a non-austintexas.gov email address will have a colored circle with no initials? that's cool w/ me.

sergiogcx: @johnclary Ah, sorry for the late reply. A user without initials would see a generic humanoid figure that is default fallback to Material UI:
https://material-ui.com/components/avatars/#fallbacks

amenity:
LSTM 👌🏻 ✨

amenity:
Not sure if it's because I'm testing with my non-COACD credentials, but the avatar on the account page is not showing my initials. 
Screen Shot 2021-02-10 at 12 08 43 PM

sergiogcx:
It's not supposed to, it always defaults to null. I can change it real quick to show initials too, though the font size will be large.

mateoclarke:
@sergiogcx I agree with Amenity. I think consistency with the avatar would be better.

sergiogcx:
@amenity @mateoclarke there is a new PR:
https://github.com/cityofaustin/atd-moped/pull/215

PR #204, opened 2021-02-08
Add .prettierignore to database subfolder

mddilley:
This PR adds a `.prettierignore` file to the DB subfolder to prevent auto-formatting that creates conflicts with changes that Hasura generates.

I tested in `/moped-database/metadata/tables.yaml` with my VSCode setup and was able to verify that format-on-save did not create any changes after this update, but this could use some testing on other machines.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+# Ignore DB code
+moped-database/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,0 @@
-# Ignore DB code
-moped-database/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+# Ignore DB config code
+moped-database/*
--- a/moped-database/.prettierignore
+++ b/moped-database/.prettierignore
@@ -1,0 +1,2 @@
+# Ignore DB code
+*


mateoclarke (approved):
After poking around, I think I figured out that in order for the .prettierignore to work with the VSCode (or whatever text editor) plugin, the ignore config needs to live at the root level. So, the update you made Mike, would only work when I opened my editor from the ./moped-database subfolder. This is great if your standard workflow preference is to open each project subfolder in separate editor windows. 

### working at the subfolder level
Screen Shot 2021-02-09 at 12 16 06 AM

### not working at the full project level (note where it expects the ignore config file)
Screen Shot 2021-02-09 at 12 15 17 AM

The fix, why not have it in both locations? I've added another .prettierignore file that will work when you have an editor window open at the project's true root with all the subfolders visible. 

Let me know if y'all think that's gross or if there should be other updates or considerations.


why_not_both

mddilley:
@mateoclarke Thanks for updating! I don't think it is gross at all if it keeps the repo tidy. Glad that you were able to get it to work in the root of the project! I'll do a quick test to see if we still need the one in `/moped-database` or if the root level is all encompassing before I merge.

PR #207, opened 2021-02-09
Project Phases enhancements

mateoclarke:
https://github.com/cityofaustin/atd-data-tech/issues/5210

### - Updates True/False label text to Yes/No with “switch” toggle component in edit mode.

**read-mode, Yes/No labels**
Screen Shot 2021-02-09 at 12 59 07 AM

**edit-mode, Material UI "Switch" component**
Screen Shot 2021-02-09 at 12 59 12 AM

### - Default sorting, sort chronologically from newest to oldest phase

_@zappfyllc I think we need to discuss this. I interpreted this to mean sort by phase_start desc. Let me know if you had another column in mind when you where thinking of a chronological sorting. This implementation seems to conflict with your other requirement:_
> Current phase stored in moped_projects should automatically reflect the newest phase chronologically where "active is true"

Screen Shot 2021-02-09 at 1 15 35 AM

### - Move add phase button from top-right default to underneath other existing items.
_Following example here: https://github.com/mbrn/material-table/issues/2133_

Screen Shot 2021-02-09 at 6 58 52 PM


zappfyllc (changes requested):
@mateoclarke Looking great! Quick comments:

- You interpreted correctly. I think sorting by newest phase, aka descending from phase_start is definitely a good default behavior. If we need to change it in the future, we can.
- I love the Material UI Switch: let's remove the "Yes/No" labels on the sides of the switch component and let the switch should speak for itself. If users are confused, we'll figure it out via testing but I think this would look the best.

Looks great, Mateo!

moped-editor/src/views/projects/projectView/ProjectTimeline.js
```diff
@@ -54,6 +79,12 @@ const ProjectTimeline = () => {
    * */
   const { projectId } = useParams();
 
+  /** addAction Ref - mutable ref object used to access add action button
+   * imperatively.
+   * @type {object} addActionRef
+   * */
+  const addActionRef = React.useRef();
+
```
sergiogcx, line 87: I love seeing these! 👍

moped-editor/src/views/projects/projectView/ProjectTimeline.js
```diff
@@ -129,6 +163,26 @@ const ProjectTimeline = () => {
...
+                      typeof props.action === typeof Function ||
+                      props.action.tooltip !== "Add"
+                    ) {
+                      return <MTableAction {...props} />;
+                    } else {
+                      return (
+                        <div
+                          ref={addActionRef}
+                          onClick={props.action.onClick}
+                        />
```
sergiogcx, line 181: Neat!

sergiogcx (approved):

amenity (approved):
Sorry - spastic button clicking earlier. 🤦🏻  This looks super, @mateoclarke!

amenity:
Will applying that 'color="primary"` prop take care of #5424?

mateoclarke:
> Will adding [that `color="primary"` prop](https://github.com/cityofaustin/atd-moped/pull/207/commits/7b8dfbfee902b66bd5d344a36f56d3b81b080db4) take care of #5424?

Yeah I think that would do it!

PR #214, opened 2021-02-10
Fixes the activity log label for project extend and geojson

sergiogcx:
Preview:
https://deploy-preview-214--atd-moped-main.netlify.app/

The problem:

The fix:

To test, run locally with `$ npm run start:staging`

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectActivityLog.js
+++ b/moped-editor/src/views/projects/projectView/ProjectActivityLog.js
@@ -6,6 +6,7 @@ import {
   getChangeIcon,
   getRecordTypeLabel,
   getHumanReadableField,
+  ProjectActivityLogGenericDescriptions,
 } from "./ProjectActivityLogTableMaps";
 
 import ProjectActivityLogDialog from "./ProjectActivityLogDialog";
@@ -29,7 +30,7 @@ import {
 
 import { PROJECT_ACTIVITY_LOG } from "../../../queries/project";
 import makeStyles from "@material-ui/core/styles/makeStyles";
-import {Alert} from "@material-ui/lab";
+import { Alert } from "@material-ui/lab";
 
 const useStyles = makeStyles(theme => ({
   table: {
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -128,11 +129,19 @@ const ProjectActivityLog = () => {
     return data?.moped_activity_log?.length ?? 0;
   };
 
+  /**
+   * Returns True if the field should be a generic type (i.e., maps, objects)
+   * @param {string} field - The field name (column name)
+   * @return {boolean} - True if the field is contained in the ProjectActivityLogGenericDescriptions object
+   */
+  const isFieldGeneric = field =>
+    field in ProjectActivityLogGenericDescriptions;
```
mddilley, line 138: I need to remember to use the `in` keyword more. 👍

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -1305,6 +1305,15 @@ export const ProjectActivityLogOperationMaps = {
   },
 };
 
+export const ProjectActivityLogGenericDescriptions = {
+  project_extent_ids: {
+    label: "Project extent updated",
+  },
+  project_extent_geojson: {
+    label: "Project geo-json updated",
```
mddilley, line 1313: Maybe this should be GeoJSON instead of geo-json?

mddilley (approved):
Looks great, Sergio! I like this way of controlling what outputs in the activity log. A nice enhancement to an already amazing feature! 😎

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
sergiogcx: I found out it works out of luck, that's the syntax in python, it totally works in JavaScript too! :)

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
sergiogcx: Good call, will change now

amenity (approved):
LSTM!! ✨

PR #215, opened 2021-02-10
Make use of user initials in the account page

sergiogcx:
This is what it should look like now:

https://deploy-preview-215--atd-moped-main.netlify.app/


amenity (approved):
Looks great!

mateoclarke:
@sergiogcx, Looks good for the initials/avatar fix. I'm wondering if we should've replaced the user's first/last name with their email. Was that a request from Product Team? Shouldn't it be possible to make the association between the cognito user and the moped_user table

mateoclarke (commented):
🚢

sergiogcx:
There are three ways to do that, one is to extend the token a little bit, another is to retrieve data via API, another is to retrieve data via GraphQL. Happy to discuss,

PR #218, opened 2021-02-12
Project Staffing | Replace New Project team table

mddilley:
Closes part of https://github.com/cityofaustin/atd-data-tech/issues/4913
Closes https://github.com/cityofaustin/atd-data-tech/issues/5263

This PR removes the previous team table in the New Project workflow and replaces it with the same table used in the Project Summary team tab. It also refactors the mutation to add all team members in a new project as a batch instead of making individual API calls for each team member.

Also, in order to align the text inputs in each row of the table, I removed the helper text ("Select Staff" and "Select Role") below the Name and Role columns.

#### New Project Team Step
Screen Shot 2021-02-22 at 12 25 48 PM

#### Project Summary after creation
Screen Shot 2021-02-22 at 12 26 15 PM

#### Add/edit row with helper text removed
Screen Shot 2021-02-22 at 12 40 21 PM


moped-editor/src/utils/materialTableHelpers.js
```diff
@@ -0,0 +1,16 @@
...
+export const filterObjectByKeys = (obj, keys) =>
+  Object.keys(obj)
+    .filter(key => !keys.includes(key))
+    .reduce(
+      (acc, key) => ({
+        ...acc,
+        [key]: obj[key],
+      }),
+      {}
+    );
```
sergiogcx, line 16: Classic mike! Filter-Reduce :)

moped-editor/src/views/projects/newProjectView/NewProjectView.js
```diff
@@ -79,21 +82,13 @@ const NewProjectView = () => {
     project_priority: "",
     project_description: "",
     project_name: "",
-    start_date: "2021-01-01",
+    start_date: moment().format("YYYY-MM-DD"),
```
sergiogcx, line 85: Isn't moment deprecated? No problem with it, I am familiar and I like it, just wondering if we want to continue working with it. :)

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -153,44 +136,95 @@ const ProjectTeamTable = ({
...
-                  status_id: 1,
-                };
-
-                addProjectPersonnel({
-                  variables: {
-                    objects: [personnelData],
-                  },
-                });
-              }
+              isNewProjectActions[isNewProject].add(newData);
```
sergiogcx, line 227: This is classy! I like how you used true and false as keys, never thought you could do that :) 👍

sergiogcx (approved):
It works as advertised!

moped-editor/src/views/projects/newProjectView/NewProjectView.js
mddilley: Yeah, it definitely is deprecated. I'm all about `date-fns`, `luxon`, or something else so we can learn a new library! There are only a few uses of `moment` in this repo so I don't think that would be hard at all. 👍

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
mddilley: Same here, it was a nice surprise that it worked out this way. 😄

amenity (approved):
LSTM!! 👌✨

mateoclarke:
Looks good Mike. One comment that might be a backlog issue for later may be to use a custom add button below the table for consistency based on feedback from the Phases PR.

Check out the 3rd point in this `WIP` PR.
https://github.com/cityofaustin/atd-moped/pull/207

amenity:
Thanks, @mateoclarke. 🙏 Hopefully we can reuse these Material Table customizations?

PR #219, opened 2021-02-19
[Enhancement] Return user to log in page upon session termination #4739

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4739

This PR will do one of two things:
1. Instead of showing the JWT Token Expired message, it shows a spinner and makes the browser reload the current page, forcing the AWS Amplify to refresh the token.
2. If AWS Amplify cannot refresh the token, then it will sign the user off.

Unfortunately, the only way to test this PR is by logging into the PR build, wait 1 hour and see the token expire.
https://deploy-preview-219--atd-moped-main.netlify.app/moped


moped-editor/src/auth/user.js
```diff
@@ -149,20 +169,27 @@ export const useUser = () => {
...
+  try {
+    return (
+      user.idToken.getExpiration() - Math.round(new Date().getTime() / 1000)
+    );
+  } catch {
+    return 0;
+  }
+};
+
+export const isSessionExpired = user => availableSessionTime(user) > 0;
```
mateoclarke, line 187: This code looks really similar to the logic you updated in `_isAuthenticated` and I don't see this new `isSessionExpired` boolean being used anywhere. Totally cool if we just leave `availableSessionTime` and `isSessionExpired` in here in this User Hook for future use but also don't want the logic in this implementation to stray from the primary logic being used in `_isAuthenticated` so maybe its better to delete the unused code to prevent future code cruft? I may be misinterpreting the code and up to you on if you want to address this comment.

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -0,0 +1,51 @@
...
+      ) : error ? (
+        <p>Error: {error.message}</p>
+      ) : (
+        props.children
+      )}
+    </>
+  );
+};
+
+export default ApolloErrorHandler;
```
mateoclarke, line 51: this seems like a handy little component for future use!

mateoclarke (approved):
🚢 💨  This looks good to me Sergio. There was quite a bit of code that looked like it had been autoformatted but the auth specific parts look really good. I really enjoyed seeing more usage of the `?.` optional chaining!

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * ApolloErrorHandler attempts to be a component that deals with the
+ * JWT Token Expired error in a sensible way. Initially, it only displays
+ * a spinner and forces the browser to refresh, allowing AWS Amplify
+ * to refresh the token.
+ * @param {Object} props - Props.error is the only prop needed at the moment
+ * @return {JSX.Element}
+ * @constructor
+ */
+const ApolloErrorHandler = props => {
```
mddilley, line 22: This is a super cool component!

mddilley (approved):
This is very cool, Sergio! There is a small concern about needing to wrap any component that fetches data, but this all makes perfect sense to me. 👏  🙌 

Maybe there is a way to handle the Apollo errors further up in the tree in the future with something like [Apollo Link](https://www.apollographql.com/docs/link/links/error/). Not exactly sure what that would look like but your ApolloErrorHandler component would definitely be reusable for that.

moped-editor/src/auth/user.js
sergiogcx: Thanks @mateoclarke I had a little debate in my mind and I left it there in case we need to revisit in the future, but now that you mention going stray from `_isAuthenticated` I've changed my mind. I am going to remove this helper method, and fix the new conflicts.

sergiogcx:
> This is very cool, Sergio! There is a small concern about needing to wrap any component that fetches data, but this all makes perfect sense to me. 👏 🙌
> 
> Maybe there is a way to handle the Apollo errors further up in the tree in the future with something like [Apollo Link](https://www.apollographql.com/docs/link/links/error/). Not exactly sure what that would look like but your ApolloErrorHandler component would definitely be reusable for that.

Thanks for sharing @mddilley I was was feeling guilty as I was wrapping the components under ApolloErrorHandler but I could not think of anything faster or more immediate that guarantees results. 

Also, I did not know Apollo Link existed, that would be far more efficient and I am excited to learn about it. Though I do not understand how it works, I think if we can fully grasp how to use it, I would be far more happy than the current solution. Maybe it could be some kind of universal modal that shows the error message? The sky is the limit with this. 👍

sergiogcx:
@mddilley by the way, I've created a new issue to look into Apollo Link:
https://app.zenhub.com/workspaces/data--technology-services-5caf7dc6ecad11531cc418ef/issues/cityofaustin/atd-data-tech/5305

PR #222, opened 2021-02-24
Data Migration | Add comments to the table columns

zappfyllc:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5019

Combing through the schema to add comments to the table columns in Moped; will also identify columns for removal in the comment itself (can be handled in subsequent issues as some of these may have an impact on the seed data and/or other Hasura metadata we're not considering)

### DIFF
--- a/moped-database/migrations/1617934601367_alter_table_public_moped_categories_update_comment/up.sql
+++ b/moped-database/migrations/1617934601367_alter_table_public_moped_categories_update_comment/up.sql
@@ -1,0 +1,8 @@
+COMMENT ON TABLE "public"."moped_categories" IS E'DELETE unless needed; categories are project types in IMPD';
+COMMENT ON TABLE "public"."moped_city_fiscal_years" IS E'DELETE';
+alter table "public"."moped_components" rename to "moped_component_types";
+COMMENT ON TABLE "public"."moped_component_types" IS E'Project facility types or component types; associated with moped_components';
+COMMENT ON TABLE "public"."moped_entity" IS E'DELETE';
+COMMENT ON TABLE "public"."moped_fund_opp" IS E'DELETE';
+COMMENT ON TABLE "public"."moped_group" IS E'DELETE';
+COMMENT ON TABLE "public"."moped_proj_categories" IS E'DELETE unless needed; categories are project types in IMPD';
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -2628,16 +2628,16 @@
       - date_added
       filter: {}
       check: null
-  event_triggers:	
-  - name: activity_log_moped_proj_personnel	
-    definition:	
-      enable_manual: false	
-      insert:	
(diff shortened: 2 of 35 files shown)


zappfyllc:
@sergiogcx I have everything running locally and your Hasura/Migrations documentation was 💯. With the project table comments alone, I generated about 15 migrations. Would you prefer I connect to the RDS instance via a bastion host to  insert table comments and avoid all of these migrations from the comments? Just let me know what you think is best and I can generate the key and provide it to you.

I also feel comfortable condensing the migrations myself into one per set of table comments.

sergiogcx:
I'm happy to hear you were able to run the cluster @zappfyllc, and yea that's a great idea to put several migrations into one. I've done that before because I think it is best practice in case you need to roll back. That's entirely up to you how you would like to go about it :)

zappfyllc:
@sergiogcx I didn't remove the tables on the local instance just to be safe and not break anything, but I did comment what should be deleted, which should drastically simplify the database as a whole. There are a few tables that have been left in with comments that will likely be addressed in coming features or additional project data that will make it into Moped, but this should provide a better view of what is actually necessary in Moped at this time. Will follow up tomorrow with further clarification on list of IMPD tables that are relevant to migration

sergiogcx:
> @sergiogcx I didn't remove the tables in this PR just to be safe and not break anything, but I did comment what should be deleted, which should drastically simplify the database as a whole. There are a few tables that have been left in with comments that will likely be addressed in coming features or additional project data that will make it into Moped, but this should provide a better view of what is actually necessary in Moped at this time. Will follow up tomorrow with further clarification on list of IMPD tables that are relevant to migration
> 
> Since we're removing many of the tables that have yet to be implemented, I didn't find it necessary to comment on every table column outside of features that are already implemented.

Thanks @zappfyllc

moped-database/migrations/1617934601367_alter_table_public_moped_categories_update_comment/up.sql
```diff
@@ -0,0 +1,8 @@
+COMMENT ON TABLE "public"."moped_categories" IS E'DELETE unless needed; categories are project types in IMPD';
+COMMENT ON TABLE "public"."moped_city_fiscal_years" IS E'DELETE';
```
sergiogcx, line 2: As a note to self: It looks like we need this table for the time being.

https://github.com/cityofaustin/atd-moped/blob/5400a1b79d4539340a234f0a7bf720ec2322bc37/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js#L36

amenity (approved):

amenity (approved):

amenity:
@sergiogcx - Can we merge this and close https://github.com/cityofaustin/atd-data-tech/issues/5019?

sergiogcx:
@amenity the database has changed a bit since this PR was created, I can close this PR and create a new one to make the updates.

PR #223, opened 2021-02-25
JWT Token: Clean-up

sergiogcx:
These are two small changes:
- Removes unused functions
- Allows more time (5 seconds) for AWS Amplify to refresh token


mddilley (approved):
👍

mateoclarke (commented):
👍

PR #226, opened 2021-02-25
User Management | Update User query

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4994

This PR updates the user query to fetch user roles and then populate in the Edit Staff form. I addressed one console warning coming from a couple of MUI components.

### Roles fetched for Test Viewer and populated in Staff Edit Form
Screen Shot 2021-02-26 at 12 58 30 PM

### Handling null roles (so we can backfill roles for existing users through the UI)
Screen Shot 2021-02-26 at 1 50 20 PM


sergiogcx (approved):
Thanks for working on this!

amenity (approved):
LSTM! ✨

PR #227, opened 2021-02-26
Moped API | File Uploads to S3

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5298

Basically, this branch extends the moped API to allow for uploads and downloads from S3. This allows us to use a react component to upload files directly from the UI into S3.

- Route `/files/download/<file-path>`:  This route will attempt to find the file in the pre-configured bucket name, and if it is found it will return a signed url that makes the file temporarily public (up to 60 seconds).
- Route `/files/request-signature?file=<file-path>`: This route will attempt to request permission to upload a file to s3. If the request is successful, the user is allowed to upload a file to that specific file path only.

This implements the file upload API, unfortunately this cannot be tested without Postman and AWS Creds, but here are the screenshots:

### DIFF
--- a/moped-api/app.py
+++ b/moped-api/app.py
@@ -12,6 +12,7 @@ MOPED_API_CURRENT_ENVIRONMENT = os.getenv("MOPED_API_CURRENT_ENVIRONMENT", "STAG
 from auth.auth import auth_blueprint
 from users.users import users_blueprint
 from events.events import events_blueprint
+from files.files import files_blueprint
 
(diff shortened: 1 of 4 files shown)


mddilley (approved):
Looks great to me! Glad to see some of the existing helpers making a nice new blueprint come to life. I ran the tests locally, and they all passed. As usual, your tests made visualizing the purpose of the each helper much easier. 👏

PR #229, opened 2021-03-03
Project Staffing | Improve fields displayed in activity log from personnel table

mddilley:
Closes part of https://github.com/cityofaustin/atd-data-tech/issues/5285

This PR updates the activity log config and field handling to accept details about lookup tables needed to translate foreign keys into their string values. `fieldValues` accepts an array so that field can be concatenated if needed like the full name of a user. I will make this update to the Activity Log Details view in a future PR.

Example project with updates:
https://deploy-preview-229--atd-moped-main.netlify.app/moped/projects/150?tab=activity_log

Example update to a field config in `ProjectActivityLogTableMaps`:
```js
role_id: {
        icon: "",
        label: "Role",
        type: "int4",
        lookup: {
          table: "moped_project_roles",
          fieldLabel: "project_role_id",
          fieldValues: ["project_role_name"],
        },
      },
```

### Project Personnel User and Role updates
Screen Shot 2021-03-08 at 12 25 59 PM


moped-editor/src/queries/project.js
```diff
@@ -204,6 +204,12 @@ export const PROJECT_ACTIVITY_LOG = gql`
         user_id
       }
     }
+    activity_log_lookup_tables: moped_activity_log(
+      where: { record_project_id: { _eq: $projectId } }
+      distinct_on: record_type
+    ) {
+      record_type
+    }
```
sergiogcx, line 212: This is very cool! I did not know distinct_on was available in Hasura

moped-editor/src/utils/activityLogHelpers.js
```diff
@@ -0,0 +1,161 @@
...
+import { useQuery, gql } from "@apollo/client";
+import { ProjectActivityLogTableMaps } from "../views/projects/projectView/ProjectActivityLogTableMaps";
+
+export const PLACEHOLDER_QUERY = gql`
+  query NotARealQuery {
+    just_a_placeholder {
+      affected_rows
+    }
+  }
+`;
```
sergiogcx, line 11: It does not really matter, but I though this may be easier on the eyes:
https://github.com/cityofaustin/atd-moped/blob/66845ca0d116da02b9a0a93b8a44ec8b45164501/moped-editor/src/components/DataTable/DataTable.js#L63

sergiogcx (approved):
Thank you for working on this @mddilley ! 🙌 

Maybe not in the scope of this issue, but the difference page still shows the numbers, not sure if that's important to you @amenity. Maybe a new issue can be created for that purpose (if at all).

moped-editor/src/queries/project.js
mddilley: Same! I just happened across it when using GraphiQL in the Hasura console.

moped-editor/src/utils/activityLogHelpers.js
mddilley: That definitely looks a lot nicer! I'll update with that placeholder. Thanks for pointing out that existing query.

mddilley:
Thanks for the review @sergiogcx 🙏  That is totally right. Extending this to the detail view that shows the differences is the next step to fully close this issue.

mateoclarke (approved):
👏 👏 👏

amenity (approved):
Looks good! 

I was a bit confused about how to test — thought that editing the "Team" page would populate activity. Mike explained that that trigger hasn't been implemented yet and that what I see in the example activity feed is staff changes during project creation. 👍

PR #230, opened 2021-03-04
Moped Editor | Project File Uploads

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5299
Closes https://github.com/cityofaustin/atd-data-tech/issues/5301

### To test, you will need to run the stack locally.
1. Run the Hasura cluster
2. Run the Moped API

### Running the Moped API
1. You will need to have your AWS variables exported, otherwise the API will fail to load.
2. Export one more environment variable: `export MOPED_API_UPLOADS_S3_BUCKET="atd-moped-editor-staging"`
3. Run: `FLASK_ENV=development flask run --without-threads`


moped-editor/src/utils/downloadFileAttachment.js
```diff
@@ -0,0 +1,45 @@
...
+            .catch(err => {
+              console.log("Error Downloading File: " + JSON.stringify(err));
+            });
+        } else {
+          // Error
+          console.log("Error: Unable to download file.");
+        }
+      })
+      .catch(err => {
+        debugger;
```
mddilley, line 41: A debugger left in here

mddilley (approved):
Works just as expected! I was able to get everything running locally thanks to your notes in the PR and test upload and download. 🙏  

One thing that I noticed when choosing a file to upload is that the file uploads to the S3 bucket regardless of whether I click "Save" or not in the modal. I think that the user experience of being able to see the file progress is really cool, but I could see users selecting the wrong file here and there. Just thinking out loud, and maybe this is something that is already on your radar or not a huge concern as far as storage goes.

Either way, 🎉 !
Screen Shot 2021-03-12 at 11 31 47 AM

moped-editor/src/utils/downloadFileAttachment.js
sergiogcx: Aah, darn it. Let me get rid of it 🙏

sergiogcx: Removed! 👍

amenity:
Y'all I don't have time to run locally, but looks wonderful! I have a few design tweaks that I'll stick in a new issue.

amenity (approved):

moped-editor/src/utils/humanReadableFileSize.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param bytes Number of bytes.
+ * @param si True to use metric (SI) units, aka powers of 1000. False to use
+ *           binary (IEC), aka powers of 1024.
+ * @param dp Number of decimal places to display.
+ *
+ * @return Formatted string.
+ */
+export default (bytes, si=false, dp=1) => {
+    const thresh = si ? 1000 : 1024;
```
mateoclarke, line 12: math

mateoclarke (commented):
Really good stuff in here and glad to see it already working in staging!

PR #232, opened 2021-03-05
Project Extent | Implement refinements to "Map project" step in new project workflow and "Project summary" page

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5376

This PR makes the following updates: 

1. Require at least one line selected (used this MUI pattern https://material-ui.com/components/steppers/#non-linear-error-step)
2. Sentence case in "Map project"
3. Remove "Polygon ID" hover
4. Set map extent to match to project extent when edit panel opens

### New Project stepper warns and disables submission if no features selected in map
validateLocation

### Map fits to feature collection on edit
zoomOnEdit

### Map fits to feature collection in new project stepper
zoomInStepper

### Sentence case
sentenceCaseMap

### DIFF
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -8,6 +8,54 @@ export const PROJECT_NAME = gql`
   }
 `;
 
+export const ADD_PROJECT = gql`
+  mutation MyMutation(
+    $project_name: String! = ""
+    $project_description: String! = ""
+    $current_phase: String! = ""
+    $current_status: String! = ""
+    $eCapris_id: String! = ""
+    $fiscal_year: String! = ""
+    $start_date: date = ""
+    $capitally_funded: Boolean! = false
+    $project_priority: String! = ""
+    $project_extent_ids: jsonb = {}
+    $project_extent_geojson: jsonb = {}
+  ) {
+    insert_moped_project(
+      objects: {
+        project_name: $project_name
+        project_description: $project_description
+        current_phase: $current_phase
+        current_status: $current_status
+        eCapris_id: $eCapris_id
+        fiscal_year: $fiscal_year
+        start_date: $start_date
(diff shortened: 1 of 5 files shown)


amenity (commented):
✅ New Project stepper warns and disables submission if no features selected in map
✅ Map fits to feature collection on edit
✅ Map fits to feature collection in new project stepper
✅ Sentence case


The [error-step](https://github.com/cityofaustin/atd-moped/pull/232/files#submit-review) looks really slick. 😎 I didn't see it at first and expect that users might miss it as well, so we might want to implement an additional warning closer to the "Finish" button. 

I am getting a bug where when I delete all the segments from a project, the map view on the Project Summary page does not update at the same time:


https://user-images.githubusercontent.com/1463708/110227059-f5092980-7eb9-11eb-9f3b-af26459fa34e.mov

mateoclarke (approved):
The code looks really good. I was able to reproduce the bug emoji caught. It looks like the state of the `featureCollection `objects in the view-only and edit-mode version gets out of sync when there is an empty `featureCollection` saved. I followed Amenity's steps and then clicked the Edit button and when you get back in the edit mode, the lines I had deleted were not selected.

All this said, this seems like it could be hairy to resolve and seems like an edge case in practice. I'd be ok if we backlog this bug and merge this PR as is. If @amenity is too.

amenity:
Yup! I’m good to merge and backlog the bug.

mddilley:
@amenity @mateoclarke Thanks for review and catching/reproducing that bug! 🙏  I'll take a quick look at it but will create an issue if it is will take too much time. I'll merge this today either way.

mddilley:
@amenity @mateoclarke Thanks for review and catching/reproducing that bug! 🙏  I'll take a quick look at it but will create an issue if it is will take too much time. I'll merge this today either way.

PR #235, opened 2021-03-08
Moped Database | Problem with Cognito Entries in DynamoDB #5412

sergiogcx:
Fixes a the database_id and workgroup_id bug in the API


mddilley (approved):
Thanks for tracking down this tough bug and adding some extra error handling in the fix! 👍

mateoclarke (approved):
🚢

PR #238, opened 2021-03-10
[Bug] Activity feed does not display all project changes as intended #5456

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5456

Live Demo:
https://deploy-preview-238--atd-moped-main.netlify.app/moped/projects/180?tab=activity_log

It looks like this:

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectActivityLog.js
+++ b/moped-editor/src/views/projects/projectView/ProjectActivityLog.js
@@ -8,6 +8,8 @@ import {
   getChangeIcon,
   getRecordTypeLabel,
   getHumanReadableField,
+  getMappedValue,
+  isFieldMapped,
   ProjectActivityLogGenericDescriptions,
 } from "./ProjectActivityLogTableMaps";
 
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -1265,17 +1265,64 @@ export const ProjectActivityLogTableMaps = {
...
+      icon: "close",
+    },
+    INSERT: {
+      label: "Created",
+      icon: "beenhere",
+    },
+    UPDATE: {
+      label: "Update",
+      icon: "create",
+    },
```
mddilley, line 1280: So cool to see this code extended to customize the UI based on the table. 🙌

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -175,8 +189,9 @@ const ProjectActivityLog = () => {
                 </TableRow>
               </TableHead>
               <TableBody>
-                {data["moped_activity_log"].map(change => (
+                {reorderEventList(data["moped_activity_log"]).map(change => (
                   <TableRow key={change.activity_id}>
+                    {console.log("Change: ", change)}
```
mddilley, line 194: This log might serve an important purpose but wanted to point it out in case it was left behind unintentionally.

mddilley (approved):
This looks so nice, @sergio! I tested by adding myself to the example in your PR notes, and it worked great. 😎

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
sergiogcx: Ah, thanks for catching this! 🙏 I will have it removed

amenity:
Everything that's in the log looks really slick, but once I make a change it goes back to robot-talk:
Screen Shot 2021-03-15 at 12 19 45 AM
😕

sergiogcx:
> Everything that's in the log looks really slick, but once I make a change it goes back to robot-talk:
> Screen Shot 2021-03-15 at 12 19 45 AM
> 😕

@amenity Ah, thanks for showing me this. I need to check what I am missing, what record is that in the picture?

amenity:
@sergiogcx - sorry, I totally missed your question. Sorry not to include that. Did you find an example or do you want me to reproduce? Or this can be a separate issue.

sergiogcx:
@amenity Got it fixed, it was kind of a one-liner thing. Since there is no reference table I had to create one in-memory in the configuration. We should expect more situations like these, but it's sort of difficult to know where these are, will have to review every table and find out which ones will get edited, and how these will output on screen.

PR #239, opened 2021-03-10
add form validation and disable default enter key behavior

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5452

- Adds form validation so a user can't submit a new team member without a username or role
- Disables default MaterialTable enter key behavior so hitting enter on a single field doesn't submit the whole form (this had to be done because MaterialTable's form validation is overridden when a user hits enter)

There's still a minor UX bump in that without being able to submit a form with the enter key, the user has to click the "save" button. Still better than a nasty error! Solution would be to reenable default enter key behavior when user_id and role_id are populated. Haven't had a chance to figure that out yet but wanted to get this submitted in time for the release! Happy to take a crack at it if there's more time.

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectTeamTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectTeamTable.js
@@ -4,7 +4,7 @@ import { useQuery, useMutation } from "@apollo/client";
 // Material
 import { CircularProgress, TextField, Typography } from "@material-ui/core";
 import { Clear as ClearIcon } from "@material-ui/icons";
-import MaterialTable from "material-table";
+import MaterialTable, { MTableEditRow } from "material-table";
 import Autocomplete from "@material-ui/lab/Autocomplete";
 import { filterObjectByKeys } from "../../../utils/materialTableHelpers";
 import typography from "../../../theme/typography";
@@ -93,6 +93,7 @@ const ProjectTeamTable = ({
       title: "Name",
       field: "user_id",
       render: personnel => getPersonnelName(personnel.user_id),
+      validate: rowData => !!rowData.user_id,
       editComponent: props => (
         <Autocomplete
           id="user_id"
@@ -116,6 +117,7 @@ const ProjectTeamTable = ({
       title: "Role",
       field: "role_id",
       render: personnel => roles[personnel.role_id],
+      validate: rowData => !!rowData.role_id,
       editComponent: props => (
         <Autocomplete
           id="role_id"
@@ -216,6 +218,13 @@ const ProjectTeamTable = ({
     <ApolloErrorHandler errors={error}>
       <MaterialTable
         columns={columns}
+        components={{
+          EditRow: (props, rowData) => <MTableEditRow {...props} onKeyDown={(e) => {
+              if (e.keyCode === 13) {
+                // Bypass default MaterialTable behavior of submitting the entire form when a user hits enter
+              }
+          }} />
+        }}
         data={isNewProject ? personnelState : personnel}
         title="Project team"
         options={{


moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -216,6 +218,13 @@ const ProjectTeamTable = ({
     <ApolloErrorHandler errors={error}>
       <MaterialTable
         columns={columns}
+        components={{
+          EditRow: (props, rowData) => <MTableEditRow {...props} onKeyDown={(e) => {
+              if (e.keyCode === 13) {
+                // Bypass default MaterialTable behavior of submitting the entire form when a user hits enter
```
mateoclarke, line 224: I'm kinda confused how this is working since no code executes in the block for keycode 13. But it must be doing something because the deploy instance looks good. If there is any links to Github Issues or Docs that would help someone understand this `EditRow` API and what default behavior it overrides, I think that could help.

mateoclarke (approved):
🚢  

I have a question about how this magic is working, but looks good to ship!

amenity (approved):
Geez! This is _so_ much nicer than I was imagining! Really really smooth. You should submit this back to Material Table, @tillyw! 😎

amenity:
> There's still a minor UX bump in that without being able to submit a form with the enter key, the user has to click the "save" button. 

I think this explains my instinct to move the checkbox/x to the far righthand side.... I imagine is a considerable refactoring, though, so we'll wait for users to validate that this is actually a hurdle for people-not-me.

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
tillyw: From what I understand, this just clears the default behavior so that nothing occurs when a user presses enter other than that a value is selected. I thought about putting a `return` statement or `e.preventDefault()` but as it was working as-is I didn't want to potentially block any desired behavior. Happy to try that out if we don't want to have an empty conditional. In case it is helpful, I pulled the workaround from this [MaterialTable PR comment](https://github.com/mbrn/material-table/pull/2008#issuecomment-662529834). The first part of the comment is for enter behavior and includes the empty conditional, the second part would be for a user hitting escape and canceling the action.

tillyw:
> > There's still a minor UX bump in that without being able to submit a form with the enter key, the user has to click the "save" button.
> 
> I think this ties in to my instinct to move the checkbox/x to the far righthand side.... I imagine that entails an impractical amount of refactoring, though, so we'll wait for users to validate that this is actually a hurdle for people-who-are-not-me.

I had the same instinct @amenity, and that would make this smoother for a keyboard user! They could just hit `tab` once or twice and then `enter` and submit the row. Not sure how much refactoring that would involve, but I would be open to trying to reorient that column.

tillyw:
> > > There's still a minor UX bump in that without being able to submit a form with the enter key, the user has to click the "save" button.
> > 
> > 
> > I think this ties in to my instinct to move the checkbox/x to the far righthand side.... I imagine that entails an impractical amount of refactoring, though, so we'll wait for users to validate that this is actually a hurdle for people-who-are-not-me.
> 
> I had the same instinct @amenity, and that would make this smoother for a keyboard user! They could just hit `tab` once or twice and then `enter` and submit the row. Not sure how much refactoring that would involve, but I would be open to trying to reorient that column.

@amenity I just reoriented the column (just had to consult the docs really quickly) and it does really smooth out the behavior as we anticipated. Happy to submit a PR if we want to do this. We could even move the plus/add entry button to the right of the table/columns (instead of above it) to make it quickly accessible with a couple of tabs in case someone is adding a lot of team members and wants to be able to do it quickly using only a keyboard (currently you'd have to tab back through all the column labels to get back to the plus button).

PR #243, opened 2021-03-16
Activity log tracking for file uploads and activity log mapping

sergiogcx:
This is a patch to allow for the activity log to track file uploads and for the activity log to handle the updates in the UI


mateoclarke (approved):
🚢

PR #246, opened 2021-03-17
User Management | Fix bug where 'date_added' is overwritten with current date in Edit Staff form

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5495

This PR removes the default value for `date_added` created in the API code to let the database handle the initial value for `date_added`. Previously, the update route in the User API would generate another datetime with a value of "now" and overwrite the existing value for `date_added`.

This can be tested by: 
1. Running the API locally from this branch 
2. Using Postman or other to hit `http://127.0.0.1:5000/users/022ddc8a-1917-4f71-b996-9b8681e77726` and pass a JWT token from logging into staging for authorization
3. Send the payload below

```json
{
    "email": "mike.dilley@austintexas.gov",
    "first_name": "Mike",
    "last_name": "Dilley",
    "roles": ["moped-admin"],
    "status_id": 1,
    "title": "Software Developer",
    "workgroup": "Data & Technology Services",
    "workgroup_id": 3
}
```

I verified the `date_added` does not update by looking for change in the timestamp in the GraphQL response before and after the above API call in the requests in `https://deploy-preview-246--atd-moped-main.netlify.app/moped/staff/edit/8`.


moped-api/README.md
```diff
@@ -39,7 +39,7 @@ $ source venv/bin/activate
 3.Now you are ready to install the requirements
 
 ```
-$ pip install -r requirements/dev.txt
+$ pip install -r requirements/development.txt
```
sergiogcx, line 42: Not a problem at all, just curious about the rename :)

moped-api/claims.py
```diff
@@ -253,4 +253,4 @@ def decrypt(fernet_key: str, content: str) -> str:
     :return str: The decrypted string
     """
     cipher_suite = Fernet(fernet_key)
-    return cipher_suite.decrypt(content.encode()).decode()
+    return cipher_suite.decrypt(content.encode()).decode()
```
sergiogcx, line 256: Not a problem here either, but the python linter is going to bark about not having an empty line at the end of the file

moped-api/users/users.py
```diff
@@ -193,6 +193,9 @@ def user_update_user(id: str, claims: list) -> (Response, int):
     if is_valid_user(current_cognito_jwt) and has_user_role("moped-admin", claims):
         cognito_client = boto3.client("cognito-idp")
 
+        # Remove date_added, if provided, so we don't reset this field
+        request.json.pop('date_added', None)
+
```
sergiogcx, line 198: 👍

sergiogcx (approved):
Amazing work as always! 🚢

moped-api/README.md
mddilley: Forgot to mention that in the notes, but I had noticed in the past that there was a mismatch between the README steps and the actual file name. Just snuck that update in here.

moped-api/claims.py
mddilley: Good call! I was battling my formatter because it was trying to make a bunch of changes. I missed that one so I'll update before merge. 🙏

mateoclarke (commented):
👍

PR #252, opened 2021-03-19
Configure activity feed timestamps to display to users in CDT #5543

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5543

Live Demo:
https://deploy-preview-252--atd-moped-main.netlify.app/moped/projects/164?tab=activity_log

It works great now!

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectActivityLogDialog.js
+++ b/moped-editor/src/views/projects/projectView/ProjectActivityLogDialog.js
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from "react";
 import { useQuery } from "@apollo/client";
+import adjustedTimezoneDateTime from "../../../utils/adjustedTimezoneDateTime";
+import { DateTime } from "luxon";
 
 import {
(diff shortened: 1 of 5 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityLogDialog.js
```diff
@@ -98,10 +100,10 @@ const ProjectActivityLogDialog = ({ activity_id, handleClose }) => {
   };
 
   const getDateTime = data => {
-    const date = data?.moped_activity_log[0]?.created_at;
+    const date = adjustedTimezoneDateTime(data?.moped_activity_log[0]?.created_at); // adjustedTimezoneDateTime(data?.moped_activity_log[0]?.created_at);
```
mateoclarke, line 103: This line is doing weird syntax highlighting so its not totally clear whats happening. But it seems like we can delete the comment right?

mateoclarke (approved):
🚢  

texas_time

PR #260, opened 2021-04-01
Sets time zone to UTC when formatting for display

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5682

Formatting display date using UTC fixes issue where incorrect date was rendered. Correct date is now rendered. @sergiogcx, can you please confirm this does not interfere with [this time zone-related PR you merged](https://github.com/cityofaustin/atd-moped/pull/254)?

https://user-images.githubusercontent.com/35410637/113343907-09441900-92f6-11eb-854f-7c66140c4170.mov

### DIFF
--- a/moped-database/migrations/1617731918952_alter_table_public_moped_project_alter_column_start_date/down.sql
+++ b/moped-database/migrations/1617731918952_alter_table_public_moped_project_alter_column_start_date/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "public"."moped_project" ALTER COLUMN "start_date" TYPE date;
--- a/moped-database/migrations/1617731918952_alter_table_public_moped_project_alter_column_start_date/down.sql
+++ b/moped-database/migrations/1617731918952_alter_table_public_moped_project_alter_column_start_date/down.sql
@@ -1,1 +1,0 @@
-ALTER TABLE "public"."moped_project" ALTER COLUMN "start_date" TYPE date;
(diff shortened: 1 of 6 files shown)


sergiogcx (changes requested):
The old code looks correct: `formattedValue = new Date(formattedValue).toLocaleDateString();`

The problem is not the UI, the problem is that the `date` column is not of type `timestamptz` so it assumes it is already in UTC. 

[The PR you are referring to](https://github.com/cityofaustin/atd-moped/pull/254) fixes this problem by changing the type from timestamp to timestamptz. In short, to fix we just have to transform the date column from type date to timestamptz.

amenity (approved):
Looks splendid to me!

mateoclarke (commented):
So this code change makes sense solving the problem from a UI perspective which was the way the issue was written. But it sounds like @sergiogcx is suggested a database type change which also make sense to me as a better solution closer to the data source and more comprehensive across the UI. 

My question would be, would this change just be to "start date" or do we need to make more changes to date types across the Project table?

sergiogcx:
@tillyw @mateoclarke what would happen if instead of

```
formattedValue = new Date(formattedValue).toLocaleDateString();
```

we use

```
formattedValue = formattedValue;
```

That would make use of the date data RAW without having to change the code.

mateoclarke (approved):
🚀

tillyw:
Reverted this back to just the UI changes, @sergiogcx can you approve so I can merge? Will circle back around with y'all tomorrow about creating a new issue for the proposed DB changes @mateoclarke

sergiogcx (approved):

sergiogcx:
Thank you @tillyw

PR #263, opened 2021-04-06
Multi-Select Project Personnel

sergiogcx:
It looks like the multi-select is stable now, let me know your thoughts.

The live demo should be functional at the new project page and the team tab:
 https://deploy-preview-263--atd-moped-main.netlify.app/moped

Adding Personnel Video

Update Personnel Video

If you are running locally, be sure to restart the hasura server so that the latest unique constraint is applied.

Closes https://github.com/cityofaustin/atd-data-tech/issues/5696

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectTeamTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectTeamTable.js
@@ -11,18 +11,13 @@ import {
 import { Clear as ClearIcon } from "@material-ui/icons";
 import MaterialTable, { MTableEditRow } from "material-table";
 import Autocomplete from "@material-ui/lab/Autocomplete";
-import { filterObjectByKeys } from "../../../utils/materialTableHelpers";
+
 import typography from "../../../theme/typography";
 
 // Error Handler
 import ApolloErrorHandler from "../../../components/ApolloErrorHandler";
 
-import {
-  TEAM_QUERY,
-  ADD_PROJECT_PERSONNEL,
-  UPDATE_PROJECT_PERSONNEL,
-  UPSERT_PROJECT_PERSONNEL,
-} from "../../../queries/project";
+import { TEAM_QUERY, UPSERT_PROJECT_PERSONNEL } from "../../../queries/project";
 
 import ProjectTeamRoleMultiselect from "./ProjectTeamRoleMultiselect";
 import makeStyles from "@material-ui/core/styles/makeStyles";
@@ -45,8 +40,7 @@ const ProjectTeamTable = ({
     variables: { projectId },
     fetchPolicy: "no-cache",
   });
-  const [addProjectPersonnel] = useMutation(ADD_PROJECT_PERSONNEL);
-  const [updateProjectPersonnel] = useMutation(UPDATE_PROJECT_PERSONNEL);
+
   const [upsertProjectPersonnel] = useMutation(UPSERT_PROJECT_PERSONNEL);
 
   if (loading || !data) return <CircularProgress />;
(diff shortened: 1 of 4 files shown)


amenity:
@sergiogcx - thanks, this is working so smoothly! 🙌 
I'm not sure about using chips to render the list of roles, because when one edits it's done through the multi-select drop down, whereas chips are usually deleted with a `x` [remove icon on the righthand side](https://material.io/components/chips#anatomy) of a chip and added via predictive typing. (For some reason I can't find any Material UI examples where there is one extra chip with a `+` sign that you use to add additional chips.) We can't rely on predictive typing for users to assign project roles; they need the whole list in front of them. 

tl;dr — this does an awesome job letting users add multiple roles per person. We may want to mess around with the way the roles are rendered/edited.

amenity (approved):

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -116,18 +148,24 @@ const ProjectTeamTable = ({
...
+      render: personnel => {
+        return personnel.role_id.map(chipRoleId => (
+          <Chip
+            className={classes.roleChip}
+            variant="outlined"
+            label={roles[chipRoleId]}
+          />
+        ));
+      },
+      // validate: rowData => Array.isArray(rowData) && rowData.length() > 0,
```
mddilley, line 160: I'm wondering how we can validate that a user added at least one role to a new project personnel entry. I tested this out, and it looks like this is handled by the upsert mutation so maybe that is the answer. 👍

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -153,62 +191,106 @@ const ProjectTeamTable = ({
...
     true: {
       add: newData => {
-        const activePersonnel = { ...newData, status_id: 1 };
+        let newDataCopy = {...newData};
+        // Aggregate into a unique set if there is stuff already there
+        const newPersonnelState = personnelState.map(item => {
+          if (item.user_id === newData.user_id) {
+            const output = {
+              user_id: item.user_id,
+              role_id: [...new Set([...item.role_id, ...newData.role_id])],
```
mddilley, line 200: Love the use of a set for uniqueness 💯

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -2,41 +2,74 @@ import React from "react";
...
+      // instantiate a new object & populate
+      personnel[`${item.user_id}`] = {};
+      personnel[`${item.user_id}`].user_id = item.user_id;
+      personnel[`${item.user_id}`].role_id = [item.role_id];
+      personnel[`${item.user_id}`].notes = item.notes;
+    } else {
+      // Aggregate role_ids, and notes.
+      personnel[`${item.user_id}`].role_id.push(item.role_id);
+      personnel[`${item.user_id}`].notes = (
+        (personnel[`${item.user_id}`].notes ?? "") +
```
mddilley, line 65: Cool to see how you handled the notes associated with each entry because I forgot how this sits in the DB. Would it make sense to create a new line for each so maybe they sit next to the matching role? Just a thought and not sure how possible that is.

mddilley (approved):
This is awesome, Sergio! Thanks for the great code comments because it made this straightforward to follow. 🙏   I really like how this solves the problem on the front end with little DB updates. The upsert mutation works so well for this. 🙌

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
sergiogcx: It's possible for sure. We would need to discuss it with a prodman.

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
sergiogcx: Oh man, good catch. I forgot to go back to this. Well, yes the mutation will fail if the data does not look good. But there will not be user feedback. Let me see if there is an obvious way to make this work.

sergiogcx: Lol, I wrote the rule once and uncommented, but it worked like charm just by uncommenting. 👍

PR #264, opened 2021-04-06
Multiple-Select Migration

sergiogcx:
This migration creates a unique index that makes sure there will not be duplicate personnel entries in the database table


mddilley (approved):
🙏  Looking forward to seeing the rest of this in action!

PR #266, opened 2021-04-08
Clears eCapris subproject ID when user untoggles Capitally Funded field

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5679

- If a boolean field in ProjectSummaryTable has a dependentField value (this would be the name of the dependent field), DataTable.js clears out the dependentField's data when a user switches the boolean field off

https://user-images.githubusercontent.com/35410637/114077068-048ae200-986d-11eb-9794-300f48191c7f.mov


mateoclarke (approved):
🚢  Looks good! Could you update the PR title to something would would make more sense to a user in the Release Notes?

tillyw:
Thanks @mateoclarke! Sorry about the title, rewrote it to something more readable. @amenity if this looks good to you, I can merge it ASAP!

amenity:
@tillyw - encountering a few issues with the new switch. 😬
- Was able to switch _on_ capital funding [here](https://deploy-preview-266--atd-moped-main.netlify.app/moped/projects/198), but not add an eCapris subproject ID on the first few tries. First I clicked on a browser-supplied value in a dropdown. Then I typed and hit "enter". The second time I typed and hit enter, this time without a period in the number, it went through. 

https://user-images.githubusercontent.com/1463708/115455433-938de780-a1e7-11eb-8569-395c26d34efb.mov


I'm also getting an error message _removing_ capital funding from a project. To reproduce, see https://deploy-preview-266--atd-moped-main.netlify.app/moped/projects/176 and try to switch off capital funding. 
Screen Shot 2021-04-20 at 2 31 41 PM

tillyw:
Ack, sorry you encountered these issues @amenity! The issue with turning off the switch was another one caused by the field name changing in the database/main branch, it was working before that changed and I neglected to update the field name in the UI to match the new field name. That should be resolved in the new preview environment!

I believe the first issue you encountered with not being able to input an ID is because the database now has a rule that each `ecapris_subproject_id` needs to be unique, so if you input one that has already been used (which I suspect was the case when you use the browser-supplied ID) it will error out. Can you please try again with the new deploy preview and a unique `ecapris_subproject_id`?

Maybe we should also consider adding a special warning message in a future PR that tells the user the ID is already being used in case this comes up for a user.

amenity (approved):
Looks great!

amenity:
PS opened a [Slack convo about that rule](https://austininnovation.slack.com/archives/CNUEPKLB1/p1618949678294200); we'll need to get rid of it.

tillyw:
Sounds good, thanks @amenity!

PR #267, opened 2021-04-08
reverse order of event list

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5599

Displays events in reverse chronological order, with event creation at the bottom. Also fixed all display dates in Moped to display in UTC time.


amenity (approved):
LSTM! ✨⤵️👍

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -145,11 +145,11 @@ const ProjectActivityLog = () => {
...
       ) {
-        // Move it to the top of the list (make it first)
-        outputList.unshift(event);
-      } else {
-        // Else, just stack it to the bottom
+        // Append it to the end of the list (make it last)
         outputList.push(event);
+      } else {
+        // Else, prepend it to the beginning of the list (so it appears in descending chronological order)
+        outputList.unshift(event);
```
mateoclarke, line 152: This function seemed to be straying from its intended purpose as documented and it seemed a bit janky to me to be ordering activity log items when we could seem to control that through reliably the query. So I made a PR using that query based approach instead of in the UI. Let me know what you think: https://github.com/cityofaustin/atd-moped/pull/283/files

mateoclarke (commented):
I think the localeDateString updates are worth keeping from this PR but I want to see if you'd consider my PR #283 instead of refactoring the `reorderEventList` function. Let me know what you think.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
tillyw: @mateoclarke I reworked the PR to order the query results in descending order, then modified the operations on the frontend to focus only on pulling out the creation event and adding it to the end of the event list that is displayed. I tried to ensure this operation would make sense and be easily readable to other devs. Please let me know if this looks good!

amenity:
@tillyw @mateoclarke -- do we have more to do here?

tillyw:
> @tillyw @mateoclarke -- do we have more to do here?

I pushed an updated PR last week and tagged @mateoclarke in a comment letting him know. Just need final review!

amenity:
@tillyw @mateoclarke - Thanks!  so should this PR be closed?

tillyw:
> @tillyw @mateoclarke - Thanks! so should this PR be closed?

No, it should be merged pending an approving review from @mateoclarke

mateoclarke (approved):
🚢

PR #268, opened 2021-04-08
Move actions column index to the right on MaterialTable instances

tillyw:
PR that would move the actions column to the right when adding to the Project Team and Timeline tables

Associated with https://github.com/cityofaustin/atd-data-tech/issues/5375

Example:
Screen Shot 2021-04-08 at 4 54 27 PM


amenity (approved):
Small change, but what a giant difference this makes! 😍 Thanks, @tillyw.

mateoclarke (approved):
🛳️

PR #269, opened 2021-04-09
Project Summary Map | Set order of layers in the map

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5685

This PR adds a `layerOrder` property to the map layer configurations. Then, when rendering the layer on the map, the order is enforced by sorting. This keeps the points on top of the lines regardless of the order they are stored in the DB.

This can be tested with these two layers:
https://deploy-preview-269--atd-moped-main.netlify.app/moped/projects/209 (Points set first)
https://deploy-preview-269--atd-moped-main.netlify.app/moped/projects/212 (Streets set first)

### Before
Screen Shot 2021-04-09 at 2 26 56 PM

### After
Screen Shot 2021-04-09 at 2 26 38 PM


moped-editor/src/utils/mapHelpers.js
```diff
@@ -296,8 +298,16 @@ export const createSummaryMapLayers = (selectedIds, geoJSON) => {
...
+    ))
+    .sort((a, b) => {
+      // The id of the Source component maps to the source layer names in mapConfig, each layer config has a set order
+      const idA = a.props.id;
+      const idB = b.props.id;
+      const orderA = mapConfig.layerConfigs[idA].layerOrder;
+      const orderB = mapConfig.layerConfigs[idB].layerOrder;
+
+      return orderA > orderB ? 1 : -1;
+    });
```
sergiogcx, line 310: Daang, it sounded a lot easier than this. The fact that you have to sort by layer order... 🙏

sergiogcx (approved):
Sorry to see about the sorting, hopefully there is a better way in the future 👍

moped-editor/src/utils/mapHelpers.js
sergiogcx: You probably already know it, but a thought crossed my wires. Is there anything [z-index](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index) can do to save us? A long shot in my head but worth asking,

mddilley: You and I are on the same page! The documented way to manage this is [beforeId](https://visgl.github.io/react-map-gl/docs/api-reference/layer#beforeid) which is pretty much like z-index for Mapbox layers. It gets a little confusing and hard to define the order because the argument is a layer id attribute, but your feedback helped me go that direction. It shouldn't be hard to refactor since there are only a few layers, and I like it a lot better! 🙏

mateoclarke (approved):
🚢  Looks good but sounds like theres another approach being tried in another PR?

mddilley:
Yep, @mateoclarke, I was going to refactor this code to use an attribute called `beforeId` (kind of like z-index for map layers), but I forgot about one limitation. If you set a `beforeId` as a layer that isn't always present in the map, the layer with the `beforeId` set has no point of reference and doesn't render. 

For example, if the `beforeId` for the streets layer is set to the points layer, the streets layer will error and not render if there are no points in a project. I'll keep this code as is since, at worst, there will only be 3 layers to sort through.

amenity (approved):
Splendid! ✨

PR #270, opened 2021-04-09
Data Model | Update values in `moped_phases` #5381

sergiogcx:
This migration makes heavy use of PL/SQL, since a simple upsert may not be sufficient. It also needs to be aware of the stage it is being run under (local, staging or prod).

The first major challenge is the existing data and the structural relationship between the `moped_proj_phases` and `moped_phases` tables. It relies on text to establish those relationships. The main problem with that is that it is difficult to make updates to the phase names with existing records because it would sever those relationships, and one would run into foreign key errors. 

To fix this this PR drops that constraint and establishes a new one based on the phase_id, so that the phase names can be renamed more easily:

Closes https://github.com/cityofaustin/atd-data-tech/issues/5381


mateoclarke (approved):
🚢

PR #272, opened 2021-04-13
Project Extent | Show drawn points in project summary map and refine UI

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5749

This PR:

- Adds the drawn points to the project summary map
- Updates the styles of drawn points when the drawing UI is active (for hovered, selected, and default states)
- Includes drawn locations in the "number of locations" total at the bottom of the map (now calculated from the project extent feature collection)
- Updates the cursor in drawing mode to `pointer`
- Adds data driven calculation of drawn point width in the draw UI based on the zoom level
- Prevents user from editing vector tile layer features by clicking while drawing with the draw UI
- Opens draw UI if a user clicks on a drawn point in the edit map

Preview project with two drawn points: https://deploy-preview-272--atd-moped-main.netlify.app/moped/projects/176

#### Drawn points in summary map
Screen Shot 2021-04-15 at 6 35 16 PM

#### Updated draw UI points styles and location total
Screen Shot 2021-04-15 at 7 24 19 PM

#### Pointer cursor, data driven drawn point width, disabled vector tile layer interaction, and opening draw UI on drawn point click
dataDriven


moped-editor/src/utils/mapDrawHelpers.js
```diff
@@ -37,27 +36,91 @@ export const MODES = [
...
+  maxZoom,
+  minPixelWidth,
+  maxPixelWidth
+) {
+  return (
+    ((currentZoom - minZoom) * (maxPixelWidth - minPixelWidth)) /
+      (maxZoom - minZoom) +
+    minPixelWidth
+  );
+}
```
sergiogcx, line 85: Check this out @johnclary, mike is out there doing science 💯 

https://matthew-brett.github.io/teaching/linear_interpolation.html

moped-editor/src/utils/mapHelpers.js
```diff
@@ -161,12 +185,38 @@ export const createZoomBbox = featureCollection => {
...
+ */
+export const getSummaryMapInteractiveIds = featureCollection => [
+  ...new Set(
+    featureCollection.features.map(
+      feature =>
+        mapConfig.layerConfigs[feature.properties.sourceLayer].layerIdName
+    )
+  ),
+];
+
```
sergiogcx, line 219: Brilliant use of sets!  🔢  ⚡  👍

moped-editor/src/utils/mapHelpers.js
```diff
@@ -263,25 +313,23 @@ export const createProjectSelectLayerConfig = (
...
+    return acc[sourceLayerName]
+      ? {
+          ...acc,
+          [sourceLayerName]: {
+            ...geoJSON,
+            features: [...acc[sourceLayerName].features, feature],
+          },
+        }
+      : { ...acc, [sourceLayerName]: { ...geoJSON, features: [feature] } };
+  }, {});
```
sergiogcx, line 332: On fire 🔥  Wouldn't it have been easier to do something like this? 

```
return {
          ...acc,
          [sourceLayerName]: {
            ...geoJSON,
            features: [...(acc[sourceLayerName] ? sourceLayerName].features : []), feature],
          }
        }
```

Maybe it would be easier to read and less code duplication

sergiogcx (approved):
Impressive work yet again! 💯 🥇

moped-editor/src/utils/mapDrawHelpers.js
```diff
@@ -37,27 +36,91 @@ export const MODES = [
...
+    if (currentZoom >= minZoom && currentZoom < maxZoom) {
+      return linearInterpolation(
+        currentZoom,
+        minZoom,
+        maxZoom,
+        minPixelWidth,
+        maxPixelWidth
+      );
+    }
+  }
```
sergiogcx, line 123: Ah, yes the for loop. I personally don't mind, but mateo might ask you why not use `map` or `forEach` ? 🤷‍♂️

moped-editor/src/utils/mapDrawHelpers.js
mateoclarke: 🔥

mateoclarke (approved):
🚢  Really great work making all these improvements to a complex mapping feature!

moped-editor/src/utils/mapDrawHelpers.js
mddilley: Just seeing this, yep, the reason that I used a `for` loop is that the loop is used to search where a value falls in some ranges. I wanted to be able to break out of the loop once the match is found and prevent unnecessary iterations since this function fires rapidly when a user is zooming in and out

moped-editor/src/utils/mapHelpers.js
mddilley: Yeah, this is a great idea! I'll update before merge. 🙏

moped-editor/src/utils/mapHelpers.js
mddilley: Totally stole this from one of your PRs 😄

moped-editor/src/utils/mapDrawHelpers.js
amenity: 🧠👨‍🔬!

amenity (approved):
WOW. Just wow. This is soooo good, @mddilley. Really love all the UI style choices you've made here, e.g. selected vs. unselected, drawn vs existing. 👨🏻‍🎨

PR #273, opened 2021-04-13
[Database] Update project statuses #5251

sergiogcx:
Should be a breeze, the state is identical across all three stages.

Closes https://github.com/cityofaustin/atd-data-tech/issues/5251


mddilley (approved):
Thanks for updating! I tested locally and forgot that the data is seeded after the migrations, but it is not a worry for staging and prod. Ran the SQL, and it works like a charm. 🙏

mateoclarke (approved):
🚢

PR #274, opened 2021-04-13
Creates new "current phase" table on timeline (user can select one active phase)

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5309

- Creates new table on summary page that displays whichever phase is set to "current" in the timeline. If there is none, "none" is displayed.
- Ensures user can only set one active phase at a time. If user adds or sets a new active phase while there is an existing active phase, the existing active phase is set to inactive. If user deletes the only "active" phase, the summary tab displays "none."
- When a user updates the active phase in the timeline table, the data is also refetched in the summary view, so the user can click the summary tab to see current information without having to refresh the application.
- If user does not turn on the "active" switch when adding a new phase, the default value displayed in the table is "no" rather than no value (you can't test this in the deploy preview, but it works locally, it's an inconsequential change so not important to test, more for aesthetic purposes so that field does not appear blank in the table). 

Some things we might want to address in future issues:
- Remove the "current phase" field option that appears when a user is creating a new project
- Set the timeline table to only allow a user to add one instance of any given phase so there are no duplicates (e.g., if there is already a "construction" phase, do not allow the user to add a new phase and select "construction" from the dropdown)
- Remove borders from and style summary table tables

https://user-images.githubusercontent.com/35410637/117221256-7e5fae00-adce-11eb-9dc5-8b893ee5ef80.mov


amenity (approved):
LSTM!! ✨🙌

amenity:
Moving this to Review/QA, @tillyw @mateoclarke ⚡✨

mateoclarke (approved):
Looks great! 🚢  

Can you update the name of our PR to something that would look better in our Release Notes please?

tillyw:
> Looks great! 🚢
> 
> Can you update the name of our PR to something that would look better in our Release Notes please?

Renamed PR, sorry about that! Trying to be more consistent about naming my PRs!

PR #276, opened 2021-04-14
Fixes bug with File Upload

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5810

Live Demo:
https://deploy-preview-276--atd-moped-main.netlify.app/moped/projects/188?tab=files

Fixes:
- File upload bug where the dialog would break after the first successful upload
- Fixes sorting for dates
- Fixes sorting for file size

### DIFF
--- a/moped-editor/src/components/FileUpload/FileUploadDialogSingle.js
+++ b/moped-editor/src/components/FileUpload/FileUploadDialogSingle.js
@@ -243,7 +243,7 @@ const FileUploadDialogSingle = props => {
           color="primary"
           variant="contained"
           startIcon={<Icon>save</Icon>}
-          disabled={!!fileReady}
+          disabled={!fileReady}
         >
           Upload
         </Button>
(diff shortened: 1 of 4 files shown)


moped-editor/src/components/FileUpload/FileUploadDialogSingle.js
```diff
@@ -88,6 +88,25 @@ const FileUploadDialogSingle = props => {
     setFileDescription(e.target.value);
   };
 
+  /**
+   * Resets all the values in the file upload component
+   */
+  const clearState = () => {
```
mddilley, line 94: Good call! The bug makes sense now. 👍

mddilley (approved):
Looks great - I couldn't reproduce the bug from the demo. Great stuff. 🙌 

I noticed one small quirk with the date sort, but I think that you might have just patched it. Looks like it is cycling through three orders here.
Apr-16-2021 10-37-16

sergiogcx:
> Looks great - I couldn't reproduce the bug from the demo. Great stuff. 🙌
> 
> I noticed one small quirk with the date sort, but I think that you might have just patched it. Looks like it is cycling through three orders here.
> Apr-16-2021 10-37-16

Yea, it seems the three states are:
- Sort low to high
- Sort high to low
- Disable sort

PR #280, opened 2021-04-17
Data Migration | Transform IMPD data to fit moped_proj_personnel table in Moped #5602

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5602

What this branch does is a necessary refactor for the actual migration happening in a different PR. Once the personnel is migrated, they are imported as inactive; however, this presents the projects team page with a problem because the graphql query excludes any inactive users causing the code to a WSOD. Also, I noticed this query gathers the entire list of active users, even if there is no personnel associated to a project, which is not performant or necessary.

** Tried renaming the branch name to have the same github issue number but GitHub it keeps nagging **

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -2537,6 +2537,10 @@
 - table:
     schema: public
     name: moped_proj_personnel
+  object_relationships:
+  - name: moped_user
+    using:
+      foreign_key_constraint_on: user_id
   insert_permissions:
   - role: moped-admin
     permission:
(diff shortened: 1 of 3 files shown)


amenity:
Another one I can't test on Netlify, prehaps? Getting 4everhang 🐒 Screen Shot 2021-04-20 at 12 10 52 AM

mateoclarke (approved):
🚢  Looks good code-wise!

mateoclarke:
> Another one I can't test on Netlify, prehaps? Getting 4everhang 🐒 

Any Pull Request that has updates to the `moped-database` files probably isn't going to work in the Netlify deploy preview instances. We can chat about why this is so and whether its enough of a priority to find other solutions to have product acceptance testing before merging PRs like this.

PR #281, opened 2021-04-17
Moped Editor | Make The Project Search Feature Great Again #5830

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5830

It appears that switching the ecapris_subproject_id type from string to number is breaking the code. This PR refactors the code so that whenever there is an unavoidable situation whenever we pass a string to a numeric field in a search it assumes a safe default value.


mateoclarke (approved):
🚢

PR #283, opened 2021-04-20
Update activity log order in query instead of UI

mateoclarke:
Alternative approach to cityofaustin/atd-data-tech#5599

### DIFF
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -259,9 +259,7 @@ export const PROJECT_ACTIVITY_LOG = gql`
         user_id
       }
     }
-    moped_users(where:{
-      status_id:{_eq: 1}
-    }) {
+    moped_users(where: { status_id: { _eq: 1 } }) {
       first_name
       last_name
       user_id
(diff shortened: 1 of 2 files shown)


tillyw:
@sergiogcx We were wondering if there is a particular need for a logical statement that isolates the "creation" event and puts it first (or last in this iteration). Do you think the same would be accomplished by ordering by `created_at` in the query, or is there something we're missing?

sergiogcx:
@tillyw @mateoclarke yes, the additional logic is needed because the events arrive at different times. Whenever a new project is created we are also creating personnel records and phases records, those are separate events and they arrive at unpredictable times to the database activity log. So I added additional logic to separate the create event for the project and move it to the top.

sergiogcx:
To follow up on the previous comment I made, here is the code that selects the INSERT event for an object of type moped_project: 

https://github.com/cityofaustin/atd-moped/blob/f6a45b10939831e977075b75b0c97b09869e1299/moped-editor/src/views/projects/projectView/ProjectActivityLog.js#L139

mateoclarke:
ok, thanks Sergio, I'll update this PR.

PR #286, opened 2021-04-21
Replaces the ecapris_subproject_id unique constraint with an unconstrained index

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5881

Replaces the ecapris_subproject_id unique constraint with an unconstrained index and allows ecapris_subproject_id to have no value.

To test, use [this deploy preview environment](https://deploy-preview-288--atd-moped-main.netlify.app/moped) rather than the one associated with this PR. The one I linked here includes a fix for an unrelated bug for a better testing experience. Not that you can't test whether you can put in the same ID twice (unique constraint) in the deploy preview environment. You would have to locally run the [code that establishes that environment](https://github.com/cityofaustin/atd-moped/pull/288).

### DIFF
--- a/moped-database/migrations/1619024083593_ecapris_replace_constraint_with_index/down.sql
+++ b/moped-database/migrations/1619024083593_ecapris_replace_constraint_with_index/down.sql
@@ -1,0 +1,6 @@
+-- Drop the unconstrained index
+DROP INDEX moped_project_ecapris_subproject_id_index;
+
+-- Create the unique index constraint
+CREATE UNIQUE INDEX moped_project_ecapris_subproject_id_uindex
+    ON moped_project (ecapris_subproject_id);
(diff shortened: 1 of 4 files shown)


mateoclarke:
So when I run this locally, I'm not getting the same `Error: Uniqueness violation` bug I was seeing. But it also seems like there's still a bug related to the ecaprisID. The new error I see is `Error: invalid input syntax for type numeric: ""`. This seemed to me like we're trying to submit an empty string for the eCaprisID even though I hand't interacted with that field at all. I went back to the first page, added a random numeric ID and then that error was gone. I think we need to address how the eCaprisID works in the first page as a part of this bug fix.

Screen Shot 2021-04-21 at 12 44 16 PM

mateoclarke (approved):
🚢

sergiogcx (approved):
🚢 🚀 🙏

amenity (approved):
As anticipated, looks ... ETSAB*  TM 😅  since I'm not running locally. Thanks for the quick fix though, @tillyw !

_*exactly the same as before_

amenity:
Addendum; I _was_ able to test the fix for an unrelated bug: It institutes validation (must be numeric) on the eCAPRIS field.

PR #287, opened 2021-04-21
Bug | Handle saving a project when no team members are added

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5879

This PR updates the handling of project personnel records when saving a project through the new project stepper. Previously, if a project contained no personnel, the code tried to map over an `undefined` returned from referencing an invalid index of an empty array. Ex. [][0] => undefined => undefined.map() => `Uncaught TypeError: Cannot read property 'map' of undefined`. We probably haven't run into this bug before since we always add team members when demoing/testing.

To test:
1. Created a new project and be sure to add be sure to eCapris ID (to avoid the [eCapris ID error](https://github.com/cityofaustin/atd-data-tech/issues/5881))
2. Don't add any personnel
3. Select a feature
4. Save the project

### DIFF
--- a/moped-editor/src/views/projects/newProjectView/NewProjectView.js
+++ b/moped-editor/src/views/projects/newProjectView/NewProjectView.js
@@ -228,23 +228,27 @@ const NewProjectView = () => {
       .then(response => {
         const { project_id } = response.data.insert_moped_project.returning[0];
 
-        const cleanedPersonnel = personnel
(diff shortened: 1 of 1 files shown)


mateoclarke (approved):
🚢  
judge_judy_bug

sergiogcx (approved):
🚢 🙏

amenity (approved):
Works like a charm. ☘️ 🙏

PR #289, opened 2021-04-22
Bug | Can't add or edit Team members #5880

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5880

Live Demo:
https://deploy-preview-289--atd-moped-main.netlify.app/moped/projects/176?tab=team

(Nodding to my professor last semester, I never thought I'd ever use tuple operations and tuple sets. )

### DIFF
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -103,6 +103,15 @@ export const TEAM_QUERY = gql`
       project_role_id
       project_role_name
     }
+    moped_users(
+      order_by: { last_name: asc }
+      where: { status_id: { _eq: 1 } }
+    ) {
+      first_name
+      last_name
+      workgroup_id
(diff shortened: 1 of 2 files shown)


mateoclarke (approved):
🚢  

I would love it if you could talk me through the tuple solution in more detail. Its a little over my head how it solves the problem presented in by this bug. But I'm looking forward to learning more from you through this one.

amenity (approved):
Works great! I did catch a baby bug where "null" actually displays in the notes field. Couldn't 100% reliably reproduce, but it only seemed to happened when it was the first record to be added. LMK if you want this to be a separate bug report, @sergiogcx 

https://user-images.githubusercontent.com/1463708/115767113-5dc73b00-a36e-11eb-9f71-34b964f592e5.mov

PR #290, opened 2021-04-22
Project Extent | Fix geocoder so click and hover events don't propagate through the input box

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5825

This PR:

1. Updates the zoom NavigationControl to no longer capture clicks and propagate them to the map below
2. Adds a[ containerRef to the Geocoder](https://github.com/SamSamskies/react-map-gl-geocoder#ignore-map-events-example) component so clicks don't propagate to the map below
3. Renders the map layer controls outside of the ReactMapGL component and adjusts the CSS so it remains in the same position on top of the map and no longer propagates mouse events to the map below
4. Memoizes both `handleViewportChange` and `handleGeocoderViewportChange` because it fixes the bug where the geocoder dropdown menu flickered and remained open after choosing a result. Plus, the docs say so. 😅 
     > All non-primitive prop values besides mapRef and containerRef should be memoized. 

This can be tested on: https://deploy-preview-290--atd-moped-main.netlify.app/moped/projects/188

#### Clicks don't propagate through zoom controls, layer controls, or geocoder controls
clickThrough

#### Dropdown no longer remains open and after choosing a result from the geocoder and no flickering
dropdownStuck


sergiogcx (approved):
🚢 🙏

amenity (approved):
🙏👍

PR #293, opened 2021-04-28
Project Components | Add moped_project_features table and relate to moped_project table

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5938
Closes https://github.com/cityofaustin/atd-data-tech/issues/5939

This PR updates the DB with a new table called `moped_proj_features` that contains rows of data about features associated with projects. It also removes the `project_extent_geojson` field from the `moped_project` table to normalized project extent records. The frontend has also been updated to query `moped_proj_features` records and build a GeoJSON feature collection to render them.

In addition, the `ADD_PROJECT` mutation has been updated with nested insertions for feature and personnel records so all data from the New Project workflow is inserted in one request.

To test, we can use the workflow dispatch to see the updates to the DB and frontend working together. Since this is a refactor, there are no new features, but the following steps can be taken to test that there are no regressions.
1. Create a new project with personnel and locations
2. Use the edit map in the New Project view to add, edit, remove, and draw features


mateoclarke:
I tried deploying this branch with the workflow dispatch, but the auth seems to be acting weird. I log in with my pw, see the landing page briefly then the screen bombs white. If I clear the local storage cache, I can get back to the login page.
```
2.9b294a9f.chunk.js:2 SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
```

On local, I was unable to get past the seeding stage. This is the error I'm seeing.
```
FATA[0024] [postgres-error] query execution failed ($.args[0].args)
[42703] FatalError: column "project_extent_geojson" of relation "moped_project" does not exist
```

I will try to revisit this in the morning

mddilley:
@mateoclarke I forgot to add a note to this PR that the seed data error is patched in https://github.com/cityofaustin/atd-moped/pull/303 which branches from this. Sorry about that. 😅  I would have done that in reverse order, but I didn't catch the issue with the seed data until after the work here.

mateoclarke (approved):
This code makes sense and I tested the branch with updated seeding data so I think this is good to go too. 

One thing I did notice in my local env running this branch is that the Mapbox basemap doesn't seem to be loading. Not sure if this is an environment thing or a bug or maybe something going on with my browser. But it does seem like a regression. As long as we don't loose track of that, I'm good with merging this refactor.

Screen Shot 2021-05-07 at 3 51 25 PM

PR #311, opened 2021-05-17
Project Components | Update data model to relate components and features #5962

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5962

**This is a database-only change. Local test only.


mateoclarke (approved):
None of thee changes raise any red flags for me but perhaps it would be good to include @johnclary as a reviewer on this one. 🚢

johnclary:
thanks @sergiogcx! i read through the migrations but did not explore in the console. thanks for knocking this out. i owe you some UI mockups. will get those to you ASAP.

PR #312, opened 2021-05-17
Create SQL View to access and filter project personnel

chiaberry:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5517

frontend & db changes, run locally or test instance

Most of the credit goes to Sergio for helping me out with the SQL View

To Test: from projects list view, toggle Advanced Search. Select "Team Member" from drop down, try searching for Mike or Amenity to see results. I also tested by filtering by more than one person.


moped-editor/src/components/GridTable/GridTable.js
```diff
@@ -309,11 +309,6 @@ const GridTable = ({ title, query }) => {
     // The first key is the outermost nested part of the graphql query
     const section = obj[keys[0]];
 
-     // Bypass value extraction if column value should be "stringified"
-    if (query.config.columns[exp]?.stringify) {
-      return JSON.stringify(section)
-    }
-
```
sergiogcx, line 316: Are we going to remove the stringify feature?

chiaberry: okay, i was torn about it, I removed it because we arent using it now, but if you think we might in the future, ill put it back

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -101,26 +101,26 @@ export const ProjectsListViewQueryConf = {
...
+            uniqueNames[fullName] = uniqueNames[fullName] + `, ${projectRole}`
           } else {
-            uniqueNames[fullName] = person.moped_project_role.project_role_name
+            uniqueNames[fullName] = projectRole
           }
         })
-
         for (const [key, value] of Object.entries(uniqueNames)) {
           personnel.push(`${key} - ${value}`)
         }
```
sergiogcx, line 126: This is very clever, I like how the value turns into a string and then the string gets broken down and parsed. There are however certain things that need change, but not deal-breakers:

1. The comma notation, that's important for me because it helps me know whenever an expression ends.
2. A few variables (if not all) can be made `const`, they are not modified and do not need to be `let` variables.
3. Great use of `namesArray.forEach` it looks great, but the `Object.entries` loop can be improved.

sergiogcx (approved):
This is great! thank you for working on it :)

amenity (approved):
Looks beautiful, @chiaberry! 🙏 ✨

PR #313, opened 2021-05-18
Fix problems with project permissions (Patch)

sergiogcx:
Cannot explain the regression, but this should take care of the permissions.

** This can be tested in the moped-test environment and local environment,


mateoclarke (approved):
🚢

PR #316, opened 2021-05-20
Project Components: Ability to draw lines #6150

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6150

** Front-end changes only, safe to use Netlify PRs, moped-test or local env.

Live Demo:
https://deploy-preview-316--atd-moped-main.netlify.app/moped/session/signin


amenity (approved):
Works great, @sergiogcx! 🙌 🗺️ 

Would it be possible to have "Drawn Lines" and "Drawn" be a single layer?

sergiogcx:
@amenity so I've re-named them to "Drawn Points" and "Drawn Lines".

The short answer is that yes it is possible. The long answer is that it will likely be a separate issue, this is because there needs to be a decent refactoring of the logic behind the selection to be based on groups and not individual items.

moped-editor/src/views/projects/projectView/ProjectSummaryMap.js
```diff
@@ -41,7 +41,16 @@ const ProjectSummaryMap = ({ projectExtentGeoJSON, setIsEditing }) => {
   const mapRef = useRef();
   const featureCount = countFeatures(projectExtentGeoJSON);
 
+  /**
+   * Make use of a custom hook that returns a vector tile layer hover event handler
+   * and the details to place and populate a tooltip.
+   */
```
johnclary, line 47: appreciate all of the doc strings you've added throughout!

johnclary (approved):
works like a charm @sergiogcx! 🙌 

Agree with @amenity's suggestion and I think the map UX can definitely improved in general, but we knew that before this enhancement and can tackle it as we move to component creating and editing.

PR #320, opened 2021-05-23
Add updated_at field to moped_project and sort by last updated

chiaberry:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6094

Database changes. Also changes to the hasura events trigger (which handles the insertion into the activity log). The changes to the hasura events trigger pass the python tests, but have not been tested in production. 

Adds the field "updated_at" to the moped_project table and includes it as the column "last modified" in the project view.

Screen Shot 2021-05-23 at 4 52 50 PM


moped-database/metadata/tables.yaml
```diff
@@ -4439,84 +4439,87 @@
...
+      - project_length
+      - project_name
+      - project_order
+      - project_priority
       - project_team_members
+      - project_uuid
+      - start_date
+      - status_id
+      - timeline_id
+      - updated_at
```
chiaberry, line 4465: the only real change is updated_at, the rest was hasura changing the order of permissions

moped-database/migrations/1621542484886_run_sql_migration/up.sql
```diff
@@ -0,0 +1,30 @@
...
+    mp.fiscal_year,
+    mp.capitally_funded,
+    mp.project_priority,
+    mp.date_added,
+    mp.added_by,
+    mp.is_retired,
+    mp.milestone_id,
+    mp.status_id,
+    string_agg(concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members,
+    mp.updated_at
```
chiaberry, line 25: adds updated_at to the sql view

moped-data-events/activity_log/MopedEvent.py
```diff
@@ -44,6 +47,9 @@ class MopedEvent:
           }) {
             affected_rows
           }
+          update_moped_project(where: {project_id: {_eq: $recordId}}, _set: {updated_at: $timestamp}) {
+            affected_rows
+          }
```
sergiogcx, line 52: Brilliant, cool that you made it a timestamptz value, those are rare in GraphQL! 👍

moped-data-events/activity_log/MopedEvent.py
```diff
@@ -304,6 +310,7 @@ def get_variables(self) -> dict:
             "description": self.get_diff(),
             "updatedBy": self.get_event_session_var(variable="x-hasura-user-id", default=None),
             "operationType": self.get_operation_type(default=None),
+            "timestamp": datetime.datetime.now(tz=pytz.utc).isoformat(),
```
sergiogcx, line 313: This is clever, straight into the variables 👍

sergiogcx (approved):
Thank you! 🙏  I know there is no real way to test this except through deployment, so feel free to merge away 🚢

mateoclarke (approved):
🚢

amenity:
Question, @sergiogcx @chiaberry — is the fact that I'm not seeing the sort order on the projects page change when I update projects because 
- All the projects ~were last updated~ share a "last modified" date because that column was populated at the same time for all when it was created?
- We're not processing time of day, only the date?
- Something specific to the Test instance DB
- A lag — (I'm also not seeing any changes reflected in the activity log)

sergiogcx:
@amenity The activity log does not work in the moped-test area at the moment (it requires SQS and the API to be deployed).

amenity:
@sergiogcx - Aha, thanks! Does the same thing explain why modifying projects doesn't actually affect the "Last modified" date?

sergiogcx:
@amenity Ah, right I forgot to mention the most important part. You are totally right, the last_updated depends on SQS and the API to work. I am feeling pretty confident chia wrote enough pytests to make sure the script runs on main, but I do not mind creating an SQS Lambda  and an API that are exclusively connected to moped-test in order to see things like this.

amenity:
@sergiogcx - thanks, I realized I could just test by creating a new project since, as @chiaberry confirmed, it does take into account timestamp. 
Screen Shot 2021-05-25 at 8 29 36 PM

amenity (approved):
LSTM! ✨  ⬇️ ⌚ ⬆️  ✨

PR #324, opened 2021-05-26
Investigate and address User API failures #5799

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5799

** This is mostly a UI change only, you can test in Netlify, moped-test and locally.

There are API changes that are meant to make sure the user name stays as lowercase in AWS.

Also, there was an upgrade for the DataGrid component where a column's className can now be customized.

Live Demo:
https://deploy-preview-324--atd-moped-main.netlify.app/moped/staff

### DIFF
--- a/moped-api/claims.py
+++ b/moped-api/claims.py
@@ -25,6 +25,18 @@ current_hasura_claims = LocalProxy(
 )
 
 
+def lower_case_email(user_email: str) -> str:
+    """
+    Attempts to lower case a user email address
+    :param user_email: The user email address in question
+    :return: 
+    """
+    try:
+        return str(user_email).lower()
+    except:
+        return user_email
+
+
 def get_claims(payload: LocalProxy) -> dict:
     """
     It's a handy way to extract the hasura claims from a valid payload.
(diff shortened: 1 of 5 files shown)


sergiogcx:
The email address is now presented without any formatting:

moped-api/claims.py
```diff
@@ -25,6 +25,18 @@
...
+    """
+    Attempts to lower case a user email address
+    :param user_email: The user email address in question
+    :return: 
+    """
+    try:
+        return str(user_email).lower()
+    except:
+        return user_email
+
```
chiaberry, line 38: 🎊

moped-editor/src/views/staff/StaffListViewQueryConf.js
```diff
@@ -140,7 +141,9 @@ export const StaffListViewQueryConf = {
     },
   },
   order_by: {},
-  where: {},
+  where: {
+    status_id: "_eq: 1"
```
chiaberry, line 145: if we are going to only get the active users, do we still need to show the column `status_id`?

chiaberry:
how can I test the user api failures? is there a way to provoke the error message?

moped-editor/src/views/staff/StaffListViewQueryConf.js
sergiogcx: We don't, good call! 👍

sergiogcx:
> how can I test the user api failures? is there a way to provoke the error message?

Create a dummy user (don't worry it will fail), with a bad password without numbers and/or special characters.

moped-editor/src/views/staff/StaffListViewQueryConf.js
mateoclarke: I agree with Chia. Doesn't seem necessary

sergiogcx: It's deleted now :)

amenity:
Thanks, @sergiogcx! This looks great.

amenity (approved):

PR #339, opened 2021-06-07
Advanced filters "is blank" bug fix

chiaberry:
https://deploy-preview-339--atd-moped-main.netlify.app/moped/

To test: try searching by "is blank" in the advanced filters for Team Member or Current Phase. 

Also I removed the option to filter by "is blank" for the fields where `NULL` isn't an option


moped-editor/src/components/GridTable/GridTable.js
```diff
@@ -222,17 +222,23 @@ const GridTable = ({ title, query }) => {
       return;
     }
 
-    // If the operator includes "is_null", then the value is always true
+    // If the operator includes "is_null", we check for empty strings
     if (gqlOperator.includes("is_null")) {
-      value = "true";
+      gqlOperator = envelope ? "_eq" : "_neq";
```
sergiogcx, line 227: Making envelope a boolean attribute seems to be an anti-pattern. I think it should remain as string type to wrap up values going into the database. I'd be ok if we switch the boolean value to something like `envelope === "true"`.

chiaberry: works for me -- I originally had it as envelope === true but it felt weird. 
do we ever use the string true to wrap a value?

chiaberry: ill change it back to the envelope being a string

moped-editor/src/components/GridTable/GridTable.js
```diff
@@ -222,17 +222,23 @@ const GridTable = ({ title, query }) => {
...
+      // Because of the way the SQL view concatanates the team members and roles
+      // " :" is always present even if there are no team members present
+      if (field === "project_team_members") {
+        value = '" :"';
+      } else {
+        value = '""';
+      }
```
sergiogcx, line 234: Very cool that it works 👍 ⚡ , although I do not love hard-coded special rules for the field `project_team_members`, GridTable needs to be re-usable an agnostic framework so we can use it in other projects.

One option could be to fix the SQL query so that it provides the output we need, or another option is to extend the javascript in a more agnostic way:

```
  string_is_null_sergios_super_special_rule: {
    operator: "_is_null",
    label: "is blank",
    description: "Special case rule",
    envelope: "true",
    specialNullValue: '" :"',   // This would then be passed down
    type: "string",
  },
  
```

Then we can use it like this in line 230, basically if there is a special null value specified then use it; otherwise, stick to the default. This way we can allow other fields to choose what kind of null value it is:

```
  value = specialNullValue ? specialNullValue : "";
```

sergiogcx (approved):
Thanks for working on this @chiaberry glad you made it work 👍 💯 ⚡

moped-editor/src/components/GridTable/GridTable.js
chiaberry: on the plus side, i thought of going down this path, so i know exactly what youre describing. 
on the minus side, i didnt do it this way so 😬 . ill update it too.

amenity (approved):
LSTM, @chiaberry ! Thanks.

mateoclarke (approved):
🚢

PR #343, opened 2021-06-09
Defines Moped Material Table theme/design customizations

tillyw:
Fully testable in [deploy preview](https://deploy-preview-343--atd-moped-main.netlify.app/)

Fixes https://github.com/cityofaustin/atd-data-tech/issues/5375 except for [Enable date-clearing with Material Table date picker #6159](https://github.com/cityofaustin/atd-data-tech/issues/6159). Suggest tackling that separately so it doesn't block the rest of the enhancements in this issue.

Some of the requirements were on date displays, which were not present in the Team table, so I incorporated those into the Timeline table.

Also sets pagination only when there are more than 25 rows.

Screen Shot 2021-06-17 at 5 37 33 PM
Screen Shot 2021-06-17 at 5 37 50 PM
Screen Shot 2021-06-17 at 5 38 08 PM


amenity (commented):
@tillyw - this looks fabulous!

One thing — are you applying extra styles to the h2 that we're using as the table title? I like the way it looks (like the comps!) but I wonder if it should be in a separate issue for global text/heading styles. 

These are the defaults I see on @chiaberry's style page (#345)
Screen Shot 2021-06-22 at 12 47 52 PM

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -371,12 +399,25 @@ const ProjectTeamTable = ({
...
           search: false,
           rowStyle: { fontFamily: typography.fontFamily },
           actionsColumnIndex: -1,
         }}
+        localization={{
+          header: {
+            actions: "",
+          },
+          body: {
+            emptyDataSourceMessage: <Typography variant="body1">No team members to display</Typography>,
```
johnclary, line 418: @tillyw would you add one of these for the Phases, Milestones, and Files tables? "No phases to display", "No milestones to display", "No files to display".

johnclary: errr, well—I guess we still need to apply all of these customizations to the other material table instances. separate issue @tillyw @amenity ?

johnclary (approved):
what a difference this makes! thank you @tillyw!

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
amenity: @johnclary @tillyw yes - sending [this into Slack](https://austininnovation.slack.com/archives/CNUEPKLB1/p1624434006172500).

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -363,6 +369,28 @@ const ProjectTeamTable = ({
...
+                  size="large"
+                  startIcon={<AddCircleIcon />}
+                  ref={addActionRef}
+                  onClick={props.action.onClick}
+                >
+                  Add team member
+                </Button>
+              );
+            }
+          },
```
sergiogcx, line 393: This is brilliant, TIL! 🙌

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -371,12 +399,25 @@ const ProjectTeamTable = ({
                 return personnel[item];
               })
         }
-        title="Project team"
+        title={
+          <Typography variant="h2" color="primary">
+            Project team
+          </Typography>
+        }
```
sergiogcx, line 406: Whoa, another one. I would have never imagined you could just pass your own component. 💯

sergiogcx (approved):
Very cool way to customize the table component, with @tillyw 's examples we can take it to another level. 😎 🥇 👍

tillyw:
> Very cool way to customize the table component, with @tillyw 's examples we can take it to another level. 😎 🥇 👍
> 
> 

Thanks @sergiogcx!

PR #348, opened 2021-06-17
Update Cognito account-creation emails to users #6219

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6219

** Can only be tested locally with postman, API change only.

From the [boto documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp.html#CognitoIdentityProvider.Client.admin_create_user): 

>MessageAction (string) -- Set to "RESEND" to resend the invitation message to a user that already exists and reset the expiration limit on the user's account. Set to "SUPPRESS" to suppress sending the message. Only one value can be specified.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp.html#CognitoIdentityProvider.Client.admin_create_user

### DIFF
--- a/moped-api/users/users.py
+++ b/moped-api/users/users.py
@@ -99,7 +99,7 @@ def user_create_user(claims: list) -> (Response, int):
                 UserPoolId=USER_POOL,
                 Username=email,
                 TemporaryPassword=password,
-                MessageAction="SUPPRESS"
+                MessageAction="SUPPRESS",
                 UserAttributes=[
                     {"Name": "email", "Value": email},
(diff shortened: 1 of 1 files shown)


sergiogcx:
I can confirm that the suppression is working, I've created an account for myself and received no email. Deleted, re-tried with the same email address and removed suppression and I did receive an email as expected.

sergiogcx:
So the email will no longer be submitted, but just in case it does we changed the template so that it will look like this with working links:

PR #349, opened 2021-06-17
Remove password reset UI from account page #6220

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6620

Live demo:
https://deploy-preview-349--atd-moped-main.netlify.app/moped/account

** UI change only, Can be tested in Netlify, Moped-test, or locally

In the code we simply remove ProfileDetails component from AccountView

PR #350, opened 2021-06-17
Patch account creation API to address bug with existing Cognito users #6372

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6372

A little late, but I was able to find a way to run the code and test it. I was able to sort out the logic in a way that it respects existing users in Cognito (re-using their Cognito ID) and to respect existing Database users in Hasura:

** Testing: This PR only works in Moped-Test environment only. Try creating a dummy account, then try creating the same account again using the same email, try using another department or other privileges. **Deleting a user works and actually deletes cognito users from Staging, careful.

https://user-images.githubusercontent.com/5282430/122988089-f83ffc80-d366-11eb-8d57-73491285a251.mp4


sergiogcx:
A little late, but I was able to find a way to run the code and test it. I was able to sort out the logic in a way that it respects existing users in Cognito (re-using their Cognito ID) and to respect existing Database users in Hasura:

Demo Run (not a real a

mateoclarke (approved):
🚢  👍 

Great work Sergio! I can see how this got complicated for something that seemed simple. It all looks good to me. One thing I noticed in my testing was that the password field doesn't really do anything. I expected it to change my password based on the new value I passed for an existing user when I submitted. Maybe we should `disabled=true` that field in situations where it truly is uneditable?

Looking at the original issue, maybe that is out of scope, but just wanted to highlight that. Not sure what the password reset UX flow is but maybe @amenity and you have already discussed.

amenity:
Thanks Sergio, it's great to get this addressed and I appreciate all the hard work. @mateoclarke - we were hoping that #349 would simplify things since we really want everyone to use their COA credentials in SSO.

amenity (approved):

PR #354, opened 2021-06-25
Prevent filter null failure when making new project

chiaberry:
Fixes bug introduced by breadcrumbs update

In theory object returned by useLocation() should always include state, but using the optional chaining there too, because one never knows.


sergiogcx (approved):
The fix worked! Thanks! 🙏 🚢 🚀

PR #356, opened 2021-06-28
Update Moped subcomponents with latest taxonomy #6526

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6526

** Database change only.

After the migration, this is what subcomponents looks like:


johnclary (approved):
ship

PR #357, opened 2021-06-29
Prevent edit/delete comment icons from appearing if not comment creator

chiaberry:
Database updates, so need to test on moped test or locally. 

Testing locally takes a bit of effort. I signed in and made some comments, signing in locally we are all JD. I changed the user_id in `src/auth/user.js` Line 104 to a different userid (ex, Mike Dilley is 8, Nathan Wilkes is 2). Sign out and sign back in and see the comments.  In this screenshot the user is Nathan. 

Screen Shot 2021-06-29 at 3 49 51 PM

I thought we had a similar issue in moped-test since when I logged in it says I am Patrick, but I see John has been able to leave comments as himself. 

edit -- now that I check for moped-admin status as well as who authored the comments, you have to sign in as the test-viewer (creds in 1password) to see if the icons are hidden. The user's highest role is based on whose password signed in, so changing the user_id in line 104 doesn't affect the role.

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -112,6 +112,7 @@ const ProjectComments = () => {
             project_note: DOMPurify.sanitize(noteText),
             project_id: projectId,
             status_id: 1,
+            added_by_user_id: Number(userSessionData.user_id),
           },
         ],
       },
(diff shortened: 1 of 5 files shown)


patrickm02L:
I'm having the same issue with login. When I use the SSO, it thinks I'm Chia.

chiaberry:
How many comments are in production? My migration would make all those comments have the "added_by_user_id" be 1, right?

chiaberry:
I need to let admins edit / delete any comment, standby....

chiaberry:
I updated the code so it checks if the logged in user is moped-admin. I used moped-test-viewer's account to test locally.

moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -203,28 +206,33 @@ const ProjectComments = () => {
...
+                                <IconButton
+                                  edge="end"
+                                  aria-label="edit"
+                                  onClick={() =>
+                                    editComment(i, item.project_note_id)
+                                  }
+                                >
+                                  <EditIcon />
+                                </IconButton>
+                              )}
```
sergiogcx, line 224: 👍 This is a good start!

sergiogcx (approved):
In the future we will want to make sure we have row-level security using advanced Hasura permissions, but this is a great start! 📓 🚢

chiaberry:
@sergiogcx can you elaborate on the row-level security? Does that mean to include check before executing the mutation that the user that initiated the mutation has the correct userid and permission level?

sergiogcx:
> @sergiogcx can you elaborate on the row-level security? Does that mean to include check before executing the mutation that the user that initiated the mutation has the correct userid and permission level?

Sure thing, and yes you got it. With this mechanism there is nothing stopping the proverbial hacker from deleting anyone's comments by simply manipulating the graphql queries. 

And yes, we would make use of Hasura's permission check mechanism:
https://hasura.io/docs/latest/graphql/core/auth/authorization/permission-rules.html#row-level-permissions

I'd be happy to pair up with you and show you how we would get our values to work.

moped-editor/src/views/projects/projectView/ProjectComments.js
tillyw, line 224: Love this!

tillyw (approved):
Tested this locally by changing user id in auth and logging as test-editor and it worked as expected!

patrickm02L (approved):

PR #359, opened 2021-07-06
[Project Components] Zoom to extent of all components when creating new component #6574

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6574

** Front-end changes only, you can use Netlify, moped-test or locally.

Live Preview:
https://deploy-preview-359--atd-moped-main.netlify.app/moped/session/signin

This PR updates the components editor so that whenever a new component is added, it initializes the map to start in the same area as the rest of the project.

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useQuery } from "@apollo/client";
 import { useParams } from "react-router-dom";
 import { makeStyles } from "@material-ui/core/styles";
@@ -61,6 +61,7 @@ const ProjectComponents = () => {
   const { projectId } = useParams();
   const classes = useStyles();
 
+  const [featureFullCollection, setFeatureFullCollection] = useState(null);
   const [selectedComp, setSelectedComp] = useState(0);
   const [mapError, setMapError] = useState(false);
   const [componentEditMode, setComponentEditMode] = useState(false);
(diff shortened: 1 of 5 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents.js
```diff
@@ -71,29 +72,30 @@ const ProjectComponents = () => {
...
+                    ...featureComponent.moped_proj_feature,
+                    project_features_components_id:
+                      featureComponent.project_features_components_id,
+                  })
+                )
+              : []),
+          ],
+          []
+        )
+      : [];
```
sergiogcx, line 98: This part looks scarier, but the only thing I did was to add the line `data && data?.moped_proj_components`, this is because it was no longer protected by `if(loading)` or `if(error)` which had to be moved down because useEffect cannot be used conditionally.

johnclary:
@sergiogcx this works perfectly! one request—would you mind passing in a max zoom to prevent the map from zooming super close (especially for point features)?

I believe we can make this change globally adding the `maxZoom` option ([docs](https://visgl.github.io/react-map-gl/docs/api-reference/web-mercator-viewport#fitboundsbounds-options-object)) when calling `featureViewport.fitBounds()` right [here](https://github.com/cityofaustin/atd-moped/blob/a8fbf0b696dfbd8a7a7793e9a339e951db950b6d/moped-editor/src/utils/mapHelpers.js#L696).

So:
```javascript
featureViewport.fitBounds(mapBounds, {
        padding: 100,
        maxZoom: 16
});
```

The maxZoom of 16 should be good IMO.

If this sends you down a rabbit hole—forget it. We can tackle it later.

johnclary:
Ah, one issue: I'm getting the WSOD when I try to add a component to a project that doesn't have one. I think this is an issue with the deploy env, right?

sergiogcx:
> @sergiogcx this works perfectly! one request—would you mind passing in a max zoom to prevent the map from zooming super close (especially for point features)?
> 
> I believe we can make this change globally by adding the `maxZoom` option ([docs](https://visgl.github.io/react-map-gl/docs/api-reference/web-mercator-viewport#fitboundsbounds-options-object)) when calling `featureViewport.fitBounds()` right [here](https://github.com/cityofaustin/atd-moped/blob/a8fbf0b696dfbd8a7a7793e9a339e951db950b6d/moped-editor/src/utils/mapHelpers.js#L696).
> 
> So:
> 
> ```js
> featureViewport.fitBounds(mapBounds, {
>         padding: 100,
>         maxZoom: 16
> });
> ```
> 
> The maxZoom of 16 should be good IMO.
> 
> If this sends you down a rabbit hole—forget it. We can tackle it later.

Sweet! Thank you for that snippet, let me patch it right away. Hmm, this time it looks like a legitimate but. I'll 2x-check and sort it out.

amenity (approved):
LSTM ✨ 🙌🏻

johnclary:
Nailed it! Thanks!

johnclary (approved):
🚢 🗺️

PR #360, opened 2021-07-06
Applies Material Table customizations used on Team table to other instances

tillyw:
Fully testable in [deploy preview environment](https://deploy-preview-360--atd-moped-main.netlify.app/).

Fixes [Apply Material Table customizations used on Team table to other instances #6294](https://github.com/cityofaustin/atd-data-tech/issues/6294)
Fixes [Rename "is_current_phase" column from "Active?" to "Current" #6617](https://github.com/cityofaustin/atd-data-tech/issues/6617)

Applies all requirements in [Define Moped Material Table theme/design customizations #5375](https://github.com/cityofaustin/atd-data-tech/issues/5375) to Phases, Milestone and Files tables. I decided not to pursue any further abstraction at this stage since this issue encompasses a lot of changes and I'd propose pursuing that in a separate issue if that is something we'd like to do.

Includes: 
 - @tillyw's form validation fix
- Tilly's moving action buttons to the far right
 - Add row button should have a specific text label — see @mateoclarke's implementation in the "Project phases" component
 - Add row button should remain in upper right
 - Add row button should have a circle-plus icon instead of a square-plus icon (see "Projects" page)
 - Hide/remove "Actions" column heading
 - Show pagination only when there are more than 25 rows.
 - Render dates as mm/dd/yyyy
 - Make the "No records to display" text context-dependent, e.g. "No team members to display" on the Team tab.
 Remove min-height on table
-  Change table title from h6 to h2
- Renames "Active?" column in Phases table to "Current Phase"
- I also moved "Current Phase" column to the end so that the table fits more aesthetically with Milestones table (date columns are aligned, Current Phase and Completed columns are at the ends of their respective tables)

Does not include:
- [Enable date-clearing on calendar picker Enable date-clearing with Material Table date picker #6159](https://github.com/cityofaustin/atd-data-tech/issues/6159)

Screen Shot 2021-07-07 at 3 52 19 PM
Screen Shot 2021-07-07 at 4 00 26 PM
Screen Shot 2021-07-07 at 3 53 54 PM


amenity (approved):
LSTM! So fresh and so clean. 🧼 ✨

chiaberry (approved):

tillyw:
Hi all, just want to check in on review status for this issue

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -194,37 +206,76 @@ const ProjectFiles = props => {
...
+                    Add file
+                  </Button>
+                );
+              }
+            },
+          }}
           icons={{ Delete: ClearIcon }}
           options={{
-            search: true,
+            ...(data.moped_project_files.length < 26 && {
```
mateoclarke, line 258: Took me a second to understand why this was 26. In general, random hardcoded values should be avoided when possible. I wonder if we could use a variable that is set somewhere else as the paging default. If not, even defining a constant as...

```js
const PAGING_DEFAULT_COUNT = 25;

...
# inside options prop
...(data.moped_project_files.length < PAGING_DEFAULT_COUNT + 1 && {
```
maybe thats easier to read?

mateoclarke (approved):
🚢

mateoclarke:
> Hi all, just want to check in on review status for this issue

Hey Tilly, just wanted to make sure you know you're free to merge with one dev and one product review. Unless I'm missing something about why you were checking on review status.

moped-editor/src/views/projects/projectView/ProjectFiles.js
tillyw: Thanks for pointing this out! Created a tables file to store this value so it can be changed globally and placed it in a new constants folder where we can store other values like this.

mateoclarke: This concept in software development is called "magic numbers" 
https://refactoring.guru/replace-magic-number-with-symbolic-constant

chiaberry: the mascot for the refactoring guru site is adorable. thanks for sharing the link

PR #364, opened 2021-07-13
[Bug] Previous 'Subtype' stays in text box after selecting new 'Type' #6616

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6616

Description:
  Refactors subtype & subcomponent logic and removes duplicates

** Front-end change only, test with Netlify

Live Demo:
https://deploy-preview-364--atd-moped-main.netlify.app/moped/session/signin

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
@@ -262,11 +262,11 @@ const ProjectComponentEdit = ({
     const selectedType = (newValue ?? e.target.value ?? "").toLowerCase();
 
     // Generates a list of available component subtypes given a component type
-    const availableSubTypes = getAvailableSubtypes(selectedType);
+    const newAvailableSubTypes = getAvailableSubtypes(selectedType);
 
     // Set The selected component type
     setSelectedComponentType(selectedType);
-    setAvailableSubtypes(availableSubTypes);
+    setAvailableSubtypes(newAvailableSubTypes);
     setSelectedComponentSubtype(null);
   };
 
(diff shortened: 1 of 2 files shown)


chiaberry:
I'm still seeing the bug: 

recording

sergiogcx:
> I'm still seeing the bug:
> 
> recording

Thank you @chiaberry I guess I did not understand the bug entirely... haha, removing reviewers for now until I fix for real.

sergiogcx:
Ready for review again, thanks @chiaberry 🙏

patrickm02L (approved):
Tried every combination I could think of to break the Types and Subtypes from holding their previous designation, but couldn't. Nice fix! Ship it.

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
```diff
@@ -262,11 +262,11 @@ const ProjectComponentEdit = ({
     const selectedType = (newValue ?? e.target.value ?? "").toLowerCase();
 
     // Generates a list of available component subtypes given a component type
-    const availableSubTypes = getAvailableSubtypes(selectedType);
+    const newAvailableSubTypes = getAvailableSubtypes(selectedType);
```
chiaberry, line 265: ah, was it conflicting with state?

sergiogcx: Thankfully not, but it made it difficult to troubleshoot because the names were nearly identical.

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
```diff
@@ -694,10 +694,12 @@ const ProjectComponentEdit = ({
...
-                  value={availableSubtypes.find(
-                    subtype =>
-                      subtype.toLowerCase() === selectedComponentSubtype
-                  )}
+                  value={
+                    availableSubtypes.find(
+                      subtype =>
+                        subtype.toLowerCase() === selectedComponentSubtype
+                    ) ?? null
+                  }
```
sergiogcx, line 702: This was the real fix for subtype, for some reason the value stays the same if `.find` returns `undefined`, but forcing it to return null instead clears the field for good.

chiaberry: interesting, it wasnt triggering a refresh if it was undefined? in any case, thank you for the explanations!

PR #367, opened 2021-07-15
Change capital funding switch component color from red to blue

jgabitto:
- Closes https://github.com/cityofaustin/atd-data-tech/issues/5424
- Add color prop to primary
- Change aria label prop from secondary to primary

### DIFF
--- a/moped-editor/.eslintrc.json
+++ b/moped-editor/.eslintrc.json
@@ -1,3 +1,4 @@
 {
-  "extends": ["plugin:prettier/recommended"]
+  "extends": ["plugin:prettier/recommended"],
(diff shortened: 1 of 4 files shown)


sergiogcx (approved):
Thanks for taking care of this one! 📦 🚀 🚢 💅

patrickm02L (approved):
Cool. Good work. Ship It!

mateoclarke (approved):
There are some updates the the linting config, package.json changes supported browsers and chart.js upgrades its version number. I'm not sure if these were intentional changes. They don't seem harmful but they seem out of scope of this update.

Either way, I think this is good to merge.

PR #373, opened 2021-07-23
Update moped theme and remove backgroundColor override

jgabitto:
- Closes https://github.com/cityofaustin/atd-data-tech/issues/6712
- Update properties in theme object
- Remove backgroundColor property from classes.root


sergiogcx:
@jgabitto It looks like it's not building because the colors variable is not being used, that needs to be fixed:

jgabitto:
> @jgabitto It looks like it's not building because the colors variable is not being used, that needs to be fixed:
> 
> 

Ok removed the colors import

sergiogcx (approved):
Looks nice! Thanks for fixing those issues. 🚢 🚀 📦

amenity:
Looks great! Thanks, @jgabitto.🚢

PR #374, opened 2021-07-23
Comment out typography import and typography property

jgabitto:
- Tees up https://github.com/cityofaustin/atd-data-tech/issues/6713
- Comment out typography import
- Comment out typography property in theme object

### DIFF
--- a/moped-editor/src/theme/index.js
+++ b/moped-editor/src/theme/index.js
@@ -1,6 +1,6 @@
 import { createMuiTheme, colors } from "@material-ui/core";
 import shadows from "./shadows";
-import typography from "./typography";
+// import typography from "./typography";
 
 const theme = createMuiTheme({
   palette: {
@@ -27,7 +27,7 @@ const theme = createMuiTheme({
     },
   },
   shadows,
-  typography,
+  // typography,
 });
 
 export default theme;


moped-editor/src/theme/index.js
```diff
@@ -27,7 +27,7 @@ const theme = createMuiTheme({
     },
   },
   shadows,
-  typography,
+  // typography,
```
sergiogcx, line 30: If topography does not need to be there, then best to remove the comments too.

sergiogcx:
@jgabitto @amenity does this seem right? The title looks disproportionally large:

amenity:
@jgabitto - yep, it's tough to stomach but this is what I asked for.

amenity:
Let's keep this open until we get other typography issues lined up to resolve it.

moped-editor/src/theme/index.js
mateoclarke: agreed

mateoclarke (commented):
I agree with Sergio, instead of commenting out a line of code, you can always just delete it. If you need the old code back, it lives in the git history. That's the beauty of version control. Otherwise you end up with a bunch of edited out lines of code that you're not sure why they are there years down the line.

amenity:
Just want to reiterate — this should not be merged into master!

amenity:
In fact I think the PR can be closed — what we really wanted was the screenshots.

PR #386, opened 2021-08-04
Renders user avatar images in activity feed #4809

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4809

* Front-end change only:
Live Demo: https://363-sg-avatars--atd-moped-main.netlify.app/moped/projects/273?tab=activity_log

Refactors the CDNAvatar component and it is applied to the activity log.

### DIFF
--- a/moped-editor/src/components/CDN/Avatar.js
+++ b/moped-editor/src/components/CDN/Avatar.js
@@ -1,35 +1,63 @@
 import React from "react";
 import { Avatar, Icon, makeStyles, Typography } from "@material-ui/core";
-import { useUser } from "../../auth/user";
 
 import config from "../../config";
 
-const useStyles = makeStyles(() => ({
+const useStyles = makeStyles(theme => ({
   root: {},
   avatar: {
     height: 100,
     width: 100,
     marginBottom: 8,
+    backgroundColor: theme.palette.grey["300"],
   },
   userInitials: {
     fontSize: "2rem",
   },
 }));
 
-const CDNAvatar = props => {
-  const classes = useStyles();
-  const { user } = useUser();
 
-  const imageSrc = props?.src ? `${config.env.APP_CLOUDFRONT}/${props.src}` : null
+/**
+ * A wrapper for Material's Avatar component.
+ * @param {Object} className - The class name override
+ * @param {String} src - The image in the CDN
+ * @param {String} initials - The initials to render within the avatar (when no image is present).
+ * @param {boolean} largeInitials - If true, it makes the initials a little larger.
+ * @param {String} userColor - The background color to render under the initials.
+ * @return {JSX.Element}
+ * @constructor
+ */
+const CDNAvatar = ({className, src, initials, largeInitials, userColor}) => {
+  const classes = useStyles();
+
+  /**
+   * The image source from CloudFront CDN
+   * @type {string|null}
+   */
+  const imageSource = src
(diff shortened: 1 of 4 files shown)


moped-editor/src/components/CDN/Avatar.js
```diff
@@ -1,36 +1,72 @@
...
+ * @constructor
+ */
+const CDNAvatar = ({
+  className,
+  src,
+  initials,
+  largeInitials,
+  userColor,
+  useGenericAvatar,
+}) => {
```
sergiogcx, line 37: For starters, here we deconstruct the props and we look for specific fields.

moped-editor/src/components/CDN/Avatar.js
```diff
@@ -1,36 +1,72 @@
...
-      <Typography className={props?.largeInitials ? classes.userInitials : null}>
-        {props?.initials ? props.initials : <Icon>user</Icon>}
-      </Typography>
+      {!!!useGenericAvatar && initials && initials.length > 0 ? (
+        <Typography className={largeInitials ? classes.userInitials : null}>
+          {initials}
+        </Typography>
+      ) : (
+        <Icon>person</Icon>
+      )}
```
sergiogcx, line 69: Basically the behavior is:
1. If there is a source file, use it to get it from the CDN.
2. If there is no src, then check if there is a generic avatar override.
3. If there is no override, then use the initials (if they are provided).
4. If there is an override or there are no initials, then use the Person icon.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -232,13 +232,13 @@ const ProjectActivityLog = () => {
...
+                            <CDNAvatar
                               className={classes.avatarSmall}
-                            >
-                              {getInitials(change?.moped_user)}
-                            </Avatar>
+                              src={change?.moped_user?.picture}
+                              initials={getInitials(change?.moped_user)}
+                              userColor={null}
+                              useGenericAvatar={true}
+                            />
```
sergiogcx, line 241: Replace Avatar with our custom component, passing all the necessary fields and overrides.

moped-editor/src/components/CDN/Avatar.js
```diff
@@ -1,36 +1,72 @@
...
-      style={{ backgroundColor: imageSrc ? "white" : user?.userColor }}
+      alt={initials}
+      className={className ? className : classes.avatar}
+      src={imageSource}
+      style={imageStyleOverride}
     >
-      <Typography className={props?.largeInitials ? classes.userInitials : null}>
-        {props?.initials ? props.initials : <Icon>user</Icon>}
-      </Typography>
+      {!!!useGenericAvatar && initials && initials.length > 0 ? (
```
tillyw, line 63: Never seen three "not" operators used in sequence, really interesteing!

sergiogcx: Basically a "negative truthy" if you will. Essentially we are not assuming useGenericAvatar is provided, so if it isn't we turn the value into a boolean which we then turn into its opposite value using the third "not" operator.

tillyw (approved):
Works great and the code makes sense to me! Thanks for solving this one, @sergiogcx

patrickm02L (approved):
Looks good. Ship it!

PR #391, opened 2021-08-12
Patch Heroku Deployment Script

sergiogcx:
This patch should prevent PostGIS from being installed in moped-test. This is only a problem in Heroku, because PostGIS comes with +5,000 rows, which puts us close to the 10k limit on the free tier and makes us get a bunch of annoying emails.

Tested to work, now the total amount of rows is 1500 (down from 7000+):


amenity:
@sergiogcx  - sounds good! I unassigned myself bc I don't feel qualified to review. ;) Maybe Chia/Tilly when they're back?

PR #394, opened 2021-08-13
[Bug] Staff WSOD

sergiogcx:
Attempts to address the WSOD in the Staff list:

### DIFF
--- a/moped-editor/src/components/GridTable/GridTable.js
+++ b/moped-editor/src/components/GridTable/GridTable.js
@@ -428,7 +428,7 @@ const GridTable = ({ title, query }) => {
         {/*Toolbar Space*/}
         <GridTableToolbar>
           <GridTableSearch
-            projectsList={data}
+            parentData={data}
             query={query}
             searchState={{
               searchParameters: search,
--- a/moped-editor/src/components/GridTable/GridTableSearch.js
+++ b/moped-editor/src/components/GridTable/GridTableSearch.js
@@ -325,7 +325,7 @@ const GridTableSearch = ({
                 <Button
                   disabled={
                     projectsList &&
-                    !projectsList["project_list_view"].length &&
+                    !(projectsList?.project_list_view ?? []).length &&
                     true
                   }
                   className={classes.downloadCsvButton}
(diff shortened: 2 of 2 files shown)


moped-editor/src/components/GridTable/GridTableSearch.js
```diff
@@ -324,9 +325,7 @@ const GridTableSearch = ({
               {query.config.showExport && (
                 <Button
                   disabled={
-                    projectsList &&
-                    !projectsList["project_list_view"].length &&
-                    true
```
sergiogcx, line 329: It looks like we had a fixed table name under a dynamic component, it works fine under the projects list, but as this framework (GridTable and GridTableSearch) is used in other places like the Staff page, the keys would will not exist creating a WSOD.

moped-editor/src/components/GridTable/GridTableSearch.js
```diff
@@ -324,9 +325,7 @@ const GridTableSearch = ({
...
+                    (parentData?.[query.config.table] ?? []).length === 0
```
sergiogcx, line 328: So the fix is quite simple, basically pass the name of the table in question to the array, then if provided calculate its length, and if it is zero then assume it's disabled.

moped-editor/src/components/GridTable/GridTable.js
```diff
@@ -428,7 +428,7 @@ const GridTable = ({ title, query }) => {
         {/*Toolbar Space*/}
         <GridTableToolbar>
           <GridTableSearch
-            projectsList={data}
+            parentData={data}
```
sergiogcx, line 431: Renamed to a generic type, the framework should not care what type it is.

mateoclarke (approved):
Good catch @sergiogcx! Instead of merging to `main`, can you point this PR agaisnt `v0.8.0-release-candidate`? 🚢

moped-editor/src/components/GridTable/GridTableSearch.js
jgabitto: That makes a lot of sense!

jgabitto (approved):
The changes you have suggested make sense as this is a reusable component!

PR #396, opened 2021-08-16
Update header size and hierachy

chiaberry:


patrickm02L:
@amenity I think you should review this one. I feel like you have more of a handle on the typography work.

PR #402, opened 2021-08-19
Resolves component geometry gaps - #6761

johnclary:
~Fixes~ Substantially addresses [#6761](https://github.com/cityofaustin/atd-data-tech/issues/6761).


moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -297,8 +322,12 @@ const NewProjectMap = ({
...
     <Box className={noPadding ? classes.mapBoxNoPadding : classes.mapBox}>
       {/* These two lines act as a conditional global override of MapBox. */}
-      {!visibleLayerIds.includes("drawnByUser") && <NewProjectDrawnPointsInvisibleStyle/>}
-      {!visibleLayerIds.includes("drawnByUserLine") && <NewProjectDrawnLinesInvisibleStyle/>}
+      {!visibleLayerIds.includes("drawnByUser") && (
+        <NewProjectDrawnPointsInvisibleStyle />
+      )}
+      {!visibleLayerIds.includes("drawnByUserLine") && (
+        <NewProjectDrawnLinesInvisibleStyle />
+      )}
```
johnclary, line 330: ☝️ linting

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -370,7 +402,7 @@ const NewProjectMap = ({
               key={config.layerIdName}
               type="vector"
               tiles={[config.layerUrl]}
-              maxzoom={config.layerMaxLOD || mapConfig.mapboxDefaultMaxZoom} // maxLOD found in vector tile layer metadata
+              maxZoom={config.layerMaxLOD || mapConfig.mapboxDefaultMaxZoom} // maxLOD found in vector tile layer metadata
```
johnclary, line 405: i noticed this typo in the maxZoom prop name. did not have any effect on UI as far as i can tell.

johnclary:
thanks @frankhereford for the tip to use turfjs/combine. that does the magic of making sure the split features get stitched together in the right order. there is a risk that we will "miss" a fragment of a split feature if it is outside the current map viewport (ie it is not currently rendered on the map). the alternative is to specify our own bounding box to use when searching for feature fragments—but how large should we make the bbox?

johnclary:
Proof that the entire component geometry as saved — compare to images in [#6761](https://github.com/cityofaustin/atd-data-tech/issues/6761)
Screen Shot 2021-08-19 at 5 04 13 PM
Screen Shot 2021-08-19 at 5 03 19 PM

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -216,8 +217,35 @@ const NewProjectMap = ({
...
+      // The alternative is to supply our own search geometry bbox
+      const renderedFeatures = mapRef.current.queryRenderedFeatures();
+
+      // It's possible to pass a filter function to queryRenderedFeatures. We
+      // could theoretically filter for features that match the ID of interest, but
+      // I could not get the expression to work. I doubt the performance gain would
+      // be significant vs our own filter call.
+      const splitFeatures = renderedFeatures.filter(
+        feature => getFeatureId(feature, layerName) === selectedFeatureId
+      );
```
sergiogcx, line 241: @johnclary Ah, this is a gem! 💯 💎 It didn't occur to me to search the rest of the feature like this. The bbox would help performance for sure.

sergiogcx (approved):
Brilliant!

mateoclarke (approved):
line_feature

amenity (approved):
Look ma, no gaps!!! 
Screen Shot 2021-08-19 at 6 34 01 PM
😊

johnclary:
[Proof](https://deploy-preview-402--atd-moped-main.netlify.app/moped/projects/290?tab=map) that this fix is not perfect. Large features are still prone to having missed fragments.

I was just testing selecting features with the map zoomed out to the entire extent of the city, and it was surprisingly performant despite running a filter operation against every map feature. I'll test out passing a bound box the feature query that encompasses the whole city. If it's lagging, we can set the bounding box size dynamically based on zoom (or something).

frankhereford (approved):
From looking at the diffs, I think this looks great. I certainly don't have the familiarity with the larger code base that the rest of the team has, so please take my thoughts with a grain of salt, but I certainly don't see anything that looks out of order. 

Nice solution to a tricky problem!

johnclary:
I did further testing to see what our options are for identifying feature fragments which are outside of the current viewport when a feature is selected.

First, I tried to pass our geocoder's bounding box to [`queryRenderedFeatures`](https://docs.mapbox.com/mapbox-gl-js/api/map/#map#queryrenderedfeatures). I thought this would query all features in Austin, but the bbox will only filter from currently rendered features. In other words, we are restricted to querying from features that already rendered (ok, duh, hence the method name).

So I moved on to [`querySourceFeatures`](https://docs.mapbox.com/mapbox-gl-js/api/map/#map#querysourcefeatures). Unfortunately, this method has a similar limitation:
> this function does not check tiles outside the currently visible viewport

Ok, this all makes sense. It's not possible to query features outside the currently rendered tiles, because this tile data is not held in memory (although it's surely caching).

If we want to solve this, the most straightforward solution I can think of is to query an external data source on-the-fly (e.g. ArcGIS Online) for the complete geometry of the clicked feature.

An another alternative would be to load the CTN geometries into the moped database and lookup/copy that geometry into the component record when a user selects it. We would then need to worry about keeping the CTN geometries in Moped in sync w/ the original source in ArcGIS.

Worth noting that although this PR doesn't completely fix the problem, it opens up a work around: the user needs to zoom out until the highlighted feature is visible on the map before selecting it. 

My suggestion would be to (1) merge this PR and (2) open a separate spike to test out querying feature geometries on-the-fly from AGOL.

johnclary:
[Related issue](https://github.com/cityofaustin/atd-data-tech/issues/7012) to fetch complete geoms from AGOL.

amenity:
Nice work! No gaps coming up for me on loooong selected stretches.

PR #403, opened 2021-08-20
Demoing GridTable with no max height

johnclary:
No issue here—just curious to see max height removed.


johnclary:
Closing—will open PR to specifically address table height and not search bar/filter sizing.

PR #405, opened 2021-08-23
Fix bug preventing changing project name

chiaberry:
Closes cityofaustin/atd-data-tech#7010

In a PEMDAS-type code error, I was incorrectly checking if a title update was empty. 
I wanted to check that `projectName.trim()===""` was not true. But `!projectName.trim()===""` was evaluating as `!projectName.trim()` equals `""`

Test here: https://7010-change-project-name--atd-moped-main.netlify.app/ by changing a project's name and saving.


amenity (approved):
You got it, @chiaberry 🙏

sergiogcx (approved):
🚢 🚀

PR #406, opened 2021-08-24
Fetch line geometries from ArcGIS Online feature service

johnclary:
WIP

### DIFF
--- a/moped-editor/src/utils/mapHelpers.js
+++ b/moped-editor/src/utils/mapHelpers.js
@@ -66,6 +66,12 @@ const basemaps = {
   },
 };
 
+/**
+ * ArcGIS Online feature service endpoint of the CTN line segments. Managed by DTS GIS team.
+ */
+const ctnAGOLEndpoint =
+  "https://services.arcgis.com/0L95CJ0VTaxqcmED/ArcGIS/rest/services/Moped_CTN_Segments_08_23_2021/FeatureServer/0";
+
 export const mapStyles = {
   statusOpacities: {
     selected: 0.75,
@@ -130,7 +136,7 @@ export const mapConfig = {
       layerIdName: "ctn-lines",
       layerIdField: "PROJECT_EXTENT_ID",
       tooltipTextProperty: "FULL_STREET_NAME",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 1,
       layerColor: theme.palette.primary.main,
       layerUrl:
@@ -177,7 +183,7 @@ export const mapConfig = {
       layerLabel: "Points",
       layerIdName: "project-component-points",
       layerIdField: "PROJECT_EXTENT_ID",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 2,
       layerColor: theme.palette.secondary.main,
       layerUrl:
@@ -216,7 +222,7 @@ export const mapConfig = {
       layerLabel: "Drawn Points",
       layerIdName: "drawnByUser",
       layerIdField: "PROJECT_EXTENT_ID",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 3,
       layerColor: theme.palette.secondary.main,
       layerMaxLOD: 12,
@@ -239,7 +245,7 @@ export const mapConfig = {
       layerLabel: "Drawn Lines",
       layerIdName: "drawnByUserLine",
       layerIdField: "PROJECT_EXTENT_ID",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 4,
       layerColor: theme.palette.primary.main,
       layerMaxLOD: 12,
@@ -265,7 +271,7 @@ export const mapConfig = {
       layerLabel: "Project Features",
       layerIdName: "projectFeatures",
       layerIdField: "PROJECT_EXTENT_ID",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 5,
       layerColor: theme.palette.grey["800"],
       layerMaxLOD: 12,
@@ -298,7 +304,7 @@ export const mapConfig = {
       layerLabel: "Project Points",
       layerIdName: "projectFeaturePoints",
       layerIdField: "PROJECT_EXTENT_ID",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
       layerOrder: 6,
       layerColor: theme.palette.grey["800"],
       layerMaxLOD: 12,
@@ -385,7 +391,7 @@ export const getLayerNames = () => Object.keys(mapConfig.layerConfigs);
  * @return {String} The ID of the polygon clicked or hovered
  */
 export const getFeatureId = (feature, layerName) =>
-  get(feature, mapConfig.layerConfigs[layerName].layerIdGetPath);
+  get(feature.properties, mapConfig.layerConfigs[layerName].featureIdProperty);
 
 /**
  * Get a feature's property that contains text to show in a tooltip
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -111,6 +111,32 @@ export const useStyles = makeStyles(theme => ({
   ...layerSelectStyles,
 }));
 
+// todo: document
+const handleFeatureCollectionUpdate = (
```
johnclary, line 115: nothing new here—just isolating this block

moped-editor/src/layouts/DashboardLayout/NavBar/MobileDropdownMenu.js
```diff
@@ -69,6 +69,7 @@ const MobileDropdownMenu = () => {
       >
         {navigationItems.map(item => (
           <MenuItem
+            key={item.href}
```
johnclary, line 72: addresses: `ESLint Error: Each child in a list should have a unique "key" prop`

moped-editor/src/utils/mapHelpers.js
```diff
@@ -128,9 +134,8 @@ export const mapConfig = {
     CTN: {
       layerLabel: "Streets",
       layerIdName: "ctn-lines",
-      layerIdField: "PROJECT_EXTENT_ID",
       tooltipTextProperty: "FULL_STREET_NAME",
-      layerIdGetPath: "properties.PROJECT_EXTENT_ID",
+      featureIdProperty: "PROJECT_EXTENT_ID",
```
johnclary, line 138: `layerID` was throwing me off, so I standardized around `featureId`. I think this term added to the the confusion that led us to have a redundant `layerIdField` prop.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -111,6 +114,48 @@ export const useStyles = makeStyles(theme => ({
   ...layerSelectStyles,
 }));
 
+const removeFeatureFromCollection = (selectedFeature, featureCollection) => {
```
johnclary, line 117: Nothing really new in these three functions—i just isolated the logic for updating the feature collection because we need to call it conditionally based on if/how we address fragmented features

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -203,6 +247,81 @@ const NewProjectMap = ({
...
+    }
+    // this feature is rendered to the edge of the viewport and may be fragmented. we cannot
+    // know with certainty, so we fetch it's complete geometry from AGOL to be safe.
+    // also considered turf.booleanWithin() and turf.booleanContains - not reliable.
+    const selectedFeatureId = getFeatureId(
+      selectedFeature,
+      selectedFeature.properties.sourceLayer
+    );
+
+    queryCtnFeatureService(selectedFeatureId).then(queriedFeatureCollection => {
```
johnclary, line 303: here's the AGOL API call—and the performance hit while fetch occurs. in a nutshell, we only query AGOL if 1) the selected feature is a CTN line and 2) it is touching the edge of the viewport and 3) the feature is being added (not removed) from the feature collection state.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -218,52 +337,9 @@ const NewProjectMap = ({
     if (!layerName || !getClickEditableLayerNames().includes(layerName)) return;
 
     let selectedFeature = getGeoJSON(e);
-    const selectedFeatureId = getFeatureId(selectedFeature, layerName);
```
johnclary, line 221: trashing the previous the attempt to re-combine split segments :/

moped-editor/src/views/projects/projectView/ProjectComponentsMapView.js
```diff
@@ -280,9 +280,9 @@ const ProjectComponentsMapView = ({
         mapStyle={mapStyleConfig}
       >
         {/* Draw Navigation controls with specific styles */}
-        <div className={classes.navStyle}>
```
johnclary, line 283: addresses an invalid DOM nesting linter error. i tested without noticing any regression.

moped-editor/src/views/projects/projectView/ProjectSummary.js
```diff
@@ -3,7 +3,6 @@ import { useParams } from "react-router-dom";
 
 import ProjectSummaryTable from "./ProjectSummaryTable";
 import ProjectSummaryMap from "./ProjectSummaryMap";
-import ProjectSummaryEditMap from "./ProjectSummaryEditMap";
```
johnclary, line 6: Removing the editing state management from the ProjectSummary page, since that functionality moved to the Map tab. And you'll see further below that the ProjectSummaryEditMap can be deleted. @sergiogcx that's fine, right?

johnclary:
If y'all would kindly test out the component editor—try zooming in close to long segments (there are many near the airport). I'm curious to hear how you feel about the performance. The AGOL fetch only happens in those unique cases so you can watch the AGOL request in your dev tools to confirm.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -203,6 +247,81 @@ const NewProjectMap = ({
     saveActionDispatch
   );
 
+  /**
+   * Handles when a new map feature is selected and updates feature collection state
+   * @param {Object} selectedFeature - The selected feature geojson
+   * @return {undefined} Side effect of calling setFeatureCollection with new state
+   */
+  const handleSelectedFeatureUpdate = selectedFeature => {
```
johnclary, line 255: this is a lot of code to run on every render—I could hoist it out of the component? `useCallback` might by the right choice?

moped-editor/src/utils/mapHelpers.js
sergiogcx: @johnclary The `layerIDGetPath` variable is meant to be used with Lodash, it's [basically a way to provide the path of an attribute in an object as a string for access](https://lodash.com/docs/4.17.15#get). 

The name of the variable is trivial but personally I'd appreciate having the word "path" to remind me what its purpose is and its relationship to Lodash, or at least a comment near the new name `featureIdProperty` that can help us in the future. 🙏 Regardless, it looks like the refactor is working.

moped-editor/src/utils/mapHelpers.js
```diff
@@ -635,15 +632,15 @@ export const createProjectViewLayerConfig = (
...
       className={className}
       style={{
         left: hoveredCoords?.x,
         top: hoveredCoords?.y,
       }}
     >
-      <div>{tooltipText.toLowerCase()}</div>
-    </div>
+      <span>{tooltipText.toLowerCase()}</span>
+    </span>
```
sergiogcx, line 643: Really liking these changes, it never occurred to me to try using a span tag instead. 🙌

moped-editor/src/utils/mapHelpers.js
```diff
@@ -1015,24 +1012,28 @@ export const useSaveActionReducer = () => {
...
+    .map(param => `${param[0]}=${encodeURIComponent(param[1])}`)
+    .join("&");
+  const url = `${ctnAGOLEndpoint}/query?${paramString}`;
+
+  return await fetch(url)
+    .then(response => response.json())
+    .catch(err => {
+      console.log("Error fetching geometry: " + JSON.stringify(err));
+      return null;
+    });
```
sergiogcx, line 1038: Brilliant! It's an amazing idea! 💯 🗺️

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
sergiogcx: @johnclary we should try useCallback, if the code really gets run every render the gains should be obvious. I am not sure hoisting will improve performance, from what I know running something like useYourComponentHook method would initialize new functions inside of it every run and would not have the referential equality performance that useCallback provides. 

In any case, this is a great opportunity to learn and exercise some react profiling. 🔍 📈

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -203,6 +247,81 @@ const NewProjectMap = ({
...
+        selectedFeature,
+        featureCollection,
+        setFeatureCollection,
+        isPresent
+      );
+    });
+
+    return;
+  };
+
```
sergiogcx, line 324: Absolutely brilliant, I personally did not experience lags and it feels seamless, but I can see this being an issue if the user's computer is slow or the internet connection is bad.

sergiogcx: @johnclary out of curiosity, how large is the size of the full CTM map?

sergiogcx (approved):
Amazing work as always! 🚢 🚀

moped-editor/src/views/projects/projectView/ProjectSummary.js
sergiogcx: Totally fine, and thank you for keeping the code lean and clean! 💇  ✨

PR #407, opened 2021-08-28
Ability to define new project and create components from signal asset

johnclary:
Fixes [#7073](https://github.com/cityofaustin/atd-data-tech/issues/7073) and [#7075](https://github.com/cityofaustin/atd-data-tech/issues/7075)


moped-editor/src/views/projects/newProjectView/NewProjectView.js
```diff
@@ -50,6 +51,73 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
+// todo: move elsewhere?
+const COMPONENT_DEFINITIONS = {
```
johnclary, line 55: This is not future proof. Alternatives:
- Query component types from DB and use as a lookup
- ?

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
```diff
@@ -0,0 +1,126 @@
...
+      },
+      geometry: signal.geometry,
+      id: featureUUID,
+    };
+    featureCollection.features.push(feature);
+  }
+  return featureCollection;
+};
+
+const SignalAutocomplete = ({
```
johnclary, line 34: If/when we want to re-use this in the components editor—we'll need to wrap some of this project create form-specific handling. tbd?

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
```diff
@@ -0,0 +1,126 @@
...
+      renderInput={params => (
+        <TextField
+          {...params}
+          error={nameError}
+          helperText="Required"
+          InputProps={{
+            ...params.InputProps,
+            endAdornment: (
+              <>
+                {loading ? (
```
johnclary, line 109: it's hard to verify the spinner renders when testing because the Socrata request (probably cachced) is faaast :) i've seen it, I swear.

johnclary:
Before this merges: would like to discuss the dependency on the Socrata dataset. Need to consider it's update frequency and downtime (two saturday evenings/month).

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
```diff
@@ -0,0 +1,126 @@
+import React from "react";
+import { TextField, CircularProgress } from "@material-ui/core";
+import { Autocomplete } from "@material-ui/lab";
+import { v4 as uuidv4 } from "uuid";
+
+const signalToFeatureCollection = signal => {
+  // Immitate a "drawn point" feature from a traffic signal feature
```
johnclary, line 7: if we change our approach to handling drawn points, this would be a dark dependency that could break.

johnclary:
This is ready for review—no rush as I'm OOO next week.

chiaberry:
https://jc-7073-new-project-from-signal--atd-moped-main.netlify.app/ to see it in action ✨

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+          setError(error.toString());
+        }
+      );
+  }, []);
+
+  return (
+    <Autocomplete
+      id="signal-id"
+      filterOptions={(options, { inputValue, getOptionLabel }) => {
+        // limits options to ensure fast rendering
```
chiaberry, line 98: I looked to see if it was possible, but we could do a virtualized list in the autocomplete if we wanted to. I can't imagine someone scrolling forever to find their signal though.

chiaberry (approved):

amenity (approved):
I thought I'd already approved this but perhaps that's pending in a forgotten tab. In any case, this is SO slick. 😎 🙏

johnclary:
thanks y'all! will hold off on merging until #409 can be merged into this. WIP.

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
johnclary: oooh cool idea—I will read up on that

johnclary:
Folks—I had intended to ask y'all to review #409 separately, but I merged `main` into that branch to avoid some conflicts, and which turned it into a massive PR.

Now this PR includes the component editing functionality from that branch, and is quite a massive PR. I'm sorry! @sergiogcx I do think it'd be good for you to review this. Happy to pair on it as well.

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
```diff
@@ -221,7 +235,7 @@ const ProjectComponentEdit = ({
                 component_subtype: component?.component_subtype ?? null,
               },
             },
-            subcomponents: componentSubcomponents,
+            subcomponents: [...currentSubcomponents, ...componentSubcomponents],
```
johnclary, line 238: @sergiogcx i uncovered a tricky bug here that we somehow hadn't noticed before. Subcomponent changes were not being saved properly for components which had a required subtype, because the subtype's subcomponents were being overwritten here. I have some ideas for a more elegant fix, but this gets it done :/

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -269,15 +270,14 @@ const NewProjectMap = ({
...
-
   /**
    * Generate a viewport configuration object
    */
   const [viewport, setViewport] = useFeatureCollectionToFitBounds(
     mapRef,
     // If this is a new feature, use the project feature collection to retrieve the area
     newFeature ? projectFeatureCollection : featureCollection,
-    false
+    isSignalComponent
```
johnclary, line 280: this causes the map to auto-zoom to selecting a signal during component editing. when creating a new component, the map will not zoom to the selected signal until the component is saved. 

all of this was quite tricky because we're forking the map editing behavior specifically for signal components.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -454,9 +454,11 @@ const NewProjectMap = ({
...
         width="100%"
         height="60vh"
-        interactiveLayerIds={!isDrawing && getEditMapInteractiveIds()}
-        onHover={!isDrawing ? handleLayerHover : null}
-        onClick={!isDrawing ? handleLayerClick : null}
+        interactiveLayerIds={
+          !isDrawing && !isSignalComponent && getEditMapInteractiveIds()
+        }
+        onHover={!isDrawing && !isSignalComponent ? handleLayerHover : null}
+        onClick={!isDrawing && !isSignalComponent ? handleLayerClick : null}
```
johnclary, line 461: these click and hover events must be disabled for signal components because we're not rendering those interactive layers. we could maybe avoid some of this condition checking by refactoring the click/hover handlers.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -486,9 +488,13 @@ const NewProjectMap = ({
         />
 
         {/* RENDER LAYERS */}
-        {Object.entries(mapConfig.layerConfigs).map(([sourceName, config]) =>
-          // If a config has a url, it is needs state to update selected/unselected layers
-          config.layerUrl ? (
+        {Object.entries(mapConfig.layerConfigs).map(([sourceName, config]) => {
+          if (isSignalComponent && sourceName !== "drawnByUser") {
```
johnclary, line 492: And here's where we filter out the CTN clickable map layers. We are forcing the behavior that if a signal asset is selected only the point geometry from Socrata data can be stored.  If the user wants to select a signal asset but use a different geometry, this should be handled upstream by editing the signal record in the Data Tracker.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -528,14 +534,14 @@ const NewProjectMap = ({
...
-        )}
+          );
+        })}
 
         {/* Street Tool Tip*/}
         {renderTooltip(featureText, hoveredCoords, classes.toolTip)}
 
         {/* Draw tools */}
-        {renderMapDrawTools(mapEditToolsContainerRef)}
+        {!isSignalComponent && renderMapDrawTools(mapEditToolsContainerRef)}
```
johnclary, line 544: Similar to above, we don't want users hand-drawing geometries for signal components.

moped-editor/src/views/projects/projectView/ProjectComponentSubcomponents.js
```diff
@@ -101,6 +101,9 @@ const ProjectComponentSubcomponents = ({
         options={availableSubcomponents}
         // The list changes depending on available subcomponents for that specific type
         getOptionLabel={option => option.subcomponent_name}
+        getOptionSelected={(option, value) =>
+          option.subcomponent_id === value.subcomponent_id
+        }
```
johnclary, line 106: fixes an MUI linter warning that was throwing on component init

moped-editor/src/utils/socrataHelpers.js
```diff
@@ -0,0 +1,31 @@
...
+            // result is a geojson feature collection
+            setFeatures(result.features);
+          }
+        },
+        error => {
+          setError(error.toString());
+        }
+      );
+  }, [url]);
+  error && console.error(error);
```
sergiogcx, line 29: I live this custom hook and how you run console.error only if there is an error. Will borrow this line in the future! 👍

moped-editor/src/utils/mapHelpers.js
```diff
@@ -1056,3 +1055,38 @@ export const queryCtnFeatureService = async function(projectExtentId) {
...
+    features: [],
+  };
+  if (signal) {
+    /* 
+    / preserves the feature's previous UUID if it's being edited. we are **not** preserving
+    / any other feature properties when the feature is edited. so, for example, if the user
+    / edits a signal component and the signal geometry in socrata has since changed, the new
+    / geometry will be saved.
+    */
+    const featureUUID = signal.id || uuidv4();
```
sergiogcx, line 1077: No action required on here, but I wonder if `signal.id` could potentially throw a WSOD in some unlikely scenario where (signal) is a value and not an object, or it is an object but does not have the "id" key. I may be mistaken, but I wonder if the use of conditional chaining is warranted in this case.

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
sergiogcx: I was wondering what would happen if a component changes type from signal to something else, but I think the code you have written should handle it just fine.

moped-editor/src/views/projects/newProjectView/NewProjectView.js
sergiogcx: ➕ Agreed. This will almost certainly have to change to a DB lookup, we are creating/editing types in the future that will affect their name and index id.

moped-editor/src/views/projects/newProjectView/NewProjectView.js
```diff
@@ -50,6 +51,89 @@ const useStyles = makeStyles(theme => ({
...
+ */
+const COMPONENT_DEFINITIONS = {
+  generic: {
+    name: "Extent",
+    description: "New Project Feature Extent",
+    component_id: 0,
+  },
+  phb: {
+    name: "PHB",
+    description: "Pedestrian ssignal",
```
sergiogcx, line 66: Typo ☝️

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
sergiogcx: @johnclary Ah, thanks for finding this bug. It makes sense, and your fix is elegant as-is imo. Happy to hear the other ideas you have.

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
```diff
@@ -0,0 +1,137 @@
...
+) => {
+  useEffect(() => {
+    const signalSubtype = signal
+      ? signal.properties.signal_type.toLowerCase()
+      : "";
+    const featureCollection = signalToFeatureCollection(signal);
+    setSelectedComponentSubtype(signalSubtype);
+    setEditFeatureCollection(featureCollection);
+  }, [signal, setSelectedComponentSubtype, setEditFeatureCollection]);
+};
```
sergiogcx, line 56: These need to be put in a different file I think? Otherwise, it becomes an anti-pattern 🤷

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
```diff
@@ -0,0 +1,137 @@
...
+      getOptionLabel={option =>
+        option
+          ? `${option.properties.signal_id}: ${option.properties.location_name}`
+          : ""
+      }
+      onChange={(e, signal) => {
+        setSignal(signal);
+      }}
+      loading={loading}
+      options={options}
```
sergiogcx, line 123: Maybe options need to be filtered to have a length of at least one character?

sergiogcx (approved):
This is amazing that we are integrating the map with socrata! 🗺️

mateoclarke (approved):
🚢 💨 ✨ 

monsters-inc-traffic-lights

moped-editor/src/utils/mapHelpers.js
johnclary: you're right. should never happen, but it's risky to rely on perfect data from socrata 100% of the time. i'll follow your suggestion.

moped-editor/src/views/projects/newProjectView/NewProjectView.js
johnclary: 👍 thanks for the nudge. i'll re-use a query from ProjectComponents/

moped-editor/src/views/projects/newProjectView/SignalAutocomplete.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+    setProjectDetails(updatedProjectDetails);
+    setSignal(signal);
+    const featureCollection = signalToFeatureCollection(signal);
+    setFeatureCollection(featureCollection);
+  };
+
+  const { features, loading, error } = useSocrataGeojson(SOCRATA_ENDPOINT);
+
+  const options = features ? [...features, ""] : [""];
```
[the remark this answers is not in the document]
johnclary: yes—that's exactly what's causing it. i spent a lot of time fussing with the `value` prop setter of MUI autocomplete but kept hitting a warning when the input value was `undefined`. 

i just retested this component and realized that specifically passing `null` as an input value will bypass the warnings. will fix!

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
johnclary: in that case I'll leave it as-is ;) glad i was able to get familiar with this reducer—it's awesome!

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
johnclary: @sergiogcx—to be clear, you're suggesting to move both of these hooks outside the component?  i'm happy to do that and am inclined to make a `signalComponentHelpers` util to avoid bloating the mapHelpers any more.

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
johnclary: i over-engineered this because i didn't understand the normal MUI autocomplete behavior. will refactor.

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
sergiogcx: Ah, yes mapHelpers is way too bloated. signalComponentHelpers would be very helpful. 🙏

johnclary:
I created [#7136](https://github.com/cityofaustin/atd-data-tech/issues/7136) to complete these follow-up tasks

PR #409, opened 2021-08-29
Ability to select a signal asset when editing project components

johnclary:
Fixes [#7075](https://github.com/cityofaustin/atd-data-tech/issues/7075).


amenity:
@johnclary - elegant execution — and it works flawlessly.👏 👏👏
Can't think of any improvements.

PR #413, opened 2021-09-06
Fix various React/DOM warnings - #7035

johnclary:
Fixes [#7035](https://github.com/cityofaustin/atd-data-tech/issues/7035)

### DIFF
--- a/moped-editor/src/components/DataTable/DataTable.js
+++ b/moped-editor/src/components/DataTable/DataTable.js
@@ -227,7 +227,8 @@ const DataTable = ({ fieldConfiguration, data, loading, error, refetch }) => {
   const executeMutation = (field = null, value = null) => {
     const mutationField = field || editField;
     const mutationValue = value !== null ? value : editValue;
-    const nullableField = fieldConfiguration.fields[mutationField].nullable ?? true;
+    const nullableField =
+      fieldConfiguration.fields[mutationField].nullable ?? true;
     // Execute mutation only if there is a new value selected, prevents user
     // from attempting to save initial value, which would be null
     // Also prevents from executing migration if value has not changed
@@ -304,7 +305,7 @@ const DataTable = ({ fieldConfiguration, data, loading, error, refetch }) => {
    */
   const handleFieldValueUpdate = value => {
     if (errorField !== "") {
-      setErrorField("")
+      setErrorField("");
     }
     setEditValue(value.target.value);
   };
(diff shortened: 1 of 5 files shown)


moped-editor/src/layouts/DashboardLayout/TopBar.js
```diff
@@ -106,7 +106,7 @@ const TopBar = ({ className, onOpen, ...rest }) => {
         </RouterLink>
         <Hidden smDown>
           <Box>
-            <Tabs className={classes.tabs}>
+            <Tabs className={classes.tabs} value={false}>
```
johnclary, line 109: we're not using the `value` prop to set tab state, but it needs a value regardless. `undefined` throws an error.

moped-editor/src/views/projects/projectView/TabPanel.js
```diff
@@ -15,7 +15,7 @@ function TabPanel(props) {
     >
       {value === index && (
         <Box p={3} className={props.className}>
-          <Typography>{children}</Typography>
+          {children}
```
johnclary, line 18: This change impacts every tab of the `ProjectView`, but as far as I can tell it only actually affects the `DataTable` component, which is only in use on `ProjectSummary`

moped-editor/src/components/DataTable/DataTable.js
```diff
@@ -534,17 +536,16 @@ const DataTable = ({ fieldConfiguration, data, loading, error, refetch }) => {
...
+                      {fieldConfiguration.fields[field].editable && !isEditing && (
+                        <span>
+                          <Icon
+                            className={classes.editIcon}
+                            onClick={() => handleFieldEdit(field)}
+                          >
+                            create
+                          </Icon>
+                        </span>
+                      )}
```
johnclary, line 548: ☝️ linting

moped-editor/src/components/DataTable/DataTable.js
```diff
@@ -388,7 +391,7 @@ const DataTable = ({ fieldConfiguration, data, loading, error, refetch }) => {
       ? lookupTablesData[lookupTable] ?? []
       : fieldConfig?.options ?? [];
 
-    const currentValue = editValue ?? initialValue;
+    const currentValue = editValue ?? (initialValue || "");
```
johnclary, line 394: the only `select` field in this component is for "Current Status". It's not possible to set it to empty, but projects can be created w/o an initial "Current Status". This may have been a regression that snuck in at some point, but in any case we will tackle it with [#7036](https://github.com/cityofaustin/atd-data-tech/issues/7036)

johnclary:
devs, this wipes out a lot of console warnings that have accumulated over time. i tried to keep the commits limited to one per error.

@amenity need your product review of the Project Summary page for this one—you should notice no change.

amenity (approved):
No problems on the Summary page!

sergiogcx (approved):
Thanks for the linting! ✨  it's definitely tech-debt in the code we need to start chipping at

mateoclarke:
Thanks for this cleanup work John!

tillyw:
Thanks, John!

PR #419, opened 2021-09-20
Restrict component features to points or lines

chiaberry:
Closes https://github.com/cityofaustin/atd-data-tech/issues/7133

Won't be merged however until I address https://github.com/cityofaustin/atd-data-tech/issues/7213

Test deployment: https://7133-restrict-component--atd-moped-main.netlify.app/moped/projects/280?tab=map

The gist of it is: check the `line_representation` of a component that is selected and toggle the map layers based on if it should be represented as a line or a point.


moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
```diff
@@ -168,7 +172,8 @@ const ProjectComponentEdit = ({
   const { saveActionState, saveActionDispatch } = useSaveActionReducer();
 
   /**
-   * Generates an initial list of component types, subtypes and counts
+   * Generates an initial list of component types, subtypes and counts (counts is total number of subtypes)
+   * todo chia: ask if this initial list ever changes? or is this a constant list of component types, subtypes
    */
   const initialTypeCounts = data // Do we have data?
```
chiaberry, line 178: does intialTypeCounts change during mapping, or is it a static array? I think its a constant list, but I wasnt sure if initial meant that it might change along the way?

mateoclarke (commented):
This all looks good. I do have a few questions from poking around in the deploy preview app. 

1. Do all of the components currently use lines for their `line_representation`? I expected that some, like Signage, would use points. As I toggled through all the options, none of them used points. Maybe this is a data issue and not a code issue. Or maybe I'm confused 🙃. But I thought it was at least worth pointing out. 
2. I was able to create a map component that was represented with both points and lines by not selecting a component type until after I had done my mapping. I think the intention was to make it so that it wouldn't be possible to mix points and lines for a component geography. I could see lots of users start the process of adding a component by plotting on the map first before adding the details about type/subtype and description in the left sidebar. Right now, it seems like we are showing all the map edit option buttons and removing some based on the type selection. What if instead,  we didn't show any of those buttons unless there is a type selected?

Screen Shot 2021-09-23 at 4 19 59 PM

amenity:
> What if instead, we didn't show any of those buttons unless there is a type selected?
Yes!

chiaberry:
@mateoclarke you aren't confused, I had incorrectly refactored my code and it was coming back as everything should be a line. I un-refactored it and now it works as intended. Pavement marking for example is a point component. 

and yes! i will make it so you can't draw until you've selected a component

sergiogcx (approved):
Amazing code, thanks for the docstrings and proper code commentary.

mateoclarke (approved):
Just reviewed this again and it looks good! 🚢 💨

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -493,6 +496,15 @@ const NewProjectMap = ({
             // hides feature selecting and drawing layers when when component is a signal
             return null;
           }
+          if (
+            drawLines === true &&
```
johnclary, line 500: do we need to check for strict equality here (and in the reverse case?)—seeing this implies to me that `drawLines ` could be something other than `true` or `false`

moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
johnclary: @chiaberry you are correct. it's static.

moped-editor/src/views/projects/projectView/ProjectComponents.js
```diff
@@ -166,13 +168,13 @@ const ProjectComponents = () => {
   };
 
   /**
-   * This is a helper variable that stores true if we have components, false otherwise.
+   * This is a helper variable that stores true if we have components associated with project, false otherwise.
```
johnclary, line 171: nice doc improvements throughout!

johnclary (approved):
awesome work!  ⛵ 🗺️

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
chiaberry: drawLines is initially null and I'm making it do double duty in a way by assuming drawLines as null means no component is selected

I could add another state variable "componentSelected", keeping drawlines as only true or false. Then it would be if drawlines is true and componentSelected is true, return null etc.

johnclary: ah, got it! it's obvi now that i looked at the code, but can you just add a comment here that mentions if `drawLines` is `null`, this ensures no map drawing is possible?

amenity (approved):

PR #422, opened 2021-09-27
Prevent drawing features until component is selected

chiaberry:
https://7213-prevent-drawing--atd-moped-main.netlify.app/moped/projects

Closes https://github.com/cityofaustin/atd-data-tech/issues/7213

In my previous  PR https://github.com/cityofaustin/atd-moped/pull/419, I introduced the variable `drawLines`, which initialized as null and then was set as true or false depending on which component was chosen. So if `drawLines` is null, it means a component hasn't been selected yet, and we shouldn't render the draw layers yet. 

~~However, this is preventing anyone from adding a new project since you need at least one feature to save a new project but since we can't choose a component in the add project flow, we can't draw on the map.~~


moped-editor/src/views/projects/projectView/ProjectComponentEdit.js
```diff
@@ -895,7 +893,7 @@ const ProjectComponentEdit = ({
                         rows={4}
                         defaultValue=""
                         variant="filled"
-                        value={componentDescription}
+                        value={componentDescription ?? ""}
```
chiaberry, line 896: did this to clear an error that was showing up in the console, not having values in fields be null

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -491,75 +501,74 @@ const NewProjectMap = ({
...
-          return config.layerUrl ? (
-            // If a config has a url, it is needs state to update selected/unselected layers
-            <Source
-              key={config.layerIdName}
-              type="vector"
-              tiles={[config.layerUrl]}
-              maxZoom={config.layerMaxLOD || mapConfig.mapboxDefaultMaxZoom} // maxLOD found in vector tile layer metadata
-            >
-              <Layer
+        {drawLines !== null &&
```
chiaberry, line 504: do not show the layers unless a component is selected and drawLines is not null

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -491,75 +501,74 @@ const NewProjectMap = ({
...
-                features: [
-                  ...featureCollection.features,
-                  ...(otherProjectFeaturesCollection?.features ?? []),
-                ].filter(
-                  feature => feature.properties.sourceLayer === sourceName
-                ),
-              }}
-            >
-              <Layer
+                type="vector"
```
chiaberry, line 523: nothing changed here except wrapping it in ` {drawLines !== null &&` but github did weirdness with highlighting differences

sergiogcx (approved):
It just works! 🚀

chiaberry:
bug to fix/investigate: "but when I was poking around in the edit component mode, once I select a component type, and go to select a point dropper or line drawer, the icons disappear and drawing seems disabled."

moped-editor/src/views/projects/newProjectView/NewProjectMap.js
```diff
@@ -491,75 +501,75 @@ const NewProjectMap = ({
...
+              </Source>
+            );
+          })}
 
         {/* Street Tool Tip*/}
         {renderTooltip(featureText, hoveredCoords, classes.toolTip)}
 
         {/* Draw tools */}
         {!isSignalComponent &&
+          drawLines !== null &&
```
chiaberry, line 568: @sergiogcx this is the only change since your last review

chiaberry: it used to also be contingent on `isDrawing`

johnclary (approved):
🚢

amenity (commented):
Beautiful! Thank you, @chiaberry🙏 ✨

Issue #424, opened 2021-09-28
[Bug] Fix moped-test API deployment


sergiogcx:
Building and deploying the API works fine in nearly all conditions, for some reason the build process is not finding one specific library:

![2021-09-28_11-36-40.png](https://images.zenhubusercontent.com/5b7edad7290aac725aec290c/96356558-a37e-4201-b962-dcd692e923de)

sergiogcx:
It appears setuptools 58 introduced braking changes which affects a couple libraries we have:

![2021-09-28_14-54-07.png](https://images.zenhubusercontent.com/5b7edad7290aac725aec290c/7db56c40-eb9a-4dfa-ae37-e286e725a018)

Downgrading to 57.5.0 fixes the issue until we know zappa and secondary libraries can handle the updates.

sergiogcx:
![2021-09-28_14-57-45.png](https://images.zenhubusercontent.com/5b7edad7290aac725aec290c/2b6ed706-caab-4c25-9d0c-d6b9661ccbd4)

PR #425, opened 2021-09-28
[Bug] Fix moped-test API deployment#424

sergiogcx:
Closes https://github.com/cityofaustin/atd-moped/issues/424


sergiogcx:
None of these approaches were needed, by downgrading setuptools we could maintain the pipeline.

PR #428, opened 2021-10-06
Condense and refine search components on "Projects" page

chiaberry:
https://7207-refine-search--atd-moped-main.netlify.app/moped/projects
To test: search using the search bar, hit enter to execute search.
Use the icon at the end of the search bar to open filters/advanced search.

I renamed the `GridTableExport` component to be `GridTableNewItem` since the component renders a new item button and doesn't export anything. Also I included a fix for the padding issue that happens when there is no "add" button (staff page gets an add staff button, but projects page does not). The fix feels a tad hacky, I can remove it if we aren't comfortable with it.

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableSearch.js
+++ b/moped-editor/src/components/GridTable/GridTableSearch.js
@@ -44,7 +44,7 @@ const useStyles = makeStyles(theme => ({
     margin: ".5rem",
   },
   searchBarContainer: {
-    padding: "24px",
+    padding: "2px",
     [theme.breakpoints.down("xs")]: {
       paddingBottom: "12px",
     },


moped-editor/src/components/GridTable/GridTableNewItem.js
```diff
@@ -26,12 +25,16 @@ const GridTableExport = ({ query }) => {
             {query.config.new_item_label}
           </Button>
         ))}
+      {// add empty div of equal height if there is no button to match styles
+      !query.config.showNewItemButton && (
+        <div style={{ height: "36.5px" }}></div>
+      )}
```
chiaberry, line 31: if we arent going to render a button, then render an empty div of button size. 

I considered conditionally adding padding to the div in GridTableSearch, but this is the component that knows whether or not it is going to render the button.

moped-editor/src/components/GridTable/GridTableSearch.js
```diff
@@ -35,14 +32,48 @@ const useStyles = makeStyles(theme => ({
...
+    },
+  },
+  advancedSearchRoot: {
+    width: "100%",
+    zIndex: "3",
+    paddingLeft: "43px",
+    paddingRight: "17px",
+    [theme.breakpoints.down("sm")]: {
+      paddingLeft: "23px",
+      paddingRight: "14px",
```
chiaberry, line 59: I could not figure out what the formula is for this padding. Basically setting the Popper to 100% width made it even wider than the screen. I can't define the width of the element in pixels because then it wouldnt be reponsive. These numbers feel random, but they work. Yes it's weird.

patrickm02L (changes requested):
https://user-images.githubusercontent.com/5566727/136459547-0b3f95ba-3509-4db4-b846-247774dce0a9.mp4

The video highlights the changes that I saw upon review. The notes below are what I pointed out in the video.

**Desktop view**
* Shrink hover effect of the Advanced search to be inside the input box
* Modal alignment when advanced search is activated
    * Gap on the left and overlap on the right
* Trash can is not showing up
* Align search button to the right (not left as I say in the video), same edge as the Download button

**Mobile View**
* Advanced search jumps to the top
    * Can we make the advanced search stick to the bottom of the (input, search button, download button) stack?
    * What is your take on why it’s sticking to the top of the page?
* All buttons should align with the search fields
* Download button pops up at different break points
* Download button is “big” in the same break points where it’s popping up above the search input

chiaberry:
I really like the video showing me what you are seeing. No clue what happened to my trash can, I will bring that back. 

The filter is sticking to the top of the page when you are actively resizing it. if you were to refresh at that size and try again, I dont think it would stick to the top. My take is that the divs are re-rendering or changing in some way and the filter modal loses its anchor. 


What screen sizes should I keep in mind when adjusting things? I was going off of ipad and iphone X, but I just checked and you cant see the activity log on an ipad, even when rotating. 

The sizing of the modal is tricky, because at my normal screen width its not all janky. I'll see if with fresh eyes tomorrow if i can figure out what pattern is to the sizing
Screen Shot 2021-10-07 at 6 37 15 PM

amenity (approved):
It seems that these alignment bugs do not show up on Windows 10 in Firefox and Chrome and we may still have some work to do on overall global grid/mobile layout patterns. So sticking them [here](https://github.com/cityofaustin/atd-data-tech/issues/7402).  🚢

chiaberry:
updated the padding to be 2px, and I'm going to merge!

PR #429, opened 2021-10-08
Update new project map placeholder

frankhereford:
This PR intends to close cityofaustin/atd-data-tech/issues/7328.

It includes a new image and content to be used as the map placeholder when no map is available on a project page.

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectSummaryMapFallback.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummaryMapFallback.js
@@ -12,7 +12,7 @@ const useStyles = makeStyles(theme => ({
   card: {
     minHeight: "12rem",
     textAlign: "center",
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectSummaryMapFallback.js
```diff
@@ -21,6 +21,12 @@ const useStyles = makeStyles(theme => ({
   paragraphGroup: {
     margin: "1rem 0",
   },
+  mapPlaceholder: {
+    color: "#848484",
```
chiaberry, line 25: this could be `theme.palette.text.secondary` so if we update that color later, we dont have to go around hunting for everywhere with secondary text

amenity (approved):
LSTM✨ 🙏 Thanks, Frank🚢

PR #435, opened 2021-10-13
UI refinements to "Role" column on "Team" table

chiaberry:
Closes: https://github.com/cityofaustin/atd-data-tech/issues/7359
also closes: https://github.com/cityofaustin/atd-data-tech/issues/7448

check it out: https://7359-team-role-ui--atd-moped-main.netlify.app/moped/projects/339?tab=team


amenity:
Sweeeet! Looks so snazzy. Is there a way to get rid of the chip-rendering altogether? I still see it in the background while I'm using the checkboxes. The ["tag" option](https://v4.mui.com/components/selects/#multiple-select) here would be better if possible.

PR #445, opened 2021-10-28
Support local cloning of Production database #7572

sergiogcx:
### Video Tour

https://www.youtube.com/watch?v=1jE1XrJ_uis

To test, first connect to the VPN then:

### Hasura-Cluster Cloning & Restoring

There may be occasions where you need to copy the data in staging or production to run it locally. Clone and restore allow for these operations to happen easily.

#### I. Configure

You will first need to configure your environment:

Example:

1. Copy the hasura_cluster template file into your ssh directory with this specific name:

   `$ cp config/hasura_cluster_template.json ~/.ssh/hasura_cluster.json`;

   Note: This script expects to find it with this name `~/.ssh/hasura_cluster.json`

2. Change permissions so only you can read/write to it:

   `chmod 600 ~/.ssh/hasura_cluster.json`

3. Edit the file with nano or vim, and fill-in the blanks:

   `nano ~/.ssh/hasura_cluster.json`
   
   -or-

   `vim ~/.ssh/hasura_cluster.json`

4. Save and exit

#### II. Clone

Cloning basically runs pg_dump for you and saves the file into the snapshots directory. To do this, you run:

This operation does not insert into your local database.

Syntax:

```bash
$./hasura-cluster clone [environment]
```

Example:

- Cloning staging or production:

```bash
./hasura-cluster clone staging
./hasura-cluster clone production
```

- Cloning a staging or production backup from S3:

```bash
/hasura-cluster clone bucket staging "2021-11-03"
/hasura-cluster clone bucket production "2021-11-03"
```

#### III. Restore

This operation loads a snapshot into your local database instance. It does not create a snapshot.

- Listing snapshot files downloaded:

```bash
./hasura-cluster list local [staging|production]
```

- Listing files in an s3 bucket:

```bash
./hasura-cluster list bucket [staging|production]
```

- Restoring a specific file:

```bash
./hasura-cluster restore [filename]
```

NOTE: If hasura isn't running, restore is going to exit with an error.

- Restarting the hasura cluster without migrations:

```bash
./hasura-cluster reload
```


moped-database/hasura-cluster
```diff
@@ -39,33 +70,187 @@ function __get_version() {
...
+  S3_BACKUP_DATE=$2;
+
+  # Check we have AWS
+  __aws_installed_or_exit;
+
+  BUCKET_NAME=$(__config_get_key "backups.${LOCAL_ENV}.bucket");
+  BUCKET_PREFIX=$(__config_get_key "backups.${LOCAL_ENV}.prefix" | sed -r "s/_staging\//_staging/g"| sed -r "s/_production\//_production/g");
+  LOCAL_FILE_PATH="./snapshots/${LOCAL_ENV}/atd_moped_${LOCAL_ENV}-full.sql.gz";
+  LOCAL_FILE_PATH_UNZIPPED="./snapshots/${LOCAL_ENV}/bucket_${LOCAL_ENV}-${S3_BACKUP_DATE}.psql";
+  aws s3 cp "s3://${BUCKET_NAME}/${BUCKET_PREFIX}/${S3_BACKUP_DATE}/atd_moped_${LOCAL_ENV}-full.sql.gz" "${LOCAL_FILE_PATH}";
```
frankhereford, line 220: Currently, the back up job that is making snapshots for the production database is using the naming scheme `atd_moped-full.sql.gz` where this script expects `atd_moped_production.sql.gz`. I'm not sure if the best solution is to change this script to swap in `full` for `production` when someone is pulling from a production backup, or to change the backup job to name those files in the way the script currently expects them.

frankhereford (changes requested):
This work is really, really high quality! I was able to get it setup and working without any trouble, and that is thanks to your documentation being so detailed. 💯 🏅 

I've tested all the new functions in the `hasura-cluster` script, and I only found a couple of tiny things that you may want to look at:

* The backup naming scheme for production backups on S3 isn't what the script expects.
* The script can't detect the SSH session providing the tunnel via the bastion and exits because of it.

Other than those two tiny things, this is really phenomenal work. Thanks for doing it - it's going to be super useful.

moped-database/hasura-cluster
```diff
@@ -120,6 +324,160 @@ function run_migration() {
...
+}
+
+#
+# Destroys an existing link
+#
+function unlink() {
+  LOCAL_PORT=$(__config_get_key "bastion.local_port");
+  if [[ $(__is_local_port_open "${LOCAL_PORT}") = "TRUE" ]]; then
+    # Find the pid
+    SSH_PID=$(ps -eo pid,command | grep ssh | grep bastion | cut -d" " -f1);
```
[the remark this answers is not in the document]
sergiogcx: @frankhereford Pretty sure it isn't, this is the output of the first command `ps -eo pid,command`, can you help me confirm?



This is the output when it's piped to -f1:

moped-database/hasura-cluster
sergiogcx: I think it's a valid observation @frankhereford; however, *.gz files are treated as temporary files and are removed after the execution is finished (see line 224: `rm "${LOCAL_FILE_PATH}";` where LOCAL_FILE_PATH is the zipped file).

In short, the name of the gz file is trivial because it gets removed. Would you agree?

sergiogcx: Ah, that is a good catch. Thanks @frankhereford

frankhereford (approved):
This is going to be so useful - thank you for making this!

moped-database/README.md
```diff
@@ -72,6 +72,96 @@ $ hasura console
...
+./hasura-cluster clone staging
+./hasura-cluster clone production
+```
+
+
+- Cloning a staging or production backup from S3:
+
+```bash
+/hasura-cluster clone bucket staging "2021-11-03"
+/hasura-cluster clone bucket production "2021-11-03"
```
mateoclarke, line 131: these need `.`s before the slash, right?

mateoclarke (approved):
Really great work and thanks for making the video explainer!

It might be useful to add to the documentation the dependencies of `awscli` and `gnu-sed` are required. I had to install those via Homebrew to get anywhere.

In my testing, at first I was having issues executing the clone staging or production commands. The error was says `Cannot establish link, you don't seem to be on VPN?` but I was connected to Cisco VPN. It seemed in the end to be an issue with my configuration file and not the network at all. Is there a command to test that the `hasura_cluster.json` file has proper details for configuration? That might be helpful to document. The error message mislead me in my troubleshooting.

Aside from that I had issues finding a snapshot that was compatible with the metadata in this branch, but the staging snapshot from `2021-11-04` finally did the trick. ✨

moped-database/README.md
sergiogcx: correct

PR #447, opened 2021-10-28
Updates 404 page with new image, language, search bar and styling

tillyw:
https://github.com/cityofaustin/atd-data-tech/issues/6604

Updates 404 page with new image, language, search bar and styling.

Fully testable [here](https://deploy-preview-447--atd-moped-main.netlify.app/).

https://user-images.githubusercontent.com/35410637/140419450-2915bbae-4dfc-4d4d-969f-34123f0a7b73.mov

### DIFF
--- a/moped-editor/public/static/images/moped-flat.svg
+++ b/moped-editor/public/static/images/moped-flat.svg
@@ -1,0 +1,165 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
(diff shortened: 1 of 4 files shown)


amenity:
Looking good! A couple of things: 

- I would expect to be able to click in the search bar anywhere to activate the field so I could begin typing. It took me a bit to realize I had to click the magnifying glass icon specifically
- It just looks a bit weird for search bar height to decrease when it's activated while the width increases. I'd keep the height the same. And that shadow/line under the active state should be removed. 
default

active

mateoclarke (approved):
I agree with Amenity's feedback about focus state and phantom bar underneath the search input.

The other weird thing that I stumbled on in testing this is that if you enter in a project ID doesn't exist, you get a WSOD. I expected that I would get rerouted to the 404. For example: https://deploy-preview-447--atd-moped-main.netlify.app/moped/projects/3511. 

Maybe thats out of scope for this issue but I'd recommend we make an issue to investigate the routing. cc/ @amenity

amenity:
Good catch @mateoclarke🙏 ... That seems like one of our top use cases for a 404 page, so would love to get it covered in this PR if possible. But we can create a follow-up issue instead. Up to you, @tillyw

amenity (approved):
Creating https://github.com/cityofaustin/atd-data-tech/issues/7719 for the bug @mateoclarke found.

tillyw:
just demoed with amenity and she gave me permission to merge, so merging! @mateoclarke sorry i didn't give it a proper name for the release notes when i created the pr. i've updated it!

PR #448, opened 2021-10-29
Add moped types and moped project types 

chiaberry:
Addresses https://github.com/cityofaustin/atd-data-tech/issues/7533

1. Adds moped_type and moped_project_types tables
2. Adds rows based on options in the issue
3. drops the moped_categories and moped_proj_categories tables.
4. update permissions and event_triggers
5. removes seed data for old tables


moped-database/metadata/tables.yaml
```diff
@@ -4136,6 +3945,104 @@
...
+    permission:
+      columns:
+      - project_id
+      - type_name
+      - project_type_id
+      - date_added
+      - added_by
+      filter: {}
+      check: {}
+  event_triggers:
```
chiaberry, line 4021: here's the event trigger for project_types

zappfyllc:
> @chiaberry apologies if I'm missing it here (I can't find it in the commits but I didn't pull this up locally), but can you add two relationships

Found the foreign key @chiaberry right after posting this, sorry about that! Leaving my comment in case there's anything missed there but looks good to me!

chiaberry:
@zappfyllc I will my PR to satisfy this: 

>  Between the tables moped_type and moped_project_types, where the type_id would point to a new column called type_id in the table moped_project_types

because its currently using type_name since I was copying over the previous schema (heres what it looks like in main right now
Screen Shot 2021-11-01 at 2 24 24 PM
 
)

zappfyllc:
@chiaberry ah, that's definitely my bad on the old schema: when I created that placeholder table, I thought project categories were going to be more broad and static, but I think because of how specific they seem to be, we will be better off utilizing the ID so we don't have to update the references throughout the related table if a `type_name` changes. I would say we can defer to that pattern moving forward (`int` over `text` for foreign keys and/or column references in relationships).

I'll get better at including these details in subsequent issues, but it's a lot to keep track of so thanks so much for your patience as we slog through these missing details.

chiaberry:
@zappfyllc thank you for your patience and I work through these!

moped-database/migrations/1635521196468_create_table_public_moped_project_types/up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE "public"."moped_project_types"("project_id" integer NOT NULL, "type_name" text NOT NULL, "project_type_id" integer NOT NULL, "date_added" timestamptz DEFAULT clock_timestamp(), "added_by" integer, PRIMARY KEY ("project_type_id") , FOREIGN KEY ("project_type_id") REFERENCES "public"."moped_types"("type_id") ON UPDATE restrict ON DELETE restrict, FOREIGN KEY ("project_id") REFERENCES "public"."moped_project"("project_id") ON UPDATE restrict ON DELETE restrict, UNIQUE ("project_type_id"));
```
frankhereford, line 1: For this table - my understanding is that it's a table that maps `moped_project` and `moped_types` in a many:many relationship.

I took the SQL above and ran it through a formatter for readability:

```
CREATE TABLE "public"."moped_project_types"(
    "project_id" integer NOT NULL,
    "type_name" text NOT NULL,
    "project_type_id" integer NOT NULL,
    "date_added" timestamptz DEFAULT clock_timestamp(),
    "added_by" integer,
    PRIMARY KEY ("project_type_id"),
    FOREIGN KEY ("project_type_id") REFERENCES "public"."moped_types"("type_id") ON UPDATE restrict ON DELETE restrict,
    FOREIGN KEY ("project_id") REFERENCES "public"."moped_project"("project_id") ON UPDATE restrict ON DELETE restrict,
    UNIQUE ("project_type_id")
);
```

There is a `text` column defined in this table called `type_name` which I think may be vestigial. The actual textual representation of a type is contained in `moped_types.type_name` and is referenced in this table by a FK ID. I think keeping this text column around isn't intended. @zappfyllc - do you think that's the case?

If so, I suggest we exclude that column in this `CREATE` statement. Keeping it around won't hurt anything, but it may be confusing to developers implementing code which uses these tables.

frankhereford (changes requested):

moped-database/migrations/1635521196468_create_table_public_moped_project_types/up.sql
zappfyllc: Yep, we can definitely omit that column @frankhereford!

frankhereford (approved):
Looks great! Thanks for removing that extra column!

mateoclarke (approved):
🚢 💨 ✨

moped-database/migrations/1635521196468_create_table_public_moped_project_types/up.sql
```diff
@@ -0,0 +1,13 @@
...
+  "date_added" timestamptz DEFAULT clock_timestamp(),
+  "added_by" integer,
+  PRIMARY KEY ("project_type_id"),
+  FOREIGN KEY ("project_type_id") REFERENCES "public"."moped_types"("type_id") ON UPDATE restrict ON DELETE restrict,
+  FOREIGN KEY ("project_id") REFERENCES "public"."moped_project"("project_id") ON UPDATE restrict ON DELETE restrict,
+  UNIQUE ("project_type_id"));
+COMMENT ON TABLE public.moped_project_types IS 'List of related project types';
+
+DROP TABLE "public"."moped_proj_categories";
+DROP TABLE "public"."moped_categories";
```
sergiogcx, line 13: Those should be fine 👍 , luckily there isn't any data in moped_proj_categories in neither staging or prod, so this migration should succeed.

sergiogcx (approved):
Yep, this should work! ⚡ 🙏

PR #449, opened 2021-11-02
Add Signal Projects Table View

chiaberry:
Closes https://github.com/cityofaustin/atd-data-tech/issues/7316

Database changes, so test locally / moped-test. Also our seed data doesn't have any signal projects, so when testing locally, you will need to create some projects with signals. 

https://moped-test.austinmobility.io/moped/dashboard

Questions: 

1. You can't click an empty cell to edit in material table. The docs recommend including a default empty value. I did that for Contractor, the word "blank". I'm definitely up for changing that. 
2. How should I style the fields that are links to where to edit them? Example: funding field, current phase

| Column      | Data Source| Note | editable |
| ----------- | ----------- | --------- | --------| 
| Name / Project Name    |  project_name    | links to project | no |
| Signal IDs   |   pulls signal ids from `signal_proj_features` | will link to knack project (https://github.com/cityofaustin/atd-data-tech/issues/7556)| no |
| Project Type    |     `moped_type` from `moped_project_types`|  |no |
| Current Phase  | current_phase     | | link to timeline tab|
| Task Order   | placeholder      | placeholder: https://github.com/cityofaustin/atd-data-tech/issues/7310| |
| Contractor/Contract    |  new string column in Moped _project | | yes, inline |
| Internal Status Note   | Status update (from Project details page)     | |no |
| Funding Source   |  |   | link to tab |
| Project DO #   | `purchase_order_number` new column in moped_project  | | edit inline |
| Project Sponsor  |    project_sponsor    |  dropdown to edit | |
| Targeted Construction Start  | moped_proj_phases where phase = Construction, display the phase start date, otherwise leave blank        |  | |
| Modified Date  | updated_at       |  | no |

#### Ship list
- [x] Code reviewed 

How to Test:
Create a project with a signal / Add a signal to an existing project
Add a Construction Moped Project Phase with a start date
Add a Project DO#, Contractor by editing the cell or the row
Add a Project Sponsor either with drop down or on the Funding tab in the Project View
Update status on project summary page


amenity:
@chiaberry - Looking awesome, thank you! 

Good convo happening in Slack and I'm going to sync with users tomorrow. In the meantime, can we try a few minimal design tweaks: 
- Change table cell font size to 14px
- Change table padding from 16px to 14px
- Min-width 300px on the notes column on desktop

zappfyllc:
@chiaberry here are my initial notes. I'll keep testing but want to get you some feedback as quickly as possible!

- Excellent job on the queries from `status update`, `current phase`, `funding source` and `targeted construction start`: they appear to be working great as I could go update those values in a project, come back to the table, and BAM! Success!
- Bulk editing a row is not working for me with saving changes to "Project DO#" and "Project sponsor" (thought "Project sponsor" works with inline editing that individual cell).
- Inline editing the cell "Project DO#" is not working for me either.

General thoughts:

@amenity @johnclary This has nothing to do with Chia's work, but testing this makes me think we're going to need a pattern other than "just link to editing that tab in the project details" haha. It was a needed decision to get here, but I am immediately frustrated as a user by too many columns pursuing that pattern: I think finding a way to deliver an experience that still adheres to good data practices is probably vital for 1.0. I like the idea of a modal similar to Knack's pattern that delivers a very slimmed down version of the UI to add multiple values/aspects of data, particularly if we are confident in the use and functionality of those tabs. For tabs we aren't super confident in *cough* timeline *cough*, I'd suggest we keep the link pattern for now and continue to draw in read-only data into this table so as not to create work that will later be removed/refactored.

chiaberry:
@zappfyllc I'll get on trying to figure out whats going on with project DO and project sponsor in the edit row functionality

zappfyllc:
Note to all: I just created #7713, which covers the UI implementation for Project types, but @chiaberry did already complete the database work in #7533. I believe we might have punted on this partially because of the work around refactoring the project summary view, or the actual UI might have been my oversight (oops!).

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -0,0 +1,383 @@
...
+   * Build data needed in Signals Material Table
+   */
+  data.moped_project.forEach(project => {
+    // project status update equivalent to most recent project note
+    project["status_update"] = "";
+    if (project?.moped_proj_notes?.length) {
+      const note = project.moped_proj_notes[0]["project_note"];
+      // Remove any HTML tags
+      project["status_update"] = note
+        ? String(note).replace(/(<([^>]+)>)/gi, "")
```
frankhereford, line 65: I think the idea of stripping out HTML tags is very important, and the code here should get anything that is bounded by `<>`s, including nested closing tags. I'm not sure what would happen here with a double `<<` though. 

This is just me throwing out an idea -- HTML detection and cleanup is one of those things like timezones that is a rabbit hole that can be run down without ever finding the bottom. Would we may want to consider out-sourcing that task to a library whose only purpose is to do that and has a test suite? I found [this library, sanitize-html](https://github.com/apostrophecms/sanitize-html), which may fit the bill if we wanted to go that route at some point.

The thought reminds me of [Little Bobby Drop Tables](https://xkcd.com/327/). :)

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -0,0 +1,397 @@
...
+        getOptionLabel={e => e.entity_name}
+        onChange={(event, value) => props.onChange(value)}
+        getOptionSelected={(option, value) =>
+          option.entity_id === value.entity_id
+        }
+        renderInput={params => (
+          <TextField {...params} variant="standard" label={null} />
+        )}
+      />
+    ),
```
chiaberry, line 308: in future issues I'm going to add more custom edit components, as other cells get their own dropdowns/fields to edit and update.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -0,0 +1,383 @@
...
+                    setTimeout(() => {
+                      projectActions.cellUpdate(
+                        newValue,
+                        oldValue,
+                        rowData,
+                        columnDef
+                      );
+                      setTimeout(() => refetch(), 501);
+                      resolve();
+                    }, 500);
```
frankhereford, line 385: I don't think there's anything wrong here, but I am curious about why the browser should wait half a second before doing the update. Is the internal setTimeout's duration of one millisecond later intended to serialize two events in a known order?

frankhereford (approved):
I think this looks great! It works really well and it fits into the rest of the app's look and style very well.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
johnclary: +1 and if sanitizing is the goal y'all have [already implemented it](https://github.com/cityofaustin/atd-moped/blob/235604e270c7423c359d7b20ef7f1ec12fefcd72/moped-editor/src/views/projects/projectView/ProjectComments.js) in project comments :)

moped-editor/src/routes.js
```diff
@@ -7,14 +7,15 @@ import AccountView from "src/views/account/AccountView/AccountView";
 import StaffListView from "src/views/staff/StaffListView";
 import NewStaffView from "src/views/staff/NewStaffView";
 import EditStaffView from "src/views/staff/EditStaffView";
-import DashboardView from "src/views/reports/DashboardView/DashboardView";
+// import DashboardView from "src/views/reports/DashboardView/DashboardView";
```
mateoclarke, line 10: should we go ahead and just delete this?

moped-editor/src/routes.js
```diff
@@ -39,7 +40,8 @@ export const routes = [
       {
         path: "dashboard",
         action: "dashboard:visit",
-        element: <DashboardView />,
+        // element: <DashboardView />,
```
mateoclarke, line 43: same here?

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
mateoclarke: In Chia's defense, I believe this is the pattern we have across the app. And I think it originally came from the example/demo in the table library. In my experience testing the app, if you reduce the timeout, you sometimes hit a race condition where the data update to the server is slower and the data in the db gets updated but the data rendered in the table is the old value. Worth more triage to understand more fully, but for now, the half second delay doesn't cause any bugs.

mateoclarke (approved):
Looks good Chia! 🚢 💨 ✨ 

On your question about how to style the links, I would just do the same blue text thats there for the project name. But kinda like the discussion was going, the way that works is kinda confusing from a user perspective.

Also, just from a bigger picture level, are we imagining updating the nav for this view? That's more of a product question.

moped-editor/src/routes.js
zappfyllc: > On your question about how to style the links, I would just do the same blue text thats there for the project name. But kinda like the discussion was going, the way that works is kinda confusing from a user perspective.

+1 on link styling (but not a hill I'll die on if anyone else feels strongly) and agreed we should figure out a good pattern for editing those tabs while keeping the table in sight for the user. We can tackle that problem in subsequent issues/PRs for sure.

> Also, just from a bigger picture level, are we imagining updating the nav for this view? That's more of a product question.

@amenity and @patrickm02L can correct me here but I think we want "Dashboard" to stay and contain this signals table view even in the long-term, although the projects table that is generated will be filtered based on the user's affiliation with the project sponsor as more groups than AMD onboard. But overall we discussed the Dashboard being a combination of a few different views on the page (e.g. statistics, notifications, and other widgets may also be added to the Dashboard, as it would be the landing page for the user instead of the page containing all projects). Does that answer the question? (Sorry if I misunderstood!)

zappfyllc:
@chiaberry apologies because there might be more of these super granular changes but the only one that jumps out to me right now is making the current phase's value in the table have its first letter capitalized (e.g. "Construction") just like it is in the "Projects" tab.

Also, the bugs I identified earlier look like they're fixed! I'll let you know if I encounter any others in testing, but aside from the UX pattern I wrote about above in regards to linking to the tabs, I actually think this is looking great and I can't find any issues to get it shipped sooner than later! Then we can address some of the issues we punted along with the UX pattern.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
sergiogcx: A promise wrapped in a promise triggered by two trivial time variables that may not work if the internet is bad or if the servers are slow?

@chiaberry This is old code (from back in May) but you may want to look here:
https://github.com/cityofaustin/atd-moped/commit/715500cb704d83647446837184c1c7616f902fbf#diff-42256b926efe0131d06eacb259af2b962b743165a68c9607c19c0e59ae6dbc00R249-R253

And here:
https://github.com/cityofaustin/atd-moped/commit/715500cb704d83647446837184c1c7616f902fbf#diff-42256b926efe0131d06eacb259af2b962b743165a68c9607c19c0e59ae6dbc00R291-R297

In short, mutations are promises, why not just return the mutation with a refetch attached to it?

sergiogcx (changes requested):
Remove commented code out (as pointed out by mateo), and if possible we need to get rid of setTimeout calls.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
chiaberry: Y'all are right. I'm fixing it now. We already have this specific pattern with the timeouts in moped in a few places where we use Material Table, ill rewrite it to match the example you linked Sergio, thanks.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
chiaberry: I'll use the project comments sanitizer, I copied the regex from the code Sergio wrote here (https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryStatusUpdate.js#L49) to ensure the status would render the same in both places, but the library sanitizer will handle it better.

chiaberry:
I've removed the commented out code, upper cased the phase name and got rid of the timeouts.

I ended up not using dom purify in the project notes field because dom purify sanitizes and keeps html in the string, whereas the regex in project notes completely removes the html. I think the projects notes here should match what is seen on the summary page.

zappfyllc (approved):

zappfyllc:
I'm approving from a product perspective (and it does seem like most dev concerns have been addressed): @johnclary and @amenity, if there's anything you want to add if you have time to quickly review, but otherwise I'd say let's move forward with shipping this and tackling some of the UX problems we've spoken about in subsequent issues. We can add any minor tweaks/modifications we've forgotten in future issues as well, but I think it's good to label this as a win and we can iterate. 

@chiaberry truly excellent job tackling the data model attached to this table (it was anything but simple) and asking all the right questions! It looks great and is providing the kind of momentum we need for getting this in front of AMD users 🛵 🚦. Thank you and cheers 🍾 🥂 !

sergiogcx (approved):

amenity (approved):
@chiaberry - truly fantastic! I agree with JD that there is product/design stuff we can sort out in further issues but the functionality is looking great. 

Approving either way, but if possible could you name the page "Signal projects" in the table and HTML title? 
Screen Shot 2021-11-15 at 3 30 30 PM

PR #455, opened 2021-11-05
7567: Add knack_project_id text field to moped_project

frankhereford:
## Associated issues

This PR intends to close [7565](https://github.com/cityofaustin/atd-data-tech/issues/7565), [7567](https://github.com/cityofaustin/atd-data-tech/issues/7567), and [7566](https://github.com/cityofaustin/atd-data-tech/issues/7566).

These issues pursue the goals laid out in the epic [7311](https://github.com/cityofaustin/atd-data-tech/issues/7311). 

[7565](https://github.com/cityofaustin/atd-data-tech/issues/7565): Not-represented in the codebase, there are instructions to add fields needed to the project object and view in the Knack based Data Tracker app found in the issue.

[7565](https://github.com/cityofaustin/atd-data-tech/issues/7565): Create a `text` field called `knack_project_id` to be added to the `moped_project` table.

[7566](https://github.com/cityofaustin/atd-data-tech/issues/7566): The actual changes to the moped-editor app to provide the user a way to initially cause data to be pushed to knack and then to update that data in knack on subsequent invocations. 

## Testing
**URL to test:** This set of changes includes changes to the database schema and should be tested on the testing instance or a local deployment. 

**Steps to Test:** Add / update projects and utilize the "Sync to Data Tracker" option in the project menu to synchronize fields to knack (upserts). 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+            }
+          )
+          .then(response => response.json())
+          .then(
+            result => {
+              if (result.errors) {
+                // knack error
+                snackbarHandler({
+                  severity: "warning",
+                  message: "Success: Data Tracker sync failed.",
```
zappfyllc, line 139:
```suggestion
                  message: "Error: Data Tracker sync failed.",
```

zappfyllc: I won't comment on all of these to avoid flooding the PR, but just a straightforward typo here on the error messaging.

zappfyllc (commented):
@frankhereford, excellent work! Clean and built for reusability. Couple of comments:

- Minor typo I think: the warning error messages all say "Success" and then that the sync failed. Change this to "Failure".
- This is not by any means important for this issue, but in the future, I imagine for reusability we would want to abstract the `knack_project_id` condition from the function that returns the HTTP method needed and utilize a variable called `knack_record_id` or something that could be passed through as a column reference so that you could potentially sync other tables in Moped.
- Similarly, I could see the same thing being done with the error messaging (so that we can pass a different Knack application name in) and I might not fully understand the buildBody function but I think potentially the Knack field keys could get passed in to make this fully reusable based on schema (although I may be wrong on this front, I only glanced at it).

Excellent work, Frank! I have not tested this locally yet, but wanted to provide some quick feedback based on what I saw in the code.

zappfyllc:
@frankhereford I think I know the answer to this, but for the sake of documenting it with this work, do you want me to create a separate issue for figuring out adding the signal_ids into the payload, or roll that requirement up into this PR? @chiaberry is going to be addressing some of the same dependencies via #466. I don't believe the Knack record id of the respective signals related to a project are currently making it into the Moped DB (there's a JSONB column in `moped_proj_features` that is currently lacking that info, but that's where we would draw the signal Knack record IDs from and place into this payload).

cc: @johnclary

frankhereford:
> @frankhereford I think I know the answer to this, but for the sake of documenting it with this work, do you want me to create a separate issue for figuring out adding the signal_ids into the payload, or roll that requirement up into this PR? 

@zappfyllc - thank you for mentioning that. I had become focused on the direction given in one of our recent meetings to work to the unchanged, original scope of the issues as they had been written, and I had let the idea of the signal ID payload slip. Adding it is quick and easy (sounds like famous last words!) - I lean towards wrapping them into this PR. If it's anything except quick and easy, I'll slack you and we can go from there - that work?

chiaberry:
@zappfyllc I have this PR where I am saving the knack ids for the signals https://github.com/cityofaustin/atd-moped/pull/466/files

sergiogcx (changes requested):
As of this moment you still have branch conflicts, plus I will be making some additional observations:

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+    {
+      project,
+      closeHandler,
+      snackbarHandler,
+      sceneNumber = 514,
+      viewNumber = 3047,
+      knackApplicationId = "6167314778435d001ea3e7cb",
+    },
+    ref
+  ) => {
```
sergiogcx, line 24: This feels like an anti-pattern:
1. The default values should probably be part of an environment variable, since I suspect we are going to have multiple stages to test this code with.
2. React.forwardRef probably doesn't belong there for this large component, I don't see ref being used or passed explicitly within this code, though I suspect MenuItem may require it. 
3. We should consider going the custom hook route for real reusability.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+        "/views/view_" +
+        viewNumber +
+        "/records";
+      if (project.moped_project[0].knack_project_id) {
+        // existing record
+        url = url + "/" + project.moped_project[0].knack_project_id;
+      }
+      return url;
+    };
+
```
sergiogcx, line 43: It has been a convention for us to only use let only for mutable variables, const for everything else.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+          .then(response => response.json())
+          .then(
+            result => {
+              if (result.errors) {
+                // knack error
+                snackbarHandler({
+                  severity: "warning",
+                  message: "Success: Data Tracker sync failed.",
+                });
+                return Promise.reject(result);
```
sergiogcx, line 116: Use `.catch()` instead of `.then(..., onRejected)`, onRejected never handles rejected promises from the same then callback, you are also prone to have race conditions if the chain is misqueued.

This implementation is not necessarily bad, it simply is cryptic to read and difficult to debug because the trace of Promise.reject gets lost in the chain.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+              return Promise.reject(error);
+            }
+          )
+          .then(knack_record => {
+            mutateProjectKnackId({
+              variables: {
+                project_id: project.moped_project[0].project_id,
+                knack_id: knack_record.record.id,
+              },
+            });
```
sergiogcx, line 195: Shouldn't this also be part of the chain? What would happen if this mutation fails?

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -4,6 +4,8 @@ import { Link as RouterLink, useParams, useLocation } from "react-router-dom";
 import { createBrowserHistory } from "history";
 import { makeStyles } from "@material-ui/core/styles";
 
+import KnackSync from "./KnackSync.js";
+
```
sergiogcx, line 8: Another small convention, we don't include the `.js` part of the file name

moped-editor/src/views/projects/projectView/KnackSync.js
johnclary: @frankhereford [here's](https://github.com/cityofaustin/atd-moped/blob/3f5539935dec8a93dd7163ff69d0adb9ff2037a5/moped-editor/src/utils/mapHelpers.js#L1060) a `.catch()` example for you.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+    ref
+  ) => {
+    /**
+     * Function to build the correct Knack URL to interact with based on properties and if there will be an
+     * update or an initial sync.
+     * @returns string
+     */
+    const buildUrl = () => {
+      let url =
+        "https://api.knack.com/v1/pages/scene_" +
```
johnclary, line 32: for string concats/substitutions, use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals):
 ```javascript
`https://api.knack.com/v1/pages/scene_${sceneNumber}/views/${viewNumber}/records`
```

File: moped-editor/src/views/projects/projectView/KnackSync.js (hunk 2 above)
Comment (johnclary) [reply moped-editor/src/views/projects/projectView/KnackSync.js:24]: i agree. @frankhereford unless i'm missing something you can just drop the forwardRef and declare the component with 
```javascript
const KnackSync = ({
      project,
      closeHandler,
      snackbarHandler,
    }
  ) => { ... }
```

johnclary: At a glance it looks like you could do away with the `project.moped_project[0]` accessor by passing the value of it as the `project` pop. I definitely prefer that vs repeating `project.moped_project[0]` throughout.

moped-editor/src/views/projects/projectView/KnackSync.js
johnclary: i think your code will be cleaner if you just assign the url to a variable every render (or use a hook), rather than calling `buildUrl()` within your fetch.  ideally, you would hoist this function out of the component and use it like so:

so
```javascript
const url = buildUrl({scene, view, projectId});
```
i'm not very consistent when deciding when to hoist functions out of components, but the basic idea is that it is more performant, since the function itself is otherwise created on every render. performance won't matter here, but the code will look nicer.

moped-editor/src/views/projects/projectView/KnackSync.js
johnclary: related to above, when you chain these together i believe you can use a single `.catch()` at the end of the chain that will catch errors at any stage.

that said, i'm seeing where Sergio was going suggesting a hook. it would be quite bad if a knack project where created but the mutation failed. it would add quite a bit of state to isolate the two separate calls with hooks, but that might make it easier to build in some resiliency. if we want to go that route i think it's worthy of a separate PR.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -148,21 +152,43 @@ const ProjectView = () => {
...
+  const handleSnackbarClose = () => {
+    setSnackbarState(DEFAULT_SNACKBAR_STATE);
+  };
+
+  /**
+   * Fuction which can be passed to child component to allow it to invoke a snackbar which
+   * will persist even after that child component has been deconstructed or otherwise hidden.
+   */
+  const handleSnackbarOpen = snackbarState => {
+    snackbarState.open = true;
```
johnclary, line 187: it looks like you're assigning a value to a state object, which will fail silently. the tick is to deep copy the state object with destructuring:

```javascsript
  const handleSnackbarOpen = snackbarState => {
    let snackbarStateNew = {...snackBarState}
    snackbarStateNew.open = true;
    setSnackbarState(snackbarStateNew);
    setTimeout(() => setSnackbarState(DEFAULT_SNACKBAR_STATE), 3000);
  };
```

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -114,6 +114,7 @@ const ProjectNameEditable = props => {
         if (props?.setIsEditing) props.setIsEditing(false);
         setIsEditing(false);
         setTimeout(() => setSnackbarState(DEFAULT_SNACKBAR_STATE), 3000);
```
johnclary, line 116: can this be avoided by using the `autoHideDuration` prop on the snackbar component?
https://mui.com/components/snackbars/

moped-editor/src/views/projects/projectView/KnackSync.js
johnclary: fwiw [here's](https://github.com/cityofaustin/atd-moped/blob/3f5539935dec8a93dd7163ff69d0adb9ff2037a5/moped-editor/src/utils/socrataHelpers.js#L8) an example of a hook-ified fetch.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+     * @returns object
+     */
+    const buildHeaders = () => {
+      let headers = {
+        "Content-Type": "application/json",
+        "X-Knack-Application-Id": knackApplicationId,
+        "X-Knack-REST-API-Key": "knack",
+      };
+      return headers;
+    };
```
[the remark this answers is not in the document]
johnclary: agree and am not sure where to build in a protection against WSOD. what do you suggest @sergiogcx ?

johnclary:
@frankhereford thanks for you work on this tricky functionality! i'm happy to pair if you want to talk through any of these requests. i think the main priority on this PR is to make sure we don't introduce any WSOD, and then branch from it for any larger re-factoring. but i defer to @sergiogcx and @mateoclarke.

frankhereford:
> @frankhereford thanks for you work on this tricky functionality! i'm happy to pair if you want to talk through any of these requests. i think the main priority on this PR is to make sure we don't introduce any WSOD, and then branch from it for any larger re-factoring. but i defer to @sergiogcx and @mateoclarke.

@johnclary Thanks! I'm jumping into these change requests now, and I'll let you know if I get snagged up on anything. I am eager to get my new code up to par - thanks for checking it out.

moped-editor/src/views/projects/projectView/KnackSync.js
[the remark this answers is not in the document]
sergiogcx: @johnclary I was thinking of something like this maybe:

```
const getHttpMethod = (project.moped_project?.[0]?.knack_project_id ?? false) ? "PUT" : "POST;
```

moped-editor/src/views/projects/projectView/KnackSync.js
mateoclarke: agree with John on both of these suggestions.

patrickm02L:
Observations about the Sync to Data Tracker feature regarding UX/UI:
- Micro-interaction to indicate that sync is happening because it lags a few second after it's selected
- Sync is hidden in the ellipses menu; make more intentional or visible.

https://user-images.githubusercontent.com/5566727/142257874-7f938186-0052-425c-89fb-1352c9bc3a2c.mp4

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Function to determine the HTTP method to use base on if there will be an update or initial post to Knack
+     * @returns string
+     */
+    const getHttpMethod = () => {
+      let method = "POST";
+      if (project.moped_project[0].knack_project_id) {
+        method = "PUT";
+      }
```
mateoclarke, line 52: You could write lines 49-52 as:
```js
const method = project.moped_project[0].knack_project_id ? "PUT" : "POST";
```

mateoclarke: Oh nvm, I see Sergio made an even better suggestion below along the same lines.

mateoclarke (commented):
Good feedback on this PR from @sergiogcx and @johnclary. I echo their suggestions.

mateoclarke:
> Observations about the UX/UI of the Sync to Data Tracker feature:
> 
> * Micro-interaction to indicate that sync is happening because it lags a few second after it's selected
> * Sync is hidden in the ellipses menu; make more intentional or visible.
> 
>  Patrick.s.Test.Project.-.17.November.2021.mp4

On Patrick's suggestion, it seems like it would be good to have a "sync in process" or pending snack bar message become visible once you click the Sync to Data Tracker button, that then gets replaced by the success (or error) message.

moped-editor/src/views/projects/projectView/KnackSync.js
mateoclarke: Screen Shot 2021-11-17 at 12 47 46 PM

mateoclarke:
This might be out of scope but seems like it would be a helpful enhancement is to keep the success Snackbar visible longer (say 5 sec) and include in the message a link to the associated project in Knack. That way, users can navigate to the page and confirm the sync performed as expected.

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: Thanks for pointing that out! I've aligned the wording to the error states, added a missing snackbar error and made sure that all sentences have a `.` at the end of them.

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: This is very helpful input - thank you both. 

I totally agree that the values that define the knack endpoint should be pulled out of the app, and I've moved them into env variables. 👍 

I also think that only handling a slice of the hasura response down as a property to the component cleans up a bunch of repetitious and ugly code, so that's a ✅ too.

Re the React.forwardRef, initially was using the vanilla function form like John suggested, but at some point, I started getting this error. 

image

I kinda blindly followed the advice it offered, and I managed to get a state where I was not introducing another error in the console, so I went the forwardRef route.

That said, I don't fully understand why it's correct, and I'd be hard pressed to justify the choice. Based on @sergiogcx's comment, I suspect he is on the right track because this component needs a way to reach up a level in the hierarchy to close the menu. I'd be eager to read up or chat with y'all about what is the real answer here for long term maintainability and reuse.

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: That makes perfect sense and is a great convention. I have changed the constant variables to use the const declaration.

moped-editor/src/views/projects/projectView/KnackSync.js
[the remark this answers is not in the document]
frankhereford: * Totally agree, the header object needed is the same for the initial sync HTTP call and for all following syncs. I've defined it as an object and not a function - great catch.

* @sergiogcx's suggested code here is really sharp and exactly what we need to get the right HTTP method. I don't think we can get away with computing the value once though and hanging on to it as a const because it's going to be a POST for the initial sync and then PUTs for all following calls. Having said that, there may be possible to rely on react to tear down the component when the menu closes and recompute the correct HTTP method when the component is instantiated when the menu is opened again.

I'll do some testing to see if that's how things work - please let me know if there's something I'm missing.

moped-editor/src/views/projects/projectView/ProjectView.js
frankhereford: Got it - thanks for pointing that out.

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: @mateoclarke - I agree. Both of these suggestions are on point, and I have adopted them both. Thanks!

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: These suggestions are helpful, thanks y'all. I have incorporated them.

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
frankhereford: Yes! Thanks for pointing this out and asking me to revisit that documentation. I had read it, seen that **_property_** and misconstrued it to be something I could put in the snackbar's state object, which obviously didn't work. I have put in the correct property now, and also adjusted the time to 5000 milliseconds as @mateoclarke suggested. Thanks for this help!

moped-editor/src/views/projects/projectView/ProjectView.js
frankhereford: I have adopted this destructuring into a deep copy, but I can't quite claim to fully understand what's going on here and why this fails silently. More reading in my near future, but for now, thank you for catching this subtle error!

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -148,21 +152,43 @@ const ProjectView = () => {
...
+  };
+
+  /**
+   * Function which can be passed to child component to allow it to invoke a snackbar which
+   * will persist even after that child component has been deconstructed or otherwise hidden.
+   */
+  const handleSnackbarOpen = snackbarState => {
+    let snackbarStateCopy = {...snackbarState};
+    snackbarStateCopy.open = true;
+    setSnackbarState(snackbarStateCopy);
```
johnclary, line 189: looks good. @frankhereford [here's](https://stackoverflow.com/a/40309023/2103784) the best explainer i could find on this. a word of warning with this destructuring approach is that iirc it will only work with shallow objects. if you had a state variable with nested objects you would need to use another method to deep copy.

johnclary:
@frankhereford at a glance the changes look good. going to ask @mateoclarke and @chiaberry to take a deeper look. i am confounded by the need for forwardRef but i haven't tested this locally.

frankhereford:
> @frankhereford at a glance the changes look good. going to ask @mateoclarke and @chiaberry to take a deeper look. i am confounded by the need for forwardRef but i haven't tested this locally.

Hi @johnclary, thanks for giving this another pass. I just returned from getting a bite to eat and getting caught up on some smaller items that I had on the TODO list from the end of last week, and I am jumping back into this issue / PR right now. I'm super eager to get it out the door, and your feedback is invaluable. I will communicate with Mateo & Chia to get their thoughts and also do some more learning via reading on my part as well. Thanks again!

chiaberry:
I was poking around to see what I could find about forwardRef in this context and found this issue: https://github.com/mui-org/material-ui/issues/15903

They recommended wrapping the functional component in a div, but I didnt find that to fix anything in this scenario. It seems to me the fowardref is needed because of the dropdown menu that uses the MaterialUI Popover component.

johnclary:
interesting. what's throwing me off is that we're not actually doing anything with the `ref`, e.g. passing it to a child component...

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: It took me a while to understand what you were pointing out, but, yes - this now makes perfect sense to me. Thank you for pointing this out, and I agree fully.

moped-editor/src/views/projects/projectView/KnackSync.js
frankhereford: It should for sure - thank you for pointing out that this wasn't handled correctly.

frankhereford:
> @frankhereford I think I know the answer to this, but for the sake of documenting it with this work, do you want me to create a separate issue for figuring out adding the signal_ids into the payload, or roll that requirement up into this PR? @chiaberry is going to be addressing some of the same dependencies via #466. I don't believe the Knack record id of the respective signals related to a project are currently making it into the Moped DB (there's a JSONB column in `moped_proj_features` that is currently lacking that info, but that's where we would draw the signal Knack record IDs from and place into this payload), but I believe @chiaberry's work in #466 will fix that.
> 
> cc: @johnclary

@zappfyllc - Hi JD, we had discussed adding the signal_id as part of this PR, but in a conversation with @mateoclarke this afternoon, he suggested that any further enhancements related to this PR be spun off into their own issue. You were wisely alluding to that in your comment here. 

Along these lines, I have made [a new issue, #7925,](https://github.com/cityofaustin/atd-data-tech/issues/7925) to capture this information and enhancement.

frankhereford:
> Observations about the UX/UI of the Sync to Data Tracker feature:
> 
> * Micro-interaction to indicate that sync is happening because it lags a few second after it's selected
> * Sync is hidden in the ellipses menu; make more intentional or visible.
> 
>  Patrick.s.Test.Project.-.17.November.2021.mp4

@patrickm02L - Thank you for this thoughtful feedback. I have created a new issue [here, #7927](https://github.com/cityofaustin/atd-data-tech/issues/7927), to track this enhancement request.

amenity (approved):
Works so beautifully, @frankhereford ! Thanks.

amenity:
@mateoclarke mentioned that you thought the code on this was good to go... Approve and then @frankhereford can merge net week?

mateoclarke:
> @mateoclarke mentioned that you thought the code on this was good to go... Approve and then @frankhereford can merge net week?

Correct, I re-reviewed the code but I didn't have a chance to test if the records were being created in the test Knack instance before I left town last week. After syncing up on it today on which Knack instance to check, sounds like you were able to follow up and product test. 👍

mateoclarke (approved):
🚢 💨 ✨

PR #456, opened 2021-11-08
Add record to moped_users for Frank in seed data

frankhereford:
This PR adds a record for Frank to the `moped_users` table. This isn't required for local development, but it is required to use the test instance. Not having a record in the moped_users table causes an infinite `Navigate to..` loop when there is no record to be found that matches the data returned by cognito. Until this PR is merged, a viable workaround for me to test is manually run a mutation to insert a record for myself in the `moped_users` table after each deployment of the testing instance.


chiaberry (approved):
added bonus that it should fix it for me too --

patrickm02L (approved):
Fixed it for me as well

frankhereford:
Here's an interesting observation to add some more information to this situation:

* Once I had gotten logged in successfully in the testing instance (having added a record by hand to the `moped_users` table, I was able to use the site as expected.
* Next, when I deployed a new branch out to the testing instance, which caused the rebuilding of the database through migrations & seed data -- **which caused my temporary record in `moped_users` to get wiped out** -- I was still able to use the site. 
* As expected, if I log-out and then try to login again, I go back into the navigate-to-loop, and require adding a temp record in again to be able to reestablish my session.

mateoclarke (approved):
No issues here. Just wondering if in general we should update the seed data vs doing this via a migration. I guess since in this case, the issue is in staging and not in production, its best to do it as seed data since that won't run in prod.

PR #450, opened 2021-11-02
Implement moped project partners #7358

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/7358

- Database changes
- Implements project partners UI

### Live Demo
Please run in moped-test

PR #483, opened 2021-11-30
Add project types to signal project dashboard and project summary views

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/7713

Added projects types dropdown edit field to the Signal Project Dashboard and the Project summary. 
I had to update the database because when I first added the project types table, I did not include a way for project types to be soft deleted (status_id).
I also rearranged the fields on the project summary view. 

## Testing
**URL to test:** 
Need to deploy to moped-test to test or run locally, there are db changes. 

**Steps to test:**

On the project summary page, add new project types to the project. Test removing a type. 

On the signals dashboard, test editing the project types by editing in the cell and by editing a whole row. 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
```diff
@@ -0,0 +1,190 @@
...
+   * Helper initial state lists
+   */
+  const typeList = data?.moped_types ?? [];
+  const typeDict = typeList.reduce(
+    (prev, curr) => ({
+      ...prev,
+      ...{ [curr.type_id]: curr.type_name },
+    }),
+    {}
+  );
```
sergiogcx, line 45: [Looking back](https://github.com/cityofaustin/atd-moped/pull/450/files#diff-b26f3e28654dd086f13a0473547d1ae187d5222c3156708a18623633ca34a270R39-R45) I wonder if I should have used a `.forEach` and something like `typeList[n.type_id] = n.type_name`.  It looks great, though! 👍  It works as expected.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
```diff
@@ -0,0 +1,190 @@
...
+   */
+  const handleProjectTypesSave = () => {
+    // Original type Ids
+    const oldTypesList = originalTypesIds;
+    // Get selected types ids
+    const newTypesList = selectedTypes;
+    // Retrieves the ids of oldTypesList that are not present in newTypesList
+    const typeIdsToDelete = oldTypesList.filter(t => !newTypesList.includes(t));
+    // Retrieves the ids of newTypesList that are not present in oldTypesList
+    const typeIdsToInsert = newTypesList.filter(t => !oldTypesList.includes(t));
```
sergiogcx, line 88: Another throwback for me, maybe there is a way to improve the O(n*n) performance, but the list is so small its kind of irrelevant. 👍  Feel free to challenge these as you code.

sergiogcx (approved):
Works great! 🚢  🚀  📦

frankhereford (approved):
I have gone over the code, but not in super detail. I have tested the intended functionality based on your (really helpful) PR initial comment, and I find everything to work exactly as one would expect. 💯 

I think the way this UI is designed and coded up feels really good to use -- I think this is great work.

amenity (changes requested):
This looks and works great! I guess that you're thinking forward to future use cases, @zappfyllc, in using a multi-select? Just noticing that all the projects currently in Knack only have one type. 

My only request, just added to [docs](https://atd-dts.gitbook.io/moped/dev-guides/design-system/component-styles#select-fields), is that we use our primary color for the boxes instead of the secondary.

chiaberry:
@amenity I will make that change. I'm going to go ahead and update the project partner checkboxes to also be blue so they both match.

amenity (approved):
LSTM✨

mateoclarke (approved):
🚢 💨 ✨

zappfyllc (approved):
Looks and works great @chiaberry!

PR #491, opened 2021-12-06
Render project status badge on Projects page

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/7827

## Testing
**URL to test:** 
https://7827-project-status-badge--atd-moped-main.netlify.app/moped/projects
**Steps to test:**

Confirm that the badges on the projects page match the badge on the project summary page.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/GridTable.js
+++ b/moped-editor/src/components/GridTable/GridTable.js
@@ -92,11 +92,14 @@ export const getSearchValue = (query, column, value) => {
  * GridTable Component for Material UI
  * @param {string} title - The title header of the component
  * @param {Object} query - The GraphQL query configuration
+ * @param {Object} referenceData - optional, static data used in presentation
  * @return {JSX.Element}
  * @constructor
  */
-const GridTable = ({ title, query, searchTerm }) => {
+const GridTable = ({ title, query, searchTerm, referenceData }) => {
   const classes = useStyles();
+
+  console.log(referenceData)
 
   /**
    * State Management for pagination
@@ -358,7 +361,7 @@ const GridTable = ({ title, query, searchTerm }) => {
       <ProjectStatusBadge
         status={statusId}
         phase={label}
-        projectStatuses={data?.moped_status ?? []}
+        projectStatuses={referenceData?.moped_status ?? []}
         condensed
       />
     ) : (
--- a/moped-editor/src/components/GridTable/GridTable.js
+++ b/moped-editor/src/components/GridTable/GridTable.js
@@ -55,9 +55,6 @@ const useStyles = makeStyles(theme => ({
     "text-transform": "capitalize",
     "white-space": "pre-wrap",
   },
-  tableChip: {
-    "text-transform": "capitalize",
-  },
 }));
 
 /**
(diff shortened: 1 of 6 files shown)


amenity (approved):
LSTM! ✨ 😍

mateoclarke (approved):
Status badges look great! ✨ 

prodigal-son

moped-editor/src/queries/project.js
```diff
@@ -102,6 +102,19 @@ export const SUMMARY_QUERY = gql`
   }
 `;
 
+export const STATUS_QUERY = gql `
+  query StatusQuery {
```
johnclary, line 106: i've been meaning to ask y'all, do we ever make use of [hasura caching](https://hasura.io/docs/latest/graphql/cloud/response-caching.html) for our lookup value queries? cc @mateoclarke @frankhereford @sergiogcx @tillyw

moped-editor/src/components/GridTable/GridTable.js
```diff
@@ -507,10 +490,13 @@ const GridTable = ({ title, query, searchTerm }) => {
                                       <>
                                         {query.config.columns[
                                           column
-                                        ].hasOwnProperty("chip")
-                                          ? buildChip(
+                                        ].hasOwnProperty("badge")
```
johnclary, line 493: can this ☝️ be:
```javascript
 query.config.columns[column]?.badge
```
?

sergiogcx (approved):

moped-editor/src/queries/project.js
sergiogcx: I didn't know it existed until now, it seems to be a feature of hasura cloud only? @johnclary 

>Query response caching is available for Hasura Cloud projects on the Standard (pay-as-you-go) tier and above

mateoclarke: We don't afaik. I believe its a Hasura Cloud (not Core) feature and only v2+.

johnclary (approved):
🌬️ ⛵

moped-editor/src/queries/project.js
johnclary: ah, my bad! thanks

mateoclarke: on second read, it seems like it is configurable in the open source engine, but not sure what setup is required.
https://hasura.io/learn/graphql/hasura-advanced/performance/1-caching/

moped-editor/src/components/GridTable/GridTable.js
sergiogcx: This is old code from vision zero, iirc. I believe hasOwnProperty is there because we need a True/False return, so it is only matter of making it a truthy statement:

moped-editor/src/queries/project.js
johnclary: i see, hasura makes use of postgres's caching abilities. however caching at the api layer is only a cloud feature.

regardless, apollo's cache should at least be caching these queries during the user's session, right?

sergiogcx: Yes, apollo caches the majority of requests locally (in the browser of the user). If we wanted server-side cache, we can put it behind CloudFront and use http headers for controlling cache hit/miss and ttl.

PR #494, opened 2021-12-08
Implement Data Tracker Task Order Integration

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/7310

This feature includes a new [Socrata dataset for Task Orders](https://data.austintexas.gov/resource/ggav-ufvc). 

A next step will be the work required to establish an ETL to keep the Task Order dataset fresh in Socrata. We will also need to follow-up on this work to put the active Task Order value on one of the Project Details tabs (Summary or Funding?)

## Testing
**URL to test:** 
This PR includes DB schema changes so testing requires deployment to the test instance or local env.

**Steps to test:**
- Go to the Signals Dashboard
- Add one or multiple Task Orders using text search, dropdown, etc.
- Delete existing Task Orders

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/utils/autocompleteHelpers.js
```diff
@@ -0,0 +1,14 @@
+/*
+/ MUI autocomplete filter function which limits number of options rendered in select menu
+*/
+export const filterOptions = (options, { inputValue, getOptionLabel }) => {
+  // limits options to ensure fast rendering
+  const limit = 40;
```
frankhereford, line 6: This is totally fine; it works and limits the length of the dropdown. It does feel like burying a magic number in a file when it's really a configuration to the application that could be stored in [moped-editor/src/config.js](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/config.js), [moped-editor/constants/signalComponent.js](https://github.com/cityofaustin/atd-moped/tree/main/moped-editor/src/constants), or similar. None-the-less, it's working great - 👍

moped-editor/src/utils/taskOrderComponentHelpers.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { TextField } from "@material-ui/core";
+
+/*
+ * Socrata Endpoint
+ */
+export const SOCRATA_ENDPOINT =
+  "https://data.austintexas.gov/resource/ggav-ufvc.json?$limit=9999";
```
frankhereford, line 8: Is the 9999 value for limit intended to prevent enormous reset sets from coming back from Socrata or to ask that it give you every conceivable task order out the `[TEST] Mobility Project Task Orders` dataset. Given that there are currently 2,650 records total in the dataset, I presume it's the latter. 

If that's the case, I believe you can eliminate the limit argument. Socrata 2.1 endpoints have a default limit value of 'unlimited,' [according to this page](https://dev.socrata.com/docs/queries/) in their documentation. Socrata indicates that this dataset is serving up data using the latest version of their API. 

No matter the answers to the questions above, this still works and is 👍

frankhereford, line 8: Will this dataset graduate to become the production dataset? If so, 👍, and if not, I'm curious to learn what the mechanism will be to configure this URL based on environment -- I'm not sure what convention is used for that in the application currently.

frankhereford (approved):
I spun this up on my local instance and tested it - works great. I've gone through the code also, and I left a couple of comments, but they are nothing that should hold up merging this at all. 

I really like how the new component works in terms of how it presents the chips neatly grouped inside the component's box -- it's immediately clear that you're editing a set and removing items from the set is easy too. I think it's implemented really well - thank you!

amenity:
Looks fantastic, @mateoclarke! 

As far as what to do when the width of the data extends excessively, @patrickm02L and I think that wrapping to the next line is the best basic option. I tried out a width of 265px here and it looks about right:
Screen Shot 2021-12-30 at 1 19 33 PM

Even better, we think, might be listing the number and showing the "name" as a tooltip would look most tidy: 
Screen Shot 2021-12-30 at 1 18 16 PM
But assuming that is complex enough to be a standalone enhancement, we are good to do the wrapping for this PR and open another issue. 

For posterity, I was initially thinking of [letting them get cut off and stay single line](https://user-images.githubusercontent.com/1463708/147785450-c21944a4-5577-43ab-87fd-73203732dcbd.png), but that hides a lot of information from the user including how many task orders there are total.

moped-editor/src/utils/taskOrderComponentHelpers.js
chiaberry: We have the same 9999 pattern in `signalComponentHelpers.js`, when we query that socrata endpoint, just fyi

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -236,7 +239,21 @@ const SignalProjectTable = () => {
     {
       title: "Task order",
       field: "task_order",
-      // placeholder for task order issue
+      customEdit: "taskOrders",
+      emptyValue: "blank",
```
chiaberry, line 243: The other emptyValues on the table are now `emptyValue: "-"`

chiaberry (approved):

sergiogcx (approved):

moped-editor/src/utils/taskOrderComponentHelpers.js
mateoclarke: Interesting what you cited about Socrata's latest API documentation Frank. I added this limit because when I was working on this it was limiting me at the first 1,000. This is what I think you'll see if you [visit this URL](https://data.austintexas.gov/resource/ggav-ufvc.json) 

Screen Shot 2022-01-04 at 3 49 00 PM

This auto-enforced limit of 1000 has been a common experience in my work with the SODA API but I would love if they changed it to `unlimited`, but I guess seeing is believing for me.

mateoclarke: Yes, I anticipate that once we are automatically updating this dataset with an ETL process, we'll go ahead and drop the `[TEST] ` prefix in the title. I don't think we'll need to worry about maintaining multiple version of the Socrata dataset depending on environment.

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
mateoclarke: Thanks, I had a feeling I guessed wrong on what the decision was for default empty value text I and I really appreciate this correction. I'll fix this in a commit.

amenity:
@mateoclarke - are you able to wrap the lines in this PR, or should I open a new PR for that and approve this one?

moped-editor/src/utils/taskOrderComponentHelpers.js
johnclary: Good catch on the docs @frankhereford - I take that to mean there is no upper limit on how many records you can request, but you must specify a limit if you want more than the default (1k). That squares w/ my experience querying Socrata.

moped-editor/src/utils/socrataHelpers.js
```diff
@@ -29,3 +29,33 @@ export const useSocrataGeojson = url => {
   error && console.error(error);
   return { features, error, loading: !features && !error };
 };
+
+/**
+ * Fetches data from a socrata json endpoint
+ * @param {String} url - The url endpoint
+ * @return {Object} data (json object), loading, and error states
+ */
+export const useSocrataJson = url => {
```
johnclary, line 38: @mateoclarke this looks identical to `useSocrataGeojson`, right? I think we can nix `useSocrataGeojson` and use this everywhere, you'll just need to swap out a few things for the signal autocomplete stuff.

moped-editor/src/utils/autocompleteHelpers.js
johnclary: thanks for abstracting this!

moped-editor/src/views/projects/signalProjectTable/TaskOrderAutocomplete.js
```diff
@@ -0,0 +1,49 @@
...
+  SOCRATA_ENDPOINT,
+} from "src/utils/taskOrderComponentHelpers";
+import { filterOptions } from "src/utils/autocompleteHelpers";
+
+/**
+ * Material Autocomplete wrapper that enables selecting a task order record from a
+ * Socrata dataset and setting it.
+ * * @param {Object} classes - MaterialUI style object
+ * * @param {Object} props - passes down props parent MUI Table
+ * * @param {Array} value - passes an array of objects, each representing a task order
```
johnclary, line 16: might be good to document here what the shape of each task order {Object} looks like. ie what properties it should have.

moped-database/migrations/1640641794224_alter_table_public_moped_project_add_column_task_order/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."moped_project" ADD COLUMN "task_order" jsonb NULL;
```
johnclary, line 1: i suggested below that we might document what this json blob looks like

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/utils/taskOrderComponentHelpers.js
frankhereford: @johnclary @mateoclarke - You're both so right! I misinterpreted the parenthetical in the "Default" column in those docs to mean that on 2.1 endpoints, the default would be unlimited, but as you pointed out, it is actually a default of 1K with the _option_ to turn it up to whatever amount of data you want returned. Thanks for clarifying that for me - TIL! 🙏

moped-editor/src/utils/socrataHelpers.js
mateoclarke: will do.

mateoclarke: I was a little stumped bc they appear identical but replacing them breaks the signal autocomplete. I think there's a smart way to abstract and make this dual purpose, but the way the geojson expects a feature attribute of the result is significant in this case. I could do some more nested if/else but I think I'll leave it for now.

Screen Shot 2022-01-11 at 1 55 04 AM

johnclary: ah, yes. i think the format of the response JSON is merely dependent on the file extension you put in the endpoint url (`.json` or `.geojson`). i'll open an issue.

PR #499, opened 2021-12-16
Drop project list view before creating new one

chiaberry:
## Associated issues

none

This fixes the release candidate not building. Recent commits to main have edited the Project List SQL View, but once it was merged with production, errors appeared. 

> (so it could be that this commit got patches into production, that never got merged into main https://github.com/cityofaustin/atd-moped/commit/a8f9f922b54c6c8d79ff33fb05670c0a80785ce9? (edited) )

Our newer Views drop a column, so I dropped the View before creating or replacing: https://dba.stackexchange.com/questions/77564/postgresql-drop-column-from-view

This PR also removes the other references to that dropped column. 

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1637823185562_alter_project_list_view/up.sql
+++ b/moped-database/migrations/1637823185562_alter_project_list_view/up.sql
@@ -1,3 +1,5 @@
+DROP VIEW "public"."project_list_view";
+
 CREATE OR REPLACE VIEW "public"."project_list_view" AS 
  SELECT mp.project_uuid,
     mp.project_id,
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -4737,9 +4737,7 @@
       - project_length
       - project_name
       - project_order
-      - project_partner
       - project_priority
-      - project_sponsor
       - project_team_members
       - project_uuid
       - start_date
@@ -4768,9 +4766,7 @@
       - project_length
       - project_name
       - project_order
-      - project_partner
       - project_priority
-      - project_sponsor
       - project_team_members
       - project_uuid
       - start_date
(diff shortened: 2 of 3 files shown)


sergiogcx (approved):

sergiogcx (commented):
I don't like the idea that both main and production will have different files. Shouldn't we be merging main into the release candidate?

moped-database/migrations/1637823185562_alter_project_list_view/up.sql
```diff
@@ -1,3 +1,5 @@
+DROP VIEW "public"."project_list_view";
```
sergiogcx, line 1: I know this is somewhat harmless, but it feels wrong. production and main should be identical. Not to mention migrations do not get re-run if they have been already applied.

chiaberry: yeah it feels weird to me too. Apparently production has had a migration since June that main hasn't had. What is another way to fix it? We need to drop a column from the view.

mateoclarke: Could we apply this line just to the new migration that hasn't been run in prod yet? `1638905763818_run_sql_migration/up.sql`

sergiogcx: We had a couple good ways to fix it, one would have been ideally a migration that is friendly to both main and production and does the exact same thing. Also, we always have the option of using hasura's cli.

Again, my only problem is that we should make production like a mirror of main, and this does the opposite. If we continue, we are gonna find ourselves troubleshooting conflicts in code and in the database.

chiaberry:
What would the migration that is friendly to both look like?

PR #500, opened 2021-12-16
Reconcile migration mismatch between production and main

chiaberry:
## Associated issues

Adds the migration that exists in production that is not in main. 

## Testing
**URL to test:** 

**Steps to test:**

I have tested locally and it runs. Will test on Test. Please pull and run locally if you can. 

---
#### Ship list
- [ ] Code reviewed 


mateoclarke (approved):
Worked for me locally and the code looks right! 🚢 💨 ✨

sergiogcx (approved):
It works in test and local, so I think that should be ok.

PR #522, opened 2022-01-12
Add Knack Signal ID to the Moped Component Payload & ETL Script

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/7925.

Please see [child issue #8213](https://github.com/cityofaustin/atd-data-tech/issues/8213) for additional work which has been requested related to this.

## Testing

https://deploy-preview-522--atd-moped-main.netlify.app/moped/

I find this easier to test locally as you may have to manipulate some records to facilitate testing, but it should work on the other deployments. 

* Visit knack builder's development database for moped ("TEST | Austin Transportation Data Tracker | MOPED DEV - 12 OCT 2021") and clear out any records that have a Moped Project ID. 
* Similarly, make sure you have no records in the Moped Database which indicate they have been sync'd to Knack by looking for projects with a `knack_project_id`
* Add some signals to a project that has not been sync'd
* Sync them using the UI element on the project summary page.
* Observe the record created in the Knack app

#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -29,18 +29,33 @@ const getHttpMethod = knackProjectId => {
   return knackProjectId ?? false ? "PUT" : "POST";
 };
 
+/**
+ * Function to map the signal IDs from a project object into an array and return the array length.
+ * @returns integer
+ */
+const countSignalsInProject = project => {
+  const signalIds = project.moped_proj_features.map(
+    feature => feature.location.properties.signal_id
```
johnclary, line 38: @frankhereford PR #553 is going to break this line (L38). the fix will be to change this:
```
feature => feature.location.properties.signal_id
```
to 
```
feature => feature.feature.properties.signal_id
```

i don't foresee any conflicts—so we can decide how to orchestrate the change after both of our branches are stable

johnclary: @frankhereford i was wrong all points—let's sync when both our branches are stable 🤪

frankhereford: Sounds good - looking forward to it! 🙃

moped-editor/.env-cmdrc
```diff
@@ -13,26 +13,36 @@
...
     "REACT_APP_NEARMAP_TOKEN": "NWE1ZTA1YTYtOTg4Yy00ZGNhLWFjMzYtMmYxOTI2M2UyODRk",
     "REACT_APP_KNACK_DATA_TRACKER_SCENE": 514,
-    "REACT_APP_KNACK_DATA_TRACKER_VIEW": 3047,
-    "REACT_APP_KNACK_DATA_TRACKER_APP_ID": "6167314778435d001ea3e7cb"
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_VIEW": 3047,
+    "REACT_APP_KNACK_DATA_TRACKER_SIGNAL_VIEW": 1483,
+    "REACT_APP_KNACK_DATA_TRACKER_APP_ID": "6167314778435d001ea3e7cb",
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_ID_FIELD": "field_3998",
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_NAME_FIELD": "field_3999",
+    "REACT_APP_KNACK_DATA_TRACKER_CURRENT_STATUS_FIELD": "field_4000",
```
frankhereford, line 21: This PR establishes a subtle (bad?) naming scheme. It's not great, but I'm not sure how to better smash what could be a single object into ENV variables.

ENV variables that fit [the pattern here](https://github.com/cityofaustin/atd-moped/blob/flh-7925-signal-id-knack-payload/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js#L88) are expected to contain a knack field as the value and are used to build the body when synchronizing to data tracker.

moped-editor/.env-cmdrc
```diff
@@ -13,26 +13,36 @@
...
+    "REACT_APP_NEARMAP_TOKEN": "NWE1ZTA1YTYtOTg4Yy00ZGNhLWFjMzYtMmYxOTI2M2UyODRk",
+    "REACT_APP_KNACK_DATA_TRACKER_SCENE": 514,
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_VIEW": 3047,
+    "REACT_APP_KNACK_DATA_TRACKER_SIGNAL_VIEW": 1483,
+    "REACT_APP_KNACK_DATA_TRACKER_APP_ID": "6167314778435d001ea3e7cb",
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_ID_FIELD": "field_3998",
+    "REACT_APP_KNACK_DATA_TRACKER_PROJECT_NAME_FIELD": "field_3999",
+    "REACT_APP_KNACK_DATA_TRACKER_CURRENT_STATUS_FIELD": "field_4000",
+    "REACT_APP_KNACK_DATA_TRACKER_SIGNAL_CONNECTION": "field_3861"
+  },
```
frankhereford, line 45: Linting caused this whole block to scoot over.

moped-editor/src/views/projects/projectView/KnackSync.js
```diff
@@ -1,205 +0,0 @@
-import React from "react";
```
frankhereford, line 1: This file was the first implementation of sync ➡️  knack. It worked, but it lacked the updates that the component that is used in the summary page itself had received. As a result, it needed to be removed, as it was a ton of duplicated code and it allowed projects to be sync'd without consideration to signals and other recent updates.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -65,33 +82,62 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
+      .forEach(regExResult => {
+        fieldMap[process.env[regExResult[0]]] = regExResult[1].toLowerCase();
+      });
 
-    Object.keys(field_map).forEach(element => {
-      if (project.currentKnackState[element] !== project[field_map[element]]) {
-        body[element] = project[field_map[element]];
-      }
+    Object.keys(fieldMap).forEach(element => {
+      body[element] = project[fieldMap[element]];
```
frankhereford, line 102: This is the consumption of the ENV variables mentioned above with the suspect naming scheme.

Thanks to @chiaberry for her advice on how to clean up the filter().map().forEach() chaining.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -100,27 +146,55 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
         .then(response => response.json())
         .then(result => {
           if (result.errors) {
             // Successful HTTP request, but knack indicates an error with the query, such as non-existent ID
-            return Promise.reject(result);
+            throw result;
           } else {
             // Successful HTTP Update request with meaningful results from Knack
-            return Promise.resolve();
+            return true;
```
frankhereford, line 197: Is this right? What is the least more true than false?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -100,27 +146,55 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
             // Successful HTTP request with meaningful results from Knack
             project.currentKnackState = result; // this assignment operates on `project` which is defined in broader scope than this function
-            return fetch(knackEndpointUrl, {
-              // fetch returns a Promise for the next step
-              method: knackHttpMethod,
-              headers: buildHeaders,
-              body: buildBody(),
-            });
+
+            return countSignalsInProject(project)
```
frankhereford, line 155: This return (and the similar one below) are the reason why this function still returns an explicit Promise.resolve()

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -469,13 +457,6 @@ const ProjectView = () => {
                             horizontal: "center",
                           }}
                         >
-                          <KnackSync
-                            project={data.moped_project[0]}
-                            closeHandler={handleMenuClose}
-                            snackbarHandler={handleSnackbarOpen}
-                            refetch={refetch}
-                          />
-
```
frankhereford, line 478: All the changes in this file are the removal of the old Knack Sync option in the drop-down menu for projects.

moped-etl/app/moped_data-tracker_sync_projects.py
```diff
@@ -64,13 +102,22 @@
...
             update_needed = True
             knack_data[knack_object_keys[key]] = moped_project[key]
+
+    knack_signals = build_signal_set_from_knack_record(knack_records[moped_project["project_id"]])
+    moped_signals = build_signal_set_from_moped_record(moped_project)
+
+    if not moped_signals == knack_signals:
+        update_needed = True
+        knack_data[KNACK_OBJECT_SIGNALS] = list(moped_signals)
+
```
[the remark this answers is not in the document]
frankhereford: Don't listen to this guy! 🙃 

We have [a plan to address the transition](https://github.com/cityofaustin/atd-data-tech/issues/8393) of the application from the development knack instance to the production one.

frankhereford:
This is ready for review - I'll mention it to y'all in #atd-moped in the morning after I give it one last pass. Thanks!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
chiaberry: this is like a galaxy brain question for me this morning

chiaberry (approved):
I tested locally a few different ways and it all works out. I don't think you need the `return true` in the promise chains (especially where the snackbarHandle is being called)

It works great!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
johnclary: if you're not doing anything useful later with `true` (you're just continuing the chain), then i think you can just nix the `else {}`  block entirely and let the function return `undefined`. imo that would be better, because i otherwise read this to mean that there's a dependency on the return value `true`.

moped-editor/.env-cmdrc
frankhereford: On more consideration, I'm less happy with this. My rule of thumb is that I'm doing something wrong if I'm encoding information into the key of a key, value pair. The human-readable name for the column is extracted from the key via regex. 👎 

I'm tremendously open to any suggestions; I'm not sure what is a better route here beyond some hackery to escape a JSON object into one ENV variable which is brittle and feels like a portability concession.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
frankhereford: @johnclary - thank you! I couldn't see the forest for the trees there. That else block used to have some wayward functionality that slowly got chipped away until it was an empty non-op.

amenity:
I am not able to open the map editor on [this project](https://deploy-preview-522--atd-moped-main.netlify.app/moped/projects/388?tab=map), @frankhereford. :\ Never seen this happen before, but when I click "Add components" button from the "Summary" tab, I see: 
Screen Shot 2022-02-08 at 4 13 33 PM

frankhereford:
I wanted to capture some ideas that John and I discussed today for this branch. Thank you @johnclary for your insights and guidance - I appreciate it mega.

* [ ] Call `countSignalsInProject` outside of the component's render method, because it is a reasonable assumption that signals are not being added or removed hyper-frequently. 
* [ ] Factor out the call to knack that figures out what knack has assigned as the IDs for signals. Take the data that is in the JSON blob in our database as canonical. 
* [ ] Add a CSS decloration to make the mouse into the little finger pointer symbol when over the knack-sync <Link>. 
* [ ] Add some comments in the code on the logic branch that contains the "update a previously sync'd" process to explain that this code is good to hang out, even though the current UI does not expose a way to execute it.
* [x] Confirm that John was correct that while we don't have a merge conflict, there is code in my branch that is affected by his work.

frankhereford:
## Latest updates

Re: the comment above, I have pushed some changes into this branch related to @johnclary's guidance. 

### :warning: There is a meaningful impact on those who test this branch now. :warning:

The Knack-Sync process no longer queries Knack to figure out what internal Knack ID a given signal has. It now simply inspects the data from our database about associated signals and uses the Knack ID on file for them. This provides for a snappier sync process, as it removes an entire request to the Knack API. Gotta go fast. 

The rub is that, for all deployments except production, it will now appear that the actual sync'ing of signals appears broken. This is a result of our storing signal's Knack IDs that come from the production Data Tracker app and them actually landing in the development knack instance, which has its own set of internal Knack signal IDs. 

Knack will happily accept these non-existing signal IDs in the payload, and may even file records in their database table which provides the association between projects and signals, but the underlying SQL join behind Knack can no longer function. Resultantly, any signals you have in a project in Moped will not show up when you look at the connection field in the projects table in Knack. 

Unfortunately, I do not have any ideas on how to provide a work-around for this, but I'm open to any suggestions of course!

johnclary:
👍 thanks @frankhereford - i think this is the best path forward for now.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -100,27 +127,35 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
             // Successful HTTP request with meaningful results from Knack
             project.currentKnackState = result; // this assignment operates on `project` which is defined in broader scope than this function
-            return fetch(knackEndpointUrl, {
-              // fetch returns a Promise for the next step
-              method: knackHttpMethod,
-              headers: buildHeaders,
-              body: buildBody(),
-            });
+
+            Promise.resolve(numberOfSignalsInProject)
```
frankhereford, line 136: I was fussing around with wrapping this up in some sort of function to avoid the explicit Promise.resolve() call, but on consideration, the exact thing I needed was a resolved promise, so I put this in here. 

Open to any suggestions if this not the best route.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -65,33 +78,47 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
    */
-  const buildBody = () => {
+  const buildBody = signalIds => {
     let body = {};
 
-    const field_map = {
-      field_3998: "project_id",
-      field_3999: "project_name",
-      field_4000: "current_status",
-    };
```
frankhereford, line 75: Goodbye magic numbers! 🪄 🔢

frankhereford:
@amenity said:
> ever seen this happen before, but when I click "Add components" button from the "Summary" tab, I see

Hi Amenity - I saw in slack that this looks like it's an issue we have on staging currently which, I think, indicates this may be an existing, unrelated issue. If you think that I may be confused here, please let me know, and I'll get chomping on this for sure. Thanks!

amenity (approved):
Works beautifully!

amenity:
And yes, per Slack discussion, the behavior I was seeing was also visible in staging (same database) and we're hoping it will be corrected by some of the outstanding PRs.

moped-editor/.env-cmdrc
mateoclarke: I don't think this is a bad pattern. I can't really suggest a more elegant solution. But I've done something like you're describing organizing Knack keys and fields into a nested JSON. Not the same context as this but maybe it helps? 🤷‍♂️  

https://github.com/cityofaustin/atd-mobile-signals-work-orders/blob/master/src/queries/api.js#L23

mateoclarke (approved):
🚢 💨 ✨

moped-editor/.env-cmdrc
frankhereford: Ah, TIL more about the cloning process and that it will preserve column names behind the scenes. That will cut down on things to keep track of. Very cool.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
```diff
@@ -58,6 +58,10 @@ const useStyles = makeStyles(theme => ({
   fieldLabelText: {
     width: "calc(100% - 2rem)",
   },
+  knackFieldLabelText: {
+    width: "calc(100% - 2rem)",
+    cursor: "pointer",
+  },
```
johnclary, line 64: 👍

PR #529, opened 2022-01-19
8002 Address deleted funding sources on signal dashboard page

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/8002. 

## Testing

https://deploy-preview-529--atd-moped-main.netlify.app/moped/projects

* Create a project from a signal asset and assign it a funding source. 
* On the signal dashboard, observe that the new funding source shows for this project.
* Delete this funding source.
* On the signal dashboard, observe that the deleted funding source is no longer shown.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/queries/funding.js
+++ b/moped-editor/src/queries/funding.js
@@ -3,7 +3,7 @@ import { gql } from "@apollo/client";
 export const FUNDING_QUERY = gql`
   query ProjectFunding($projectId: Int) {
     moped_proj_funding(
-      where: { project_id: { _eq: $projectId }, funding_status_id: { _gt: 0 } }
+      where: { project_id: { _eq: $projectId }, funding_status_id: { _neq: 0 } }
     ) {
       proj_funding_id
       added_by
--- a/moped-editor/src/queries/signals.js
+++ b/moped-editor/src/queries/signals.js
@@ -46,7 +46,7 @@ export const SIGNAL_PROJECTS_QUERY = gql`
         feature_id
         location
       }
-      moped_proj_funding {
+      moped_proj_funding (where: {funding_status_id : { _neq: 0 } }) {
         moped_fund_source {
           funding_source_name
         }
--- a/moped-etl/app/process/logging.py
+++ b/moped-etl/app/process/logging.py
@@ -10,5 +10,4 @@ def getLogger(name, level=logging.INFO):
     handler.setFormatter(formatter)
     logger.addHandler(handler)
     logger.setLevel(level)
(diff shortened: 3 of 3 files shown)


johnclary:
@frankhereford thanks for fixing this up — i missed this query when i triaged the issue. would you update the `where` condition [this query](https://github.com/cityofaustin/atd-moped/blob/2f87f38fa9bd706b5e30677fa6e87b6301985d02/moped-editor/src/queries/funding.js#L6) to match the `_neq` in the one you modified (or vice versa).

Maybe a topic for broader discussion—we have business logic (the status ID) encoded in these queries that is not transparent. Let's add some docstrings to the queries that explains the importance of the filter conditions? When I've been working through the Moped migration its always a pain to remember what status ID means what. Is there a better way?

johnclary (approved):
⛵

frankhereford:
> @johnclary says:
> When I've been working through the Knack > Moped migration its always a pain to remember what status IDs means what. Is there a better way?

I agree that we are embedding magic numbers in the codebase. I think we have a convention going that false values are "deleted", but we'd be in a bind if we wanted to support multiple states that needed to not show up on lists because they were "deleted."

One thing that the CRIS data does well in VZ is having a look-up table for almost everything. I think we would benefit from creating similar lookable tables for any column that stores categorical data using an integer. Using SQL joins, we can easily link our integer values to, say, a slug varchar() and a human readable label varchar(). These strings would be prime candidates for what we expose via graphql-engine.

Thinking of `funding_status` in particular, if we had a look-up table, we could add a field like `is_soft_delete` and be able to support multiple, not-deleted or deleted states as needed.

This could be done by refactoring a bunch of queries which would be a good bit of work, or we could use database views to hide those joins from graphql-engine -- still not trivial by any means, but it's an idea.

amenity (approved):
Beautiful, thanks @frankhereford !

chiaberry (approved):

moped-editor/src/queries/funding.js
```diff
@@ -3,7 +3,7 @@ import { gql } from "@apollo/client";
 export const FUNDING_QUERY = gql`
   query ProjectFunding($projectId: Int) {
     moped_proj_funding(
-      where: { project_id: { _eq: $projectId }, funding_status_id: { _gt: 0 } }
+      where: { project_id: { _eq: $projectId }, funding_status_id: { _neq: 0 } }
```
mateoclarke, line 6: 👍

PR #531, opened 2022-01-20
Add financial code tracking

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/4530

I added a lookup table named `moped_funds` to reference the funds. The dept-unit is in a socrata dataset: https://data.austintexas.gov/dataset/Transportation-Project-Financial-Codes/bgrt-2m2z. 

I also deleted a few views that we never used in moped. (CustomerListView, ProductListView and SettingsView). 

## Testing
**URL to test:** 
Run on moped test or run locally as there are db changes. 
**Steps to test:**

Go to a project's funding tab. Add a funding source and add a fund and/or dept-unit. Save and then try editing. 

Screen Shot 2022-01-21 at 10 09 10 AM

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -183,7 +183,7 @@ const ProjectSummaryProjectECapris = ({
             text={
               (isValidNumber(eCapris) && (
                 <>
-                  <Typography variant={"span"} color={"primary"}>
+                  <Typography variant={"inherit"} color={"primary"}>
```
chiaberry, line 186: updated this to fix an error in the console, Typography has no variant "span"

moped-editor/src/views/projects/projectView/FundingDeptUnitAutocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+ * Socrata dataset and setting it.
+ * * @param {Object} classes - MaterialUI style object
+ * * @param {Object} props - passes down props parent MUI Table
+ * * @param {Array} value - passes an array of objects, each representing a dept unit
+ *  @return {JSX.Element}
+ */
+const FundingDeptUnitAutocomplete = ({ classes, props, value }) => {
+  const { data, loading, error } = useSocrataJson(SOCRATA_ENDPOINT);
+  console.log(data)
+  console.log(classes)
```
frankhereford, line 36: Would you consider changing these to `console.info()`'s or removing them?

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -152,7 +167,7 @@ const ProjectFundingTable = () => {
 
   const { loading, error, data, refetch } = useQuery(FUNDING_QUERY, {
     // sending a null projectId will cause a graphql error
-    // id 0 used when creating a new project, no project personnel will be returned
+    // id 0 used when creating a new project, no project funding will be returned
     variables: {
       projectId: projectId ?? 0,
     },
```
frankhereford, line 173: Please correct me if I'm misunderstanding, but I am under the impression that a user should never be at this part of an app without there being a `projectId` defined and available to provide to the query. 

If that's the case, and if a user somehow does get to this page without a particular project (I really don't think this is a possible route), I would suggest we issue a redirect or somehow send them back to the project listing and not try to render anything with inevitably null or blank data.

I think this query *should* break if there isn't a projectId, and putting a `0` in there as a default value opens up the possibility for the app to think we're dealing with project `0` and have potential knock on effects. 

That said, I don't think this is a big deal, just a thought to consider.

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -209,7 +224,18 @@ const ProjectFundingTable = () => {
       },
     })
       .then(() => refetch())
-      .catch(() => {});
+      .catch(error => {
```
frankhereford, line 227: 👍 🙏

frankhereford (approved):
I've looked at the DB schema, in particular the new table this creates, and I think it looks good. While reviewing the code, I have made two small suggestions, but neither should hold up merging in my opinion. Thank you for making this change!

🌊 🚢  💯 🌊

moped-editor/src/views/projects/projectView/FundingDeptUnitAutocomplete.js
chiaberry: i forgot to remove them! thanks for catching it

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
chiaberry: You're absolutely right. This is a hold over from when the add project flow contained more tables than just the title and description. The same pattern is on the personnel table. 

If I'm remembering correctly, if a null projectId was in the query, it would fail with an error, but sending 0 would return empty data.

johnclary:
I have not given this a full review but this looks really good @chiaberry! thanks for the extra cleanup and thanks @frankhereford for the good suggestions.

amenity (approved):
Wow, works so beautifully, @chiaberry !! Thank you.

mateoclarke (approved):
🚢 💨 ✨ 

Only one minor suggestion. I noticed Fund name renders in uppercase while Dept-unit are in capital case/sentence case. I assume this is because of the way the text values are stored in the DB and in Socrata, but maybe its worth it to style consistency on the front end?

Screen Shot 2022-01-24 at 6 11 50 PM

amenity:
Good eye, @mateoclarke. I don't mind the combo case for now but will keep that in mind for future work. cc/ @patrickm02L

PR #533, opened 2022-01-21
Perform miscellaneous database updates

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/8058

## Testing
**URL to test:** 

Database changes, so must be local or on moped test. 

**Steps to test:**

To see the changes on the moped entity table: Go to a projects summary page and select the partner dropdown. COA ATD Vision Zero is now an option. (Note, when I tested locally, that option moved to the bottom of the list. I'm not sure if the order in the dropdown matters, but if it does we could order by enitity_id)

To see the moped components change, go the map tab, add a new component and look for a Crosswalk with only 2 s. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1642785096000_update_moped_components_table/down.sql
+++ b/moped-database/migrations/1642785096000_update_moped_components_table/down.sql
@@ -1,0 +1,5 @@
+UPDATE public.moped_components SET component_name = 'Crossswalk' WHERE component_id = 13;
+UPDATE public.moped_components SET component_name = 'Crossswalk' WHERE component_id = 14;
+UPDATE public.moped_components SET component_name = 'Crossswalk' WHERE component_id = 15;
+
+UPDATE public.moped_entity SET entity_name = 'COA ATD Intersection Safety', workgroup_id = NULL WHERE entity_id = 11;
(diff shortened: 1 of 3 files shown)


frankhereford (approved):
ezgif-3-cb685c971f

amenity (commented):
@chiaberry - thanks! Please update these dropdowns so the option appears in the right spot:
Screen Shot 2022-01-23 at 8 17 25 PM
/1463708/150711726-f1f633f6-c5b8-45ec-a80a-74cab90e99fd.png)

chiaberry:
@amenity, I updated the code to sort the entities by entity_id

amenity:
@chiaberry - thanks! Does this mean that we'll need to add a `sort` column if we want to insert values down the line?

chiaberry:
oh, like if for example there will the COA Fancy Bond Group and you would want it to appear between Aviation and Corridor? Yeah, we would have to come up with a different way to sort/display them then

mateoclarke:
> oh, like if for example there will the COA Fancy Bond Group and you would want it to appear between Aviation and Corridor? Yeah, we would have to come up with a different way to sort/display them then

We could create a `sort` column if we want to control sort order for these values consistently across the UI.

mateoclarke (approved):
🚢 💨 ✨ 

crossswalk

amenity (approved):
Nice, thanks!

PR #534, opened 2022-01-21
Update filters on signal dashboard query

chiaberry:
## Associated issues

the query used to be: 

``` 
where: {
        _or: [
          {
            moped_proj_components: {
              moped_components: { component_name: { _ilike: "signal" } }
            }
            status_id: { _is_null: true }
          }
          {
            moped_proj_components: {
              moped_components: { component_name: { _ilike: "signal" } }
            }
            status_id: { _neq: 4 }
          }
        ],
        status_id: {_neq: 6}
      }
```

which was preventing projects with a null status from showing up. Which seem to be existing projects / seed projects that then get a signal added to it. (I guess you can't both be is_null is true and _neq: 6?)

now it will be signal projects with either null status, or not 3 or not 6. 

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/queries/signals.js
+++ b/moped-editor/src/queries/signals.js
@@ -1,5 +1,8 @@
 import { gql } from "@apollo/client";
 
+// Status ID 6 is archived (soft deleted)
+// Status ID 3 is cancelled
+
 export const SIGNAL_PROJECTS_QUERY = gql`
   query SignalProjectsQuery {
     moped_project(
@@ -15,10 +18,15 @@ export const SIGNAL_PROJECTS_QUERY = gql`
             moped_proj_components: {
               moped_components: { component_name: { _ilike: "signal" } }
             }
-            status_id: { _neq: 4 }
+            status_id: { _neq: 6 }
           }
-        ],
-        status_id: {_neq: 6}
+          {
+            moped_proj_components: {
+              moped_components: { component_name: { _ilike: "signal" } }
+            }
+            status_id: { _neq: 3 }
+          }
+        ]
       }
       order_by: {updated_at: desc_nulls_last}
     ) {


mateoclarke (approved):
🚢 💨 ✨ 

kill_bug

mateoclarke:
I mentioned this to @amenity over voice, but I was thinking that instead of having these hardcoded ID values for each instance of statuses we want hidden, what if we renumbered our status IDs to have any hidden status as negative integer IDs? And then make the query a greater than 0 filter

frankhereford:
@mateoclarke said:
> I mentioned this to @amenity over voice, but I was thinking that instead of having these hardcoded ID values for each instance of statuses we want hidden, what if we renumbered our status IDs to have any hidden status as negative integer IDs? And then make the query a greater than 0 filter

My general rule of thumb is to want to have ID columns in a database be nothing more than a unique integer that identifies the record. Doing so allows you to make it an unsigned int which is the absolute best-case scenario in terms of database index performance, which is key to writing performant `JOIN` statements. That said, there are worse sins than a signed int of course.

To your comment about not hard-coding magic numbers into the codebase, I am massively in support of that notion - I believe that is a extra-anti-pattern to include hard-coded values.

An approach that we we may want to discuss is adding a column to the table that contains the records, `moped_proj_components` in this case, and add a column named `is_soft_deleted` typed as a boolean. Then we can `JOIN` the table `ON (moped_proj_components.is_soft_deleted is false)` or the graphql equivalent like:

```
      where: {
        _or: [
          {
            moped_proj_components: {
              moped_components: { component_name: { _ilike: "signal" } }
            }
            status_id: { _is_null: true }
          }
          {
            moped_proj_components: {
              moped_components: { component_name: { _ilike: "signal" } }
            }
            is_soft_delete: { _eq: 1 }
          }
          {
            moped_proj_components: {
              moped_components: { component_name: { _ilike: "signal" } }
            }
            is_soft_delete: { _eq: 0 }
          }
        ]
      }
```

I picked used `1` and `0` for my boolean in the above gql because I'm not sure if graphql-engine has a boolean comparison operator.

amenity:
THANKS! Good to ship, @chiaberry!

PR #469, opened 2021-11-20
Add detection and support for M1 ARM64 architecture to moped-database

frankhereford:
## Associated issues

This PR intends to close [#7810](https://github.com/cityofaustin/atd-data-tech/issues/7810).

## Testing
Follow the normal instructions to spin up the database on either platform and verify that it works for you.
- [x] X86_64
- [x] ARM64

### PostgreSQL server image
~* I have increased the amount of time that the cluster has to start before the `hasura-cluster` script executes migrations, metadata and seeds from 60 seconds to 120 seconds. The cluster starts slower now because it's running via the qemu emulation layer, but once it's running, the performance appears to be nominal. If Hasura were to start to ship multi-platform images, this would become a non-issue.~

* I have switched the postgres image over to a native build for arm64. I built this today using the postgis docker build repo without modification; it's super easy. This results in database rebuild times being reduced to 25% of the original duration for me when testing the VZ DB (which has millions of records.) For Moped, the database size is still small, so that's not an issue, but cluster start-up times are cut in half if not more. At the time of writing, a cluster spin up takes 1 minute and two seconds (1:02.54) of wall time on this computer. 🕐 

### graphql-engine image
~* I have also built the graphql-engine v1.3.3 from source personally, and the image can be found [here](https://hub.docker.com/repository/docker/frankinaustin/graphql-engine-arm64).~

* There was group consensus to use `fedormelexin`'s build of `graphql-engine` v1.3.3, however, I would like to propose that we stick with my build. I have zero reason to believe that Fedor's build isn't completely fine; it's that I have used mine personally for development for a few months now and it has performed without a single issue. It's just the only one of the two that I have personal experience with, and it's totally solid. 🪨 

---
#### Ship list
- [ ] Code reviewed

PR #537, opened 2022-01-25
Support ARM64 in Moped database stack

frankhereford:
## Associated issues

This PR intends to close [#7810](https://github.com/cityofaustin/atd-data-tech/issues/7810).

This PR has been created to supplant PR #469 which bears age-related complications outside of the scope of these changes.

## Review
* @amenity had removed herself from reviewing [#469](https://github.com/cityofaustin/atd-moped/pull/469), so I am honoring that again on this PR. Amenity, please review if you would like to, of course!

## Testing
Follow the normal instructions to spin up the database on either platform and verify that it works for you.
- [x] X86_64
- [x] ARM64

Ideally, I would welcome a test from all moped developers of this branch to better ensure that there are no breaking changes to your ability to run a local database cluster. Please feel free to comment on the outcome of your tests. You should not have to change any portion of your workflow in any way; just spin it up and make sure it's working as expected.

### PostgreSQL server image

* I have switched the postgres image over to a native build for arm64 (for arm64 users only, of course). I built this over the weekend using the postgis docker build repo without modification; it's super easy. This results in database rebuild times being reduced to 25% of the original duration for me when testing the VZ DB (which has millions of records.) For Moped, the database size is still small, so that's not an issue, but cluster start-up times are cut in half, if not more. At the time of writing, a cluster spin up takes 1 minute and two seconds (1:02.54) of wall time on this computer. 🕐 

### graphql-engine image

* There was group consensus to use `fedormelexin`'s build of `graphql-engine` v1.3.3, however, I would like to propose that we stick with my build. I have zero reason to believe that Fedor's build isn't completely fine; it's that I have used mine personally for development for a few months now and it has performed without a single issue. It's just the only one of the two that I have personal experience with, and it's totally solid. 🪨 

### docker image selection arguments to `hasura-cluster`

* The group had agreed upon adding an argument to `hasura-cluster` to allow a developer to override the images being used for the database and / or graphql-engine server. 
  * As it stands, `hasura-cluster` uses the entire argument array, `$@`, to determine what operation it's going to carry out, and it will take a little refactoring to support multiple arguments. 
  * I would like to spin that off as an issue to work in the future, if the group feels it's a valuable feature.
  * I believe it will become useful when we look at upgrading `graphql-engine`, `postgresql`, or `postgis`.

---
#### Ship list
- [ ] Code reviewed


mateoclarke (approved):
Works as expected on my old school intel machine. 👴 

intel-pentium-inside

mateoclarke:
Looks like it builds perfecting in the Netlify preview as well. ✨ 🧙

PR #538, opened 2022-01-25
Patch project sponsor disappearing bug

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/8244

## Testing
**URL to test:** 
Deploy to test, to test!

**Steps to test:**

Missing edit hyphen for some projects -- save a blank project DO#, by entering only empty spaces and saving. it should still render as a - instead of the blank spaces

To test the disappearing project sponsor. First, try resizing the screen. confirm the column texts dont disappear. Try updating other fields and the project sponsor shouldn't go away. 

I think this should also fix the project DO# not wanting to save. I can't see in the video of the bug, but I suspect the project sponsor had disappeared so when the project DO was trying to save, it was sending "undefined" as the project sponsor, thus preventing the update from completing. That is my conjecture though, since I wasn't able to reproduce. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
+++ b/moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
@@ -118,7 +118,7 @@ const SignalProjectTable = () => {
     project["project_types"] = project_types;
 
     // project sponsor
-    project["project_sponsor"] = entityList.find(
+    project["project_sponsor_object"] = entityList.find(
       e => e.entity_id === project?.project_sponsor
     );
 
@@ -294,12 +294,12 @@ const SignalProjectTable = () => {
     },
     {
       title: "Project sponsor",
-      field: "project_sponsor",
+      field: "project_sponsor_object",
       render: entry => (
         <Typography className={classes.tableTypography}>
-          {entry?.project_sponsor?.entity_name === "None"
+          {entry?.project_sponsor_object?.entity_name === "None"
             ? "-"
-            : entry?.project_sponsor?.entity_name}
+            : entry?.project_sponsor_object?.entity_name}
         </Typography>
       ),
       customEdit: "projectSponsor",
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -287,15 +287,19 @@ const SignalProjectTable = () => {
       title: "Project DO#",
       field: "purchase_order_number",
       emptyValue: "-",
+      render: entry =>
+        entry.purchase_order_number.trim().length === 0
+          ? "-"
+          : entry.purchase_order_number,
```
chiaberry, line 293: this allows blank project DO's to still be editable

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
```diff
@@ -118,7 +118,7 @@ const SignalProjectTable = () => {
     project["project_types"] = project_types;
 
     // project sponsor
-    project["project_sponsor"] = entityList.find(
+    project["project_sponsor_object"] = entityList.find(
       e => e.entity_id === project?.project_sponsor
     );
```
chiaberry, line 123: using the same variable name was the fatal flaw here. After adding the windowResize hook, the table was rerendering, but not querying the data. So it was trying to match this to an object instead of a number, which would then result in undefined/empty.

mateoclarke: 🙃  Thanks for figuring this one out!

moped-editor/src/views/projects/signalProjectTable/SignalProjectTable.js
mateoclarke: ✂️  `trim()` FTW!

mateoclarke (approved):
🚢 💨 ✨ 

bug-bugs

amenity:
@chiaberry - we'll need another PR for main, right?

amenity:
@chiaberry - I'm still seeing the "DO #" edit-in-place disappear when I delete the value from the signals dashboard.

https://user-images.githubusercontent.com/1463708/151038634-e7dd544b-2b1d-43db-aade-36283bc65585.mov

amenity (approved):
⚡  bug iz gone! ⚡

PR #539, opened 2022-01-25
6233: Remove unused tables and rework FK relationships to allow cascading deletions of child table records

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/6233.

## Considerations

This PR changes the way that the database will handle a `DELETE` command on `moped_project` and its child tables. It configures the database to `CASCADE` the delete command down to get all related data of a project when a project is deleted. 

This PR allows for a project and all related data to be deleted with a single SQL command or mutation.

It also includes `DROP`ing a number of unused tables.

## Testing

This PR is entirely related to database changes, so testing must occur locally or on the _testing_ deployment. 

For those who do run moped locally, I recommend doing a round of local testing and then doing a round of testing on the _test_ instance. Testing on _test_ may require use of the bastion host to access the DB instance as it is firewalled behind it.

A portion of the testing will include issuing a `DELETE` command against the `moped_project` table, and I think we need to be mindful of preserving that data on the shared environments until it needs to be deleted. Let's keep communicating about database changes that remove or alter data outside of our local instances.

All that said, the testing URL is: https://moped-test.austinmobility.io/moped/

**Steps to test:**

* Ensure migrations run when you spin up an instance of the database. This tests the changes to the seed data.
* Ensure that just the new migrations run against a restoration of data into your local database stack from both staging and production.
* Ensure that the correct tables have been removed.
* Look at the tables that have the foreign key constraints changed to use cascading deletes. Make sure there are no tables that contain information that is part of the app (like a list of departments, divisions, etc.) These deletes should only cascade into tables which contain information about a project or related, user-entered, data.
* Make sure that you can `DELETE` any `moped_project` that you want or can construct in a single SQL statement without errors.

---
#### Ship list
- [x] Code reviewed 


moped-database/migrations/1643301492425_rework_fks_for_projects_and_children/up.sql
```diff
@@ -0,0 +1,76 @@
+-- moped_proj_phases.project_id => moped_project
```
johnclary, line 1: these comments are super helpful!

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -264,41 +264,6 @@ export const ProjectActivityLogTableMaps = {
       },
     },
   },
-  moped_proj_fund_opp: {
```
johnclary, line 267: interesting—so we have activity log handling set up for this table, but the table itself was never implemented?

moped-database/metadata/tables.yaml
```diff
@@ -3849,53 +2636,49 @@
       - project_priority
       - project_sponsor
       - project_uuid
-      - project_website
       - purchase_order_number
+      - project_website
       - start_date
       - status_id
-      - task_order
```
johnclary, line 3856: @frankhereford it's kind of hard to assess the changes  to `tables.yaml` to in the github UI, but an noticing `task_order` permissions are being revoked. could this be an oversight?

johnclary:
@frankhereford this looks excellent and please open up to the rest of the devs for review. please also deploy to test when the time is right. thank you!

amenity (approved):
I'm not able to inspect the DB, but I did an abbreviated end-to-end test of basic functionality and everything appears to be working great.

moped-database/metadata/tables.yaml
frankhereford: @johnclary I am very grateful that you caught that! 🦅 👁️  The way the file get shuffled around when it's written out by `graphql-engine` makes reading diffs really tricky, I agree.

You are exactly right that I had inadvertently revoked access to `task_order` as well as `knack_project_id` for the moped_projects table. I went and double checked all the other tables to see if I had done it in other places, and I did find that the same thing had happened to `fund` and `dept_unit` in the `moped_proj_funding` tables.

TBH, I am not sure exactly how this happened, but I definitely did it here in [this commit](https://github.com/cityofaustin/atd-moped/pull/539/commits/1cd53ddeebdfbf3391424b3cb1b1cbf8c9e16624). The best I can think of is that I made a mistake and a `graphql-engine` daemon running while I was jumping branches with `git`.  Thanks @mateoclarke for brainstorming on this with me, too.

I've pushed up two commits that restore these columns' permissions. Thank you again!

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
frankhereford: @johnclary I'm glad you asked this question! There were 6 more entries in this object that needed to be removed to align with tables that were dropped. I've pushed up commits to do that into this branch / PR. 

To answer your question, yes, we did have triggers in place in the database for this table to have changes tracked in the activity log. I have audited which tables are tracked by looking for those triggers, and the results are below. Nothing sticks out to me as out of place, but others may have see some that we should enable tracking on.

```
Tracked Tables:
 moped_proj_entities 
 moped_proj_financials
 moped_proj_funding
 moped_proj_notes
 moped_proj_partners
 moped_proj_personnel
 moped_proj_phases
 moped_project
 moped_project_files
 moped_project_types

Untracked Tables:
 moped_activity_log
 moped_city_fiscal_years
 moped_components
 moped_department
 moped_entity
 moped_fund_programs
 moped_fund_sources
 moped_fund_status
 moped_funds
 moped_milestones
 moped_organization
 moped_phases
 moped_proj_components
 moped_proj_components_subcomponents
 moped_proj_features
 moped_proj_features_components
 moped_proj_fiscal_years
 moped_proj_milestones
 moped_project_roles
 moped_status
 moped_subcomponents
 moped_subphases
 moped_types
 moped_users
 moped_workgroup
```

Thanks again!

mateoclarke (approved):
air-high-five-good-job

PR #553, opened 2022-02-03
Refactor component <> feature relationship from many-to-many to many-to-one

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8389

Todo:
- [x] Test/update: links to signals from dashboard, project summary
- [x] Create a migration to backfill existing projects
- [x] Test that fetching the complete geometry from agol is working
- [x] Think about the use of `PROJECT_EXTENT_ID` in this code: i'd like to change up the way we use this property, but can leave it out of scope. Chia took care of all the immediate needs in #546 
- [x] Test/update: Knack sync: Holding off until #522 can be incorporated
- [ ] Not doing this: Delete all projects in staging - will do before merging [8157-new-tile-endpoint](https://github.com/cityofaustin/atd-moped/tree/8157-new-tile-endpoint) to main

## Summary of changes

- deprecate `moped_proj_features_components` associative table
- remove fk relationship between `moped_proj_features` and `moped_project`
- deprecate `COMPONENT_DETAILS_QUERY`, and move the source for moped_components and moped_subcomonents up one level in the tree.
- refactor the data munging involved with parsing and packaging up mutation payloads, which are simpler now that we don't have the associative table, moped_proj_features_components.
- hookify and move up the tree the functions that index moped_components and moped_subcomponents

## Testing
**URL to test:** [Moped Test](https://moped-test.austinmobility.io/moped/)

**Steps to test:**
Make sure there are no UI regressions. Create, edit, and delete lots of components, with subcomponents.
Use **multiple features** per component. Including:
- A signal component created through the new project workflow
- A selected point(s) component (create/edit/delete)
- A selected line(s) component (create/edit/delete)
- A drawn line(s) component (create/edit/delete)
- A drawn point(s) component (create/edit/delete)
- A mix of drawn/selected points
- A mix of drawn/selected lines
- Verify the summary map renders correctly
- Verify the map tab renders correctly
- Verify that the signals dashboard (and signal links) renders correctly
- Verify that the project summary (and signal links) renders correctly

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager approved
- [x] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#gid=776973707)

PR #555, opened 2022-02-06
Update working branch with changes from #8251 -  data tracker links

johnclary:
I need these changes so that I can make a few tweaks to property paths in order to accommodate #553.


johnclary:
@chiaberry @frankhereford - no conflicts here 😅  - but do you think it would be better to merge main into my branch? i _think_ this is the right way to go but i'm not confident. sorry for the mess!

frankhereford:
> no conflicts here 😅 - but do you think it would be better to merge main into my branch? i _think_ this is the right way to go but i'm not confident. sorry for the mess!

@johnclary - I think both routes are valid, but these changes should be in main, so if I had to pick, I'd merge main in. I'm not great at keeping the git-dance straight in my head, so please take me with a big grain of salt. 🧂

johnclary:
thanks y'all—merging this one

PR #559, opened 2022-02-08
Displays an alert banner in the staging environment

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/8234

## Testing
URL to test:
https://deploy-preview-559--atd-moped-main.netlify.app/ 

**Steps to test:**
not sure how to test this because the banner will only display in the staging environment. @mateoclarke or @chiaberry , do you have any thoughts? will add @amenity for review once we sort that out.

---
#### Ship list
- [x] Code reviewed 
- [x] ~Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#gid=776973707)~

### DIFF
--- a/moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
+++ b/moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
@@ -17,7 +17,9 @@ const useStyles = makeStyles(theme => ({
     display: "flex",
     flex: "1 1 auto",
     overflow: "hidden",
-    paddingTop: 64,
+    // If in staging environment, add extra padding
+    // to make room for staging environment info alert
+    paddingTop: process.env.REACT_APP_HASURA_ENV === "staging" ? 114 : 64,
   },
   contentContainer: {
     display: "flex",
@@ -32,9 +34,6 @@ const useStyles = makeStyles(theme => ({
 }));
 
 const DashboardLayout = () => {
-  console.log(process.env.NODE_ENV);
-  console.log(process.env.REACT_APP_HASURA_ENV);
-  console.log(process.env.APP_ENVIRONMENT);
   const classes = useStyles();
   const [isOpen, setOpen] = useState(false);
   const { user } = useUser();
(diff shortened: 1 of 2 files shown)


amenity:
@tillyw - a few screenshots at different screen widths would work for me!

chiaberry (approved):
I tested locally by adding the option if the environment was local, to show the Alert. Looks good at different sizes too. 

My ideas for ways to Amenity to test would be to have one version on netlify with the env variable being netlify (but then isnt merged to main). Or screen share demo.

moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
```diff
@@ -17,7 +17,9 @@ const useStyles = makeStyles(theme => ({
     display: "flex",
     flex: "1 1 auto",
     overflow: "hidden",
-    paddingTop: 64,
+    // If in staging environment, add extra padding
+    // to make room for staging environment info alert
+    paddingTop: process.env.REACT_APP_HASURA_ENV === "staging" ? 114 : 64,
```
mateoclarke, line 22: These magic numbers make me a little uneasy. Also seems like they should have a unit, right? It all looks good in staging currently so not asking for any changes, just pointing it out.

mateoclarke (commented):
👍

moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
tillyw: thanks for pointing that out, @mateoclarke ! I was wondering if there's a better solution, would be curios to know the best way to handle styling in these different environments!

mateoclarke: Sometimes, especially with things like padding, magic numbers are unavoidable. The most basic thing to do is something like:

```js
const stagingBannerPadding = "114px"
const standardPadding = "64px"
...
  paddingTop: process.env.REACT_APP_HASURA_ENV === "staging" ? stagingBannerPadding : standardPadding,
...
```
This is a way of making the code and variable assignments serve a similar or complementary role to code comments.

tillyw: that is much cleaner, thank you! i'll do that next time or maybe make these tweaks in my next pull request!

PR #567, opened 2022-02-10
WIP: Inspect and set appropriate sequence values for integer primary keys

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/8439.

## Testing

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1644473675208_set_sequence_pointers/down.sql
```diff
@@ -0,0 +1 @@
+-- intentionally left blank
```
frankhereford, line 1: There is no rolling back a set sequence value. At or about the time this PR is to be merged, the values in the up migration are right or they are not right. Moving them to another value in a down migration is not logically viable. Also, this PR can go stale, any inserts into any tables will require it to be re-inspected and updated.

This is OK right now because we're resetting the database, and would take real consideration and a maintenance window if we had live users and data.

moped-toolbox/produce_current_pk_sequences/script/gen_pk_sequences.pl
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/perl
```
frankhereford, line 1: This started as a throw-away script. I can port it to python if needed or removed from this PR.

moped-toolbox/produce_current_pk_sequences/script/gen_pk_sequences.pl
```diff
@@ -0,0 +1,96 @@
...
+  # non-int pk
+  next if $table->{'table_name'} =~ /^moped_city_fiscal_years$/;
+
+  # missing sequence
+  next if $table->{'table_name'} =~ /^moped_phases$/;
+  next if $table->{'table_name'} =~ /^moped_status$/;
+
+  # missing PK
+  next if $table->{'table_name'} =~ /^moped_fund_programs$/;
+  next if $table->{'table_name'} =~ /^moped_proj_fiscal_years$/;
```
frankhereford, line 35: These are tables with unusual primary keys, or a lack thereof.

moped-toolbox/produce_current_pk_sequences/script/gen_pk_sequences.pl
```diff
@@ -0,0 +1,96 @@
...
+      and tco.table_name = ?
+    order by kcu.table_schema,
+             kcu.table_name,
+             position;
+             ";
+  my $query = $pg->prepare($sql);
+  $query->execute($table->{'table_name'});
+  my $pk = $query->fetchrow_hashref;
+
+  print "FREAK OUT: no pk in " . $table->{'table_name'}, "\n" unless $pk->{'key_column'};
```
chiaberry, line 59: 😱

amenity (approved):

moped-database/seeds/1602292389297_initial_seed_staging.sql
```diff
@@ -998,204 +998,6 @@ INSERT INTO public.moped_project_files (project_file_id, project_id, file_key, f
...
-
---
--- TOC entry 6146 (class 0 OID 0)
--- Dependencies: 315
--- Name: moped_workgroup_workgroup_id_seq; Type: SEQUENCE SET; Schema: public; Owner: atd_moped
---
-
-SELECT pg_catalog.setval('public.moped_workgroup_workgroup_id_seq', 1, true);
-
-
```
johnclary, line 1198: @frankhereford - do i understand correctly: here is where you nix the manual setting of the sequence pointer so as to let the DB handle this automatically?

this only affects our local and heroku builds—since these migrations have already been applied to stating and prod. the pointers in staging and prod are being tweaked in your new migration within this PR

moped-database/migrations/1644473675208_set_sequence_pointers/up.sql
```diff
@@ -0,0 +1,29 @@
+SELECT pg_catalog.setval('public.moped_components_component_id_seq', 59, true);
+SELECT pg_catalog.setval('public.moped_department_department_id_seq', 10, true);
+SELECT pg_catalog.setval('public.moped_entities_entity_id_seq', 36, true);
+SELECT pg_catalog.setval('public.moped_fund_sources_funding_source_id_seq', 25, true);
```
chiaberry, line 4: when I spin up locally and check the funding source table, I don't see any source with id 25, it goes from 24 Williamson County to 26 None

chiaberry (approved):
spun up locally, inserted dummy data into some tables in dbeaver, all checks out.

moped-database/migrations/1644473675208_set_sequence_pointers/up.sql
```diff
@@ -0,0 +1,29 @@
+SELECT pg_catalog.setval('public.moped_components_component_id_seq', 59, true);
```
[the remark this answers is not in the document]
frankhereford: Actually, on second thought, I'm going to pass on that. I am not sure that graphql-engine handles comments correctly. It doesn't IIRC, it's doesn't just pipe these commands into `psql`.

PR #570, opened 2022-02-10
Inspect and set appropriate sequence values for integer primary keys & add funding source (RC)

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/8439.

## Testing

It's another database centric PR. Testing should be done locally so you can access a database console and make consequence free `INSERT`s.

**Steps to test:**

* Be able to spin up the DB
* Check that the new funding source is available
* Check that you can insert records, without specifying a ID integer, into any table except for the particular ones called out as having idiosyncratic primary keys:
  * `moped_city_fiscal_years`
  * `moped_phases`
  * `moped_status`
  * `moped_fund_programs`
  * `moped_proj_fiscal_years`

## ⚠️ Considerations for future migrations

* In our migrations, when we need to make insertions into tables, when we do so and must include a specific ID for the record, we should also explicitly increment the sequence to allow the next insert that doesn't have an ID to succeed. 

---
#### Ship list
- [ ] Code reviewed 


chiaberry (approved):
i approved the sibling! so i approve this one as well.

mateoclarke (approved):
🚢 💨 ✨ 

freak_out

amenity (approved):

PR #579, opened 2022-02-14
Refine the tables being tracked in the Moped activity log

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/8055.

## Testing

Testing for this PR is constrained to the `test` instance. Local deployments will lack the AWS based infrastructure. 

**Steps to test:**
Deploy, make any change you can come up with to a project and observe that the changes are reflected in the activity log.

## Deployment considerations
* These changes require a file to be deployed to S3 that maps database tables to primary keys. It needs to be swapped out in conjunction with the deployment of this code.

---
#### Ship list
- [x] Code reviewed 
- [x] ~Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#gid=776973707)~
- [x] Add S3 file operation to deployment checklist


moped-database/config/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.*
+!.gitignore
+!primary_key_map.json
+!hasura.local.yaml
+!hasura_cluster_template.json
+!README.md
```
frankhereford, line 6: This .gitignore makes it really hard (TM) to commit a file in this directory. This is due to a previous version of the README.md saying that this directory is for local configuration files and to be careful about putting things in here in git. This was sage advice, however, we are storing some configuration templates in here, so I have amended the README with more information but continue to honor the old intent. I don't want to cause people to accidentally start committing secrets.

moped-database/config/primary_key_map.json
```diff
@@ -0,0 +1,36 @@
+{
```
frankhereford, line 1: This file needs to be placed in S3 in a particular location, with one copy that reflects the current schema per deployment stage. Right now, we have one for `test`, `staging` and `production`.

moped-database/metadata/tables.yaml
```diff
@@ -824,6 +824,32 @@
       - status_id
       filter: {}
       check: {}
+  event_triggers:
```
frankhereford, line 827: Changes in this file bring new tables into the activity log system and also makes /all/ fields in the tracked tables tracked fields. We were missing a few here and there.

amenity:
Nice, I'm seeing things logging... but I am getting a WSOD on the "Activity log" pretty regularly as I navigate back and forth to check. Loads fine if I refresh. 🤔

Also, is it in scope to make basic refinements to these log entries so they're parseable by users? If not I'll make another issue!
Screen Shot 2022-02-24 at 12 14 41 AM

frankhereford:
@amenity said:
> Nice, I'm seeing things logging... but I am getting a WSOD on the "Activity log" pretty regularly as I navigate back and forth to check. Loads fine if I refresh. 🤔

Hmm, the WSOD is no good. I didn't make any changes to the activity log page in this PR, but .. this PR does cause new data from new tables to be displayed on that page, so it's likely related. Even more, with the given deployment in test, the WSOD is very reproducible using the same project you show in your screenshot. Thank you for that - I'm going to learn what's up, and I'll create an issue to track getting it fixed, if that's OK with you.

> 
> Also, is it in scope to make basic refinements to these log entries so they're parseable by users? Seeing these:
> 
> * [ ]  Task orders should show actual values; [now showing `[object Object]`](https://user-images.githubusercontent.com/1463708/155468885-fa29162c-5ed1-4f8b-8e84-de6a5ba86d87.png)
> * [ ]  Components should specify the component type, now showing [just the action](https://user-images.githubusercontent.com/1463708/155469780-d1803db0-a91c-4404-acd3-a2e18d47196e.png)
> 

I think doing those changes in this PR will be pretty muddy, but I think doing those changes ASAP is a great idea. May I submit a ticket to you to track that work? I can write it up pretty quick and reference some similar work we did to VZ IIRC.

Thanks for checking it out!

amenity:
@frankhereford - my suggestions here... 

screencapture-moped-test-austinmobility-io-moped-projects-227-2022-03-01-12_33_33


Notes: 
- I'm not passionate about removing the "Project" prepending if that complicates things. Just looks redundant. 
- The blue lines point out areas where I'd expect us to provide a value instead of null or a lookup table ID. 
- Even though I marked them in the screenshot, I hesitate to beautify the funding log entries too much because the upkeep is going to be annoying (we'll have to add task orders and funding codes pretty regularly, so this will be a second place to remember to update those) and the feature still seems to be in flux (we'll hopefully be pulling in some of this data from eCAPRIS, for example)

frankhereford:
Hi @amenity,

Thanks again for the really detailed feedback on what you were looking for. It helped a ton.

I have gone back over this and put a good bit of work into catching various cases that make the activity logs hard to read and end up formatted with unexpected punctuation. I also trimmed labels down to the smallest meaningful strings that I could and eliminated a bunch of extransious capitalization. 

The final output is not going to be exactly what your mark-up called for, but it should be a bunch closer -- I can't wait to see what you think.

Here's what that same project would look like under the new formatting. When I put up the new code on test just now, it reset the DB, but if you need, I did keep a copy of the testing DB that has this particular change log entries in it.

Thanks!

image

amenity:
These little bundles are really nice @frankhereford :
Screen Shot 2022-03-03 at 3 38 05 PM

moped-data-events/activity_log/deploy_lambda.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cd venv/lib/python3.8/site-packages
+zip -r9 ${OLDPWD}/lambda.zip .
+cd $OLDPWD
+zip -g lambda.zip app.py config.py MopedEvent.py
+aws lambda update-function-code --function-name atd-moped-events-activity_log_test --zip-file fileb://lambda.zip
```
frankhereford, line 7: This is a small script / utility that has been useful in developing on these lambda functions. It pushes up a new package for this particular lambda function, avoiding having to do a whole fresh test deployment to heroku. 

I'm open to removing this since it's a little out of place here, but I did want to get it on the books in git in case we need to dig it up again.

amenity (approved):
All of the new fields are being tracked, and the log is 100x more legible than before. Thanks, @frankhereford!

Per discussion in Dev Sync, we can continue to refine the human readable translations of activity in the future.

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -1070,11 +874,54 @@ export const ProjectActivityLogCreateDescriptions = {
       return `'${phaseName}' as Project Phase with start date as '${recordData.phase_start}' and end date as '${recordData.phase_end}'`;
     },
   },
+
```
chiaberry, line 877: I was going to be nit-picky and question the empty lines in the objects, but i ran prettier on the file and it kept the empty lines, so I guess its just personal preference and can totally stay in.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -328,29 +340,20 @@ const ProjectActivityLog = () => {
...
-                                          {isFieldMapped(
+                                        {isFieldMapped(
                                             change.record_type,
                                             changeItem.field
                                           )
                                             ? getMappedValue(
                                                 change.record_type,
                                                 changeItem.field,
-                                                String(changeItem.new)
+                                                String(changeItem.old)
```
chiaberry, line 354: should this be `String(changeItem.new)` ?

moped-data-events/activity_log/deploy_lambda.sh
chiaberry: Do you want us to test this script? It might be useful to include some of this explanation in the script.

chiaberry (approved):
I left a question about a variable, but other than that I say after running prettier on some of the projectView/ files, it will be good to go.

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
frankhereford: Now that you mention it, I'm also surprised the linter didn't kick that back. I'm inclined to go with your instincts here; putting those line feeds in there makes it look like a bunch of independent functions and obfuscates the object definition. 🙏

moped-data-events/activity_log/deploy_lambda.sh
frankhereford: I don't think it needs testing, it's only ever really handy to a developer working with that lambda function, but it may at least save the next developer from having to script it.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
frankhereford: You were correct - I just tested this out on `test`.

PR #597, opened 2022-03-03
Skip version check in `hasura` CLI in Github actions

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/8622. 

## Testing

Testing is sort of free on this one. Deploying this to the `test` instance and subsequent deployments to `staging` and `production`, if successful, should be considered an acceptable test. 

- [x] Deploy to `test`

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/.github/workflows/aws-heroku-helper.sh
+++ b/.github/workflows/aws-heroku-helper.sh
@@ -146,18 +146,18 @@ function run_database_migration() {
   cat config.yaml;
 
   print_header "Apply Migrations";
-  hasura migrate apply \
+  hasura --skip-update-check migrate apply \
     --admin-secret "${ATD_MOPED_DEVSTAGE_HASURA_GRAPHQL_ADMIN_SECRET}" \
     --endpoint "${HASURA_SERVER_ENDPOINT}";
 
   print_header "Apply Metadata";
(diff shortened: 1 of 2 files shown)


chiaberry (commented):
could we also get the flag set in the hasura-cluster commands?  L351

frankhereford:
@chiaberry wrote:
> could we also get the flag set in the hasura-cluster commands? L351

@mateoclarke and I posted some quick notes on the issue about this, and he suggested that we might leave it as a reminder that we're slipping away from current versions, which I'm inclined to agree with. What do you think is best? Nothing is set in stone, for sure.

One thing we can do individually though is update the CLI version installed on our laptops' OS, which will accomplish both suppressing the notice message _and_ testing out updated CLI tools with our v.1.3.3 instance. That's why I'm doing, and it's worked so far, although, I've only spun up and down the DB a few times.

amenity (approved):
Thanks, Frank!

chiaberry:
Makes sense to me! I retract my comment

mateoclarke (approved):
Skip it! 🚢 💨 ✨ 

skip_it

PR #612, opened 2022-03-23
add new project types

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/8632

## Testing
**URL to test:** 
https://moped-test.austinmobility.io/moped/

**Steps to test:**
in test env, go to an existing project and edit project type

---
#### Ship list
- [ ] Code reviewed 

Screen Shot 2022-03-23 at 6 20 14 PM

### DIFF
--- a/moped-database/migrations/1648075868116_insert_moped_types/down.sql
+++ b/moped-database/migrations/1648075868116_insert_moped_types/down.sql
@@ -1,1 +1,1 @@
-DELETE FROM public.moped_types WHERE type_order BETWEEN '6' AND '9';
+DELETE FROM public.moped_types WHERE type_name IN ('Fiber - Mod', 'Fiber - New', 'Fiber - Repair');
--- a/moped-database/migrations/1648075868116_insert_moped_types/down.sql
+++ b/moped-database/migrations/1648075868116_insert_moped_types/down.sql
@@ -1,0 +1,1 @@
+DELETE FROM public.moped_types WHERE type_order BETWEEN '6' AND '9';
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -103,7 +103,7 @@ export const SUMMARY_QUERY = gql`
       entity_id
       entity_name
     }
-    moped_types {
+    moped_types(order_by: { type_name: asc }) {
       type_id
       type_name
     }
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
@@ -44,6 +44,8 @@ const ProjectSummaryProjectTypes = ({
     {}
   );
 
+  console.log(typeDict);
+
   // Original Types: array of moped_type objects
   const originalTypes = data?.moped_project[0]?.moped_project_types ?? [];
   // Original Types Ids: array of ids (ints)
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
@@ -44,8 +44,6 @@ const ProjectSummaryProjectTypes = ({
     {}
   );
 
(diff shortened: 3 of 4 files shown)


amenity (approved):
- [x] [Signals projects page](https://moped-test.austinmobility.io/moped/dashboard)
- [x] [Project page](https://moped-test.austinmobility.io/moped/projects/161/)

@tillyw looks good in both instances! What do you think about alphabetizing the types in this PR?

moped-database/migrations/1648075868116_insert_moped_types/down.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM public.moped_types WHERE type_order BETWEEN '6' AND '9';
```
frankhereford, line 1: 🙏 for writing the migrations in an ID agnostic way and letting the DB sequencing stay in sync with the data itself. For this down migration, would you consider using the `type_name` field as the unique identifier for the records, since it's the name that is representing the real-world meaning of each record, not the order in which we choose to display them? Like, we wouldn't ever have two moped_types of "Fiber - New", so it's a good "unique" field to use. Doing this also helps with query readability, FWIW.

```
DELETE FROM public.moped_types
WHERE type_name in ('Fiber - Mod', 'Fiber - New', 'Fiber - Repair');
```

If not, would you consider explicitly defining the field values you want to use in an `IN` clause.  It's been my experience that `BETWEEN` is generally intended to be used to compare integers and to include (`<=` & `>=`) both values given to it as arguments. This query is using it here in string context, which interprets characters by their ASCII or Unicode values, maybe? I'm not sure, but `select 'R' BETWEEN '=' and 't';` evaluates to true. 🤷. The query also could match a record with `type_order` of `9`, which is not in the set of the records you're inserting into in the `up.sql` migration.

frankhereford (approved):
The `up.sql` migration looks great, and it's what does the heavy lifting for this issue.

Would you mind taking a look at the `down.sql` comment? I proposed a slightly different way to define set of records that it would be deleting. Thanks! Let me know if what I wrote doesn't make sense - thank you!!

frankhereford (changes requested):
Sorry Tilly, I meant to set my previous review as "request changes." I included a note about the down migration and how it might be changed to not use BETWEEN in string context or utilize the name field to identify records.

tillyw:
thanks @frankhereford! that's a huge improvement. i pushed an update, please rereview and i'll merge!

frankhereford (approved):
Thanks Tilly!

hammer-top

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectTypes.js
```diff
@@ -44,6 +44,8 @@ const ProjectSummaryProjectTypes = ({
     {}
   );
 
+  console.log(typeDict);
```
frankhereford: Hi Tilly - Before you merge this, do you think it's alright to remove this console.log() call?

frankhereford (changes requested):

mateoclarke (approved):
👍

moped-editor/src/queries/project.js
```diff
@@ -103,7 +103,7 @@ export const SUMMARY_QUERY = gql`
       entity_id
       entity_name
     }
-    moped_types {
+    moped_types(order_by: { type_name: asc }) {
```
mateoclarke, line 106: The only thing that was confusing is that we order by `type_name` when we also have a `type_order` column in the DB. In reading through, it seemed like @amenity called for that change. Just hope its not confusing why we have that `type_order` column in the future.

amenity:
@tillyw - I've tried to deploy this several times but can't get your latest up on test. Seems to choke on [deploying the database](https://github.com/cityofaustin/atd-moped/runs/5722393878?check_suite_focus=true).

moped-editor/src/queries/project.js
mateoclarke: wanna make a `Tech Debt` issue to clean this one up?

amenity:
Yay! Alphabetized looks fab. Thanks, @tillyw! 🚢

PR #624, opened 2022-04-05
Migrate project list table to material table core from gridtable

chiaberry:
This PR changes the project list page to use Material Table Core as its display table instead of Grid Table. It keeps GridTable's query building and filtering and pagination. 

Changing to Material Table will give us more flexibility when it comes to toggling visible columns and potentially other updates. 

## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/8773

## Testing
**URL to test:** 

https://deploy-preview-624--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

I've been going through to match functionality with what GridTable does. Searching, filtering, pagination etc. Nothing should seem different to the user compared to what currently exists in production/staging. 

---
#### Ship list
- [x] Code reviewed 


amenity (commented):
Looks super, @chiaberry! 

You asked a great question about where to merge this, given that it started as a spike... Do you see any _downsides_ to swapping out the tables and getting this in `main`? 

I think that the ultimate goal would actually be to include the advanced search on the ["Signal projects" page](https://deploy-preview-624--atd-moped-main.netlify.app/moped/dashboard). Based on your experience here, do you think that's feasible? If so, I'll create a new issue for that.

chiaberry:
@amenity I can't think of any downsides off the top of my head. I'll keep thinking about it. 

You want to set up the GridTable search/filter functionality on the signals table? It is definitely feasible, but will take some more refactoring to get the query set up in GridTable's configuration. We may have to set up a SQL View like we did for the projects table. I'd have to play around with it. I'd say its a 5.

mateoclarke (approved):
🚢 💨 ✨

chiaberry:
@mateoclarke I've added the grid table pagination functionality to this branch, So I'm going to re-request a review. thanks!

moped-editor/package-lock.json
```diff
@@ -5857,29 +5857,22 @@
       }
     },
     "@material-ui/core": {
```
frankhereford, line 5859: Non-issue, more of a discussion reminder. 💭 

I'm not sure I 100% know the best practices about when and how to check in the package lockfile. I am not aware of anyone doing anything related to updating libraries, so I think this form of the lockfile will be fine to move into main. Maybe this is something we can touch on in an upcoming dev sync - that would be helpful to me.

frankhereford (approved):
Tested this on the netlify PR, and I wasn't able to break it. 💪 I've gone over the code, and nothing sticks out to me. I compared it to `ProjectListView.js`, and everything makes sense to me. 👍 Thank you for making this tricky problem look easy, Chia! 😎

My comment on the lock file is just a placeholder for a conversation I'd like to have to get mofamiliarier with how we handle that as a group, not pointing out anything wrong with this particular version of the file. 🙏

moped-editor/package-lock.json
chiaberry: I checked in the package lockfile because I had to update the version of material ui, which means the lockfile changed -- this is a good discussion to have because I just figured if the package.json was changed, then the package-lock needed to also be included, but I don't know if thats best practice or what the alternative are.

mateoclarke (approved):
👏 👏 👏 

Approving again!

PR #626, opened 2022-04-07
Change loading icon and data refreshing on comments tab

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/8001

## Testing
**URL to test:** 

https://8001-comments--atd-moped-main.netlify.app/moped/projects/1825?tab=comments

**Steps to test:**

This is a tricky one to test, because I couldn't get the bug to consistently occur before my changes. But periodically, quickly toggling between comment type buttons would result in just a loading icon that would go away once you refresh the page. 

I believe it was because of two things: The loading icon appeared if `loading` was true or `!data` was true. Not always, but sometimes we would fall into the state of loading being false yet having no data. The network request tab did have a successful request, so something was getting stuck. I moved the loading icon to appear under the input since the input doesn't need to disappear when the comments are loading. 

~~Also I changed it so that if there is no data, instead of there being a loading icon, the "no comments to display message appears". The message still appeared sporadically even when there were comments, so I moved `data` into the useEffect call so if there is a change in data the page will re-render.~~

Spurred on by John's comments pointing out we are firing off the same graphql query multiple times, I refactored the code to only query if something has been updated and instead use local state to filter the comment types when toggling between different types. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -92,7 +92,7 @@ const ProjectComments = props => {
   const [commentAddSuccess, setCommentAddSuccess] = useState(false);
   const [editingComment, setEditingComment] = useState(false);
   const [commentId, setCommentId] = useState(null);
-  const [displayNotes, setDisplayNotes] = useState({});
+  const [displayNotes, setDisplayNotes] = useState([]);
   const [noteType, setNoteType] = useState(0);
 
   const { loading, error, data, refetch } = useQuery(COMMENTS_QUERY, {
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -198,6 +198,8 @@ const ProjectComments = props => {
    */
   const filterNoteType = typeId => setNoteType(Number(typeId));
 
+  const mopedProjNotes = data?.moped_proj_notes;
+
   /**
    * Whenever noteType changes, we change the query conditions
    */
@@ -210,7 +212,7 @@ const ProjectComments = props => {
       });
     }
     refetch();
-  }, [noteType, setNoteTypeConditions, refetch, data]);
+  }, [noteType, setNoteTypeConditions, refetch, mopedProjNotes]);
 
   if (error) return console.log(error);
 
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -99,8 +99,6 @@ const ProjectComments = props => {
   const [noteType, setNoteType] = useState(0);
   const [noteTypeConditions, setNoteTypeConditions] = useState({});
 
-  console.log(noteType, " the noteType is <------")
-
   const { loading, error, data, refetch } = useQuery(COMMENTS_QUERY, {
     variables: {
       projectNoteConditions: {
@@ -215,17 +213,15 @@ const ProjectComments = props => {
         project_note_type: { _eq: noteType },
       });
     }
-    console.log("im refetching")
     refetch();
   }, [noteType, setNoteTypeConditions, refetch]);
 
-  console.log("loading ", loading)
-  console.log("DATA ", data)
+  console.log("loading ", loading, "DATA ", data)
 
+  if (error) return console.log(error);
   // If the query is loading or data object is undefined,
   // stop here and just render the spinner.
   if (loading || !data) return <CircularProgress />;
-  if (error) return console.log(error);
 
   /**
    * Defines the CommentButton with a toggle style-change behavior.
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -274,16 +266,17 @@ const ProjectComments = props => {
...
                 <List className={classes.root}>
                   {data.moped_proj_notes.map((item, i) => {
                     const isNotLastItem = i < data.moped_proj_notes.length - 1;
                     const editableComment =
                       userSessionData.user_id === item.added_by_user_id ||
                       userHighestRole === "moped-admin";
-
                     return (
-                      <>
+                      <React.Fragment key={item.project_note_id}>
```
chiaberry, line 279: fixes a react warning about not having keys on list items

moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -214,11 +210,8 @@ const ProjectComments = props => {
       });
     }
     refetch();
-  }, [noteType, setNoteTypeConditions, refetch]);
+  }, [noteType, setNoteTypeConditions, refetch, data]);
```
mateoclarke, line 213: we talked about potentially putting a more specific variable `data?.moped_proj_notes` here instead of `data`. I don't really have a preference but kinda curious if that would work too.

mateoclarke (approved):
🚢 💨 ✨

moped-editor/src/views/projects/projectView/ProjectComments.js
chiaberry: so I did move it out into its own separate variable, instead of just checking all of data

johnclary: hmm, i'm kinda surprised this isn't causing an infinite hook loop, since I would expect `data` to change after `refetch()`. i wonder if  apollo caching is keeping the data object reference persistent and preventing infinity... but in any case this doesn't look ideal to me.

running locally, i noticed we're firing off the same graphql query three times when you toggle between note types. i believe those are:

- `useQuery` fires after `filterNoteType()` changes the query params
- `refetch()` fires in `useEffect` b/c of `noteType` dependency
- `refetch()` fires again in `useEffect` because the `data` dependency changed

haven't debugged much, but yeah i'm stumped as to why this bug is happening at all.

johnclary: could the `setTimeout`s be messing up state in an unexpected way? we use those just to force the animations to persist for a sec?

frankhereford (approved):
Tested on the netlify deployment, and the responsiveness improvement is very obvious. Thank you for taking the time to return to do this in state - it's improved as much again!

🚢 🚢

amenity (approved):
👍 🙏 Thanks, @chiaberry!

johnclary:
Thanks @chiaberry! I really like the approach of simplifying the query + state. I think you can go even further—there's not a lot of rendering going on here, so I *think* you could drop the `useEffect`s and add the filter condition to your comment `<List>` renderer around L282.

Something like
```js
displayNotes.filter(
      n => noteType === 0 || n.project_note_type === noteType
    )
    .map(item, i => { ....
```

In which case, I'm not sure you need `useState` for `displayNotes`, you can just assign it on every render if you're tired of typing `data.moped_proj_notes`:

```js
const displayNotes = data?.moped_proj_notes
```

moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -96,20 +92,21 @@ const ProjectComments = props => {
   const [commentAddSuccess, setCommentAddSuccess] = useState(false);
   const [editingComment, setEditingComment] = useState(false);
   const [commentId, setCommentId] = useState(null);
+  const [displayNotes, setDisplayNotes] = useState({});
```
johnclary, line 95: if you you're gonna keep this state hook, i think it should initialize as an empty array or `null` instead of an empty object?

johnclary (approved):

moped-editor/src/views/projects/projectView/ProjectComments.js
mateoclarke: Seems like a bit of code smell to have the `key` prop on `<React.Fragment>` instead of on `<ListItem>` component but I think I'm seeing why its necessary. I also have a slight preference for the `<>` fragment syntax so it feels like a step backward to have to spell it out. Is it because line 377 needs to stay inside the Fragment? Or could that line move outside the fragment but still inside the closing bracket on 379? Guess it doesn't really matter a whole bunch either way. Sorry for my rambling.

moped-editor/src/views/projects/projectView/ProjectComments.js
mateoclarke: good catch. Array seems best right? Since that's what we expect `displayNotes` to be?

mateoclarke (approved):
SO much better in terms of UX performance! Code review notes are super minor.

bunny-floss-dancing

chiaberry:
I experimented with using a filter expression like John suggested, but ran into some complexity with the array when editing comments and handling the `isNotLastItem` logic. I am going to merge as is.

johnclary:
thanks for looking into it!

PR #627, opened 2022-04-07
sets project_length field default to null, drops compute_project_length function

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/8862

## Testing
**URL to test:** 

**Steps to test:**
on test instance, create a new project

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1649349626445_alter_project_length_field_drop_compute_project_length_function/down.sql
```diff
@@ -0,0 +1,21 @@
...
+then date(now()) - moped_project.start_date
+else moped_project.end_date - moped_project.start_date
+end as project_length
+FROM moped_project
+$$;
+
+-- Drop null default and set default to compute_project_length
+ALTER TABLE "public"."moped_project"
+    ALTER COLUMN "project_length" DROP DEFAULT,
+    ALTER COLUMN "project_length" SET DEFAULT public.compute_project_length();
```
frankhereford, line 21: This down.sql file does not seem to restore the database to the state that it would have been in prior to running the associated up.sql commands. Also, I'm not sure these changes fully capture the full task, as I understand it. If you have clearer direction than what I'm aware of, please forgive this feedback:

The up.sql does:
* Drop the default value of the `project_length` field 🆗  
* Set the default value to `null` for the same field 🆗 
* Drop the function `compute_project_length ✅ 
* It does not drop the `end_date` column ❌ 
* It does not drop the `project_length` column ❌ 

Then the down.sql:
* Adds back the `start_date` column ❓ 
  * This is not a restoration of something happened in the up.sql. If you needed to do schema changes, they should come in matching pairs and in separatete migrations.
* recreates the function, which is good and should be in there. ✅ 
* Restores the function as the default value for `project_length` ✅

frankhereford, line 21: EOF line feed if that's alright

moped-database/migrations/1649349626445_alter_project_length_field_drop_compute_project_length_function/up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "public"."moped_project"
+    ALTER COLUMN "project_length" DROP DEFAULT,
+    ALTER COLUMN "project_length" SET DEFAULT null;
+
+DROP FUNCTION IF EXISTS "public"."compute_project_length";
```
frankhereford, line 5: EOF line feed

frankhereford (changes requested):
Generally, the pair of sql files don't do the opposite actions of each other. Splitting the migration into two may allow more flexibility and to stay within the migration system, depending on what operations you really want to do in this PR.

Thanks! 🙏

moped-database/migrations/1649349626445_alter_project_length_field_drop_compute_project_length_function/down.sql
chiaberry: I agree with Frank that this file needs to be updated to not add back the start_date, since that wasn't addressed in its corresponding up.sql file. 

Personally I think the end_date and project_length columns aren't needed, but I do remember that your slack comment about this issue stated you weren't going to address that in this PR, but rather we would discuss as a team in the future.

mateoclarke: I gave Tilly the direction to leave `end_date` and `project_length` in the database for now. This was after asking @amenity if they were needed. She said yes, that other divisions may require those fields when we import their project data. I proposed we make a backlog issue to address project `end_date` and `project_length` later.

mateoclarke:
> Generally, the pair of sql files don't do the opposite actions of each other. Splitting the migration into two may allow more flexibility and to stay within the migration system, depending on what operations you really want to do in this PR.
> 
> Thanks! 🙏

I told Tilly it was fine to put both of these changes in the same migration. Sure, it may allow more flexibility within the migration system, but it didn't seem necessary.

mateoclarke (approved):
I've address the PR feedback and I'm going to merge on Tilly's behalf so that this patch gets into staging. Please request further changes in a new Github issue.

frankhereford (approved):

frankhereford:
@mateoclarke wrote:
> I told Tilly it was fine to put both of these changes in the same migration. Sure, it may allow more flexibility within the migration system, but it didn't seem necessary.

I'm sorry, I was unclear. I intended to refer to the operations related to the `start_date` field in being in one migration and the two new operations (dropping the function & altering `project_length`'s default value) in a second migration. Thanks for pointing out that unclear observation 🙏

mateoclarke:
Deployed to test and the patch fixes the regression. 👍

PR #630, opened 2022-04-12
sets funding_status_id to user selection or tentative if no selection

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/8721

## Testing
**URL to test:** 
https://deploy-preview-630--atd-moped-main.netlify.app/

**Steps to test:**
1) create a new funding source setting a status value
2) create a new funding source without setting a status value

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
@@ -314,7 +314,7 @@ const ProjectFundingTable = () => {
     <Select
       style={{ minWidth: "8em" }}
       id={props.name}
-      value={props.value || ""}
+      value={props.value || props.defaultValue}
     >
       {props.data.map(item => (
         <MenuItem
@@ -413,6 +413,7 @@ const ProjectFundingTable = () => {
         <LookupSelectComponent
           {...props}
           name={"funding_source"}
+          defaultValue={""}
           data={data.moped_fund_sources}
         />
       ),
@@ -431,6 +432,7 @@ const ProjectFundingTable = () => {
         <LookupSelectComponent
           {...props}
           name={"funding_program"}
+          defaultValue={""}
           data={data.moped_fund_programs}
         />
       ),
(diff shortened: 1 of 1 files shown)


tillyw:
in addressing this issue, had a conversation with @amenity about setting value to null (displayed as blank value in ui) if user does not select a value but decided to set value to 1 ("tentative") in these cases. setting value as null creates conflicts with the way projects are archived (or soft deleted) because they are not displayed in the ui if they have a null value. graphql query for displaying records only displayed records with value not equal to 0, which in this case appears to include null values. @frankhereford suggested changing the way we archive or soft delete projects by altering schema to include an "archived" boolean or similar stragegy.

mateoclarke (approved):
🚢 💨 ✨ 

betty-white-dab

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -676,7 +676,7 @@ const ProjectFundingTable = () => {
               }),
           onRowUpdate: (newData, oldData) => {
             const updateProjectFundingData = newData;
-
+
```
frankhereford, line 679: Do you mind running this through prettier? 🙏

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -657,7 +657,7 @@ const ProjectFundingTable = () => {
                   ...newData,
                   project_id: projectId,
                   added_by: getDatabaseId(user),
-                  funding_status_id: 1,
+                  funding_status_id: newData.funding_status_id || 1,
```
frankhereford, line 660: This works great with the unspecified default being tentative or user input if provided. 🙌

frankhereford (approved):
🎆 🙏 🚢

amenity:
Nice! My only potential tweak would be if we could show "Tentative" already populated when they add a new line. So they aren't surprised when they submit and it's filled in:

tillyw:
@amenity updated pr so that dropdown is automatically set to tentative unless user selects another option. please take a look!

mateoclarke (approved):
👍

amenity (approved):
LGTM! Thanks, Tilly!

PR #631, opened 2022-04-14
Add columns to project view

chiaberry:
## Associated issues

This PR adds task order, contractor and project DO# to the project list. You can also filter projects by those fields. 

## Testing
**URL to test:** 

https://moped-test.austinmobility.io/moped/ 

There are database changes, so please use moped test or run locally. 

**Steps to test:**

Screen Shot 2022-04-15 at 4 43 28 PM

Toggle columns to show or hide. Using advanced search for the columns works, however whenever the page reloads, the default column view is shown. 

---
#### Ship list
- [x] Code reviewed 


moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    mp.purchase_order_number
+   FROM ((((((public.moped_project mp
+     LEFT JOIN public.moped_proj_personnel mpp ON (((mp.project_id = mpp.project_id) AND (mpp.status_id = 1))))
+     LEFT JOIN public.moped_users mu ON ((mpp.user_id = mu.user_id)))
+     LEFT JOIN public.moped_project_roles mpr ON ((mpp.role_id = mpr.project_role_id)))
+     LEFT JOIN public.moped_entity me ON ((me.entity_id = mp.project_sponsor)))
+     LEFT JOIN public.moped_proj_partners mpp2 ON (((mp.project_id = mpp2.project_id) AND (mpp2.status_id = 1))))
+     LEFT JOIN public.moped_entity me2 ON ((mpp2.entity_id = me2.entity_id)))
+     left join jsonb_array_elements(mp.task_order) as task_order_filter on true
+  GROUP BY mp.project_uuid, mp.project_id, mp.project_name, mp.project_description, mp.project_description_public, mp.ecapris_subproject_id, mp.project_importance, mp.project_order, mp.current_status, mp.timeline_id, mp.current_phase, mp.end_date, mp.project_length, mp.fiscal_year, mp.capitally_funded, mp.project_priority, mp.date_added, mp.added_by, mp.is_retired, mp.milestone_id, mp.status_id, me.entity_name, mp.updated_at, mp.task_order, mp.contractor, mp.purchase_order_number;
```
frankhereford, line 46: I think with this many fields in the `GROUP BY` clause, I would delimit the fields over new lines, the same way you have done above in your `SELECT` clause.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    string_agg(DISTINCT me2.entity_name, ', '::text) AS project_partner,
+        CASE
+            WHEN ((mp.status_id = 0) OR (mp.status_id IS NULL)) THEN NULL::text
+            WHEN (mp.status_id = 1) THEN mp.current_phase
+            ELSE mp.current_status
+        END AS status_name,
+    string_agg(task_order_filter->>'display_name', ',') as task_order_name,
+    mp.contractor,
+    mp.purchase_order_number
+   FROM ((((((public.moped_project mp
```
frankhereford, line 38: `((((((`

This `FROM` clause is using a lot of parenthetical grouping to make explicit the default order and binding of the `JOIN` operators. In my opinion, unless you're changing the order of operations, it's more clear to use include the precedence operators. (keywords? syntax? I'm not sure what you call what parentheses are.) 

I think these could be removed to increase clarity and not modify the way the query works.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    mp.is_retired,
+    mp.milestone_id,
+    mp.task_order,
+    COALESCE(mp.status_id, 0) AS status_id,
+    string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members,
+    mp.updated_at,
+    me.entity_name AS project_sponsor,
+    string_agg(DISTINCT me2.entity_name, ', '::text) AS project_partner,
+        CASE
+            WHEN ((mp.status_id = 0) OR (mp.status_id IS NULL)) THEN NULL::text
```
frankhereford, line 31: `NULL::text` seems like it should just be `NULL`, similar to above.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    mp.date_added,
+    mp.added_by,
+    mp.is_retired,
+    mp.milestone_id,
+    mp.task_order,
+    COALESCE(mp.status_id, 0) AS status_id,
+    string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members,
+    mp.updated_at,
+    me.entity_name AS project_sponsor,
+    string_agg(DISTINCT me2.entity_name, ', '::text) AS project_partner,
```
frankhereford, line 29: I'm curious about the `::text`s. It looks like the string `,` is being cast into a text type, which it should already have been. Can it be removed?

moped-database/migrations/1649874244000_alter_project_list_view/down.sql
```diff
@@ -0,0 +1,41 @@
+drop view project_list_view;
+
+CREATE VIEW public.project_list_view AS
```
frankhereford, line 3: This is the `down.sql` which is putting the old view back in place, so I totally am on board with just taking the definition as spit out of a DB client, `graphql-engine` or `pg_dump` most ideally, but if it's written by hand, the same sort of suggestions as the `up.sql` apply.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+        CASE
+            WHEN ((mp.status_id = 0) OR (mp.status_id IS NULL)) THEN NULL::text
+            WHEN (mp.status_id = 1) THEN mp.current_phase
+            ELSE mp.current_status
+        END AS status_name,
+    string_agg(task_order_filter->>'display_name', ',') as task_order_name,
```
frankhereford, line 35: Do you want to put a space after the `,` so that the list is more `one fish, two fish, red fish, blue fish` and less jammed together?

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -122,6 +122,45 @@ export const ProjectsListViewFiltersConf = {
...
+      name: "task_order_name",
+      label: "Task orders",
+      placeholder: "Task order",
+      type: "string",
+      operators: [
+        "string_contains_case_insensitive",
+        "string_begins_with_case_insensitive",
+        "string_ends_with_case_insensitive",
+        "string_is_null",
+        "string_is_not_null",
```
frankhereford, line 135: I see that the view is wrapping up 1 or more task order names into a comma delimited list. The result of that *is* a string, so this should work fine, but I wonder if we might ever want to expand these operators to do special stuff that handles arrays of values aggregated by the postgres database. This isn't a request for a change here, just a place maker for the idea of expanding the comparison operators.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    mp.contractor,
+    mp.purchase_order_number
+   FROM ((((((public.moped_project mp
+     LEFT JOIN public.moped_proj_personnel mpp ON (((mp.project_id = mpp.project_id) AND (mpp.status_id = 1))))
+     LEFT JOIN public.moped_users mu ON ((mpp.user_id = mu.user_id)))
+     LEFT JOIN public.moped_project_roles mpr ON ((mpp.role_id = mpr.project_role_id)))
+     LEFT JOIN public.moped_entity me ON ((me.entity_id = mp.project_sponsor)))
+     LEFT JOIN public.moped_proj_partners mpp2 ON (((mp.project_id = mpp2.project_id) AND (mpp2.status_id = 1))))
+     LEFT JOIN public.moped_entity me2 ON ((mpp2.entity_id = me2.entity_id)))
+     left join jsonb_array_elements(mp.task_order) as task_order_filter on true
```
frankhereford, line 45: This is powerful! ✨🎯

What a nice way to wrap up a 1:many relationship in a way that is just ready to roll and be parsed and sliced in the app.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -298,8 +298,50 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
...
+
+        return <div style={{ maxWidth: "265px" }}>{content}</div>;
+      },
+    },
+    {
+      title: "Contractor/Contract",
+      field: "contractor",
+      hidden: true,
+      emptyValue: "-",
+      render: entry => (entry.contractor === "" ? "-" : entry.contractor),
```
frankhereford, line 326: The render line seems to re-implement the expected behavior from the `emptyValue` configuration. Not a problem, just seemed to stick out.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -298,8 +298,50 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
...
+      render: entry =>
+        entry.purchase_order_number.trim().length === 0
+          ? "-"
+          : entry.purchase_order_number,
+    },
   ];
 
+  if (data) {
+    console.log(data[query.table])
+    console.log(query.query)
```
frankhereford, line 342: Do you want to make these `.info()`s or take them out?

frankhereford (approved):
@chiaberry, thank you again for this. 

Also, thank you for your time to discuss many of the issues noted below - please forgive the slew of comments, it's all way to picky. This is great work!


https://user-images.githubusercontent.com/3653206/164051432-e5bb4798-e14e-448b-8033-9e47090e9e6e.mp4

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: I want to take them out! thank you for flagging them

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: this is a quirk of material table, if it is null or undefined then it will render the emptyValue "-"
But if it is an empty string, material table doesn't identify that as empty so it would render the empty string. This way it looks consistent. There isn't a way for someone to delete a contractor besides saving an empty string.

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
chiaberry: I do want to put a space!

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
chiaberry: removed them and does not seem to affect the query behavior one bit. thank you

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
chiaberry: yes!

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
chiaberry: I didn't change this one because the other ones were strings being cast into text, but in this case its NULL being cast into text. does this return `"null"` vs `null`?

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
chiaberry: much easier to read

moped-database/migrations/1649874244000_alter_project_list_view/up.sql
frankhereford: I waffled on mentioning this one too. I should have waffled on the side of not saying anything 🧇 . 

```
select
    null as plain_null,
    null::text as text_casted_null,
    pg_typeof(null) as type_of_plain_null,
    pg_typeof(null::text) as type_of_text_casted_null;
```
Yields:

image

Thank you for catching this and asking I dig deeper. 🙏 👍

chiaberry: no, thank you!

amenity (approved):
Beautiful! Thank you, @chiaberry

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -351,6 +389,7 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
                       // is conflicting with the search/filter dropdown
                       zIndex: 1,
                     },
+                    columnsButton: true,
```
mateoclarke, line 392: ✨

mateoclarke (approved):
Hooray for more columns! 🏛️

sega-colums-1990

PR #636, opened 2022-04-20
Adds Project "Following" feature to Database schema [DB] and Project Summary page [UI]

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/8910
fixes https://github.com/cityofaustin/atd-data-tech/issues/8909
closes https://github.com/cityofaustin/atd-data-tech/issues/8919#

## Testing
**URL to test:** 

**Steps to test:**
go to a project and select and deselect the bookmark icon in the top right

---
#### Ship list
- [x] Code reviewed 

https://user-images.githubusercontent.com/35410637/165636613-16b063dd-3c18-437e-8a1f-bbc507dcc090.mov


mateoclarke:
@tillyw Since this PR addressed both the front and backend issues, could you update the PR title to reflect that please?

tillyw:
> @tillyw Since this PR addresses both the front and backend issues, could you update the PR title to reflect that please?

I have update the title. Thanks for pointing that out!

moped-database/migrations/1651095287013_squashed/up.sql
```diff
@@ -0,0 +1,18 @@
+
+CREATE TABLE "public"."moped_user_followed_projects"("id" integer NOT NULL, "created_at" Timestamp NOT NULL DEFAULT now(), "project_id" Integer NOT NULL, "user_id" integer NOT NULL, PRIMARY KEY ("id") , FOREIGN KEY ("user_id") REFERENCES "public"."moped_project"("project_id") ON UPDATE restrict ON DELETE restrict, FOREIGN KEY ("project_id") REFERENCES "public"."moped_users"("user_id") ON UPDATE restrict ON DELETE restrict, UNIQUE ("id"));
+
+alter table "public"."moped_user_followed_projects" drop constraint "moped_user_followed_projects_user_id_fkey";
+
+alter table "public"."moped_user_followed_projects" drop constraint "moped_user_followed_projects_project_id_fkey";
+
+ALTER TABLE "public"."moped_user_followed_projects" ADD COLUMN "id2" serial NOT NULL UNIQUE;
```
frankhereford, line 8: This statement ends up making a key designation for `id2` along with a `unique` constraint that are orphaned at the end of this migration.
Screen Shot 2022-04-28 at 2 00 37 PM

moped-editor/src/queries/project.js
```diff
@@ -114,6 +114,13 @@ export const SUMMARY_QUERY = gql`
       status_id
       status_name
     }
+    moped_user_followed_projects(
+      where: { project_id: { _eq: $projectId }, user_id: { _eq: $userId} } 
+    ) {
+      project_id
+      user_id
+      id
```
frankhereford, line 122: Would you consider not requesting back this ID from `graphql-engine`? It doesn't look like it gets used. I think not using it is correct -- that is to say, the way you delete a "following" record is perfect.

moped-editor/src/queries/project.js
```diff
@@ -401,6 +408,32 @@ export const ADD_PROJECT_MILESTONE = gql`
...
+`;
+
+export const PROJECT_UNFOLLOW = gql`
+  mutation UnfollowProject($project_id: Int!, $user_id: Int!) {
+    delete_moped_user_followed_projects(where: { project_id: { _eq: $project_id }, user_id: { _eq: $user_id } }) {
+      returning {
+        id
+        project_id
+        user_id
+      }
```
frankhereford, line 432: I don't think it makes sense to ask for `graphql-engine` to return data about a record that was deleted. Can this be removed?

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -103,6 +110,17 @@ const useStyles = makeStyles(theme => ({
...
+    float: "right",
+  },
+  unfollowIcon: {
+    color: theme.palette.primary.main,
+    fontSize: "2rem",
+  },
+  followIcon: {
+    color: theme.palette.text.secondary,
+    fontSize: "2rem",
+  },
```
frankhereford, line 123: These icons turned out really great. 💯 I love how they look - they communicate what the feature is and how to use it with so little - it's very elegant.

frankhereford (changes requested):
I mentioned it below, but I love how this PR turned out. The bookmark / following functionality is 💯 🔥 👍.

I have a few requests regarding the migrations and graphql queries, but they are focused around clarity of the code and not functionality at all.

I've tested this locally and reviewed the code. This is working great, and I'll really grateful for your work to get this ready - thanks again!

moped-editor/src/views/projects/projectView/ProjectView.js
tillyw: thanks, @frankhereford! I'm starting down here and working my way up

moped-editor/src/queries/project.js
tillyw: @frankhereford I see your point but I don't know if it's possible not to return any data. At least every way I have tried results in a syntax error. This [article](https://stackoverflow.com/questions/44737043/is-it-possible-to-not-return-any-data-when-using-a-graphql-mutation) makes it seem a bit complicated. Can you please advise?

tillyw: Actually I was able to rework it like this. Is this closer to what you have in mind?

```
  mutation UnfollowProject($project_id: Int!, $user_id: Int!) {
    delete_moped_user_followed_projects(where: { project_id: { _eq: $project_id }, user_id: { _eq: $user_id } }) {
      affected_rows
    }
  }
```

moped-database/migrations/1651095287013_squashed/up.sql
tillyw: I believe that reworking the `create_table` migration according to what you suggest below resolves this issue. Can you please confirm?

moped-editor/src/queries/project.js
```diff
@@ -401,6 +407,27 @@ export const ADD_PROJECT_MILESTONE = gql`
...
+    }
+  }
+`;
+
+export const PROJECT_UNFOLLOW = gql`
+  mutation UnfollowProject($project_id: Int!, $user_id: Int!) {
+    delete_moped_user_followed_projects(where: { project_id: { _eq: $project_id }, user_id: { _eq: $user_id } }) {
+      affected_rows
```
frankhereford, line 426: Perfect!

frankhereford (approved):
This is all right on point. Thanks Tilly! 💯

moped-database/migrations/1651095287013_squashed/up.sql
frankhereford: Yep, this one is all squared away. Thanks!

moped-editor/src/queries/project.js
frankhereford: I noted it I think in another comment, but to your question, this is perfect IMHO.

moped-database/metadata/actions.graphql
```diff
@@ -0,0 +1,2 @@
+
+
```
mateoclarke, line 2: What's going on with this file? Seems weird that two blank lines were added.

moped-database/metadata/tables.yaml
```diff
@@ -2614,6 +2614,7 @@
   event_triggers:
   - name: activity_log_moped_project
     definition:
+      enable_manual: false
```
mateoclarke, line 2617: I'm seeing a lot of these `enable_manual` updates to several tables in this file and along with other trivial changes. I assume these metadata changes are mostly automated but I don't think we should commit changes unless they are related to the feature. Do you know what the `enable_manual` flag does? It seems like the only changes that should've been committed are the ones related to `moped_user_followed_projects` around line 3344. The rest just seems like unnecessary noise.

moped-database/migrations/1651095287013_create_table_moped_user_followed_projects/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "public"."moped_user_followed_projects" (
+    "id" serial primary key,
+    "project_id" integer not null,
+    "user_id" integer not null,
+    "created_at" timestamp default now()
```
mateoclarke, line 5: 👍

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -411,6 +458,24 @@ const ProjectView = () => {
...
+                            <Tooltip title="Unfollow">
+                              <BookmarkIcon className={classes.unfollowIcon} />
+                            </Tooltip>
+                          ) : (
+                            <Tooltip title="Follow">
+                              <BookmarkBorderIcon
+                                className={classes.followIcon}
+                              />
+                            </Tooltip>
+                          )}
```
mateoclarke, line 476: One way to make this code more concise would be to only put a ternary operator on the differences in the props instead of the whole nested components. Something like the following:

```js
<Tooltip title={isFollowing ? "Unfollow" : "Follow" }>
  <BookmarkIcon className={isFollowing ? classes.unfollowIcon : classes.followIcon} />
</Tooltip>
```

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -402,7 +449,7 @@ const ProjectView = () => {
                 <div className={classes.root}>
                   <Box p={4} pb={2}>
                     <Grid container>
-                      <Grid item xs={12}>
+                      <Grid item xs={11}>
```
mateoclarke, line 452: I'm curious why this change was needed.

moped-editor/src/queries/project.js
```diff
@@ -401,6 +407,27 @@ export const ADD_PROJECT_MILESTONE = gql`
   }
 `;
 
+export const PROJECT_FOLLOW = gql`
+  mutation FollowProject(
+    $objects: [moped_user_followed_projects_insert_input!]!
+  ) {
+    insert_moped_user_followed_projects(objects: $objects) {
```
mateoclarke, line 414: I'm not sure if it really matters and I don't know if its worth refactoring but since we are only ever inserting one row of data, I think the `insert_moped_user_followed_projects_one` mutation would be most appropriate. It only takes one object as its argument instead of an array of objects.

Screen Shot 2022-04-28 at 6 45 02 PM

moped-editor/src/views/projects/projectView/ProjectView.js
mateoclarke: I think you could replace lines 466-476 with just those 3 lines.

mateoclarke (approved):
Great job implementing this Follow feature from front to back! It works great!

following-the-leader-following

moped-database/metadata/actions.graphql
frankhereford: I have observed this happen more than a handful of times. With some rhyme or reason all to itself, the `graphql-engine` process will write out these two lines when the console is used to create migrations / change the schema. I don't think they do any harm; i think of them as similar to leaving a console.log() in a JS app. I'd be genuinely surprised if this hasn't been fixed up in a more recent version of the graphql endpoint daemon.

amenity (approved):
Works beautifully! Thanks, @tillyw .

moped-editor/src/views/projects/projectView/ProjectView.js
tillyw: This is so much better, thank you for the suggestion!

moped-database/metadata/actions.graphql
mateoclarke: Yes, as a general rule, I would only commit files or chunks of code that I've actively changed and intend to be in the PR.

moped-editor/src/views/projects/projectView/ProjectView.js
tillyw: The icon type needs to be conditionally rendered as well, so the final code looks like this:

`                          <Tooltip title={isFollowing ? "Unfollow" : "Follow"}>
                            {isFollowing ? (
                              <BookmarkIcon className={classes.unfollowIcon} />
                            ) : (
                              <BookmarkBorderIcon
                                className={classes.followIcon}
                              />
                            )}
                          </Tooltip>`

moped-editor/src/views/projects/projectView/ProjectView.js
tillyw: It was not, thank you for catching that!

moped-database/metadata/tables.yaml
```diff
@@ -736,17 +736,17 @@
   - name: moped_proj_components_subcomponents
     using:
       foreign_key_constraint_on:
```
tillyw, line 738: wondering if these changes related to `foreign_key_constraint_on` are relevant

moped-database/metadata/tables.yaml
```diff
@@ -736,17 +736,17 @@
...
+        column: project_component_id
         table:
-          name: moped_proj_components_subcomponents
           schema: public
-        column: project_component_id
+          name: moped_proj_components_subcomponents
   - name: moped_proj_features
     using:
       foreign_key_constraint_on:
```
tillyw, line 745: same with these, since the reshuffling seems like it may impact the way things are nested?

chiaberry: the nesting is the same, just shuffled around

moped-database/metadata/tables.yaml
chiaberry: This happened to me too once when i was changing things in the hasura console. Also often times all of the permissions are shuffled around if i change them via the console. It annoys the heck out of me, and I've manually changed the table so only the pertinent changes are applied.

tillyw: thanks for your perspective on this @mateoclarke and @chiaberry! i've been combing through this and agree the only relevant changes are related to permissions for the new `moped_user_followed_projects` table and everything else is just shuffling around existing information. i'm going to revert to the original file and manually add in the permissions updates and ping y'all when that's done!

tillyw: Going to mark this as resolved per our conversation during dev sync where we decided to go ahead and merge this file in as-is but be more selective about which changes to this file to commit in the future.

moped-editor/src/views/projects/projectView/ProjectView.js
tillyw: Actually removing that caused formatting issues. I set that grid item to 11 so I could create another grid item for the following button and set that to 1. I'm returning it to that state but please let me know if I'm missing something @mateoclarke

moped-editor/src/queries/project.js
tillyw: Done!

PR #640, opened 2022-04-25
Add Signal IDs and project type columns to project list view

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/8930

Adds two new columns to be able to toggle on and off in the project list view. 

Does not add either of them to advance filtering. 

I had to change the sql view to use a common table expression (CTE) because of unexpected repeat type_names based on how many people were in the personnel table. Thank you to @frankhereford for helping me understand what was happening. 

## Testing
**URL to test:** 

test in moped-test or run locally

**Steps to test:**

Add records with signal ids and set project types. The signal IDs column is on by default, the project types column is hidden.  See them in the table and toggle columns on and off. 

Screen Shot 2022-04-25 at 3 38 47 PM

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -90,10 +90,8 @@ export const ProjectsListViewFiltersConf = {
         "string_contains_case_insensitive",
         "string_begins_with_case_insensitive",
         "string_ends_with_case_insensitive",
-        // Because of the way the SQL view concatanates the team members and roles
-        // " :" is always present even if there are no team members present
-        "string_is_null_special_case",
-        "string_is_not_null_special_case",
+        "string_is_null",
+        "string_is_not_null",
```
chiaberry, line 94: we don't need to handle a special case where no team members were showing up as " : "

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,89 @@
...
+            WHEN (mp.status_id = 1) THEN mp.current_phase
+            ELSE mp.current_status
+        END AS status_name,
+    string_agg(task_order_filter->>'display_name', ',') as task_order_name,
+    mp.contractor,
+    mp.purchase_order_number,
+    json_agg(mpf.feature) as project_feature,
+    mt.type_name
+   FROM public.moped_project mp
+   left join moped_team_members mtm on (mp.project_id = mtm.project_id)
```
frankhereford, line 53: I might kick this line in a couple of spaces and use the SQL convention of capital letters for keywords, but there's nothing wrong with this at all from a technical perspective.

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,89 @@
+drop view project_list_view;
+
+CREATE VIEW public.project_list_view AS
+with moped_team_members as (
```
frankhereford, line 4: The CTE is doing just what it needs to do:

* Create a list of all personnel, per project, with `mpp.status_id` of `1`
* Collect their full names and various roles into an aggregated, comma delimited string of people, for each project
* And make this collection available to the main query below.

Would you consider changing the name of the ephemeral, in memory, table to perhaps `project_person_list_lookup` or something like that? I would love to do something like `project__person_list__lookup` or `project_personList_lookup` -- trying to find a way to show it's `projects` and `person_lists` that are the two things being correlated, not `projects`, `persons`, and `lists`. My suggested name could be improved, i'm sure.

My idea is make it as clear as possible what that construction is doing when reading the joins down below. These never hit the disk nor are available outside the context of this query, so I think dropping the `moped_` prefix is actually helpful and clear here, as well. It differentiates them from the tables on disk.

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
frankhereford, line 53: I might kick this line in a couple of spaces and use the SQL convention of capital letters for keywords, but there's nothing wrong with this at all from a technical perspective.

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,89 @@
...
+  select
+    mpp.project_id,
+    string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ',') AS project_team_members
+   FROM 
+    public.moped_proj_personnel mpp
+        LEFT JOIN public.moped_users mu ON ((mpp.user_id = mu.user_id))
+        LEFT JOIN public.moped_project_roles mpr ON ((mpp.role_id = mpr.project_role_id))
```
frankhereford, line 11: Can we be assured that the records in `moped_users` and `moped_project_roles` are going to be available? If so, you may consider using a vanilla `JOIN`.

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,89 @@
...
+    string_agg(task_order_filter->>'display_name', ',') as task_order_name,
+    mp.contractor,
+    mp.purchase_order_number,
+    json_agg(mpf.feature) as project_feature,
+    mt.type_name
+   FROM public.moped_project mp
+   left join moped_team_members mtm on (mp.project_id = mtm.project_id)
+     LEFT JOIN public.moped_entity me ON ((me.entity_id = mp.project_sponsor))
+     LEFT JOIN public.moped_proj_partners mpp2 ON (((mp.project_id = mpp2.project_id) AND (mpp2.status_id = 1)))
+     LEFT JOIN public.moped_entity me2 ON ((mpp2.entity_id = me2.entity_id))
```
frankhereford, line 56: There are extra `()`s in use here; I think the ones that are not actually changing the order of operations can come out? Just a readability thing; the query compiler will drop them if they stay.

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
```diff
@@ -0,0 +1,89 @@
+drop view project_list_view;
+
+CREATE VIEW public.project_list_view AS
+with moped_team_members as (
+  select
+    mpp.project_id,
+    string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ',') AS project_team_members
+   FROM 
+    public.moped_proj_personnel mpp
```
frankhereford, line 9: The pattern of using very abbreviated table aliases is a common one in SQL constructions, so I may be an outlier here, but I've always found it easier to debug my statements when I return to them later if I alias long table names down to a real word, like:

`moped_projects` → `projects`
`moped_proj_personnel` → `personnel`
`moped_users` → `users`
`moped_project_roles` → `roles`

This is a personal style, so please don't adapt to it unless you find it useful in your own workflows!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -90,10 +90,8 @@ export const ProjectsListViewFiltersConf = {
         "string_contains_case_insensitive",
         "string_begins_with_case_insensitive",
         "string_ends_with_case_insensitive",
-        // Because of the way the SQL view concatanates the team members and roles
-        // " :" is always present even if there are no team members present
```
frankhereford, line 94: You solved this! Thank you!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -299,6 +300,28 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
       field: "updated_at",
       render: entry => new Date(entry.updated_at).toLocaleDateString("en-US"),
     },
+    {
+      title: "Signal IDs",
+      field: "project_feature",
+      render: entry => {
```
frankhereford, line 306: Being able to supply a custom rendering function here is top-shelf. This is a great use of it.

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -1,5 +1,5 @@
 export const filterProjectTeamMembers = value => {
-  if (value === " :") {
+  if (!value || value === " :") {
```
frankhereford, line 2: Can we get rid of the `value === " :"` check here since you solved it with your query above?

frankhereford (approved):
This is great. I have some very minor comments on style and readability on the SQL -- forgive me on that and disregard anything you don't agree with. By my read, your view is doing just what it needs to do and I think you're establishing the powerful pattern really well of doing complex RDBMS things in the database and exposing those results out via the graphql endpoint.

I've tested this on my local machine, and the final results look great. This is fantastic work!

moped-database/migrations/1650301717000_alter_project_list_view/up.sql
chiaberry: I think we can be assured the records will be available. ill update to using join, im going to read up on why using `join` is preferable to `left join` in this instance.

frankhereford: I think a `left join` and a `join` here are going to do the exact same operation. It's kind of like extra parentheses -- if you don't need them, it can be argued that it's clearer to not include them. One thing for sure though, this is a very minor issue and since it's working, it's a good solution!

mateoclarke (approved):
🚢 💨 ✨

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -177,20 +177,26 @@ export const ProjectsListViewQueryConf = {
       filter: value => new Date(value).toLocaleDateString("en-US"),
       type: "date_iso",
     },
-    // the following column configurations only work with material table and do not have the 
+    // the following column configurations only work with material table and do not have the
     // necessary details to work with Grid Table
     task_order: {
-      type: "string",
+      type: "Array",
```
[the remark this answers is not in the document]
chiaberry: good question -- I capitalized it because I was thinking that's the official type declaration, but `String` is also capitalized officially. we don't use these types in the config (the types are used in GridTable), but I should keep the formatting consistent.

amenity (approved):
Works great, thank you @chiaberry !

PR #641, opened 2022-05-02
Add funding source column to project list view

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9006

Adds column as well as option to filter on column. 

I also removed `project_importance`, `project_length` and `project_priority` from the SQL View since they are not needed for the Projects List View page.

## Testing
**URL to test:** 

moped test or test locally. 

Note -- ~~I branched this branch (https://github.com/cityofaustin/atd-moped/pull/642) off of this branch, so if you deploy PR #642 to test, you will be able to test both of these enhancements in one deployment~~ You can now deploy the branch from https://github.com/cityofaustin/atd-moped/pull/645 if you want to test the funding source column, the status updat column and the project type bug fix in one build.  cc: @amenity 

**Steps to test:**

Add multiple funding sources to a project https://moped-test.austinmobility.io/moped/projects/174?tab=funding

Go back to https://moped-test.austinmobility.io/moped/projects/, toggle the column on and see the sources. 

---
#### Ship list
- [x] Code reviewed 


moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
```diff
@@ -0,0 +1,96 @@
...
+),
+ funding_sources_lookup AS (
+  SELECT
+    mpf.project_id,
+    string_agg(mfs.funding_source_name, ', ') AS funding_source_name
+  FROM
+    moped_proj_funding mpf
+    LEFT JOIN moped_fund_sources mfs ON mpf.funding_source_id = mfs.funding_source_id
+    GROUP BY mpf.project_id
+)
```
chiaberry, line 23: I added another common table expression because a normal left join was behaving similar to the team members, where if a record had more than one funding source it would make the project_features repeat.

moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
```diff
@@ -0,0 +1,96 @@
+drop view project_list_view;
+
+CREATE VIEW public.project_list_view AS
```
frankhereford, line 3: ## TLDR
This query is very fast.

## Considerations
Related to our quick conversation about SQL this morning, I am going to make a point to start to be more aware of execution times for queries that are being embedded as views (just like we would for a graphql query.)

One of the very first things we need to consider in the pursuit of a snappy, responsive app is a snappy, responsive database as it is our data source for like almost everything. To quantify a query's performance, I want to propose that we can use execution time on the read replica as a way to make apples to apples comparisons and track query performance over time.

I say this because:

1) both the RR and the production database operate on the exact same data at any given time -- this is the nature of the read replica, and
2) the RR is a lessor powered instance that the production database. If it's fast to compile and execute on the RR, it should be at least that fast on the real deal.

Caveat: This tendency will break down if, say, the production database were under continual load above what it is capable of keeping up with, but we are _very, very_ far from that being the case right now. Historical load graphs are available in the RDS section of AWS which we can use to ensure that remains the case.

Anyway, all that said, when I ask postgres to `EXPLAIN ANALYZE` this query for me, it shows that the database is able to find and prepare the results to this query in

Screen Shot 2022-05-05 at 3 10 11 PM

That's super-fast in my opinion.⚡🔥 That's down on the order of local network latency let alone the internet, so that's really flying.

moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
```diff
@@ -0,0 +1,96 @@
...
+    COALESCE(mp.status_id, 0) AS status_id,
+    mp.updated_at,
+    ppll.project_team_members,
+    me.entity_name AS project_sponsor,
+    string_agg(DISTINCT me2.entity_name, ', ') AS project_partner,
+        CASE
+            WHEN (mp.status_id = 0 OR mp.status_id IS NULL) THEN NULL::text
+            WHEN (mp.status_id = 1) THEN mp.current_phase
+            ELSE mp.current_status
+        END AS status_name,
```
frankhereford, line 53: I think you may want to outdent this back to the same column that the fields in the `SELECT` clause lay in. The `WHEN`s and `ELSE` should be in from the `CASE` but the `CASE` is a field-like object.

moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
frankhereford: I think you've picked the best way to roll up the 1:many relationship here for displaying in a table with another CTE. 

I don't think that having multiple CTEs is an indicator of code smell, _unless_, the number of records returned by one is either VERY large. If that is the case, the condition that is evaluated when that CTE is joined into the query in the main `SELECT` statement has to be executed once for every row out of the CTE's set and they, by definition, are in memory without the benefits of an index. This is a common pitfall.

I think what I'm trying to say with too many words is:

Query smell comes from slowness, not length. As long as it's fast, it's probably what you want. If it's a complex query, use comments liberally and ask the database for its execution plan during development to help understand what is fast and what is slow in the query.

frankhereford (approved):
🚢 👍 

Thanks for including a comment in your query that explains what the things are you did in the PR. It's really helpful, as our migration system doesn't let us easily use the `--` comment syntax in the SQL itself and have them accumulate in the files in main. 🙏 It helps me know what to laser in on when reading the SQL, but also, I guess there is value in taking a look at it holistically often too - since it's the datasource for one of most important pages.

I've reviewed the code with a focus on the SQL. I have tested it locally, and it works as advertised. I'd like to highlight one suggestion regarding repeating values below, @amenity, but I don't think that it's a show-stopper in any way.

moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
```diff
@@ -0,0 +1,96 @@
...
+    public.moped_proj_personnel mpp
+        JOIN public.moped_users mu ON mpp.user_id = mu.user_id
+        JOIN public.moped_project_roles mpr ON mpp.role_id = mpr.project_role_id
+      WHERE (mpp.status_id = 1)
+      GROUP BY mpp.project_id
+),
+ funding_sources_lookup AS (
+  SELECT
+    mpf.project_id,
+    string_agg(mfs.funding_source_name, ', ') AS funding_source_name
```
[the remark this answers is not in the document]
chiaberry: Thank you for testing that! I like the idea of using DISTINCT. I don't know if users will be including the same funding source on multiple lines, but just in case I think we should write code to anticipate that.

mateoclarke (approved):
🚢 💨 ✨ 

I appreciate all the suggestions and thoughtful testing @frankhereford. I need to level up my SQL.

patrickm02L (changes requested):
Funding Source remain in projects column after deleted from funding tab. See video at 1:30

**Note:** The funding sources are not duplicating they're just there from an earlier test I did. I chose the exact same sources by coincidence. 

https://user-images.githubusercontent.com/5566727/168609493-48be5fcb-10e9-40f4-94fe-9d0da7282977.mp4

moped-database/migrations/1651001224000_alter_project_list_view_funding/up.sql
```diff
@@ -0,0 +1,97 @@
...
+  FROM
+    moped_proj_funding mpf
+    LEFT JOIN moped_fund_sources mfs ON mpf.funding_source_id = mfs.funding_source_id
+    WHERE mpf.funding_status_id = 1
```
chiaberry, line 22: added this line to ignore funding sources that have been soft-deleted.

mateoclarke:
Added "Add multiple funding sources to a project, Go back to the Projects page, make sure to toggle the "Funding source" column is on and verify you see the comma-separated sources." to QA Script.

PR #643, opened 2022-05-03
Use GridTable sort function in Material Table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9038

Takes the same sort functions that we were using in Grid Table and overrides the Material Table sorting function to use ours. 

While testing this PR, I noticed that if you toggle columns and then open the advanced filter dropdown, your column configuration gets reset. I lifted the advancedSearchAnchor's state up one level so we can listen for any changes to it and store the column configuration then as well. 

## Testing
**URL to test:** 

https://9038-custom-sorting--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**

Click the arrow next the column header to sort by column. initially sorts asc, then if clicked again will sort desc. 

Signal IDs cannot be sorted as is, because of the way the data is in the material table. If you try sorting on the signal dashboard, signal ids could not be sorted there either. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableFilters.js
+++ b/moped-editor/src/components/GridTable/GridTableFilters.js
@@ -397,7 +397,7 @@ const GridTableFilters = ({
 
   return (
     <Grid>
-      <Grid container justify={"flex-end"}>
+      <Grid container justifyContent={"flex-end"}>
         <IconButton
           onClick={handleAdvancedSearchClose}
           className={classes.closeButton}
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -137,6 +137,10 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
     column: "",
   });
 
+  // anchor element for advanced search popper in GridTableSearch to "attach" to
+  // State is handled here so we can listen for changes in a useeffect in this component
+  const [advancedSearchAnchor, setAdvancedSearchAnchor] = useState(null);
+
   // create URLSearchParams from url
   const filterQuery = new URLSearchParams(useLocation().search);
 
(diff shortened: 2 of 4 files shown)


moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -397,7 +397,7 @@ const GridTableFilters = ({
 
   return (
     <Grid>
-      <Grid container justify={"flex-end"}>
+      <Grid container justifyContent={"flex-end"}>
```
chiaberry, line 400: got an error message that justify has been deprecated

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -336,6 +361,7 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
       title: "Signal IDs",
       field: "project_feature",
       hidden: hiddenColumns["project_feature"],
+      sorting: false,
```
frankhereford, line 364: Thanks for disabling the ordering here. It makes sense to me that because we have one-or-more signal ids that can come back in this column, ordering isn't really a sensical operation. 👍

frankhereford (approved):
🚢 👍 

I've read over the code and tested on the PR deployment. Thank you for this work!

mateoclarke (approved):
This code looked tricky but I'm really glad the sorting is working well with the advanced filters.

It may be out of scope for this PR but how tricky would it be to get the arrow icon to flip up with asc and down with desc?

🚢 💨 ✨

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -478,6 +545,13 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
                         />
                       );
                     },
+                    Header: props => (
+                      <MTableHeader
+                        {...props}
+                        onOrderChange={handleTableHeaderClick}
+                        orderDirection={sort.order}
```
chiaberry, line 552: turns out all I needed to do to inform arrow direction was pass along this sort order, and material table does the column id checking etc. 

It does change all of the arrows' directions, but that is the same behavior as the normal sort icon in default material table.

chiaberry:
> This code looked tricky but I'm really glad the sorting is working well with the advanced filters.
> 
> It may be out of scope for this PR but how tricky would it be to get the arrow icon to flip up with asc and down with desc?
> 
> 🚢 💨 ✨

thanks mateo! I figured out what prop was needed to get the arrow icon to flip. left a comment on the part of the code I updated.

amenity (approved):
Works beautifully. 👌🏻 ✨

johnclary:
Slick!!

PR #645, opened 2022-05-04
Fix duplicating project type bug

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9070

## Testing
**URL to test:** 

needs to test on test. I have again branched from another PR, so you could in theory only test on this branch and see the status update and funding source column.

**Steps to test:**

Add and remove multiple project types to a project. Go back to list view and the project should not be duplicated. Basically you dont want to see this:
Screen Shot 2022-05-04 at 4 39 08 PM

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1651692507000_alter_project_list_view_type_bug/up.sql
```diff
@@ -0,0 +1,111 @@
...
+),
+project_type_lookup as (
+  SELECT
+    mpt.project_id,
+    string_agg(mt.type_name, ', ') as type_name
+  FROM
+     public.moped_project_types mpt
+     LEFT JOIN public.moped_types mt ON mpt.project_type_id = mt.type_id AND mpt.status_id = 1
+     GROUP BY mpt.project_id
+)
```
chiaberry, line 32: Here is the fix for the bug. Its a one to many relationship, I was previously treating it as a one to one. Also if someone removes a project type, its a soft delete, so we need to only get the ones that are active. 

I don't think we need to maintain it as a soft delete, I'm open to discussing if it should stay as a soft delete or if we should change it.

moped-database/migrations/1651692507000_alter_project_list_view_type_bug/down.sql
```diff
@@ -0,0 +1,103 @@
...
+     LEFT JOIN project_person_list_lookup ppll ON mp.project_id = ppll.project_id
+     LEFT JOIN funding_sources_lookup fsl ON fsl.project_id = mp.project_id
+     LEFT JOIN public.moped_entity me ON me.entity_id = mp.project_sponsor
+     LEFT JOIN public.moped_proj_partners mpp2 ON (mp.project_id = mpp2.project_id AND mpp2.status_id = 1)
+     LEFT JOIN public.moped_entity me2 ON mpp2.entity_id = me2.entity_id
+     LEFT JOIN jsonb_array_elements(mp.task_order) as task_order_filter ON true
+     LEFT JOIN moped_proj_components mpc ON mpc.project_id = mp.project_id and mpc.status_id = 1
+     LEFT JOIN moped_proj_features mpf ON mpc.project_component_id = mpf.project_component_id
+     LEFT JOIN public.moped_project_types mpt ON mpt.project_id = mp.project_id
+     LEFT JOIN public.moped_types mt ON mpt.project_type_id = mt.type_id
```
chiaberry, line 71: previous way of getting the moped type names.

chiaberry: (lines 70 - 71)

moped-database/migrations/1651692507000_alter_project_list_view_type_bug/up.sql
chiaberry: (lines 24-32)

frankhereford: By my read, you're building up a CTE with two columns (project ID and list of types) and using that as a lookup table to get that list associated to the projects being returned as rows in the main query. 👍 I think using a CTE here is both clear in terms of readability and fast as it avoids any unnecessary looping in the query plan. 

Testing this as-written, without moving the sub-select up into the select clause as suggested in the review to #642, still yields a query that executes in ~30 ms, which is great.

mateoclarke: I don't really have an opinion on if this specific case should be a soft delete. It seems like we adopted that `status_id=1` pattern all over the place. And we all agree now that a better approach would be a boolean value column specifically for soft deletes. But the more we add in conditions like this, buried in a migration file, the harder it may be for us to refactor away from this pattern in the future. No solutions, just thoughts.

mateoclarke (approved):
🚢 💨 ✨

chiaberry:
@mateoclarke this is the PR that includes all the changes from the previous db PRs. Testing this one will cover the changes in PR #642 and #641

patrickm02L (approved):
Works perfectly! Well done! 

Uploading Fix duplicating project type bug by chiaberry · Pull Request #645 · cityofaustin_atd-moped - 16 May 2022.mp4…

mateoclarke:
Added "Add multiple project types to a project. Verify the project type displays correctly within the column." to the QA script.

PR #656, opened 2022-05-24
9249 proj components soft del

mddilley:
## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
+++ b/moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
@@ -1,1 +1,7 @@
 ALTER TABLE "public"."moped_proj_components_subcomponents" ADD COLUMN "is_deleted" boolean NOT NULL DEFAULT false;
+
+COMMENT ON COLUMN moped_proj_components_subcomponents.is_deleted IS 'Indicates soft deletion';
+
+UPDATE moped_proj_components_subcomponents
+SET is_deleted = TRUE
+WHERE status_id = 0;
--- a/moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
+++ b/moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "public"."moped_proj_components_subcomponents" ADD COLUMN "is_deleted" boolean NOT NULL DEFAULT false;
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -2,199 +2,206 @@
     schema: public
     name: moped_activity_log
   object_relationships:
-  - name: moped_user
-    using:
-      manual_configuration:
-        remote_table:
-          schema: public
-          name: moped_users
-        column_mapping:
-          updated_by: cognito_user_id
+    - name: moped_user
+      using:
+        manual_configuration:
+          remote_table:
+            schema: public
+            name: moped_users
+          column_mapping:
+            updated_by: cognito_user_id
   select_permissions:
-  - role: moped-admin
-    permission:
-      columns:
-      - activity_id
-      - created_at
-      - description
-      - operation_type
-      - record_data
-      - record_id
-      - record_project_id
-      - record_type
-      - updated_by
-      filter: {}
-      allow_aggregations: true
-  - role: moped-editor
-    permission:
-      columns:
-      - activity_id
-      - created_at
-      - description
-      - operation_type
-      - record_data
-      - record_id
-      - record_project_id
-      - record_type
-      - updated_by
-      filter: {}
-      allow_aggregations: true
-  - role: moped-viewer
-    permission:
-      columns:
-      - activity_id
-      - created_at
-      - description
-      - operation_type
-      - record_data
-      - record_id
-      - record_project_id
-      - record_type
-      - updated_by
-      filter: {}
-      allow_aggregations: true
+    - role: moped-admin
+      permission:
+        columns:
+          - activity_id
+          - created_at
+          - description
+          - operation_type
+          - record_data
+          - record_id
+          - record_project_id
+          - record_type
+          - updated_by
+        filter: {}
+        allow_aggregations: true
+    - role: moped-editor
+      permission:
+        columns:
+          - activity_id
+          - created_at
+          - description
+          - operation_type
+          - record_data
+          - record_id
+          - record_project_id
+          - record_type
+          - updated_by
+        filter: {}
+        allow_aggregations: true
+    - role: moped-viewer
+      permission:
+        columns:
+          - activity_id
+          - created_at
+          - description
+          - operation_type
+          - record_data
+          - record_id
+          - record_project_id
+          - record_type
+          - updated_by
+        filter: {}
+        allow_aggregations: true
 - table:
     schema: public
     name: moped_city_fiscal_years
   insert_permissions:
-  - role: moped-admin
-    permission:
-      check: {}
-      columns:
-      - fiscal_year_value
-      - fiscal_year_start_date
-      - fiscal_year_end_date
-      - active_fy
-      backend_only: false
-  - role: moped-editor
-    permission:
-      check: {}
-      columns:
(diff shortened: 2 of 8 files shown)


moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -414,7 +414,7 @@ const ProjectComponentsMapEdit = ({
       */
       name: "",
       project_id: Number.parseInt(projectId),
-      status_id: 1,
+      is_deleted: false,
```
mddilley, line 417: Not sure if we should keep this as part of the line since `is_deleted` is false by default in the database so having this in the payload of the mutation just applies the same value.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -291,7 +291,6 @@ const ProjectComponentsMapEdit = ({
     const projectFeaturesToCreateOrUpdate = editFeatureCollection.features.map(
       feature => {
         let projectFeature = {};
-        projectFeature.status_id = 1;
```
mddilley, line 294: We no longer need to set status_id to 1 (not soft deleted) here since the default value of `is_deleted` is false on record creation.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -414,7 +413,6 @@ const ProjectComponentsMapEdit = ({
       */
       name: "",
       project_id: Number.parseInt(projectId),
-      status_id: 1,
```
mddilley, line 417: Same here - there is no longer a need initialize `status_id` for components on creation.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
mddilley: I ended up deciding to remove `is_deleted` from the payload here since its default value is already `false`.

johnclary (approved):
thanks @mddilley for finding these needles in the haystack!

it looks like [this use of status ID](https://github.com/cityofaustin/atd-moped/blob/9b8afb5e883f1b37f35c5b48e6b5815411ff045d/moped-editor/src/utils/signalComponentHelpers.js#L219) in `signalComponentHelpers` could also be removed.

have y'all discussed going ahead and dropping the unused `status_id` columns from these tables?

mddilley:
@johnclary Thanks for pointing out the signal components part. That actually helped me catch a couple of things in my other PR that I overlooked. 

We haven't fully discussed dropping the `status_id` that yet, and I'm on the fence about dropping them in this epic since `status_id` is pulling double duty for some tables with soft deletion and statuses so I want to dive in some more and not destroy data. I can definitely be convinced, though, because it makes sense to take care of them while I'm in here. 👍

chiaberry (approved):
tested it locally, checked the database, everything works as stated!

johnclary:
> We haven't fully discussed dropping the status_id yet, and I'm on the fence about dropping them in this epic since status_id is pulling double duty for some tables with soft deletion and statuses so I want to dive in some more and not destroy data. I can definitely be convinced, though, because it makes sense to take care of them while I'm in here. 👍

👍 makes sense to me to leave that outside the scope of this

moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE "public"."moped_proj_components_subcomponents" ADD COLUMN "is_deleted" boolean NOT NULL DEFAULT false;
+
+COMMENT ON COLUMN moped_proj_components_subcomponents.is_deleted IS 'Indicates soft deletion';
+COMMENT ON COLUMN moped_proj_components_subcomponents.status_id IS 'Deprecated soft deletion column';
+
+-- Update new is_deleted columns to preserve existing soft deletes that used status_id = 0
+UPDATE moped_proj_components_subcomponents
+SET is_deleted = TRUE
+WHERE status_id = 0;
```
frankhereford, line 9: Super not a big deal, but maybe add a EOF line feed to these migrations. I realize it's probably the `graphql-engine` console omitting them, no worries!

frankhereford (changes requested):
Will you take a peek at the comment below re new `moped_proj_components_subcomponents`? Other than that, this is awesome - thank you so much! Let me know if I'm wrong in my comment, and I'll re-review & approve pronto.

moped-database/migrations/1653414071962_alter_table_public_moped_proj_components_subcomponents_add_column_is_deleted/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE "public"."moped_proj_components_subcomponents" ADD COLUMN "is_deleted" boolean NOT NULL DEFAULT false;
```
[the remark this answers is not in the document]
mddilley: Glad that you took a closer look at this! I'm going to test this out more because I may have misinterpreted the comments around this code. 🙏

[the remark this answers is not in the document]
mddilley: Update: Yep, I did. I mistook some logic around checking for new or existing subcomponents by their ids for logic that was checking soft deleted or not soft deleted. Updated the code and will push with the other changes to the migrations. 🙌

mddilley:
@frankhereford I addressed your feedback above in the latest commit and will add some notes about the fix for the subcomponents being soft deleted on creation bug. 🙏

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -338,17 +337,21 @@ const ProjectComponentsMapEdit = ({
...
+          removeAllSubcomponents ||
+          !selectedSubcomponents.find(
+            newSubcomponent =>
+              // Return true if we have found the old subcomponent in this list
+              oldSubcomponent.subcomponent_id ===
+              newSubcomponent.subcomponent_id
+          )
+      )
+      // Mark remove list for deletion
+      .map(subcomponent => ({ ...subcomponent, is_deleted: true }));
```
mddilley, line 354: I refactored this code a bit to add the `is_deleted: true` k/v pair while generating the list of subcomponents that need removal. This way, we don't have to worry about using more logic below to search for where to add the flag since we know which ones need deletion right here.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -360,11 +363,6 @@ const ProjectComponentsMapEdit = ({
     // Generate output, clean up & return
     return (
       [...insertionList, ...removalList] // Mix both insertion and removal list
-        // For each subcomponent, mark for deletion (status_id = 0) if component_subcomponent_id has a number
-        .map(subcomponent => ({
-          ...subcomponent,
-          status_id: isNaN(subcomponent?.component_subcomponent_id) ? 1 : 0,
-        }))
```
mddilley, line 367: This iteration is no longer need since we add `is_deleted: true` when creating the `removalList` array above.

frankhereford (approved):
Thanks for considering my comments and knocking it out of the park!!

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
johnclary: 👍 much easier to grok!

amenity:
Besides that error ☝🏻 , everything works great. :) I also tested with points, and drawn points.

mddilley:
Just like the last one, this last commit contains changes from these steps:
1. Pulled in `main`
2. Resolved any conflicts
3. Incorporated changes to the `project_list_view` from my branch into the version controlled SQL file
4. Updated view up migration in PR to use that version of the `project_list_view` and the down migration to roll back to current view state
5. Updated migrations introduced in this branch with fresh timestamps in sequence by milliseconds (to keep local working)

frankhereford (approved):
Looks great - thanks for working through these and establishing the record tracking on the form of the view.

mateoclarke (commented):
👍

PR #660, opened 2022-05-31
Add `is_deleted` column for project partners and types soft deletes

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9250

_Note: This PR looks larger because it has up and down migrations for a small update to the SQL view that powers the project table._

## Testing
**URL to test:** 
_This PR needs to be tested either locally or on moped-test._

**Steps to test:**

**Project type**
1. Create a new project and then go to the Summary tab of the project to add a few project types from the dropdown and then click the check icon to save
2. When you add types, you should see them listed in the Projects table list under the column named Project Types
3. Remove one of the types and you should see it removed from the project
4. Remove all of the types and you will see the text "None" under Project Types

**Project partners**
1. Repeat the above steps with project partners in the project Summary tab

---
#### Ship list
- [x] Code reviewed 
- [ ] ~Product manager approved~


moped-database/migrations/1654025878154_update_proj_partner_filter/up.sql
```diff
@@ -0,0 +1,97 @@
...
+             JOIN moped_users users ON ((personnel.user_id = users.user_id)))
+             JOIN moped_project_roles roles ON ((personnel.role_id = roles.project_role_id)))
+          WHERE ((1 = 1) AND (roles.project_role_name = 'Designer'::text) AND (personnel.status_id = 1) AND (personnel.project_id = mp.project_id))
+          GROUP BY personnel.project_id) AS project_designer
+   FROM (((((((moped_project mp
+     LEFT JOIN project_person_list_lookup ppll ON ((mp.project_id = ppll.project_id)))
+     LEFT JOIN funding_sources_lookup fsl ON ((fsl.project_id = mp.project_id)))
+     LEFT JOIN project_type_lookup ptl ON ((ptl.project_id = mp.project_id)))
+     LEFT JOIN moped_entity me ON ((me.entity_id = mp.project_sponsor)))
+     LEFT JOIN moped_proj_partners mpp2 ON (((mp.project_id = mpp2.project_id) AND (mpp2.is_deleted = false))))
```
mddilley, line 94: This is the only update to the `project_list_view` (checking for partner records with `is_deleted = false` instead of `status_id = 1`)

moped-database/migrations/1654025878154_update_proj_partner_filter/up.sql
```diff
@@ -0,0 +1,97 @@
...
+            string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
+           FROM (moped_proj_funding mpf_1
+             LEFT JOIN moped_fund_sources mfs ON ((mpf_1.funding_source_id = mfs.funding_source_id)))
+          WHERE (mpf_1.funding_status_id = 1)
+          GROUP BY mpf_1.project_id
+        ), project_type_lookup AS (
+         SELECT mpt.project_id,
+            string_agg(mt.type_name, ', '::text) AS type_name
+           FROM (moped_project_types mpt
+             LEFT JOIN moped_types mt ON (((mpt.project_type_id = mt.type_id) AND (mpt.is_deleted = false))))
```
mddilley, line 21: Update to show for project type records with `is_deleted = false` instead of `status_id = 1`

moped-database/migrations/1653505845044_alter_table_public_moped_proj_partners_add_column_is_deleted/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE "public"."moped_proj_partners" ADD COLUMN "is_deleted" boolean NOT NULL DEFAULT false;
+
+COMMENT ON COLUMN moped_proj_partners.is_deleted IS 'Indicates soft deletion';
+COMMENT ON COLUMN moped_proj_partners.status_id IS 'Deprecated soft deletion column';
+
+-- Update new is_deleted columns to preserve existing soft deletes that used status_id = 0
+UPDATE moped_proj_partners
+SET is_deleted = TRUE
+WHERE status_id = 0;
```
frankhereford, line 9: EOF line feed, but no big deal

moped-database/migrations/1654025878154_update_proj_partner_filter/up.sql
frankhereford: Thanks for the explanation. Did I say how thankful I am for your push to better document the evolution of this view because without your comment, this diff wasn't going to cut it! 🙏 💯

frankhereford (approved):
## PR Review

@mddilley, this looks perfect. I tested locally and reviewed the code - this is great. Thank you for this awesome work.

## New, but related, issue suggestion

@amenity, while testing this PR, I observed that we have a "Partner" option called "None." It's used in a way that makes "None" not the absense of a partner, but rather, an actual, real partner, in the same usage that we might call "Active Transportation" a partner.

I would like to offer the idea of removing the "None" partner option, and displaying "None" when no partners have been designated. The intent of this suggestion is to mirror the way that "Project Types" works. 

I think this possible partner state best summarizes this suggestion:

Screen Shot 2022-06-03 at 11 21 05 AM

Thanks - let me know if my description doesn't make sense 🙃 🙏

chiaberry (approved):

johnclary (approved):
happy to see another batch of these soft deletes. thanks @mddilley! 🚢

johnclary:
@frankhereford good call - i created https://github.com/cityofaustin/atd-data-tech/issues/9377.

moped-database/migrations/1655490225419_update_proj_partner_filter/up.sql
```diff
@@ -0,0 +1,149 @@
+-- public.project_list_view source
```
frankhereford, line 1: No problem with this here, especially in a migration that isn't intended to be human readable, but just out of curiosity - what does this comment indicate?

frankhereford (approved):
This looks good - I have gone over the most recent changes, and I am glad you're pushing these through.

moped-database/migrations/1655490225419_update_proj_partner_filter/up.sql
mddilley: It is a Hasura generated thing so it probably can go! Just an oversight so I'm glad that you pointed it out.

mddilley: Actually, I just noticed it is in the version controlled file. I'll get that out of there so we don't keep pushing it into migrations. Thanks!

mateoclarke (commented):
👍

PR #662, opened 2022-06-02
Add `is_deleted` column for project and project file soft deletes

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9300

This PR focuses on renaming `is_retired` in the `moped_project` and `moped_project_files` tables along with making that change in:
- Queries for projects and files
- Mutations for project and file soft deletes
- Seed data that inserts project records

I also added a new directory for SQL view version control and added a file for the `project_list_view` to it.

## Testing
**URL to test:** 
_This PR needs to be tested either locally or on moped-test._ Uploading a file from your local instance may not work as intended due to there not being the correlating lambda functions available. This should work however on the test instance.

**Steps to test:**

**Project**
1. Create a new project and go to the project list view to make sure that you can see it listed
2. Go back to the new project's summary page, delete it, and you will be redirected to the list view where it will no longer be present

**Project files**
1. Go to the Files tab and add a file
2. From the file list, click on the trash can icon to soft delete the file and you will see the file removed from the list

---
#### Ship list
- [x] Code reviewed 


moped-database/database_views/project_list_view.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.current_status,
+    mp.timeline_id,
+    mp.current_phase,
+    mp.end_date,
+    mp.project_length,
+    mp.fiscal_year,
+    mp.capitally_funded,
+    mp.date_added,
+    mp.added_by,
+    mp.is_deleted,
```
mddilley, line 46: Renamed `mp.is_retired` to `mp.is_deleted` here

moped-database/database_views/project_list_view.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.project_order, 
+    mp.current_status, 
+    mp.timeline_id, 
+    mp.current_phase, 
+    mp.end_date, 
+    mp.fiscal_year, 
+    mp.capitally_funded, 
+    mp.date_added, 
+    mp.added_by, 
+    mp.is_deleted,
```
mddilley, line 140: Same here - renamed `mp.is_retired` to `mp.is_deleted`

moped-database/migrations/1654057324421_alter_table_public_moped_project_files_alter_column_is_retired/up.sql
```diff
@@ -0,0 +1,2 @@
+COMMENT ON COLUMN "public"."moped_project_files"."is_retired" IS E'Indicates soft deletion';
```
frankhereford, line 1: Your use of the comment embedded in the database schema here is so good. I support this pattern, and I think we should keep doing it whenever possible. ✅

moped-database/migrations/1654124226247_proj_list_view_project_soft_del/up.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.project_order, 
+    mp.current_status, 
+    mp.timeline_id, 
+    mp.current_phase, 
+    mp.end_date, 
+    mp.fiscal_year, 
+    mp.capitally_funded, 
+    mp.date_added, 
+    mp.added_by, 
+    mp.is_deleted,
```
frankhereford, line 140: I love how this PR identifies a problem that you just wouldn't know that the line above on 140 was changed since every migration is its own file .. **AND** then goes on to also include a solution to the problem. 💯 💯

moped-editor/src/queries/signals.js
```diff
@@ -6,7 +6,7 @@ export const SIGNAL_PROJECTS_QUERY = gql`
   query SignalProjectsQuery {
     moped_project(
       where: {
-        is_retired: { _neq: true }
+        is_deleted: { _eq: false }
```
frankhereford, line 9: Interesting how you flipped the operator and the operand for a net non-change, but I think this is more readable. Great catch.

frankhereford (approved):
Broadly, I want to support the work here to better version control around the source of the view, especially in as much as it allows us to better maintain commented SQL code. 🧠

Lasering in on one point, I'd like to also comment that I think we should be clear that the migrations are and should remain the sole, authoritative definition of the view, and that the SQL files recorded in this new directory are never intended to be executed in any sort of production or automated workflow. Rather, they fill in a gap in terms of documentation and ease of reference as we go back through file versions.

The result of this is that we have to agree as a team to keep them up to date, which I strongly support we do.  I think we should request changes on any PR that fails to keep this (and any future view) up to date with the state of the database itself. 

Big +1 to @mateoclarke's idea about a README that lays out what we're doing and how to do it and even why -- future developers will know _why_ we do something, not just procedure by wrote.

🔥 🧨 🚢 

This is awesome, thank you @mddilley!

moped-database/migrations/1654057324421_alter_table_public_moped_project_files_alter_column_is_retired/down.sql
```diff
@@ -0,0 +1,2 @@
+COMMENT ON COLUMN "public"."moped_project_files"."is_retired" IS E'';
```
chiaberry, line 1: TIL about E

mddilley: Same! Hasura auto-generated this so I had to look it up myself.

chiaberry (approved):

mateoclarke (approved):
🚢 💨 ✨ 

delete-anime-press

johnclary (approved):
🚢 🚢 🚢 🚢

mddilley:
Just pushed the README updates for the DB views along with renaming the `/database_views/` directory to `/views/`. These do not need additional review and we can continue to enrich docs around this process as it evolves.

amenity (approved):
👍🏻 🗑️  🙏🏻 🚢

moped-database/database_views/project_list_view.sql
```diff
@@ -0,0 +1,149 @@
+-- public.project_list_view source
```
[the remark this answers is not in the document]
mddilley: Yes, I've now updated this folder so no more repetition there. 🙌

PR #666, opened 2022-06-06
Update to graphql-engine 2.7.0+

frankhereford:
This PR began as a lighthearted conversation about what tech debt we might be able to address with this, um, conspicuous PR number. 

## TLDR

This PR represents multiple things:

1) Changing our local development environments to use the latest `graphql-engine` from Hasura
1) Committing to move forward to roll out the same version in `test`, `staging`, and `production` as soon as is reasonably prudent & possible

⚠️ Please withhold your approval until you are comfortable with both of these goals.

## Approval process

Because of the reaching effects of such a change, I intend to seek more approvals than normal, with a goal of getting sign-offs from the bulk of the development team. 

## Running `graphql-engine` 2.7+ locally prior to this being merged

Developers can use the latest `graphql-engine` during other work if they want. This can be accomplished by making [this line](https://github.com/cityofaustin/atd-moped/blob/main/moped-database/docker-compose.x86_64.yml#L20) look like [this line](https://github.com/cityofaustin/atd-moped/blob/the-devil-is-in-the-details/moped-database/docker-compose.yml#L6) and restarting the DB cluster. 

## General notes

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/9385. 

## Special testing considerations

Please read, at the very least, each of the "breaking changes" enumerated by Hasura [here](https://hasura.io/docs/latest/graphql/core/guides/upgrade-hasura-v2/#hasura-v2-behaviour-changes). Carefully consider each one in the scope of the entire Moped stack, and become comfortable that each either:

1) Does not have potential to be breaking to our application stack, or
1) Does have potential to be breaking to our stack, but that the change is found to be non-breaking in our particular case.

Additional consideration of the behavior and breaking changes in the subsequent minor releases to `graphql-engine` would be helpful as well.

## Testing
Testing on `test` should work as normal, rolling out `2.7.0`, and this is important to do. 

Also equally important, testing on our local development environments, is needed. Please: 

1) Spin up the DB from this branch.
1) Spin up the app and try any which way you can to break it. Testing anything and everything is fair game.
1) Use it while you do your normal development process as you're able.

### Particular code to consider and test

- [ ] Our `GQLAbstract` library takes a configuration which it uses to build a graphql query. I believe we have two distinct configurations which we feed into the library, and they can be found [here](https://github.com/cityofaustin/atd-moped/blob/the-devil-is-in-the-details/moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js) and [here](https://github.com/cityofaustin/atd-moped/blob/the-devil-is-in-the-details/moped-editor/src/views/staff/StaffListViewQueryConf.js). Do you agree that these should work and do work?

This PR contains comments in the code that helped me answer the question above. I intend to take those comments back out of this PR prior to merging. They are only helpful during this PR evaluation, I believe.

---
#### Ship list
- [x] Code reviewed 
- [x] Application _passes_ the [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#)
- [ ] Comfort with activity log functionality


mateoclarke:
bobsburgers-evil

.github/workflows/atd_moped_build_stage.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           architecture: "x64"
       - uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: "16"
```
frankhereford, line 24: Linting

moped-api/README.md
```diff
@@ -27,15 +29,18 @@ requirements/
 
 Install the requirements in your machine, run these in order:
 
-1.First create a virutal environment in the API root folder:
+1.First create a virtual environment in the API root folder:
```
frankhereford, line 32: This was the only change I intended to commit from this file, the rest of from my lint-on-save behavior that I have turned on in my code editor.

That said, I think the rest of the changes are valid markdown linting and should stay.

moped-database/docker-compose.yml
```diff
@@ -1,9 +1,26 @@
 # Architecture Independent Docker Stack Components for Moped Development Environment
 
 version: '3.7'
+services:
+    hasura:
+        image: hasura/graphql-engine:latest
```
frankhereford, line 6: Hasura's images of `graphql-engine` are built multi-arch so they run natively on ARM and X86 chips alike. No need for us to manually build an ARM image anymore. 😓

moped-editor/src/libs/GQLAbstract.js
```diff
@@ -448,6 +452,8 @@ class GQLAbstract {
         "}"
     );
 
+    // Previous comment indicated we don't support a second tier ordering directive,
+    // but the following code appears to support it.
```
frankhereford, line 456: Here is my change. It's only a comment, and I would want to take it out of the PR prior to merging, but it's helpful in in the review process. 

Everything else in this file is linting.

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -217,6 +217,7 @@ export const ProjectsListViewQueryConf = {
       type: "string",
     },
   },
+  // This object gets consumed into the GQLAbstract system, and here is the single, un-nested order_by directive. ✅
```
frankhereford, line 220: This comment would go too.

moped-editor/src/views/staff/StaffListViewQueryConf.js
```diff
@@ -121,6 +121,7 @@ export const StaffListViewQueryConf = {
       type: "String",
     },
   },
+  // A single, un-nested order by directive ✅
```
frankhereford, line 124: This comment would also go prior to merge.

frankhereford:
I've updated this PR to include changes to the github workflow that now allow it to be deployed to test, and to have the intended version of `graphql-engine` applied to test. 

I have also done a little editing on the comment above to be more explicit about the part of the application that I think had the potential to be affected by the upgrade. It's my opinion that this will work unchanged however, but I am very eager to get second and third opinions. 

I think this is ready for everyone to review when they are able. Thank you @mateoclarke for suggesting that we integrate this into our workflows when it's possible to move this forward. 🙏

.github/workflows/aws-heroku-helper.sh
```diff
@@ -60,6 +60,12 @@ function build_database() {
   print_header "Cloning Hasura Engine";
   git clone https://github.com/hasura/graphql-engine-heroku;
 
+  echo "Patching graphql-engine-heroku/Dockerfile, before:";
+  tail -n 4 graphql-engine-heroku/Dockerfile;
+  sed -ri "s/DATABASE_URL graphql-engine/DATABASE_URL HASURA_GRAPHQL_DATABASE_URL=\$DATABASE_URL graphql-engine/g" graphql-engine-heroku/Dockerfile;
+  echo "Patching graphql-engine-heroku/Dockerfile, after:";
+  tail -n 4 graphql-engine-heroku/Dockerfile;
```
frankhereford, line 67: Modify the Dockerfile in `https://github.com/hasura/graphql-engine-heroku` to specify which database we'll use by default and to connect to on `graphql-engine` startup.

mateoclarke (approved):
Screen Shot 2022-06-14 at 12 11 26 AM

mario

mateoclarke:
Looks great! Tested in both test and local. We should make sure to pull main into this branch before merging to our staging instance. Since this PR has been open for a week, I don't want us to have any merge conflicts or unintended overlap.

johnclary (approved):
@frankhereford thanks for doing this. i ran locally without issue.

i have reviewed the change log and have not identified anything beyond the ordering issue you picked up on.

it looks like we should ideally also upgrade to config `v3`, described [here](https://hasura.io/docs/latest/graphql/core/migrations/upgrade-v3/#upgrade-steps).

based on my understanding, we could do this as a separate step since config v1 is still supported. if y'all agree, i think we should table that until we upgrade the engine?

frankhereford:
@johnclary wrote:
> @frankhereford should we pin the graphq-engine version to `2.x` ? 

Thanks John for asking about this. I've thought about it, and while I'm not 100% sure I have the correct idea,  here's my take. 

In the file you noted in your comment, that's the `graphql-engine` version which is used in our development environments on our own laptops. I think this is the lowest risk place to run `:latest`, and I'm comfortable with us being as up-to-date as we want. We will slowly and organically test our way into the cutting edge as we develop. Hasura seems pretty good at pushing quality images too. To those who would prefer to stick with a known-working version locally, this also is a suitable configuration for them, as `docker` won't pull an updated version of the image until you `docker pull` instruct it to do so.

That said, there is another place where we can pin `graphql-engine`, and this is a place that I would be much more inclined to do so if others on the team were to desire it. I'm 51% sure it's better to let it float, but that is to say, very unsure what is the best route. That other place it could be pinned is implied in this line: https://github.com/cityofaustin/atd-moped/blob/the-devil-is-in-the-details/.github/workflows/aws-heroku-helper.sh#L61. That `git clone` pulls down a `Dockerfile` that itself depends on `hasura/graphql-engine:latest`, and we could rewrite that `IMAGE` line to pin `graphql-engine` if we wanted. This is a technique which Sergio used in the past to pin us to version 1.3.3. 

One place that I feel is verrrry important to pin the version is the deployments on staging and production. Those are intended to act-like-servers with persistence and reliability, so we should be intentional about what we're running on them.

johnclary:
> We will slowly and organically test our way into the cutting edge as we develop
i thought that might be where you're going with this! i'm all for anything that bakes in upgrades. 

let's discuss in dev sync? two process questions come to mind:
1. how do we make sure other devs are testing our branch on the same version of Hasura version? i can definitely see myself forgetting which version i last pulled
2. we'll still want to be pro-active about bumping the image version on staging/prod - so how do we make sure we say on top of that?

.github/workflows/aws-heroku-helper.sh
```diff
@@ -59,10 +59,18 @@ function build_database() {
 
   print_header "Cloning Hasura Engine";
   git clone https://github.com/hasura/graphql-engine-heroku;
+  cd graphql-engine-heroku;
+  git checkout e947f74cf51ec586670140fc1181394c3d2f3300;
```
frankhereford, line 63: In response to @johnclary's question about pinning `test` to a certain version of `graphql-engine`, this line is added to make sure we're running [exactly this commit](https://github.com/hasura/graphql-engine-heroku/commit/e947f74cf51ec586670140fc1181394c3d2f3300) out of the hasura/graphql-engine-heroku repo. This point-in-time mandates version 2.7.0 of `graphql-engine`.

mddilley (approved):
I tested this locally and tried every UI actions that I could and was not able to break things. Super excited to see this update happening, and I appreciate how thorough your testing and research on this has been.

I searched for any other configs for the GQL abstract, and I couldn't find any other besides the two you listed. I agree that the existing configs should be unaffected by this update.

I do agree with @johnclary that we should [upgrade to config v3](https://hasura.io/docs/latest/graphql/core/migrations/upgrade-v3/) in a separate PR from this. I ran into several errors/warnings about updating to config v3 on Discovery Day when trying to use the latest version Hasura CLI for migrations.

BTW I'm really excited for the about the first bullet point in the [breaking changes](https://hasura.io/docs/latest/graphql/core/guides/upgrade-hasura-v2/#hasura-v2-behaviour-changes) because I'm almost certain that we ran into that exact issue several times while developing VZE (and unintentionally updating every record instead of one record because `where: { id: {_eq: null }`).

tillyw:
read through the code and it all makes sense to me! going to test now

frankhereford:
I spoke with @patrickm02L, and product testing normally only happens on staging, so the failure on the activity log tests is alright in my opinion to let slide, with the caveat that we will be retest this functionality when 2.7 is on staging (prior to being put on production.)

patrickm02L (approved):
🤘😈🤘

PR #667, opened 2022-06-06
Add `is_deleted` column for personnel soft deletes

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9251

## Testing
**URL to test:** 
_This PR needs to be tested either locally or on moped-test._

**Steps to test:**
1. Create a new project and go to the **Team tab**
2. Add:
   - One team member with role of **Inspector**
   - One team member with role of **Designer**
   - Yourself with **any other role** (or JD Maccombs if you are testing locally since we are all JD locally 👤 )
3. Go the the project table and you should see the Inspector and Designer in the **Team Members** column of your new project's row
4. Go to the Dashboard view and you should see your new project under the **My Projects** tab
5. Go back to your project's **Team tab** and remove all members from the project
6. In the project table, the Designer and Inspector should now be removed from the **Team Members** columns of your project's row
7. In the Dashboard view's **My Projects** tab, your new project should no longer be listed

---
#### Ship list
- [x] Code reviewed 


mddilley:
Need to update https://github.com/cityofaustin/atd-moped/blob/df65843ec4954b5040dbbea8ae2f2527f2a0b9da/moped-editor/src/queries/dashboard.js#L27 after pulling `main` into this branch

moped-editor/src/queries/project.js
```diff
@@ -199,45 +199,14 @@ export const UPSERT_PROJECT_PERSONNEL = gql`
...
-        update_columns: [project_id, user_id, role_id, status_id, notes]
+        update_columns: [project_id, user_id, role_id, is_deleted, notes]
       }
     ) {
       affected_rows
     }
   }
 `;
 
-export const UPDATE_PROJECT_PERSONNEL = gql`
```
mddilley, line 210: This query isn't used anywhere in the UI, and any updates we make to personnel records are handle with `UPSERT_PROJECT_PERSONNEL`. There might be a good reason to keep it so let me know, and I'll put it back.

mddilley:
I just pushed one last commit to remove some `status_id` values that I missed. All good to review now. 🙏

moped-editor/src/queries/project.js
johnclary: 🙌

chiaberry (approved):
Thanks for the clear steps to test! Also have the same question as John, do we still need the status_id column?

mddilley:
@johnclary @chiaberry The reason that I've held off on removing the `status_id` fields is that I'm fearful since a couple of them were being used for soft deletes and another purpose. In case of any regressions, we would still have the `status_id` values after this is all merged. Then, after we are clear of regressions, I can make one PR that drops all of them.

Do y'all think this is overly cautious? I'm willing to be convinced that it is since it would be nice to just drop it here.

johnclary:
👍 you're knee-deep in this stuff, so whatever you think is best!

mddilley:
Alright, I'm going to go ahead and merge this and kick dropping the `status_id` down the road 🥾

PR #669, opened 2022-06-07
Fixes wsod on signal projects table

tillyw:
## Associated issues

fixes https://github.com/cityofaustin/atd-data-tech/issues/9403

## Testing
**URL to test:** 

**Steps to test:**
add new signals project and open signals project view

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/newProjectView/NewProjectView.js
+++ b/moped-editor/src/views/projects/newProjectView/NewProjectView.js
@@ -153,6 +153,7 @@ const NewProjectView = () => {
           moped_proj_phases: {
             data: [
               {
+                phase_name: "potential",
                 phase_id: 1,
                 is_current_phase: true,
                 status_id: 1,
--- a/moped-editor/src/queries/signals.js
+++ b/moped-editor/src/queries/signals.js
@@ -41,6 +41,7 @@ export const SIGNAL_PROJECTS_QUERY = gql`
         project_note
       }
       moped_proj_phases {
+        phase_name
         phase_id
         is_current_phase
         phase_start


moped-editor/src/views/projects/newProjectView/NewProjectView.js
```diff
@@ -153,6 +153,7 @@ const NewProjectView = () => {
           moped_proj_phases: {
             data: [
               {
+                phase_name: "potential",
```
chiaberry, line 156: was this creating another bug?

tillyw: yes, that's why i had to redeploy

tillyw: from what I understand, when you create a new project in the new project view without this default value, the default `phase_name` is null and it causes an error.

mateoclarke: a separate bug from https://github.com/cityofaustin/atd-data-tech/issues/9403? Or related?

tillyw: same bug. it didn't work until i added this line

mddilley (commented):
This may be what Chia was talking about in dev sync about another way to do this, but I wanted to make sure that we eventually use `phase_id` here in the signals table since there is now a foreign key relationship. 

That way we don't have a straggler view/query that uses `phase_name` (which is the reason that `phase_name` was taken out of the payload of the `ADD_PROJECT` mutation). I could be wrong and maybe this is part of https://github.com/cityofaustin/atd-data-tech/issues/9350?

mateoclarke:
I'm good with this but I liked Chia suggested, I think we should bookmark this and make sure to address with DB normalization in the future.

chiaberry:
I linked to this PR on the normalization issue https://github.com/cityofaustin/atd-data-tech/issues/9350

amenity (commented):
Fixed! Thanks for the speedy work, @tillyw !

PR #675, opened 2022-06-08
[RC patch] Fix funding source filter in project_list_view to filter sources not soft deleted

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9406

## Testing
**URL to test:** 
This needs to be tested on moped-test. I deployed it just a moment ago so it will be there if no one else needs it in the meantime.

**Steps to test:**
1. Create a new non-signal project
2. Add a funding source with a status that is anything other than **Tentative**
3. Check the Project table view and see that the funding source is visible
4. Try every other funding source status, and you should see it visible in the table
5. Delete the funding source and you should see it removed from the Project table view

---
#### Ship list
- [x] Code reviewed 


mateoclarke (approved):
🚢 💨 ✨

PR #680, opened 2022-06-09
Fixes the migration issue around `project_list_view` and `is_deleted` in the project table

mddilley:
## Associated issues
I'll create a bug issue to document this better, but this fixes a bug where migrations were applied incorrectly which caused staging to fail.

See:
https://austininnovation.slack.com/archives/CNUEPKLB1/p1654806894668499
https://austininnovation.slack.com/archives/CNUEPKLB1/p1654811027646039

## Testing
**URL to test:** 
This can be tested locally or on moped-test, but the true test will be on staging when merged.

**Steps to test:**
1. Try to interact with the backend through the UI
 
---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1654646507541_proj_list_view_fund_src_filter/up.sql
+++ b/moped-database/migrations/1654646507541_proj_list_view_fund_src_filter/up.sql
@@ -14,7 +14,7 @@ CREATE OR REPLACE VIEW "public"."project_list_view" AS
             string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
            FROM (moped_proj_funding mpf_1
              LEFT JOIN moped_fund_sources mfs ON ((mpf_1.funding_source_id = mfs.funding_source_id)))
-          WHERE (mpf_1.funding_status_id = 1)
+          WHERE (mpf_1.funding_status_id != 0)
           GROUP BY mpf_1.project_id
         ), project_type_lookup AS (
          SELECT mpt.project_id,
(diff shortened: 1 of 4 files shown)


moped-database/migrations/1654646507541_proj_list_view_fund_src_filter/up.sql
```diff
@@ -39,7 +39,7 @@ CREATE OR REPLACE VIEW "public"."project_list_view" AS
     mp.capitally_funded,
     mp.date_added,
     mp.added_by,
-    mp.is_deleted,
+    mp.is_retired,
```
mddilley, line 42: I needed to return this migration that was edited back to the state in which it was applied to staging. This migration never applied when merged to `main` so the `project_list_view` did not agree with the permissions metadata in `tables.yaml` that contains `is_deleted` instead of `is_retired`. 

The migration below will make the change so the state of `project_list_view` will agree with the tables permissions metadata.

moped-database/migrations/1654813490291_proj_list_view_proj_soft_del/up.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.current_status,
+    mp.timeline_id,
+    mp.current_phase,
+    mp.end_date,
+    mp.project_length,
+    mp.fiscal_year,
+    mp.capitally_funded,
+    mp.date_added,
+    mp.added_by,
+    mp.is_deleted,
```
mddilley, line 46: `is_retired` to `is_deleted` here

moped-database/migrations/1654813490291_proj_list_view_proj_soft_del/up.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.project_order, 
+    mp.current_status, 
+    mp.timeline_id, 
+    mp.current_phase, 
+    mp.end_date, 
+    mp.fiscal_year, 
+    mp.capitally_funded, 
+    mp.date_added, 
+    mp.added_by, 
+    mp.is_deleted,
```
mddilley, line 140: and `is_retired` to `is_deleted` here

moped-database/migrations/1654813490291_proj_list_view_proj_soft_del/down.sql
```diff
@@ -0,0 +1,149 @@
...
+    mp.current_status,
+    mp.timeline_id,
+    mp.current_phase,
+    mp.end_date,
+    mp.project_length,
+    mp.fiscal_year,
+    mp.capitally_funded,
+    mp.date_added,
+    mp.added_by,
+    mp.is_retired,
```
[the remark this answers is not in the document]
mddilley: I agree with this since this would roll back to the actual current state of the view. 👍

frankhereford (approved):

moped-database/migrations/1654813490291_proj_list_view_proj_soft_del/down.sql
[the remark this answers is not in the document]
johnclary: clever! nice work y'all!

mateoclarke:
🙏

PR #681, opened 2022-06-13
Implement "Project status" edit modal

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8806

## Testing
**URL to test:** 

https://8806-status-edit-modal--atd-moped-main.netlify.app/moped/dashboard

**Steps to test:**
1. If you are not already following a project or in a project, add projects to your dashboard
2. Click on the status and the comment modal should appear (~~question, should the cursor change? its a text cursor now when you hover~~. Also, if there is no status, I'm rendering a `-` would appreciate feedback on this decision)
3. Add a status / edit a status. Upon completion, the modal should automatically close
4. your updated most recent status should be visible in the table
5. Also check that normal comments (https://8806-status-edit-modal--atd-moped-main.netlify.app/moped/projects/1827?tab=comments) still work as expected. 

My package-lock.json changed from when moment.js was added to the project and then taken away. Let me know if this should not be committed. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/layouts/DashboardLayout/Footer.js
```diff
@@ -27,8 +27,10 @@ const Footer = () => {
       "https://api.github.com/repos/cityofaustin/atd-moped/tags";
     const response = await fetch(githubTagsApiEndpoint);
     const data = await response.json();
-    if (data) {
+    if (data[0]) {
       setMostRecentTag(data[0].name);
```
chiaberry, line 31: this line was erroring when the response from the api endpoint included an error message. There was data being returned, but not an array of tags.

mateoclarke (approved):
🚢 💨 ✨ 
The code looks great. I think we could see this feature grow and cause complexity, but for this moment, its just what we need.

One suggestion for this feature:
- mouse hover state on the Status update cell should be a mickey mouse glove.

One broader observation about the Comment feature (maybe a future enhancement):
- You can make a status update with rich text from this feature. You can make an internal note with rich text from the Comments tab. But depending on the view (dashboard, summary, or comments tab) the rich text may not display.

moped-editor/src/layouts/DashboardLayout/Footer.js
```diff
@@ -27,8 +27,10 @@ const Footer = () => {
       "https://api.github.com/repos/cityofaustin/atd-moped/tags";
     const response = await fetch(githubTagsApiEndpoint);
     const data = await response.json();
-    if (data) {
+    if (data[0]) {
```
mddilley, line 30: Not sure if `data` can ever be `null` or `undefined` here, but, if so, this could break. Looks like we've used lodash `get` elsewhere in the app so that could be a safer way.

chiaberry:
@mateoclarke I changed the cursor to be the glove when hovering over the status update
 and excellent point about the rich text.  Should rich text be turned off when inputing from the modal? 
You can also currently edit a status from the comments page and add rich text there, so that edge case does currently exist in production.

moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -86,7 +86,8 @@ const useStyles = makeStyles(theme => ({
 const projectNoteTypes = ["", "Internal Note", "Status Update"];
 
 const ProjectComments = props => {
-  const { projectId } = useParams();
+  const isStatusEditModal = props.modal;
```
mddilley, line 89: This is just a different point of view, but I could see a world where two separate components named `ProjectComments` and something like `DashboardStatusEditModal` exist. Then, each component would reuse smaller components that have the rich text editor, etc. 

Not a huge concern here, but I'm just thinking back on how I made some of the project maps pull double-duty with shared logic and things broke down a bit.

mddilley (approved):
Works just as described and looks super clean - I am a fan! 📝 🙌  

About the `-`, it feels a lot like VZE since that was a pattern used for "no data" there, but, in Moped, the Timeline tab uses blank white space in the table for "no data." But, it is hard for me to think of how to communicate to Moped users through the UI that clicking a blank space gives them the ability to open a modal and add a status update so maybe having something there is the answer. Discoverability is hard.

I'll echo what Mateo said about the status updates which is definitely a tech debt backlog issue unrelated to the work here. Notes entered through this modal look like `<p>test</p>` in the database (just like internal notes entered through the Comments tab) while those entered through the Summary tab look like `test`.

patrickm02L (changes requested):
**Test steps feedback**

1. Pass. Followed a project with an empty status, used that to test. 
2. Pass/**Request change**. Modal looks good. Use the `+` icon pattern we've established with tooltip hover that we've establish in other sections instead of `-`.
3. Pass/**Request change**. Modal close automatically makes it hard to add/delete multiple status. Recommend modal stays open even after saving. Let users use `X` in top right to close.
4. Pass/**Request change**. When I deleted most recent status in modal they still appeared in table. If deleted, recommend updating status with previous status update. 
5. Pass. This worked as expected.

### Requested Changes

**PART 1**

2. `-` alternative @ 1:00-1:41
3. modal closing @ 1:42-3:49

https://user-images.githubusercontent.com/5566727/173709989-16b70a6d-9d19-4703-80c3-0dfea03d8b4d.mp4

**PART 2**

4. deleted status remains in table @ 1:38-4:55

https://user-images.githubusercontent.com/5566727/173710087-b727f1bf-0b82-4ad8-878e-505a39ba8b46.mp4

chiaberry:
@patrickm02L I will implement the + icon and tooltip instead of the -, and look into what's going on with the deleted statuses still appearing

I do wonder about keeping the modal open after submitting a status, if modal's intent is to be able to update statuses on multiple projects quickly, i think closing it after save works. I don't know if users will add more than one status at a time. Should we check with AMD to see which they would prefer?

moped-editor/src/views/projects/projectView/ProjectComments.js
johnclary: 👍 definitely a pretty big refactor to extract out some of the mutation callbacks and what have you.

sort of related, i see our `<Page/>` component is getting wrapped up in this modal - which updates `<head>` with the page title "Project Notes" when the modal is open. 

i think we can drop `<Page/>` altogether, since afaik none of the other summary tabs use it - and it's kind of weird that the page title changes to "Project Notes" when you're viewing the project comments tab?

johnclary:
@chiaberry this feature looks reaaally nice—thank you!

i'm coming out in favor of auto-closing the modal on submit - i like things snappy. but i agree with Patrick it's unexpected to have the modal auto-close when you delete or edit an existing comment. maybe we should persist the modal except when a new comment is added?

two more things - which might be out of scope:
- it's possible to create an empty comment. ideally i think the "Save" button should be disabled if the comment is empty. 
- is it possible to force the comment input to `autofocus`? there are a handful of other places in Moped where i'd [like to see us make use of it](https://github.com/cityofaustin/atd-data-tech/issues/9144)

johnclary (approved):

patrickm02L:
In response to @johnclary' comments 
> i'm coming out in favor of auto-closing the modal on submit - i like things snappy. but i agree with Patrick it's unexpected to have the modal auto-close when you delete or edit an existing comment. maybe we should persist the modal except when a new comment is added?

Talked with @chiaberry about the modal auto-closing scenarios and we agreed that on Save it should close. We talked about close on Deleting a status as well and also agreed it should close. If the intent of a modal is to keep things snappy then a Save, Edit, or Delete make sense to create an auto-closing action. 

Note: For those who want to bulk delete statuses that action can happen in the Comments tab of the project.

> two more things - which might be out of scope:
> 
> * it's possible to create an empty comment. ideally i think the "Save" button should be disabled if the comment is empty.
> * is it possible to force the comment input to `autofocus`? there are a handful of other places in Moped where i'd [like to see us make use of it](https://github.com/cityofaustin/atd-data-tech/issues/9144)

We also talked about the "Save" button being disabled and agreed that greying out the button until something was entered in the input box made sense. 

We did not talk about autofocus.

moped-editor/src/queries/dashboard.js
```diff
@@ -32,7 +32,7 @@ export const USER_PERSONNEL_PROJECTS_QUERY = gql`
           project_name
           current_phase
           current_status
-          moped_proj_notes(where: {project_note_type: {_eq: 2}}, order_by: {date_created: desc}) {
+          moped_proj_notes(where: {project_note_type: {_eq: 2}, status_id: {_eq:1}}, order_by: {date_created: desc}) {
```
chiaberry, line 35: @mddilley heads up mike, I had to include this line to prevent deleted statuses from showing up on the dashboard.

moped-editor/src/views/projects/projectView/ProjectComments.js
chiaberry: I removed the Page component and Container and replaced it with CardContent to match the other Project tabs

chiaberry:
Here are the changes I've made in response to some reviews, these have happened for the most part post dev approvals
- replaced the `-` with the ControlPointIcon and tooltip
- adjusted the query to only get comments with status = 1 (aka not delated)
- removed the Page component from ProjectComments and replaced Container with CardContent. This looks like bigger changes in the diff, because the indentation shifted.

patrickm02L (approved):
Updates look great. Everything works beautifully!

Screen Shot 2022-06-15 at 6 50 55 PM

johnclary:
🙌

PR #686, opened 2022-06-15
Backfill projects with template milestones

johnclary:
## Associated issues
This will resolve https://github.com/cityofaustin/atd-data-tech/issues/9257 once it's approved and tested in staging and run against production.

## Testing
**URL to test:** Local

**Steps to test:**
1. You'll need a python 3.x environment with `requests` installed.
2. Use the Moped editor to create a new project. Do not use the milestone template when you create the project.
3. Edit the project type of your project to one of "Signal - Mod", "Signal - New", "PHB - Mod", "PHB - New"
4. Run `create_milestones.py`, and set `--env (-e)` to `local` and the `max-date-added (-d)` arg date to be in the future

```
python create_milestones.py -e local -d '2022-06-17 00:00:00'
```

5. Navigate to your project timeline can confirm the template milestones have been added.

Other things you can test:
- Repeat the above steps, but use a milestone template from the new project workflow. Run the script. No milestones should be added to the project.
- Repeat the above steps, but set the project type to "Fiber - Mod". Run the script. No milestones should be added to the project.
- Repeat the above steps, but add a "Completed" phase to your project and set it as the current phase. Run the script. Your project should **not** have milestones added. Completed and post-construction projects are excluded.

---
#### Ship list
- [ ] Code reviewed 
- [ ] Run against staging environment
- [ ] Run against production environment

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -1512,6 +1512,9 @@
   - name: moped_milestone
     using:
       foreign_key_constraint_on: milestone_id
+  - name: moped_project
+    using:
+      foreign_key_constraint_on: project_id
   insert_permissions:
   - role: moped-admin
     permission:
@@ -2372,6 +2375,13 @@
         table:
           schema: public
           name: moped_project_types
+  - name: moped_proj_milestones
+    using:
+      foreign_key_constraint_on:
+        table:
+          name: moped_proj_milestones
+          schema: public
+        column: project_id
   insert_permissions:
   - role: moped-admin
     permission:
@@ -2387,7 +2397,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2422,7 +2432,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2457,7 +2467,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2491,7 +2501,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2525,7 +2535,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2560,7 +2570,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2595,7 +2605,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
       - knack_project_id
       - milestone_id
       - project_description
@@ -2628,7 +2638,7 @@
       update:
         columns:
         - capitally_funded
-        - is_deleted
+        - is_retired
         - end_date
         - added_by
         - milestone_id
@@ -2682,7 +2692,7 @@
     permission:
       check: {}
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2701,7 +2711,7 @@
     permission:
       check: {}
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2730,7 +2740,7 @@
       - file_permissions
       - file_size
       - file_type
-      - is_deleted
+      - is_retired
       - is_scanned
       - project_file_id
       - project_id
@@ -2738,7 +2748,7 @@
   - role: moped-editor
     permission:
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2756,7 +2766,7 @@
   - role: moped-viewer
     permission:
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2775,7 +2785,7 @@
   - role: moped-admin
     permission:
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2794,7 +2804,7 @@
   - role: moped-editor
     permission:
       columns:
-      - is_deleted
+      - is_retired
       - is_scanned
       - created_by
       - file_size
@@ -2820,7 +2830,7 @@
         columns: '*'
       update:
         columns:
-        - is_deleted
+        - is_retired
         - is_scanned
         - created_by
         - file_size
@@ -3659,7 +3669,7 @@
       - capitally_funded
       - date_added
       - added_by
-      - is_deleted
+      - is_retired
       - milestone_id
       - status_id
       - project_team_members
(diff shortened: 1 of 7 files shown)


moped-toolbox/amd_milestones_backfill/create_milestones.py
```diff
@@ -0,0 +1,104 @@
...
+        "-d",
+        "--max-date-added",
+        type=str,
+        required=True,
+        help=f"The maximum project date_added to include in this update in format YYYY-MM-DD (UTC). Must be '2022-06-09' in prod",
+    )
+
+    args = parser.parse_args()
+
+    if args.env == "prod" and args.max_date_added != "2022-06-09":
```
johnclary, line 96: i added this date argument so that it's easier to test locally against newly created projects.

moped-toolbox/amd_milestones_backfill/queries.py
```diff
@@ -0,0 +1,61 @@
...
+		project_name
+		date_added
+		moped_project_types {
+			moped_type {
+				type_name
+			}
+		}
+	}
+	# we don't currently have a hasura relationship between projects and milestones :/
+	# so download them all so we can join them manually
```
johnclary, line 44: this PR includes a patch to hasura metadata to track the `moped_project` <> `moped_proj_milestones` relationship, but i assume we'll want to get this backfill done without waiting for the next prod release.

moped-toolbox/amd_milestones_backfill/create_milestones.py
```diff
@@ -0,0 +1,104 @@
...
+        existing_milestone_ids = [
+            m["moped_milestone"]["milestone_id"]
+            for m in all_milestones
+            if m["project_id"] == project_id
+        ]
+
+        proj_milestones_new = []
+
+        for m in TEMPLATE_MILESTONES:
+            if m["milestone_id"] not in existing_milestone_ids:
```
johnclary, line 45: this script can be run repeatedly with same args without concern for duplicating milestones. if a project already has a milestone ID specified in the template, that milestone is skipped.

the script is making the assumption that the `moped_milestone` IDs in local, staging, and project all match (they do)

moped-database/metadata/tables.yaml
```diff
@@ -2387,7 +2397,7 @@
       - ecapris_subproject_id
       - end_date
       - fiscal_year
-      - is_deleted
+      - is_retired
```
johnclary, line 2400: i have no idea how these ended up in my file

johnclary: i quit

moped-database/metadata/tables.yaml
```diff
@@ -2372,6 +2375,13 @@
         table:
           schema: public
           name: moped_project_types
+  - name: moped_proj_milestones
+    using:
+      foreign_key_constraint_on:
+        table:
+          name: moped_proj_milestones
+          schema: public
+        column: project_id
```
johnclary, line 2384: my script does *not* make use of this relationship, but it would have been nice to have it in place for this task.

moped-toolbox/amd_milestones_backfill/create_milestones.py
```diff
@@ -0,0 +1,104 @@
...
+    update_count = 0
+
+    for project in projects:
+        project_id = project["project_id"]
+
+        existing_milestone_ids = [
+            m["moped_milestone"]["milestone_id"]
+            for m in all_milestones
+            if m["project_id"] == project_id
+        ]
```
mddilley, line 40: This is pretty slick here 🔥

moped-toolbox/amd_milestones_backfill/create_milestones.py
```diff
@@ -0,0 +1,104 @@
...
+from secrets import HASURA
+
+
+def main(env, max_date_added):
+
+    logger.info(f"Gettings projects added on or before {max_date_added} in {env} env")
+
+    data = make_hasura_request(
+        query=PROJECTS_QUERY,
+        variables={"max_date_added": max_date_added},
```
mddilley, line 21: I noticed that the `-d` arg didn't work as expected for projects that I started creating at `2022-06-15T19:06:18.128427+00:00`. I had to pass `2022-06-16` for them to be picked up. Just holding off on approval because of that since it seems having the target date range query as expected is important here. I could be missing something so please let me know if so!

Screen Shot 2022-06-15 at 2 08 54 PM

mddilley (commented):
This script looks great, and I tested all of the options in the steps. All of the local projects that were expected to backfill succeeded. 🙌  Holding off on approval because of one question that I had about the date arg.

moped-toolbox/amd_milestones_backfill/create_milestones.py
johnclary: thanks for banging on that @mddilley! i thought i understood the expected behavior with taking the shortcut of a partial timestamp, and i clearly didn't.

i'm going to drop this check altogether and update the docs to reflect that the input date string is going to be handled according to postgres's [parsing rules](https://www.postgresql.org/docs/current/datetime-input-rules.html) and suggest that you manually test the graphql query before running this in prod.

moped-toolbox/amd_milestones_backfill/create_milestones.py
johnclary: 🐍

mddilley (approved):
Awesome - works beautifully! 🐍 🚢

chiaberry (approved):

moped-toolbox/amd_milestones_backfill/create_milestones.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    parser.add_argument(
+        "-d",
+        "--max-date-added",
+        type=str,
+        required=True,
+        help=f"The maximum project date_added to include in this update. Formatted as a postgres-compliant timestamptz string",
+    )
+
+    args = parser.parse_args()
```
frankhereford, line 94: Your script is a really helpful example of how to write a tidy piece of python. I always put stuff like this up at the top of the file right after imports. I'm going to adapt this style into my future work - thanks!

moped-toolbox/amd_milestones_backfill/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    res = requests.post(endpoint, json=payload, headers=headers)
+    res.raise_for_status()
+    data = res.json()
+    try:
+        return data["data"]
+    except KeyError:
+        raise ValueError(data)
+
+
+# copied from atd-moped/moped-editor/src/utils/timelineTemplates.js
```
frankhereford, line 30: It looks like this data also has a relationship to the database table `moped_milestones`. You might want to add an entry for this script in `moped-toolbox/README.md` and note what the script does and that a future user needs to make sure this data aligns to their database or they will get unexpected results. Not a major blocker by any means though.

frankhereford (approved):
🐍 🐍 ⛵ This is great, I am 100% in on this getting executed on staging and production in turn.

moped-toolbox/amd_milestones_backfill/utils.py
johnclary: i totally missed that README. will do!

moped-toolbox/amd_milestones_backfill/queries.py
```diff
@@ -0,0 +1,64 @@
...
+						{ current_phase: { _nilike: "complete" } }
+						{ current_phase: { _nilike: "post-construction" } }
+					]
+				}
+				# must contain at least one not-deleted project type related to signals or PHBs
+				{
+					moped_project_types: {
+						_and: [
+							{ moped_type: { type_name: { _similar: "%(Signal|PHB)%" } } }
+							{ status_id: { _eq: 1 } }
```
johnclary, line 28: while testing against staging, i uncovered a bug stemming from the fact i overlooked that there are "deleted" `moped_project_types` in the DB. i added `{ status_id: { _eq: 1 } }` here to ensure we ignored those.

mateoclarke (commented):
👍

PR #687, opened 2022-06-16
Remove "Projects" header from project list

johnclary:
## Associated issues
Resolves https://github.com/cityofaustin/atd-data-tech/issues/9515

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableNewItem.js
+++ b/moped-editor/src/components/GridTable/GridTableNewItem.js
@@ -13,21 +13,16 @@ import { NavLink as RouterLink } from "react-router-dom";
 const GridTableNewItem = ({ query }) => {
   return (
     <Box display="flex" justifyContent="flex-end">
-      {(query.config.showNewItemButton && query.config.customNewItemButton) ||
-        (query.config.showNewItemButton && (
-          <Button
-            color="primary"
-            variant="contained"
-            component={RouterLink}
-            to={query.config.new_item}
-            startIcon={<Icon>add_circle</Icon>}
-          >
-            {query.config.new_item_label}
-          </Button>
-        ))}
-      {// add empty div of equal height if there is no button to match styles
-      !query.config.showNewItemButton && (
-        <div style={{ height: "36.5px" }}></div>
+      {query.config.customNewItemButton || (
+        <Button
+          color="primary"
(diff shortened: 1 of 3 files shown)


moped-editor/src/components/GridTable/GridTableNewItem.js
```diff
@@ -13,21 +13,16 @@ import { NavLink as RouterLink } from "react-router-dom";
 const GridTableNewItem = ({ query }) => {
   return (
     <Box display="flex" justifyContent="flex-end">
-      {(query.config.showNewItemButton && query.config.customNewItemButton) ||
```
johnclary, line 16: It looks like we built in future support for "new item buttons", however this is only used in the staff listing, and the staff listing uses the `customNewItemButton`. 

in any case, because this render logic was inside the component, it was adding an empty div to the top of the project list.

Screen Shot 2022-06-16 at 2 05 09 PM

johnclary:
no rush here, y'all! i def want AA to review. i spun up this issue out of nowhere b/c i noticed the weird empty div while learning MUI.

moped-editor/src/components/GridTable/GridTableNewItem.js
chiaberry: the add new project button used to be in the same place as the add new staff button, but when we moved the new project button to the header, the spacing was messed up. I thought the empty div was supposed to help the spacing, but now looking at projects page vs staff page, the spacing still doesnt match 😅

johnclary: ah, that makes sense 😅

frankhereford (approved):
Looks great, thank you John. Saving real-estate, one element at a time.

moped-editor/src/components/GridTable/GridTableNewItem.js
mateoclarke: I agree with Chia, and I think if we're going to make this change to Projects, we should make sure we're consistent and apply this to the Staff page as well.

mateoclarke (approved):
👍

amenity (approved):
🪄 Thanks, @johnclary!

PR #691, opened 2022-06-21
Upgrade to react-router v6.3

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9536

## Summary of changes
1. Migrate to lockfile v2 ([discussion](https://austininnovation.slack.com/archives/CHZE6BC6L/p1655911113444509))
2. Tweak routes to make them compatible with API changes
3. Implement a `<NavLink/>` wrapper component to preserve the `activeClassName` prop
4. Remove side navigation components which we made hidden some time ago.
5. Forcibly bump `react-map-gl-draw`'s peer dependency of React from v16 to v17 (`package-lock.json`)
6. Remove react-hook-form dev tools, which had package conflicts

## Testing
**URL to test:** Netlify and local

**Steps to test:**
We're testing to make sure all of the navigation links work normally. 
1. Test the top nav links: dashboard, projects, staff
6. Click on a project from the project list—make sure all project tabs load
7. Sign out from the drop-down nav

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/routes.js
```diff
@@ -20,20 +20,20 @@ import DashboardView from "./views/dashboard/DashboardView";
...
     ],
   },
   {
-    path: "moped",
+    path: "/moped",
     action: "moped:visit",
     element: <DashboardLayout />,
     children: [
       {
-        path: "/",
```
johnclary, line 36: react-router now throws an error if you try to set the root `/` as a child. tbh i'm not sure why this was working in the first place :/

moped-editor/src/routes.js
```diff
@@ -20,20 +20,20 @@ import DashboardView from "./views/dashboard/DashboardView";
...
+    path: "/moped/session",
     element: <MainLayout />,
     children: [
       { path: "signin", element: <LoginView /> },
       { path: "*", element: <Navigate to="/signin" /> },
     ],
   },
   {
-    path: "moped",
+    path: "/moped",
```
johnclary, line 31: this path must be absolute in order for the redirect from `/moped` to `/moped/projects` to work.

johnclary:
You'll need to look at `package-lock.json` locally to see that I manually bumped the React peer dependency of  `react-map-gl-draw` from React 16 to 17.

moped-editor/src/components/NavLink.js
```diff
@@ -0,0 +1,26 @@
...
+        style={({ isActive }) => ({
+          ...props.style,
+          ...(isActive ? activeStyle : null),
+        })}
+      />
+    );
+  }
+);
+
+export default NavLink;
```
johnclary, line 26: I tried a few different ways to deal with the [deprecated `activeClassName` prop](https://reactrouter.com/docs/en/v6/upgrading/v5#remove-activeclassname-and-activestyle-props-from-navlink-). 

Probably my own ineptitude, but this was the only one I got working. Also considered:
- updating our MUI classes to use an `active` pseudo-selector
- doing a global component override in the [MUI theme](https://mui.com/material-ui/guides/routing/#global-theme-link) 
- using a custom `className` function

moped-editor/src/index.js
```diff
@@ -48,7 +48,7 @@ Amplify.configure({
 });
 
 ReactDOM.render(
-  <BrowserRouter basename={"/moped"}>
+  <BrowserRouter>
```
johnclary, line 51: it seems that in react-router `6.0.0-beta.0`, the `basename` prop was being ignored. that's fixed, and so
'moped' was being duplicated in all route paths.

it would seem more elegant to use the `basepath` prop and get the "/moped" cruft out of our paths. i did run with that and get things working with that approach. however, if we go with the `basename` approach, it's not possible to redirect the root path `/` to `/moped`, which means that in local dev and deploy previews you'd have to manually append "/moped" to your base URL. anyway, if that sounds preferable to y'all it's a minor code change.

johnclary:
OK—actually ready for review. thank you!

moped-editor/package.json
```diff
@@ -85,15 +85,14 @@
...
-    "react-router-dom": "^6.0.0-beta.0",
+    "react-router": "^6.3.0",
+    "react-router-dom": "^6.3.0",
     "react-scripts": "^3.4.1",
     "styled-components": "^5.3.0",
     "uuid": "^8.3.2",
     "yup": "^0.29.3"
   },
   "devDependencies": {
-    "@hookform/devtools": "^2.2.1",
```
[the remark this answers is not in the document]
mddilley: Let's toss it! I added this back in the day, and it doesn't need to stay here since we are not actively developing with React Hook Form.

moped-editor/src/components/NavLink.js
mddilley: This makes sense to me, and it is an interesting problem. I like fixes like this that give kind of an under-the-hood view at what is going on.

moped-editor/src/index.js
mddilley: Might be more elegant but I think explicit routes can be nice too since we had some trouble tracking down what was going on with the previous config the other day.

moped-editor/src/layouts/DashboardLayout/NavBar/NavBar.js
```diff
@@ -1,105 +0,0 @@
-import React, { useEffect } from "react";
```
mddilley, line 1: 🗑️ 👍

mddilley (approved):
I hunted down each and every link in the deploy preview that I could find, and they all worked. Took me a second the other day to realize that RR6 is the real deal now. Thanks for navigating the gotchas with the upgrade! 🎉

amenity (approved):
LGTM! 👍🏻 Thanks, @johnclary!

moped-editor/package.json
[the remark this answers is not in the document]
chiaberry: throw it away!

moped-editor/src/components/NavLink.js
```diff
@@ -0,0 +1,26 @@
...
+// see: https://reactrouter.com/docs/en/v6/upgrading/v5#remove-activeclassname-and-activestyle-props-from-navlink-
+const NavLink = React.forwardRef(
+  ({ activeClassName, activeStyle, ...props }, ref) => {
+    return (
+      <BaseNavLink
+        ref={ref}
+        {...props}
+        className={({ isActive }) =>
+          [props.className, isActive ? activeClassName : null]
+            .filter(Boolean)
```
chiaberry, line 14: this is a cool way of doing this

chiaberry (approved):
tested links, checked that the 404 page shows up, all looks good!

johnclary:
Thanks y'all!

PR #703, opened 2022-07-01
Add `is_deleted` column for moped users

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9553

## Testing
**URL to test:** 
_This needs to be tested on moped-test._

**Steps to test:**
_I'm using JD's account on purpose so that we don't mess up our own Moped accounts. Sergio's account is also good for testing. You might see Cognito errors return from the Moped API for other test accounts since the staging user pool data is not consistent with the `moped_users` table in the moped-test DB._

_Moped-test uses staging's DynamoDB table for roles and Cognito user pool so you could desync those data sources from the staging DB if you modify your own account - so be careful!_

1. Go to https://moped-test.austinmobility.io/moped/staff/edit/1
2. Click **Inactivate User** to inactivate JD.
3. You should be redirected to the user table and no longer see JD's account listed in the table.
4. Try to add JD to a new or existing project team and you should not see him in the personnel list
5. Go back to the Staff table and toggle the **Show Inactive Accounts** switch to show JD in the list.
6. Click the pencil in JD's row, add a password to the edit form, and then click **Activate User** (you will get a warning if the password isn't complex enough)
7. Try to add JD to a new or existing project team and you should see him listed in the personnel list

---
#### Ship list
- [x] Code reviewed 
- [x] ~Product manager approved~


moped-editor/src/views/staff/StaffForm.js
```diff
@@ -175,6 +172,9 @@ const StaffForm = ({ editFormData = null, userCognitoId }) => {
       delete data.password;
     }
 
+    // Reactivation needs this value to pass with programmatic submission in handleActivateConfirm
```
mddilley, line 175: Not sure if this is the best place for this. When a deactivated user was activated previously, the submission payload took advantage of the hidden field that I removed further down in the diff. That hidden field really isn't needed since `is_deleted` initializes as `false` for new users (it used to need `status_id = 1` in the payload).

Also, I'm making a backlog issue to split out the create and edit logic into two components. Create and edit actions (like deactivate/activate) are pretty tangled here. I'm all over the blame for this code so I'm happy to improve this when there is priority for this. 🖐️ 😄

mateoclarke: 👍👍👍

mateoclarke (approved):
This all looks really good. Thanks for moving all this dirt and refactoring away this tech debt. 

One thing I'm noting (and this is probably out of scope to address in this PR) is that once you deactivate a user, their name in the Team listed becomes "Unknown User". I think ideally, even if a user is soft deleted, we'd be able to see their name listed in the team. Even better would be some indicator that this team member has an inactive status. If @amenity agrees, I think we should backlog an issue to improve the way inactive users are displayed on the Team tab.

🚢 💨✨

johnclary (approved):
🚢 🔒 💨 

it's awesome to see this get patched up, Mike. thanks for tackling it—and i appreciate getting to learn via your work how these pieces fit together!

johnclary:
r.e Mateo's comment
> ideally, even if a user is soft deleted, we'd be able to see their name listed in the team

totally agree this should be the behavior, and its reminiscent of https://github.com/cityofaustin/atd-data-tech/issues/9673

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -84,16 +83,14 @@ const staffValidationSchema = (isNewUser, userStatusId) =>
       .lowercase(),
     password: yup.mixed().when({
       // If we are editing a user, password is optional
-      is: () => isNewUser || userStatusId !== 1,
+      is: () => isNewUser,
       then: yup.string().required(),
```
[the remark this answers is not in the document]
mddilley: I just checked it out, and Cognito, the API, and the UI have password validation and it seems that errors can come back from all three of those. I added making sure they are aligned to https://github.com/cityofaustin/atd-data-tech/issues/9697

mddilley:
@johnclary @mateoclarke @amenity I think that John is right about https://github.com/cityofaustin/atd-data-tech/issues/9673. I added some notes to that issue so that we can show active and inactive users in project teams.

PR #696, opened 2022-06-27
Add link to parent project from subproject summary page

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9330

## Testing
**URL to test:** 

https://9330-link-to-parent--atd-moped-main.netlify.app/moped/projects/1841/

**Steps to test:**

1. Add a child project to an existing project. Use the link from the subprojects table to visit summary page of child project. 
2.  Child project should have a link to parent project
3.  if there is no parent project, i.e. the project is not a subproject, the field in the summary page should not render. (You can tell a project is not a subproject by checking to see if the subprojects table is showing)

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager approved
- [x] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #704, opened 2022-07-05
Disassociate child projects when parent is deleted

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9329

## Testing
**URL to test:** 

https://9329-disassociate-subprojects--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

1. Add subprojects to a project (since https://github.com/cityofaustin/atd-moped/pull/696 isnt merged yet, the only way to see the relationship in the UI right now is via the parent project page) Remember the project_id of the project. 
2. Delete the project (three dots under the following icon, select delete)
3. Moped will redirect you to the projects page. Navigate back to the deleted project's page (using the project_id you remembered in step 1), the subprojects table should be empty. The only indication the project has been deleted is you can't see it on the projects list table. 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/queries/subprojects.js
```diff
@@ -53,3 +57,14 @@ export const DELETE_PROJECT_SUBPROJECT = gql`
...
+export const CLEAR_PARENT_PROJECT = gql`
+  mutation ClearParentProject($projectId: Int!) {
+    update_moped_project(
+      where: { parent_project_id: { _eq: $projectId } }
+      _set: { parent_project_id: null }
+    ) {
+      affected_rows
+    }
+  }
+`;
```
mddilley, line 70: This makes total sense here, but I wondered if this could belong in the `PROJECT_ARCHIVE` mutation and look something like below. The differences that I see is that it would happen in the same API call as the archive which could help make sure this happens each and every time, and also it means less code. It does use [an alias](https://graphql.org/learn/queries/) in the mutation since there is a conflict in the `update_moped_project` field names in the mutation.

```
export const PROJECT_ARCHIVE = gql`
  mutation ArchiveMopedProject($projectId: Int!) {
    update_moped_project(
      where: { project_id: { _eq: $projectId } }
      _set: { is_deleted: true }
    ) {
      affected_rows
    }
    clear_parent_project: update_moped_project(
      where: { parent_project_id: { _eq: $projectId } }
      _set: { parent_project_id: null }
    ) {
      affected_rows
    }
  }
`;
```

mddilley (commented):
Awesome! This is that missing piece that brings this all together! 🧩 

I made one comment about a different approach, and I'm wondering if it makes sense at all. Happy to approve because this works great just the way it is but wanted to see what you thought.

moped-editor/src/queries/subprojects.js
```diff
@@ -1,24 +1,25 @@
...
+    subprojectOptions: moped_project(
+      where: {
+        _and: [
+          { is_deleted: { _eq: false } }
+          { parent_project_id: { _is_null: true } }
+          { project_id: { _neq: $projectId } }
+        ]
+        _not: { moped_projects: {} }
+      }
+    ) {
```
johnclary, line 22: observing that this is merely linting 👍

moped-editor/src/queries/subprojects.js
johnclary: +1 for doing the subproject updates in the archive mutation!

moped-editor/src/queries/subprojects.js
johnclary: ...also observing that the `subprojectOptions` typename aliasing is cool and i forgot that it was possible/necessary
https://hasura.io/blog/customising-alias-graphql-fields-with-hasura/

johnclary (approved):
looks great / works great

moped-editor/src/queries/subprojects.js
chiaberry: Agree!! I thought it maybe would be better to have them separate in case the subproject update failed it wouldn't rollback all the changes, but now that I think about it, thats not a good reason at all, and I'd rather have the rollback. I'm going to clean up this code and use your mutation @mddilley , thank you

mddilley: Awesome - just tested again and it works wonderfully! Going to add that ✅ now

mddilley (approved):

amenity:
Note - I tested this backward, deleting the subproject. I'd expect the subproject to either disappear from the parent projects' "Subprojects" list or, preferably, still appear but in grey italic with "Deleted" as the status.

amenity (approved):
Works! Thanks, @chiaberry

chiaberry:
@amenity you aren't missing something, I did not query for only not-deleted subprojects. I'll fix that .

mateoclarke (approved):
🚢 💨✨

This is the type of muppet™️ gif you find when use search "Disassociate." 
shocked-big-bird

frankhereford (commented):
Thank you Chia for coding up this change; it works great and the code looks totally on point. You're comment about how there only real behavior we get when something is deleted is to hide it in the listing got me thinking. 

I would hate for someone to get a link to a project sent to them in like an email or what-have-you, and then time go by, and then the project gets deleted. If that link were to be followed by someone, they would have no idea that they are looking at a non-functional, deleted project. 

@amenity Would it be something we might want to consider to put a bold "Deleted" notification at the top of project summary pages that tell users they are looking at a relic? Just thinking out loud - thanks!

chiaberry:
https://github.com/cityofaustin/atd-data-tech/issues/9525 Theres an issue for that very thing!

PR #695, opened 2022-06-23
Upgrade to react-scripts v5

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9561

## Steps I took (for reference):
1. `$ nvm use` 😉 
2. delete `/node_modules` and `package-lock.json`
3. edit `package.json` to remove `react-map-gl-draw`
4. edit `package.json` to remove all `devDependencies`
6. manually set `react-scripts` in `package.json` to `^5.0.0`
7. re-install node modules and rebuild `package-lock.json`: `$ npm install`
8. install `react-map-gl-draw` again, forcibly: `$ npm install react-map-gl-draw@^0.22.3 --force`
9. manually bump react/react-dom `peerDependencies` of `react-map-gl-draw` to `^17.x` in `package-lock.json` 
10. install babel import plugin: `$ npm install --save-dev babel-plugin-import`
11. remove `eslintConfig` in `package.json` and update `.eslintrc.json`
12. code changes to get file upload and components map to compile
13. code changes to stop exporting anonymous objects
14. `$ npm run test` ⚙️   
13. `$ npm run start` ✨ 

## About our ESLint config

It's been hard for me to untangle the effects of this upgrade on linting. We're migrating from ESLint v6 to v8, as well as upgrading a number of related parsers and configs.

Also, as far as I can tell, the `react-app`  ESLint config was previously being ignored, because it was defined in `package.json`, and it was being overwritten by our `.eslintrc` file ([docs](https://eslint.org/docs/latest/user-guide/configuring/configuration-files)).

One issue I immediately encountered were errors with anonymous default exports. That rule has been in place [for years](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-anonymous-default-export.md), but was maybe only enabled by `react-app`? 🤔. Anyway, I addressed those errors in this PR.

The `react-app`  ESLint config is Create React App's base lint configuration, and they recommend that we do not ignore it. It's now enabled. Besides a few anonymous default export problems, the app is (surprisingly?) compiling without errors. 

Additionally, we had a number of ESLint plugins through [`eslint-config-airbnb`](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb). However, none of these configs were enabled in our `.eslintrc` file. I removed them:

```
 'eslint-plugin-import': '^2.25.3',
  'eslint-plugin-jsx-a11y': '^6.5.1',
  'eslint-plugin-react': '^7.28.0',
[...]

PR #706, opened 2022-07-06
Overrides react-map-gl-draw react dependency

johnclary:
As discussed [here](https://github.com/cityofaustin/atd-moped/pull/695#issue-1282937937) and [in slack](https://austininnovation.slack.com/archives/CHZE6BC6L/p1655417502246739?thread_ts=1655413287.536209&cid=CHZE6BC6L), the `react-map-gl-draw` package has an outdated react dependency (`v16`) - which requires that we force install that package.

This PR adds a dependency override ([docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides)) to allow us to run a clean, lock-free install of Moped as necessary.

We'll need to keep an eye on this override as it may become undesired/outdated in the future.

## Testing

This change has no effect on our installed packages and does not need to be UI tested. To see the effects:

1. Delete `package-lock.json` and `node_modules/`
2. `$ npm install`

Moped should install without error.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -94,5 +94,11 @@
   },
   "devDependencies": {
     "babel-plugin-import": "^1.13.5"
+  },
+  "overrides": {
+    "react-map-gl-draw": {
+      "react": "^17.0.2",
+      "react-dom": "^17.0.2"
+    }
   }
 }


chiaberry (approved):

johnclary:
thanks y'all!

mateoclarke (commented):
👍

frankhereford (commented):
👍 💯 TIL - thank you!

PR #708, opened 2022-07-07
Fix "is blank" filtering bug

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8950
https://github.com/cityofaustin/atd-data-tech/issues/9417

Previously when fields were "empty", we seemed to have been getting `""` empty strings. But now most of the fields come back with `null`. So I've switched up the query construction to go back to using `is_null` instead of `_eq` for those cases. 

This PR also fixes the advanced search filter for the Status, by removing Current status and renaming Current phase to be Status. Theres some more background in the comments of issue 9417. 

## Testing
**URL to test:** 

https://8950-is-blank--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

The fields that have `null` are 
Team member
Project Partners
Task orders
Contractor
Project DO
Funding source
Status update
Designer
Inspector

When there is no Project Sponsor, the db returns `None`. If Status is blank, the db returns `""`. 

These are all the fields that you can search by is blank or is not blank. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -32,7 +32,7 @@ import parse from "html-react-parser";
 /**
  * GridTable Style
  */
-const useStyles = makeStyles(theme => ({
+const useStyles = makeStyles((theme) => ({
   root: {
     width: "100%",
   },
@@ -195,11 +195,10 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
      * add the to the Or list.
      */
     Object.keys(query.config.columns)
-      .filter(column => query.config.columns[column]?.searchable)
-      .forEach(column => {
-        const { operator, quoted, envelope } = query.config.columns[
-          column
-        ].search;
+      .filter((column) => query.config.columns[column]?.searchable)
+      .forEach((column) => {
+        const { operator, quoted, envelope } =
+          query.config.columns[column].search;
         const searchValue = getSearchValue(query, column, search.value);
         const graphqlSearchValue = quoted
           ? `"${envelope.replace("{VALUE}", searchValue)}"`
@@ -211,15 +210,9 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
 
   // For each filter added to state, add a where clause in GraphQL
   // Advanced Search
-  Object.keys(filters).forEach(filter => {
-    let {
-      envelope,
-      field,
-      gqlOperator,
-      value,
-      type,
-      specialNullValue,
-    } = filters[filter];
+  Object.keys(filters).forEach((filter) => {
+    let { envelope, field, gqlOperator, value, type, specialNullValue } =
+      filters[filter];
 
     // If we have no operator, then there is nothing we can do.
     if (field === null || gqlOperator === null) {
@@ -231,8 +224,7 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
       if (specialNullValue) {
         gqlOperator = envelope === "true" ? "_eq" : "_neq";
         value = specialNullValue;
-      }
-      else {
+      } else {
         value = envelope;
       }
     } else {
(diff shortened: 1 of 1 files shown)


mateoclarke (approved):
🛳️💨✨

moped-editor/src/components/GridTable/GridTableFiltersCommonOperators.js
```diff
@@ -29,24 +29,40 @@ export const GridTableFiltersCommonOperators = {
...
+    specialNullValue: '"None"',
     type: "string",
   },
   string_is_not_null_special_case: {
     operator: "_is_null",
     label: "is not blank",
     description: "String field does not match special null value",
     envelope: "false",
-    specialNullValue: '" :"',
+    specialNullValue: '"None"',
```
frankhereford, line 65: Given that there is a way to handle a `specialNullValue`, this isn't a blocker by any means, but we may want to consider unifying all the columns' behavior. I think when there is a representation of the null set of something, we have the DB return`null`. If we want to show "None" in the UI, no problem, but the actual data we get back, we may want to conform to a agreed upon way we do things across the board.

frankhereford (approved):
I think this change is great, and not just because it absolutely achieves those two issues' needs. This is the sort of change that makes the app behave more like peoples' shared expectations of how an app should work, and in doing things like this, the UX of Moped just goes up and up.

Thank you @chiaberry for your work doing this, but also to you and @johnclary and everyone else who identified and shaped this refinement. These little things are what add up to a really great final product. 🙏 

I've tested this on the provided netlify deployment and the code looks great.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -558,7 +565,7 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
                     Body: props => {
                       const indexedData = data["project_list_view"].map(
                         (row, index) => ({
-                          tableData: { id: index },
+                          tableData: { id: index, uuid:row.project_id },
```
mddilley, line 568: Curious about this addition here - I couldn't figure out what the uuid is for and I couldn't find anything about it in the material-table docs. It's also been a while since I've worked with these components so bear with me 😅

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -485,6 +490,8 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
     }
   }, [data, advancedSearchAnchor]);
 
+    console.log(query.query)
```
mddilley, line 493: Maybe this is leftover from testing or it is important feedback - not a big deal at all

mddilley (approved):
I tested all of the fields in the test steps and saw the appropriate records. 👍  So glad that you fixed this up so the filters work as expected now! 🥳 🛵

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
johnclary: well now i'm curious, too! it seems to fix a unique key prop warning. i checked material-table docs (original and fork) and found nothing about this. Chia, the ultimate 🕵️‍♀️ ??

johnclary (approved):
🚢  nice improvement — works as promised - thank you!!

chiaberry:
Screen Shot 2022-07-08 at 3 28 53 PM

this is the error that showed up in the console. I think post react-scripts upgrade? but I don't remember seeing it before

So I looked to see where the key is being set, and I found this: https://github.com/material-table-core/core/blob/master/src/components/m-table-body.js#L116

I previously had messed with the table Data because of some slicing weirdness (https://github.com/cityofaustin/atd-moped/pull/639), so this could be why the key isnt set automatically. But in any case, we need to put a key in there. cc: @mddilley @johnclary

mddilley:
awesome - thanks @chiaberry for diving into the source 🤿 I like it!

amenity (approved):
👍🏻 🙏🏻 Thanks, @chiaberry!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: cleaned it up! thank you for pointing it out.

PR #712, opened 2022-07-12
Add ability to track Interim DB project and component IDs

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9569

## Testing
**URL to test:** 

local or use moped-test 

Screen Shot 2022-07-13 at 11 31 42 AM

**Steps to test:**

Add a ID to a project. (It is numeric only.)
remove the ID. edit ID. 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryInterimID.js
```diff
@@ -0,0 +1,143 @@
...
+ * @constructor
+ */
+const ProjectSummaryInterimID = ({
+  projectId,
+  data,
+  refetch,
+  classes,
+  snackbarHandle,
+}) => {
+  // Instantiate Original Value
```
chiaberry, line 29: this component is a copy of the other project summary components

moped-database/migrations/1657564405501_alter_table_public_moped_project_add_column_interim_project_id/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."moped_project" add column "interim_project_id" integer
+ null;
```
frankhereford, line 2: No problem here, but I wonder why there is a linefeed chopping off one keyword from this `alter` statement. Works fine, just curious. Maybe that's right at 80 characters. Same in the other migration.

moped-database/migrations/1657565379012_alter_table_public_moped_proj_components_add_column_interim_project_compoennt_id/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."moped_proj_components" add column "interim_project_component_id" integer
```
frankhereford, line 1: The DB is configured to only store integers, which feels correct. The front-end is a bit looser, and will let you submit numbers, and sometimes they throw a toast error about the the graphql endpoint not accepting it, and sometimes they do and silently fail to save it. Here are some example inputs that give results:

`24601` → Green toast, works fine
`24.601` → 
image
`2460e1` → Similar red toast error
`2.4e601` -> Green, but doesn't save a value, which is confusing to me.

frankhereford (approved):
I think this works great, and I've included an observation about input validation, but I don't think it's a show stopper. Thank you @chiaberry for this work!

I've tested it locally and read over the code, and I think it's 🚢 🚀

moped-database/migrations/1657565379012_alter_table_public_moped_proj_components_add_column_interim_project_compoennt_id/up.sql
mddilley: This reminds me of the issue that I had with project funding amounts. I made a component called `FundingAmountIntegerField` that might be helpful here? Feel free to make that component name more general to integer inputs if they end up aligning closely.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryInterimID.js
```diff
@@ -0,0 +1,143 @@
...
+  const [originalValue, setOriginalValue] = useState(
```
mddilley, line 30: _This could totally be a spin-off issue, and I don't think this blocks 🛳️-ing_

I know this is a copy an older component so I think there is a chance to simplify the logic a little. From what I'm seeing, the two pieces of state called `originalValue` and `interimId` can be combined. I did a quick test, and this seems to work but I could be missing a reason for the `useEffect`.

**The way this and the eCapris component work**
`props.data.moped_project[0].interim_project_id` -> initializes value of `originalValue` -> useEffect sets `interimId` when `originalValue` initializes or changes after through the UI

**Without the state called `originalValue` and the `useEffect`**
`props.data.moped_project[0].interim_project_id` -> initializes value of `interimId` and `setInterimId` updates `interimId` through the UI

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryInterimID.js
```diff
@@ -0,0 +1,143 @@
...
+    data?.moped_project?.[0]?.interim_project_id ?? null
```
mddilley, line 31: This is tricky because, if you stay in the summary view, this seems to work. If you navigate away and return, the value clears because the `interim_project_id` is missing from `SUMMARY_QUERY`. 

It's a straightforward fix but I noticed that the value still doesn't populate. ~It might have to do with the duplicate state above? Not sure.~ I _think_ the next comment is where this is happening.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryInterimID.js
```diff
@@ -0,0 +1,143 @@
...
+              className={classes.editIconConfirm}
+              onClick={handleProjectInterimIdClose}
+            >
+              close
+            </Icon>
+          </>
+        )}
+        {!editMode && (
+          <ProjectSummaryLabel
+            text={(interimId ?? "").length > 0 ? interimId : "None"}
```
mddilley, line 133: This could also be a place to check for why the `interimId` isn't showing in the UI when set.

mddilley (changes requested):
Just a couple small updates will get this over the line. The way the project summary view makes this tricky as heck to catch, and I almost overlooked the comments I left. 🔍

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryInterimID.js
chiaberry: you're right, it was the next comment where it was happening, you cant find the length of a number i.e. `44.length`, thats undefined. so it was always displaying None

mddilley (approved):
Tested this out locally, and it works great! 🚀 🚀 🚀

patrickm02L (commented):
There were a few little things that made me pause. Would like to see if you can resolve them before approval.

- You can enter an "e" as part of the ID for some reason. When saved, a green pop-up message confirms the submission, but it does not save in the input area. 
- 1password icon is hovering over the number picker. When my 1password was unlocked it was autofilling the input area (not shown in vid). Do we even need a number picker?

https://user-images.githubusercontent.com/5566727/180452947-abc41f32-dcb0-456e-8a72-3d243bf55e47.mp4

chiaberry:
@patrickm02L I've updated the code to use the functions Mike wrote and recommended. For some reason the utility functions mess up with the numeric inputmode, so I removed that which should hopefully prevent the 1password extension from popping up.

PR #713, opened 2022-07-12
Updates activity event validation schema 

johnclary:
## Associated issues
Fixes https://github.com/cityofaustin/atd-data-tech/issues/9723

## Testing
run moped-data-events tests locally with Python

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #714, opened 2022-07-12
Updates activity event validation schema (production patch)

johnclary:
Applies https://github.com/cityofaustin/atd-moped/pull/713 as a production hotfix.


mddilley (approved):
🤞

PR #721, opened 2022-07-15
Show inactive users in activity log and team table

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9673

This PR fixes a bug where inactivated users were showing as "Unknown User" in the team table and the activity log. The "Unknown User" came from two separate components and was implemented in two different ways so I pulled out some helpers to align how we are getting user full names and initials from the Hasura response or Local Storage and added some tests.

Let me know if some of these changes are not a good path forward. I think the two helpers cover our bases but I could also be looking at it naively. 🙏 

## Testing
**URL to test:** 
https://deploy-preview-721--atd-moped-main.netlify.app/moped/projects/1859?tab=summary

**Steps to test:**

_The user in these steps is actually a soft deleted user account record with first name of **Inactive** and last name **User**. I just realized this is a little confusing since we have fallback text that says **Unknown User**_

1. Go to the above link and look at the **Team** tab
2. You will see me and "Inactive User" in the table - "Inactive User" is soft deleted and should be italicized and have " - Inactive" on the end
3. Make sure that you cannot select "Inactive User" from the dropdown you use to add a new team member
4. Go to the activity log and you should see a change row where "Inactive User" was added to the team
5. Check the following locations to make sure that user initials still show as expected:
   - Avatar in top right of app (if you don't have an uploaded profile image)
   - Avatar on the account page (if you don't have an uploaded profile image)

There is still one bug where **Unknown User** shows in the **User** column of the **Activity feed** which will be fixed in https://github.com/cityofaustin/atd-data-tech/issues/9423

Screenshot of an inactive user in the Team table
Screen Shot 2022-07-18 at 10 12 21 AM

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/test/userNames.test.js
+++ b/moped-editor/src/utils/test/userNames.test.js
@@ -24,23 +24,51 @@ const invalidUserObject = {
 describe("getFullUserName()", () => {
   it("returns a full name string from a user object", () => {
     const fullNameFromHasura = getUserFullName(userObjectFromHasura);
-    const fullNameFromLocalStorage = getUserFullName(atdMopedUserDbData);
-    const fullNameFromInvalidUserObject = getUserFullName(invalidUserObject);
 
     expect(fullNameFromHasura).toBe("Fozzie Bear");
+  });
+
+  it("returns a full name string from a Local Storage object", () => {
+    const fullNameFromLocalStorage = getUserFullName(atdMopedUserDbData);
+
     expect(fullNameFromLocalStorage).toBe("Bunsen Honeydew");
+  });
+
+  it("returns a fallback string from an invalid object", () => {
+    const fullNameFromInvalidUserObject = getUserFullName(invalidUserObject);
+
     expect(fullNameFromInvalidUserObject).toBe(unknownUserNameValue);
+  });
+
+  it("returns a fallback string from null", () => {
+    const fullNameFromNull = getUserFullName(null);
+
+    expect(fullNameFromNull).toBe(unknownUserNameValue);
   });
 });
 
 describe("getInitials()", () => {
   it("returns initials from a user object", () => {
     const initialsFromHasura = getInitials(userObjectFromHasura);
-    const initialsFromLocalStorage = getInitials(atdMopedUserDbData);
-    const initialsFromInvalidUserObject = getInitials(invalidUserObject);
 
     expect(initialsFromHasura).toBe("FB");
+  });
+
+  it("returns initials from a user object", () => {
+    const initialsFromLocalStorage = getInitials(atdMopedUserDbData);
+
     expect(initialsFromLocalStorage).toBe("BH");
+  });
+
+  it("returns initials from a user object", () => {
+    const initialsFromInvalidUserObject = getInitials(invalidUserObject);
+
     expect(initialsFromInvalidUserObject).toBe(null);
+  });
+
+  it("returns initials from a user object", () => {
+    const initialsFromNull = getInitials(null);
+
+    expect(initialsFromNull).toBe(null);
   });
 });
--- a/moped-editor/src/utils/test/userNames.test.js
+++ b/moped-editor/src/utils/test/userNames.test.js
@@ -1,0 +1,46 @@
+import {
+  unknownUserNameValue,
+  getUserFullName,
+  getInitials,
+} from "../userNames";
+
+// Some places where user names are stored
+const userObjectFromHasura = {
+  first_name: "Fozzie",
+  last_name: "Bear",
+};
+
+// This comes from Local Storage
+const atdMopedUserDbData = {
+  first_name: "Bunsen",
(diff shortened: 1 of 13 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -104,43 +104,6 @@ const ProjectActivityLog = () => {
...
-    return name
-      .replace(/[^A-Za-z0-9À-ÿ ]/gi, "")
-      .replace(/ +/gi, " ")
-      .split(/ /)
-      .reduce((acc, item) => acc + item[0], "")
-      .concat(name.substr(1))
-      .concat(name)
-      .substr(0, 2)
-      .toUpperCase();
-  };
```
mddilley, line 142: I refactored this so that we go `moped_user` ▶️ initials instead of `moped_user` ▶️ full name ▶️ initials. It is moved to `userNames.js` further up in the diff. Maybe there is a better names for that file?

moped-editor/src/utils/userNames.js
```diff
@@ -0,0 +1,30 @@
+// Fallback value for user without name set
+export const unknownUserNameValue = "Unknown User";
```
mddilley, line 2: Not sure this will ever really happen since we validate first name and last name are present on user creation but it could be good to be ready for abnormal data if it were to ever happen.

frankhereford (approved):
Thanks for the note saying the user account being demo'd in your test was named "Inactive User"! That was puzzling for a second.

I think this looks great, thank you @mddilley!

roseeichelmann (approved):
Tried this and it all looks good on my end!

moped-editor/src/utils/test/userNames.test.js
```diff
@@ -0,0 +1,86 @@
...
+// Some places where user names are stored
+const userObjectFromHasura = {
+  first_name: "Fozzie",
+  last_name: "Bear",
+};
+
+// This comes from Local Storage
+const atdMopedUserDbData = {
+  first_name: "Bunsen",
+  last_name: "Honeydew",
```
chiaberry, line 16: ❤️

chiaberry (approved):

mddilley:
@amenity @patrickm02L I have a few approvals here so this should be ready for PM review when there is time. Happy to update how the inactive user rows look in the Team table. That can be expanded beyond the italicized name style, but I wanted to start simple. 🙏

patrickm02L:
**Cannot delete users from the Project Team**

I tried inactivating a user in order to see if they would show up as inactive on Project Team, they did - I used the Test Testing user to do this. 

Then I reactivated Test Testing and issued a new password in order to see if that changed them back to active on the Project Team, it did. 

But now I can't delete Test Testing from the Project Team. I added Rebecca to see if the problem with deleting was specific to Test Testing because I inactivate/activated them, but I was unable to delete her from the Team either. 


https://user-images.githubusercontent.com/5566727/180237194-c0234610-7263-4014-b72a-a6e1dd679d92.mp4

amenity:
Nice catch, @patrickm02L! Since I'm out until Tuesday, I'll leave it to you to ✔️ **Product manager approved**. 🙏🏻

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -73,34 +77,51 @@ const ProjectTeamTable = ({ projectId }) => {
...
-    }
-
-    return null; // No need to return anything...
-  });
+  const makeListOfActivePersonnel = (personnelArray) => {
+    // 1. Multiple roles per user comes from multiple rows in the proj_personnel table
+    // so we have to dedupe project personnel and aggregate the roles to appear
+    // as multiple roles per personnel row in the UI
+    // 2. Similarly, personnel notes are concatenated into one string to show in the UI
+    // 3. Soft deleted personnel are filtered
```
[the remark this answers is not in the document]
mddilley: Filtering the soft deleted project personnel could happen in the query too so let me know if anyone has a preference. 🙏

mddilley:
@patrickm02L I fixed up the bug that you found. Thanks again for uncovering it! 🙏 🐛 🕵️  We should be able to add/remove project personnel and see an accurate list in the table.

patrickm02L (approved):
Works like a dream! 🤘 

https://user-images.githubusercontent.com/5566727/180326615-1626e5ee-9d82-421e-b281-ecd6deb8776c.mp4

patrickm02L:
I also checked out this the following steps from your PR instructions, and everything worked as expected. 

> 3. Make sure that you cannot select "Inactive User" from the dropdown you use to add a new team member
> 4. Go to the activity log and you should see a change row where "Inactive User" was added to the team
> 5. Check the following locations to make sure that user initials still show as expected

PR #726, opened 2022-07-21
Remove "Archived" row in `moped_status` table

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8238

This PR removes a `moped_status` row that was originally intended for project soft deletes. Those deletes now use `is_deleted` (see https://github.com/cityofaustin/atd-data-tech/issues/9300) so we no longer need this row.

## Testing
**URL to test:** 
https://deploy-preview-726--atd-moped-main.netlify.app/moped/projects/new

_The database migration can be tested locally_

**Steps to test:**
Devs - I checked the staging and production read replicas and seed data for any projects that have `status_id` of 6, but I'd like if someone else can back me up that this update will not break any existing records.

1. Create a new project using a Timeline template to test that the change to the `ADD_PROJECT` mutation didn't break things

---
#### Ship list
- [x] Code reviewed 
- [ ] ~Product manager approved~

### DIFF
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -25,7 +25,6 @@ export const ADD_PROJECT = gql`
       }
       moped_project_types {
         project_type_id
-        status_id
       }
     }
   }
--- a/moped-database/migrations/1658432462800_delete_moped_status_6_archived/down.sql
+++ b/moped-database/migrations/1658432462800_delete_moped_status_6_archived/down.sql
@@ -1,0 +1,2 @@
+INSERT INTO moped_status(status_name, status_id, status_flag, status_priority, status_description, status_order) 
+VALUES ("Archived", 6, NULL, NULL, "Project is deleted", 6)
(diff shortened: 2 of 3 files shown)


moped-editor/src/queries/project.js
```diff
@@ -25,7 +25,6 @@ export const ADD_PROJECT = gql`
       }
       moped_project_types {
         project_type_id
-        status_id
```
mddilley, line 28: I missed removing this when updating `moped_project_types` with `is_deleted`

chiaberry (approved):

moped-editor/src/queries/project.js
frankhereford: 👍 , but just checking that I'm not missing something -- this is a tag-along change that could have happened in another PR addressing another issue? No worries if so, just checking my understanding.

frankhereford (approved):
To your request for a second pair of eyes on the production DB: I just peeked, and there currently is a record with `status_id = 6` in production and in the staging databases. They both match the `down.sql` content as well. ✅

This code has been reviewed, and the database checked to spin-up locally. 🚢 🚢

moped-editor/src/queries/project.js
mddilley: That is totally right. ✅  I should have linked the issue and PR - this change should have happened for https://github.com/cityofaustin/atd-data-tech/issues/9250 and merged in with https://github.com/cityofaustin/atd-moped/pull/660

PR #728, opened 2022-07-24
Add/remove Moped test activity log

mddilley:
Sibling to https://github.com/cityofaustin/atd-prefect/pull/22


moped-etl/prefect/tasks/api.py
```diff
@@ -99,7 +101,7 @@ def create_zappa_config(basename, config_secret_arn):
                 "AWS_COGNITO_DYNAMO_SECRET_NAME": AWS_STAGING_DYNAMO_DB_ENCRYPT_KEY_SECRET_NAME,
                 # Look at Moped API events.py to see how this key is used
                 "MOPED_API_HASURA_APIKEY": MOPED_API_HASURA_APIKEY,
-                "MOPED_API_HASURA_SQS_URL": MOPED_API_HASURA_SQS_URL,
+                "MOPED_API_HASURA_SQS_URL": create_activity_log_queue_url(basename),
```
mddilley, line 104: We can get away with creating the SQS enpoint from the `basename` so it doesn't rely on the activity log deployment.

moped-etl/prefect/tasks/ecs.py
```diff
@@ -328,6 +330,9 @@ def create_task_definition(basename, database):
                         "value": HASURA_GRAPHQL_DATABASE_URL,
                     },
                     {"name": "HASURA_ADMIN_SECRET", "value": HASURA_ADMIN_SECRET},
+                    #  This depends on the Moped API endpoint returned from API commission tasks, add /events/ to end
+                    # {"name": "MOPED_API_EVENTS_URL", "value": MOPED_API_EVENTS_URL},
+                    {"name": "MOPED_API_APIKEY", "value": MOPED_API_APIKEY},
```
mddilley, line 335: We can tie this up when we start stitching things together - it is getting close!

moped-etl/prefect/tasks/activity_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+def create_activity_log_lambda_config(
+    basename,
+):
+    return {
+        "Description": f"AWS Moped Data Event: atd-moped-events-activity_log_{basename}",
+        "Environment": {
+            "Variables": {
+                # We could probably create a helper so this and the ECS Route53 CNAME always match
+                "HASURA_ENDPOINT": f"https://{basename}-graphql.moped-test.austinmobility.io",
```
mddilley, line 37: I added this comment so we could align the endpoints generated in the ECS and activity log modules. Not super important.

.github/workflows/aws-moped-sqs-helper.sh
```diff
@@ -178,3 +178,59 @@ function deploy_event_function {
   cd $MAIN_DIR;
   echo "Exit, current path: ${PWD}";
 }
+
+#
+# Builds & Deploys Event Function for Moped Test Instances
```
mddilley, line 183: I went too far down the road of recycling this bash script, and I ended up having to add functions and make some modifications. I could have used boto3 here, but I'm not sure what would have been faster. 🤷

frankhereford: I'd say whatever works is a great solution. I think of a prefect task having a single job to do, and as long as it works, no worries on what gets it done. 👍 Tasks are like the API of getting-stuff-done.

moped-etl/prefect/tasks/api.py
```diff
@@ -140,8 +142,8 @@ def create_moped_api_deploy_command(basename, config_secret_arn):
 
     # zappa deploy requires an active virtual environment
     # then use a subshell to zappa deploy in moped-api project folder
-    command = f"""python3 -m venv venv;
-    . venv/bin/activate;
+    command = f"""python3 -m virtualenv venv;
+    source venv/bin/activate;
```
frankhereford, line 146: I don't see any harm in it, but I wonder why this needs to be executed in a subshell. This question should be taken in the context that I've never used subshell notation in bash scripts, but I don't see what environment changes or what-have-you need to be protected from leaking out of this part of the task. 

But anyway - all good! Just wanted to flag it for conversation or future learning on my part. 👍

frankhereford (approved):
This is awesome - again - can't wait to stitch this new work together! Thanks @mddilley!!

moped-etl/prefect/tasks/api.py
mddilley: I'm glad that you checked in on this. This was born out of a Stack Overflow answer and looking back on it - there isn't a good reason for it. I'll update and test when I do the same for https://github.com/cityofaustin/atd-moped/pull/730#discussion_r929000845

PR #730, opened 2022-07-25
Populate moped test database with either staging or prod data

mddilley:
Sibling to https://github.com/cityofaustin/atd-prefect/pull/24

### DIFF
--- a/moped-etl/prefect/tasks/database.py
+++ b/moped-etl/prefect/tasks/database.py
@@ -92,11 +92,11 @@ def remove_database(basename):
 
 
 populate_database_with_data_task = ShellTask(
-    name="Run populate DB with data bash command", stream_output=True, return_all=True
+    name="Run populate DB with data bash command", stream_output=True
 )
 
 
-@task
+@task(name="Create populate database with data bash command")
 def populate_database_with_data_command(basename, stage="staging"):
     logger.info(f"Creating populate with {stage} data command for database {basename}")
 
@@ -116,9 +116,8 @@ def populate_database_with_data_command(basename, stage="staging"):
     )
 
     # Set up for the commands
-    version = "12-alpine"
+    postgres_version = "12-alpine"
     dump_filename = f"moped_{stage}.sql"
-    print(dump_filename)
 
     dump_command = f"pg_dump -d {dump_conn_string} \
     --no-owner --no-privileges --verbose > {dump_filename}"
@@ -127,7 +126,7 @@ def populate_database_with_data_command(basename, stage="staging"):
     -f /tmp/{dump_filename}"
 
     command = f"""
-        docker pull postgres:{version} &&
+        docker pull postgres:{postgres_version} &&
         cd ~/ &&
         cd ../tmp &&
         docker run --rm postgres {dump_command} &&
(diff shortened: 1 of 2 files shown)


moped-etl/prefect/tasks/database.py
```diff
@@ -84,9 +91,46 @@ def remove_database(basename):
...
+    psql_command = f"psql {psql_conn_string} \
+    -f /tmp/{dump_filename}"
+
+    command = f"""
+        docker pull postgres:{postgres_version} &&
+        cd ~/ &&
+        cd ../tmp &&
+        docker run --rm postgres {dump_command} &&
+        docker run --rm -v /tmp:/tmp postgres {psql_command}
+    """
```
mddilley, line 134: @frankhereford Curious about what you think about these commands - it felt weird `cd`'ing out of the working directory, but this is the only way that I could get it to work. 🐳

frankhereford, line 134: @mddilley - great question. This works, but I think you could even maybe keep the database dump off the disk entirely. I would maybe try something like this:

```
    dump_command = f"pg_dump -d {dump_conn_string} --no-owner --no-privileges --verbose"

    psql_command = f"psql {psql_conn_string}"

    command = f"""
docker pull postgres:{postgres_version}  &&
docker run --rm postgres {dump_command} | docker run --rm postgres {psql_command}
"""
```

This would be plugging STDIN of the `psql` command into the STDOUT of the `pg_dump` command. Plumbing with pipes :)

moped-etl/prefect/tasks/database.py
```diff
@@ -1,13 +1,20 @@
...
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
 DATABASE_HOST = os.environ["MOPED_TEST_HOSTNAME"]
 DATABASE_USER = os.environ["MOPED_TEST_USER"]
 DATABASE_PASSWORD = os.environ["MOPED_TEST_PASSWORD"]
+MOPED_READ_REPLICA_HOST = os.environ["MOPED_READ_REPLICA_HOST"]
+MOPED_READ_REPLICA_USER = os.environ["MOPED_READ_REPLICA_USER"]
+MOPED_READ_REPLICA_PASSWORD = os.environ["MOPED_READ_REPLICA_PASSWORD"]
+
+MOPED_READ_REPLICA_HOST
```
frankhereford, line 17: Straggler here that I think can get the boot.

frankhereford (approved):
This looks awesome! I can't wait to connect it up to the existing work, and we'll be looking at a deployed Moped branch real soon.

moped-etl/prefect/tasks/database.py
frankhereford: I added a `-t` to the commands above. There needs to be terminal support for those pipes to work, _I think_.

mddilley: Thanks! I knew there had to be another way! I'll make this update and test with the flag while also looking at https://github.com/cityofaustin/atd-moped/pull/728#discussion_r929018815

moped-etl/prefect/tasks/database.py
mddilley: Yes, this definitely needs to go 🥾

moped-etl/prefect/tasks/database.py
mddilley: Got this working so now there is no write to disk 🔧 🙏

PR #731, opened 2022-07-25
Add new components to moped-components table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9795

## Testing
**URL to test:** 
Test locally or deploy to moped-test

**Steps to test:**

Screen Shot 2022-07-25 at 3 19 13 PM

In the dropdown, check to see if you can now see Lighting > Street lighting and Lighting > Intersection lighting. Also under Transit > Bus stop. 

Project extent - generic should now be drawn as lines. 

I identified a bug that currently exists in production (https://github.com/cityofaustin/atd-data-tech/issues/9841) where points vs line draw tools are being set by the component type and not subtype. 

---
#### Ship list
- [x] Code reviewed 


mddilley (approved):
Tested this locally, and I see all of the new components types. 👍 💡  🚌

mddilley:
@chiaberry I just caught the note about the **Project extent - generic** now being drawn as lines. Is that a change that happened in this PR?

moped-database/migrations/1658526083849_insert_into_public_moped_components/up.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO "public"."moped_components"(
+	"component_name", "status_id", "component_subtype", "line_representation") 
+VALUES 
+	(E'Lighting', 1, E'Street lighting', true),
+	(E'Lighting', 1, E'Intersection lighting', false),
+	(E'Transit', 1, E'Bus stop', false);
```
frankhereford, line 6: I always have to remind myself what Postgres Escape Notation `E''` does when I see it. It causes it to respect the usual suspects like `\n`, `\t` to get interpolated into the actual newline or tab character.

moped-database/migrations/1658526083849_insert_into_public_moped_components/up.sql
```diff
@@ -0,0 +1,10 @@
...
+
+UPDATE "public"."moped_components"
+SET line_representation = true
+WHERE component_id = 0;
```
johnclary, line 10: @mddilley here's where the Project Extent component is changing to line representation.

I'm pretty sure Chia checked, but I'll double check to see if we have any existing project extent components stored as points in the DB. We'd need to convert those to lines

amenity (changes requested):
"Intersection lighting" should be a point, so give the user the ability to select an existing point or add a new one.

johnclary:
Looks like we currently have two projects in production using the **Project extent** component as points.

- [#246](https://mobility.austin.gov/moped/projects/246) - Looks like Fernando's project. Probably best to have them delete and re-add the extent component after we make this change
- [#159](https://mobility.austin.gov/moped/projects/159)  - This project was created and cancelled by Amenity - don't need to worry about it

chiaberry:
@amenity The bug where the drawing type is dictated by the type and not the subtype is fixed in this PR https://github.com/cityofaustin/atd-moped/pull/732. Because Street Lighting is drawn by lines, it overrides Intersection Lighting. The same thing happens right now in production for Highway type.

amenity (approved):
Now that I understand [PR #732](https://github.com/cityofaustin/atd-moped/pull/732) (thanks, @chiaberry !) I'm good to go on this.

PR #732, opened 2022-07-26
Fix point vs line subtype drawing bug 

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9841

In prod right now, the way we decide if the drawing tool should be line or point is we have an array of the component types that are represented by lines and then compare the type selected to that array. The problem is that Highway is usually line except for the Access Ramp subtype, which is a point. However when you try to draw the access ramp, it still forces lines. 

So instead of an array, this PR makes an object with the keys being component type + component subtype, and when a type or subtype is selected, we lookup on the object. 

```
	"guardrail":true,
	"highwayaccessramp":false,
	"highwayaddedcapacity/lanes":true,
	"highwaycollectordistributor":true,
	"highwayflyover":true,
	"highwayintersectiongradeseparation":true,
	"highwaymanagedlane":true,
	"highwaytollroad":true,
	"landscaping":true,
	"placemaking":true,
	"refugeislandbike":false,
	"refugeislandped":false,
	"refugeislandbike/ped":false,
```

I investigated just accessing the attribute directly and not having to compare to a list, but the way we it set up now is selecting from a string list of options in the Autocomplete dropdown and not having the whole object for the type accessible.  I'm hesitant to work on a big refactor here since we will be updating our mapping soon anyway (https://github.com/cityofaustin/atd-data-tech/issues/9675) and I dont know what changes will be needed there. 

## Testing
**URL to test:** 
https://deploy-preview-732--atd-moped-main.netlify.app/moped/projects/1862?tab=map

**Steps to test:**
Add a new component to the map. 
Select Highway as your type. 
Select some subtypes, and see that they are lines except access ramp. 

Try around other types and subtypes toggling back and forth. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/.github/workflows/aws-heroku-helper.sh
+++ b/.github/workflows/aws-heroku-helper.sh
@@ -115,7 +115,7 @@ function hasura_server_ready() {
 }
 
 function wait_server_ready() {
-  INITIALIZATION_TIMEOUT_LIMIT=300
+  INITIALIZATION_TIMEOUT_LIMIT=60
   INITIALIZATION_COUNT=1
 
   while [ "$(hasura_server_ready)" = "FALSE" ]; do
--- a/moped-database/hasura-cluster
+++ b/moped-database/hasura-cluster
@@ -293,7 +293,7 @@ function __hasura_server_ready() {
 # if not, it waits 1 second and tries again (up to 60 seconds)
 #
 function __wait_server_ready() {
-  INITIALIZATION_TIMEOUT_LIMIT=120
+  INITIALIZATION_TIMEOUT_LIMIT=60
   INITIALIZATION_COUNT=1
   while [ "$(__hasura_server_ready)" = "FALSE" ]; do
     echo -ne "Hasura server not ready, please wait. ${INITIALIZATION_COUNT}/${INITIALIZATION_TIMEOUT_LIMIT} 🕔\r";


moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -722,7 +723,7 @@ const ProjectComponentsMapEdit = ({
                         id="moped-component-description"
                         label="Description"
                         multiline
-                        rows={4}
+                        minRows={4}
```
chiaberry, line 726: fixing a warning/error in the console about deprecated props

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -257,9 +256,11 @@ const ProjectComponentsMapEdit = ({
    * @param {String} newValue - The new value from the autocomplete selector
    */
   const handleComponentSubtypeSelect = (e, newValue) => {
-    setSelectedComponentSubtype(
-      (newValue ?? e.target.value ?? "").toLowerCase()
```
chiaberry, line 261: I removed `e.target.value` because our `<li>` does not have a value attribute and e.target.value is `0`. If there wasn't a newValue, we would be doing `0.toLowerCase()` which results in an error.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -257,9 +256,11 @@ const ProjectComponentsMapEdit = ({
...
-    );
+    const componentName =
+      selectedComponentType.toLowerCase().replaceAll(" ", "") +
+      newValue.toLowerCase().replaceAll(" ", "");
```
frankhereford, line 261: Thank you for explaining to me the reasoning behind doing this comparison using the combined names as a key to lookup instead of pushing this back to the database to resolve.

frankhereford (approved):
Thank you for finding and fixing this -- and for squashing a console error too! 🚀 

I've looked at the code and tested this locally and it works just as you described.

moped-editor/src/utils/projectComponentHelpers.js
```diff
@@ -48,15 +48,13 @@ export const useLineRepresentable = mopedComponents => {
...
-            ? moped_component.component_name.toLowerCase()
-            : null
-        )
-      ),
-    ].filter(item => item);
+    const lineRepresentableNew = {}
+    mopedComponents.forEach(moped_component => {
+      const componentNameKey =
+        moped_component.component_name.toLowerCase().replaceAll(" ", "") +
+        (moped_component.component_subtype ?? "").toLowerCase().replaceAll(" ", "");
```
mddilley, line 55: This feels like a good place to create a helper that describes how this translation works to help ID components/subtypes, but I think that you are right to hold off on too much work here before potential refactoring. 👍

mddilley (approved):
Works as described, and TIL how this part of the app works. 🙏

chiaberry:
After merging it's parent branch to main, this PR can now be tested on netlify: https://deploy-preview-732--atd-moped-main.netlify.app/moped/projects/1862?tab=map @patrickm02L

patrickm02L (approved):
It's good to go! 🤘

https://user-images.githubusercontent.com/5566727/183949676-d9988a4a-90fa-4255-877f-7e47ee3e726e.mp4

PR #736, opened 2022-07-28
Reset Hasura metadata for graphql-engine v2

johnclary:
## Associated issues
This PR resets our Hasura metadata `tables.yaml` to match Hasura's ordering in graphql-engine v2.

This change is the result of running `hasura metadata export` against the DB.

This PR also resolves https://github.com/cityofaustin/atd-data-tech/issues/9643 by commiting `network.yaml` to version control.

## Testing
**URL to test:** local or https://moped-test.austinmobility.io/moped
**Steps to test:**
This is tech debt only. It would be good to kick the tires on basic Moped CRUD operations to make sure nothing looks amiss.

---
#### Ship list
- [ ] Code reviewed 


johnclary:
Do y'all think I should try to write some Python to validate that the tables.yaml diff is purely cosmetic? It's a lil scary to merge this carte blanche.

frankhereford:
@johnclary wrote:

> Do y'all think I should try to write some Python to validate that the tables.yaml diff is purely cosmetic? It's a lil scary to merge this carte blanche.

Really good question. I'm comfortable if there the file's line count didn't change or changed in an expected way.

I looked, and it looks like the `tables.yaml` file is going from 3965 lines to 3907 lines. Does that sound right to you? 

A script would be neat, but I wouldn't stress it, provided you're comfortable with the above length change. If you're able to spin up the database locally and have the site load and navigate around, then I'd feel OK.

johnclary:
thanks @frankhereford.

I found [this tool](https://www.yamldiff.com/) which does a great job.

Hasura snuck in a few changes:
- removed `backend_only: false` from a few permissions
- added `insertion_order: null` from a few permissions

I'll take a look at the docs to understand this better.

mateoclarke (approved):
I'm good with merging this without over analyzing the diff. I trust Hasura and if there are any hangups, its just config as yml so it would be super easy to rollback. More-so than unexpected DB schema changes.

johnclary:
[Docs](https://hasura.io/docs/latest/auth/authorization/permission-rules/#backend-only-permissions) about `backend_only`. The default is `false`. I can't tell when/why Hasura decides to add/remove this to the metadata. Support for it recently expanded to updated and deletes in v 2.8. That may have something to do with it.

[Docs](https://hasura.io/docs/latest/api-reference/syntax-defs/#objrelusingmanualmapping) about `insertion_order`. It's `null` everywhere, which means we're using the default behavior. Again, I can't tell when Hasura decides to add this.

TLDR: these changes will have no affect on our current Moped behavior.

frankhereford (approved):
leslie-knope-parks

This is awesome. Thanks for taking care of this - the increased readability is huge and not having to step around the network.yaml file is too.

mddilley (approved):
Tested this locally and did a lot of CRUDing wherever I could in the app, and it held up just fine. Thanks for including that pesky `network.yaml` in this PR! 😄

johnclary:
thanks y'all! merging into Chia's branch

PR #737, opened 2022-07-28
Adds hash links and related phase column to lookups view

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9871

## Testing
**URL to test:** local or moped-test

**Steps to test:**
1. Navigate to `/moped/dev/lookups/` 
2. Click on the link icon next to the **Moped phases** table
3. Observe that the table may scroll slightly into view, and the URL in your browser updates with the table's hash (`#moped-phases`)
4. Do the same thing with the **Moped milestones** table
5. Confirm that the **Related phase** column renders on the milestones table
6. Copy the full url with `#moped-milestones` into a new tab and confirm that the page scrolls to that table.

Screen Shot 2022-07-29 at 6 13 33 PM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1659028303595_set_fk_public_moped_milestones_related_phase_id/up.sql
+++ b/moped-database/migrations/1659028303595_set_fk_public_moped_milestones_related_phase_id/up.sql
@@ -1,0 +1,5 @@
+alter table "public"."moped_milestones"
+  add constraint "moped_milestones_related_phase_id_fkey"
+  foreign key ("related_phase_id")
+  references "public"."moped_phases"
+  ("phase_id") on update cascade on delete set null;
--- a/moped-editor/src/queries/timelineLookups.js
+++ b/moped-editor/src/queries/timelineLookups.js
@@ -9,7 +9,6 @@ export const TIMELINE_LOOKUPS_QUERY = gql`
       phase_order
       moped_subphases(order_by: { subphase_order: asc }) {
         subphase_id
-        related_phase_id
         subphase_description
         subphase_name
         subphase_order
@@ -20,7 +19,9 @@ export const TIMELINE_LOOKUPS_QUERY = gql`
       milestone_name
       milestone_description
       milestone_order
-      related_phase_id
+      moped_phase {
+        phase_name
+      }
     }
   }
 `;
--- a/moped-editor/src/views/dev/LookupsView/index.js
+++ b/moped-editor/src/views/dev/LookupsView/index.js
@@ -3,7 +3,6 @@ import ApolloErrorHandler from "src/components/ApolloErrorHandler";
 import { useQuery } from "@apollo/client";
 import { useLocation, Link } from "react-router-dom";
 import { createBrowserHistory } from "history";
-import CircularProgress from "@material-ui/core/CircularProgress";
 import Container from "@material-ui/core/Container";
 import Grid from "@material-ui/core/Grid";
 import Paper from "@material-ui/core/Paper";
@@ -62,16 +61,17 @@ const LookupsView = () => {
   );
 
   /**
-   * Use the record hash from the URL, if present. This only happens on page load.
+   * Use the record hash from the URL, if present. This only happens once after
+   * data fetch.
    * */
   const { hash: recordKeyHash } = useLocation();
   useEffect(() => {
-    if (!recordKeyHash) {
+    if (!recordKeyHash || loading) {
       return;
     }
     const recordKey = recordKeyHash.replace("#", "").replace("-", "_");
     setSelectedRecordKey(recordKey);
-  }, [recordKeyHash]);
+  }, [recordKeyHash, loading]);
 
   /**
    * Listens for changes to the selected record key and scrolls to its table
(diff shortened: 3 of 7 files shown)


moped-editor/src/queries/timelineLookups.js
```diff
@@ -9,7 +9,6 @@ export const TIMELINE_LOOKUPS_QUERY = gql`
       phase_order
       moped_subphases(order_by: { subphase_order: asc }) {
         subphase_id
-        related_phase_id
```
johnclary, line 12: we weren't using this column

moped-database/migrations/1659028303595_set_fk_public_moped_milestones_related_phase_id/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."moped_milestones"
+  add constraint "moped_milestones_related_phase_id_fkey"
+  foreign key ("related_phase_id")
+  references "public"."moped_phases"
+  ("phase_id") on update cascade on delete set null;
```
johnclary, line 5: the foreign key existed, we just weren't referencing it as a proper constraint.

moped-editor/src/views/dev/LookupsView/index.js
```diff
@@ -31,6 +40,52 @@ const LookupsView = () => {
...
+  /**
+   * Listens for changes to the selected record key and scrolls to its table
+   */
+  useEffect(() => {
+    if (!selectedRecordKey) {
+      return;
+    }
+    const ref = refs?.[selectedRecordKey];
+    if (ref?.current) {
+      ref.current.scrollIntoView({ behavior: "smooth" });
```
johnclary, line 85: smooth scrolling isn't everyone's cup of tea—happy to change it to `auto`.

mateoclarke:
I'm not able to visit the page in the Preview App. Could it be some issue with the routing? I can test on local but I figured it was worth flagging potential issues in other environments:

https://deploy-preview-737--atd-moped-main.netlify.app/moped/dev/lookups/

moped-editor/src/views/dev/LookupsView/index.js
mateoclarke: i dig it. nice touch 💅

mateoclarke (approved):
🚢💨✨

moped-editor/src/views/dev/LookupsView/index.js
mddilley, line 85: I noticed that, if I paste `https://localhost:3000/moped/dev/lookups#moped-milestones` into a new tab, the scroll to the milestones section doesn't happen as expected.

I added `console.log("here is that ref", ref, ref?.current);` between lines 83 & 84 and ref.current remains `null` so the scroll never fires.

I think it might be the fact that [updates to `ref.current` don't trigger a re-render](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/), but I also noticed that removing the dependency array on line 61 from the `useMemo` fixes this (even though that doesn't seem like the right fix). I'm going to think about it some more.

mddilley (commented):
Step 6 didn't quite work as I expected - I opened a fresh tab and didn't see scroll to the hash in the URL but maybe I'm not testing the right way. You were totally right about how this is harder than it should be in React btw. 👍

moped-editor/src/views/dev/LookupsView/index.js
johnclary: ah, fascinating. now i'm seeing a probable race condition b/t when we set ` selectedRecordKey` from the URL and when those refs are hydrated with `.current` after rendering. 

the underlying problem, i think, is that we should be using the `useLocation()` hook to have a stateful reference to the current browser path, but things get really wonky when you start using  the react-router method (`navigate` irrc) to push to history. 

anyway, looking forward to getting back to this later today. thanks Mike!

johnclary: ok, i _think_ what's happening is that first effect hook fires on first mount, but the `loading` is blocking the ref'd table components from rendering. so hence the `null` value in `ref.current`. i'm not sure why this was working locally for me—`loading` is always `true` on the first render.

in any case, i moved the loading spinner down into the record tables, and the ref'd tabled components should always render on the first mount.

i'm open to other ideas!

johnclary: oh, i also added `loading` as a dependency to that first hook. i haven't seen it happen, but page might not scroll far enough if it's scrolling to empty tables. better to wait until we have data to scroll.

mddilley: This makes sense to me - thanks for the explanation! 🙏

mddilley (approved):
Works great! 🔖 🛵

frankhereford (approved):
Right on target! This will be very helpful.

Thank you John! This was tested locally and read through your code - it's interesting to follow the discussion regarding the auto-scrolling. So much to learn for me!

johnclary:
I'm going to merge this — @amenity happy to re-work when you have a chance to give feedback.

PR #740, opened 2022-07-29
Make email and name validation closer to O365

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9731

## Testing
**URL to test:** 

**Steps to test:**
TBD Need to test with moped-test first

---
#### Ship list
- [ ] Code reviewed 


mateoclarke (approved):
🚢 💨✨

johnclary:
👍 @mddilley works as promised. thanks for fixing this up!

not saying we should change it, but i'm curious to understand why we're validating first/last name. would something bad happen if we accepted any string?

johnclary (approved):

mddilley:
@johnclary That is definitely worth thinking about IMO. Overall, I see the validation as a good way to protect the database (string length, etc.) and make sure that data is complete with required fields, but I don't see a huge need for the regex on first & last names since only admins can create/edit user accounts. 

Plus, this PR proves that the first and last name validation can create a hurdle to onboarding a valid user. Maybe it is worth a backlog issue to make a pass through the schema? Looks like workgroup also has a regex and that comes straight from the DB user table... speaking of, that could use some normalization since we store both `workgroup` and `workgroup_id` there.

patrickm02L (approved):
It works g'reat! 🤘

https://user-images.githubusercontent.com/5566727/182479353-71685789-9d56-4646-a880-2aadc31ae21b.mp4

frankhereford (approved):
https://xkcd.com/208/

😄

chiaberry (approved):

PR #743, opened 2022-08-09
Integrate flows to be able to spin up a branch

frankhereford:
## Test Tooling early-demo

Sibling PR to https://github.com/cityofaustin/atd-prefect/pull/30.

This PR is now in a state where it can be given a branch name and it will spin up a moped of it. It's fit for use "by hand" and has not been registered into the Prefect cloud yet. 

## Next Steps
@mddilley I wanted to button up this PR before I start on the next few things that we may put next on the docket

Thinking of what could be next, I'm interested in:
* Seeing if we can carve out some more space in the 64 character limited lambda names for the branch name
* Combining our independent flows into a single flow for commissioning and decommissioning an instance
* A lot of testing - start up an issue or an epic to keep track of 🐞🐛🕷. Not sure what is best - maybe a question for product. @patrickm02L, what do you think?
* Get the new, combined flows registered and parameratized
* Improve or functionalize the branch name slugging system WTR to the PG requirements of not starting with numerals and lambda names prohibiting numerals outright
* Add more error detection as possible to tasks
* Look into threaded agent systems -- there are multiple long strings of tasks which do not depend on each other and I think we can cut the deploy time down to 25%


moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -239,55 +249,80 @@
...
-    # ecs_decommission.run(parameters=dict(basename=basename))
-    # print("\n🤖 Comissioning ECS\n")
-    # ecs_commission.run(parameters=dict(basename=basename, database=database))
+    metadata = "metadata"
 
-    # print("💡 Comissioning Netlify Build & Deploy\n")
-    # netlify_commission.run(parameters=dict(basename=basename))
+    config = ecs.create_graphql_engine_config_contents(
+        graphql_endpoint=graphql_endpoint, access_key=access_key, metadata=metadata
+    )
```
frankhereford, line 275: This section of the diff is very mangled - check the final file for something easier to grok: https://github.com/cityofaustin/atd-moped/blob/integrate-flows/moped-etl/prefect/moped_test_create_flow.py#L257

mddilley (approved):
This all looks great to me! I'll drop a comment to get back to you on the discussion points in the PR. 🚀

mddilley:
Thinking of what could be next, I'm interested in:

> Seeing if we can carve out some more space in the 64 character limited lambda names for the branch name
Combining our independent flows into a single flow for commissioning and decommissioning an instance

We can definitely shorten the Moped API Lambda name. I think that the `atd-moped-test-prefect` and `atd-moped-events-activity-log` prefixes for the Lambdas is overkill, and I'm all about shortening them. I think that seeing the branch name (or eventual slug) and maybe at least the word moped would make it clear what they are for.

> A lot of testing - start up an issue or an epic to keep track of 🐞🐛🕷.

I agree that a refinement/bug fix epic is in order. I think that is a great way to track what work is left.

> Improve or functionalize the branch name slugging system WTR to the PG requirements of not starting with numerals and lambda names prohibiting numerals outright

Agreed. It will be nice to find a way to name the resources in a unified way that works around things like the PG and Lambda 
naming limitations.

> Add more error detection as possible to tasks

Strongly agree here. It will be interesting to see what errors pop up, and I'm glad that Prefect gives us ways to address these right out of the box.

> Look into threaded agent systems -- there are multiple long strings of tasks which do not depend on each other and I think we can cut the deploy time down to 25%

Isn't it nice to be in control of the whole test system like this? 🎉 🚀 🚀 🚀 Really exciting to be able to make an improvement like that.

mddilley:
@frankhereford I also noticed that the activity log doesn't seem to be working. Looking in the AWS console, it seems like the two keys that the Moped API look for aren't the same in the ECS config and Moped API env vars. I also noticed that the activity log SQS shows the Lambda's status as `Disabled`.

You might be aware of this already, and I'm sure we can talk about it when we sync up.

frankhereford:
Hey Mike - just seeing this comment. I was not aware of this, and your comment makes a lot of sense in terms of possible causes. Let's have that overdue sync real soon and figure out what's the best way to get this fixed up. Thanks!

PR #745, opened 2022-08-10
Define ALB health check location and valid responses

frankhereford:
Sibling PR: https://github.com/cityofaustin/atd-prefect/pull/32

Small change:

* Configures the health check on the ALB's target group so it can accurately tell if a graphql-engine is up and running nominally. 

This prevents the ALB from flapping the ECS task instance.


mddilley (approved):
🚀

PR #746, opened 2022-08-10
Enable testing of `test` with other moped feature branches

frankhereford:
In advance of the upcoming test system, there are a set of 3 change which allow the test deployment to trigger a netlify build and to set environment variables. Of note, one is to allow the deployment to have the graphql-engine endpoint defined.

These changes are non-ops in our current environment, and they will merge in cleanly when we do get the test instance work into main. 

Being able to get these into main now, or at least in a branch i can merge into other branches that I want to use to test `test`, will be helpful.


frankhereford:
This is already in `main`! It went in with Chia's other branch I had pushed these same changes too. Thanks!

frankhereford:
I'm holding onto this branch, however, it's helpful when setting up feature branches for testing `test`.

PR #748, opened 2022-08-11
Fix status badge Active phase bug

chiaberry:
# Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9848

I tried to repeat the steps Meredith took and was unable to reproduce an "Active" badge by following the steps in the activity log for the project mentioned in the issue. However, there is a bug in production where if you change the status name of a non-current status, the badge reverts to "Active" instead of keeping the name of the current status. 

I moved around some of the logic so the badge name doesnt change unless the current status is being edited. 

## Testing
**URL to test:** 

https://deploy-preview-748--atd-moped-main.netlify.app/moped/projects/1866?tab=timeline

**Steps to test:**

1. Create a new project
2. navigate to Timeline Tab, Potential is the current status. Badge should say Potential. 
3. Toggle the Potential status current switch to false. Badge should be green and Active
4. Add a few more statuses. Mark one as Current. That status name should be on the badge. 
5. Change the name of an existing (non-current) status. Badge name should not change.
6. Change the name of the existing current status. Badge name should change to match the new name. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectTimeline.js
```diff
@@ -232,7 +232,7 @@ const ProjectTimeline = ({ refetch: refetchSummary }) => {
     <Grid component="label" container alignItems="center" spacing={1}>
       <Grid item>
         <Switch
-          checked={props.value}
+          checked={props.value ?? false}
```
chiaberry, line 235: fixing an error in the console

johnclary (approved):
works as promised! I think this code is easier to understand as well. thank you for sorting this out!

🚢 🚢 🚢

johnclary:
I do wonder if we should enforce that there's always a current phase associated with a project. So,

- Can't delete phase when there is only one phase
- Can never toggle off a current phase - can only mark another phase as current

It's just weird that we have this obscure "Active" status that has no meaning in terms of real life project statuses.

Totally out of scope; just wondering aloud....

patrickm02L (approved):
Couldn't find any bugs. Ship it! 🤘

https://user-images.githubusercontent.com/5566727/184663802-2722924c-c943-4776-baca-cf7707757e08.mp4

moped-editor/src/views/projects/projectView/ProjectTimeline.js
```diff
@@ -614,29 +616,33 @@ const ProjectTimeline = ({ refetch: refetchSummary }) => {
...
-                                        current_phase_id: 0,
-                                      },
-                                },
-                              })
-                            : true // No change in project, safely ignore
-                      )
+                      .then(() => {
+                        // if the phase being updated is toggled as the current phase
+                        // update moped_project with new current_phase, updating the status badge
+                        if (isCurrentPhase) {
```
mddilley, line 622: ❤️ the if/else blocks - I think this is a lot easier to read than the chained ternaries

mddilley (approved):
Tested on the deploy preview and all the test steps worked perfectly. 🙌

frankhereford (commented):
Thank you @chiaberry! I agree with the comments about the improved readability - super helpful!

PR #749, opened 2022-08-11
Adds three new milestones for TED / PDD

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/9973

## Testing
**URL to test:** 
needs to be tested locally or set up on a new test instance. 

Here's a test deployment: https://9973-add-three-new-milestones--atd-moped-main.netlify.app/moped

**Steps to test:**
go to a project's timeline, select add a new timeline and note the last three entries

---
#### Ship list
- [ ] Code reviewed 

Screen Shot 2022-08-11 at 5 58 58 PM


moped-database/migrations/1660258008607_insert_moped_milestones/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Delete added moped_milestone records
+DELETE FROM public.moped_milestones WHERE milestone_name = 'Draft Study Submitted';
+DELETE FROM public.moped_milestones WHERE milestone_name = 'Draft Preliminary Schematic Submitted';
+DELETE FROM public.moped_milestones WHERE milestone_name = 'Need to Assign Study';
```
johnclary, line 4: I tested the down migration, and if any of these milestones are in use on a project it will fail:

```sql
update or delete on table "moped_milestones" violates foreign key constraint "moped_project_milestone_milestone_id_fkey" on table "moped_proj_milestones". Key (milestone_id)=(54) is still referenced from table "moped_proj_milestones"
```

Question for the group—do we want to update the FK constraint on `moped_proj_milestones` to `ON UPDATE CASCADE` and `ON DELETE SET NULL` - ? I believe that's the best practice we've started to adopt elsewhere.


Screen Shot 2022-08-12 at 10 42 20 AM

johnclary (approved):
thanks @tillyw for jumping on this so quickly! looks great. i don't think the failing down migration is a deal-breaker, but would be nice to hear from others if we want to address the underlying problem now.

chiaberry (approved):

moped-database/migrations/1660258008607_insert_moped_milestones/down.sql
tillyw: thanks @johnclary for pointing that out! i tested the down migration but i didn't think to do it with one of the milestones in use. i have added the migration files that update the fk as you recommended

tillyw: @johnclary i updated the fk relationship for moped_proj_milestones and restarted the cluster, added one of the new milestones to a project and ran the down migration and it still seems to be failing

do you have any recommendations on what to do?

Screen Shot 2022-08-16 at 3 13 47 PM

moped-database/migrations/1660258008607_insert_moped_milestones/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Insert new milestone records
+INSERT INTO public.moped_milestones (
+        milestone_name,
+        milestone_description,
+        related_phase_id
+    ) VALUES
+        ('Draft Study Submitted', 'Draft report submitted to ATD and being reviewed', 4),
+        ('Draft Preliminary Schematic Submitted', 'Concept schematic submitted to ATD and being reviewed', 4),
+        ('Need to Assign Study', 'Identified priority location is ready for scoping study and preliminary schematic assignment', 4);
```
frankhereford, line 9: These appear in the drop-down - thanks @tillyw

frankhereford (approved):
Looks great - thank you!

moped-database/migrations/1660258008607_insert_moped_milestones/down.sql
johnclary: thanks for adding the constraints! i'll take a look 👀

chiaberry: How did you test the down migration @tillyw? Did you run both of them in order?

patrickm02L (approved):
Checks out! Ship it! 🤘

https://user-images.githubusercontent.com/5566727/184986616-a11946cf-46ac-41e1-8be8-e840035fe19c.mp4

PR #755, opened 2022-08-14
🌪 Unify flows, commission over existing deployment, slug cleanup, integrate prefect cloud

frankhereford:
## Related Items
Sibling PR: https://github.com/cityofaustin/atd-prefect/pull/33

## Issue progress
This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/9997. 
* 🤷🏻‍♂️ No `slugify` and still ugly, but I think this is as clean as we can ask for given the variety of naming constraints we're under. I want to call this ✅.

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/9994. 
* ✅

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/9996. 
* 🤷🏻‍♂️ I want to spin off some ideas from this into new issues, particularly around deployment. As it stands, Prefect cloud is very much part of this now.

## Reviewing
😢 This had to have `main` merged into it, which makes it harder to review. We can totally merge `main` into `md_moped_test_create_flow` to make most of this diff wash out. Feel free - no review needed!

I was developing on and deploying the `unify-flows` branch, which was based off our collector branch. That collector was a good bit behind main, which was making it impossible to migrate and update the metadata, as the staging database was ahead of what the code was at. This branch will bring the collector branch up to date or very close to it.

Using the GitHub file browser in the "Files Changed" tab is helping me look for changes. The nature of the changes to the main flow file still make the GitHub diff a disaster. For that file at least, I think it's easier to read the final state. 

## Features
This branch:

* [x] Unifies the work into two flows, an up and a down flow, which are parameterized, registered, and intended now to be controlled via the prefect cloud ui
* [x] This supports the concept of being able to commission a test instance **_over_** an existing test instance. There are a number of our AWS calls which do not support this natively, and have to have a task run to detect deployment status, remove that particular part of the deployment, and then allow it to be recommissioned. These constructs generally are formed with the `with case():` and `merge()` pattern.
* [x] This moves helper functions into a stand-along little library, simplifying their importing. It was too easy to have a circular import failure.
* [x] This adds numerous checks for various parts of the deployment being functional, not just "in flight" based on the return values from AWS
* [x] This moves the migration related tasks into their own subsystem file
* [x] This supports the Dask executor scheme with no modification; it's sooo much faster to deploy with it 
* [x] Lots of logging and lots of little fiddly things


moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -38,316 +44,304 @@
...
+    )  # remove leading numbers
+    internal_number_free_underscore_basename = "".join(
+        [i for i in database if not i.isdigit()]
+    )
+    awslambda = internal_number_free_underscore_basename[0:16]
 
-    basename = Parameter("basename")
-    database = Parameter("database")
-    api_endpoint = Parameter("api_endpoint")
+    slug = {"basename": basename, "database": database, "awslambda": awslambda}
```
mddilley, line 59: 👍

moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -38,316 +44,304 @@
...
-    removed_load_balancer = ecs.remove_load_balancer(
-        basename=basename, no_cluster_token=no_cluster
+    create_api_config_secret_arn = api.create_moped_api_secrets_entry(
+        slug=slug, ready_for_secret=ready_for_secret
     )
 
-    removed_hostname = ecs.remove_route53_cname(
-        basename=basename, removed_load_balancer_token=removed_load_balancer
-    )
+    # test if api is deployed, remove if so
```
mddilley, line 128: I like removing and deploying again if the API exists because it means less code, but there is a [zappa update](https://github.com/zappa/Zappa#updates) command. I don't know if there is really any advantage (might be faster?), but I wanted to throw it out there so we have more info - not vouching to change this at all.

mddilley (approved):
🚀 Loving the new modules for shared naming helpers and the migrations. This process is really growing up! Handling updates and all! 🙌 ✨

PR #750, opened 2022-08-12
Changes related to long branch names WRT TLS certificate issuance

frankhereford:
## ⚠️ Do not merge this PR. 

This PR is to make record of the work related to long branch names WRT TLS certificates. These changes do need to get into the new test system, but they are going to be the merge-conflict of all merge-conflict with the unification of flows into one flow. As a result, I will merge these in by hand in the very near future when it's easiest to do so. 

I will close this PR at that time.

@mddilley I put a request for review on here as a formality, but please don't put any big effort into this one. I just wanted to give you a chance to 👀 it if you wanted. There will be another chance to look at this when I do actually make a branch against the unified flow work, and I'll PR that of course. There will be some polish in putting that together, this is like the most wild-west work so far.

PR #756, opened 2022-08-15
Long branch name certificate support

frankhereford:
## Associations
Sibling PR: https://github.com/cityofaustin/atd-prefect/pull/34 

This PR intends to supercede and close PR #750. 👋

## Git woes
Learning more about `git` every day. This was branched off the `unify-flows` branch which had had `main` merged into it. I had expected since the target branch and the source branch have common commits, the GitHub "Files Changed" diff view would not show the common changes. Not the case. 📚

This link will show you the sum of the changes which are different between these two branches -- I hope this is helpful for reviewing. 

https://github.com/cityofaustin/atd-moped/pull/756/files/569e759cb1363a91e642d0f757019bafffc0093c..e3833ab01eeea54d207c6a9393f84b92cbb9cbec

## Features

This PR:

- [x] Defines a shortened basename for use in deploying and reaping ELB and ELB target groups
- [x] Adds support for maintaining the full basename in the graphql endpoint by using a dual hostname certificate from AWS. This uses the mapping functionality of prefect to parallelize certificate validation.
- [ ] Uses the mapping functionality to clean up CNAMES after the singular validation event. This is a low-impact regression in CNAME handling. I will spin this off as a low-priority item.
- [x] Adds a 30 second sleep event after a cert is requested to be deleted. This doesn't feel very prefect-like, but I didn't see a better thing to spin on for that event to be completed.
- [x] Configures the graphql endpoint being given to the API to form the correct, full hostname


mddilley (approved):
Thanks for the diff link - that was super helpful! Looks good to me. That slug dict is such a good way to collect the different naming standards across the flows. 🎉

PR #762, opened 2022-08-17
Slugging, Seed application, API rebuild, & API correct version

frankhereford:
Sibling PR: https://github.com/cityofaustin/atd-prefect/pull/36

### ✨ Major Items
- [x] Rework slugging function into a more readable, maintainable pattern
- [x] Fix failing seed application by adding sleeping intervals between invocations of hasura CLI client
  - This was failing silently and was maddening 
- [x] Cause the API to be built and deployed out of a moped repo checkout of the branch being deployed, not the codebase which has the ETL script in it
- [x] Improve detection of an already deployed state and enable steps to be taken for redeployment

### Minor items
- [x] Add 3 new versions of the slug (`graphql_endpoint`, `activity_log_slug`, `ecs_cluster_name`
- [x] Lots of new `logger.info()` calls, can't have enough when viewing the schematic → task view in the web UI
- [x] Make the executions of migrations controllable via the flow parameters at invocation
- [x] Uppercase the `FUNCTION_NAME` variable defined in the global scope of `api.py`.
  - Considering if this should go into the environment?
- [x] Minor refactor of using the `slug` object directly instead of pulling a single key out and defining it in the functions scope. Readability thing; I want to remind us where to look for the upstream definition of the slugged values. 
- [x] Pin `graphql-engine` to align with production

## General comment
This is a messy PR, and in hindsight would have been better to break up. The practicality of doing that would have been questionable given how the tools have been used over the last few days -- but it does make it a bit more of a mess to review. Sorry about that! I am going to be much better about breaking them out into PRs as much as possible!

### DIFF
--- a/moped-etl/prefect/moped_test_create_flow.py
+++ b/moped-etl/prefect/moped_test_create_flow.py
@@ -50,7 +50,7 @@ def slug_branch_name(basename):
 
     elb_basename = basename[0:32].replace("_", "-")
 
-    graphql_endpoint = re.search("^[\d_]*(.*)", basename).group(1).replace("_", "-")
+    graphql_endpoint = re.search("^[\d_-]*(.*)", basename).group(1).replace("_", "-")
 
     short_tls_basename = graphql_endpoint[0:27]
 
--- a/moped-etl/prefect/moped_test_create_flow.py
+++ b/moped-etl/prefect/moped_test_create_flow.py
@@ -57,7 +57,7 @@ def slug_branch_name(basename):
     activity_log_slug = basename[0:34].replace("_", "-").strip("0123456789-_")
     database = basename[0:63].replace("-", "_").strip("0123456789-_.")
     awslambda = re.sub(
-        r"[^a-zA-Z0-9]", "", basename[0:16].replace("_", "-").strip("0123456789-_")
+        r"[^a-zA-Z0-9-]", "", basename[0:16].replace("_", "-").strip("0123456789-_")
     )
     ecs_cluster_name = basename.strip("0123456789-_.")
     # zappa_stage_name = re.sub(r"[^a-zA-Z0-9]", "", basename)
@@ -184,7 +184,6 @@ with Flow("Moped Test Instance Commission", executor=executor) as test_commissio
 
     deploy_api = api.create_api_task(command=commission_api_command)
     api_endpoint = api.get_endpoint_from_deploy_output(deploy_api)
-
 
     ## Commission the ECS cluster
 
--- a/moped-etl/prefect/moped_test_create_flow.py
+++ b/moped-etl/prefect/moped_test_create_flow.py
@@ -83,7 +83,7 @@ with Flow("Moped Test Instance Commission", executor=executor) as test_commissio
 
     slug = slug_branch_name(branch)
 
-    ## Get github checkout 
+    ## Get github checkout
 
     rm_clone = "rm -fr /tmp/atd-moped"
     cleaned = migrations.remove_moped_checkout(command=rm_clone)
@@ -94,7 +94,7 @@ with Flow("Moped Test Instance Commission", executor=executor) as test_commissio
     checkout_branch = migrations.get_git_checkout_command(slug=slug)
     git_repo_checked_out = migrations.checkout_target_branch(
         command=checkout_branch, upstream_tasks=[cloned]
-    ) 
+    )
 
     ## Commission the database
 
@@ -243,13 +243,10 @@ with Flow("Moped Test Instance Commission", executor=executor) as test_commissio
         command=commission_activity_log_command
     )
 
-
-
     ## Apply Migrations, metadata and optional seed data
     graphql_endpoint = "https://" + migrations.get_graphql_engine_hostname(slug=slug)
 
     access_key = migrations.get_graphql_engine_access_key(slug=slug)
-
 
     metadata = "metadata"
 
--- a/moped-etl/prefect/moped_test_create_flow.py
+++ b/moped-etl/prefect/moped_test_create_flow.py
@@ -156,7 +156,9 @@ with Flow("Moped Test Instance Commission", executor=executor) as test_commissio
     is_deployed = api.check_if_api_is_deployed(slug=slug)
     with case(is_deployed, True):
         decommission_api_command = api.create_moped_api_undeploy_command(
-            slug=slug, config_secret_arn=remove_api_config_secret_arn
+            slug=slug,
+            config_secret_arn=remove_api_config_secret_arn,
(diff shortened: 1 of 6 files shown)


moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -47,24 +48,34 @@
...
-    underscore_basename = basename.replace("-", "_")
-    database = re.search("^[\d_]*(.*)", underscore_basename).group(
-        1
-    )  # remove leading numbers
-    internal_number_free_underscore_basename = "".join(
-        [i for i in database if not i.isdigit()]
+
+    graphql_endpoint = (
+        basename.replace("_", "-").replace(".", "-").strip("0123456789-_")
+    )
```
frankhereford, line 54: I almost want to disable `black` formatting on these lines so they all are a single line and to have them each carefully hand indented so you get columns of similar operations. 🤷🏻‍♂️

moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -90,9 +101,23 @@ def drain_service(slug):
...
+    cleaned = migrations.remove_moped_checkout(command=rm_clone)
+
+    git_clone = f"git clone {GIT_REPOSITORY} /tmp/atd-moped"
+    cloned = migrations.clone_moped_repo(command=git_clone, upstream_tasks=[cleaned])
+
+    checkout_branch = migrations.get_git_checkout_command(slug=slug)
+    git_repo_checked_out = migrations.checkout_target_branch(
+        command=checkout_branch, upstream_tasks=[cloned]
+    )
+
```
frankhereford, line 120: Haul this way up the file because it's a pre-req for lambda deployments.

moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -244,71 +268,55 @@ def drain_service(slug):
...
-    consistent_metadata = migrations.check_for_consistent_metadata(
-        command=consistency_cmd, upstream_tasks=[migrate]
-    )
-
-    settled_metadata = migrations.sleep_fifteen_seconds(
-        consistent_metadata=consistent_metadata
-    )
+    with case(do_migrations, True):
+        # what are these parentheses doing?
+        migrate_cmd = "(cd /tmp/atd-moped/moped-database; hasura --skip-update-check migrate apply; sleep 15;)"
```
frankhereford, line 293: Even when I'm 1,000% positive that the script is not starting one `hasura <thing> apply` operation before the previous one is good and finished, if there isn't some amount of "cool off" time, the `graphql-engine` will get in an unworkable state. So, here we are with `sleep()`s in the subshell.

moped-etl/prefect/moped_test_create_flow.py
```diff
@@ -244,71 +268,55 @@ def drain_service(slug):
...
+        consistent_metadata = migrations.check_for_consistent_metadata(
+            command=consistency_cmd, upstream_tasks=[migrate]
+        )
 
-    with case(use_seed_data, True):
-        apply_seed_cmd = (
-            "(cd /tmp/atd-moped/moped-database; hasura --skip-update-check seed apply;)"
+        # Should this sleep come out or should the bash sleeps be done like this?
+        settled_metadata = migrations.sleep_fifteen_seconds(
+            consistent_metadata=consistent_metadata
```
frankhereford, line 305: This is another way to do the sleeps, and this "looks" more prefecty, but it's a lot of hoop to jump through for a sleep.

moped-etl/prefect/tasks/activity_log.py
```diff
@@ -16,16 +16,14 @@
 SHA_SALT = os.environ["SHA_SALT"]
 
 # The folder name of the event function and used in naming the AWS function
-function_name = "activity_log"
+FUNCTION_NAME = "activity_log"
```
frankhereford, line 19: I was clobbering this variable in a function inadvertently. As I uppercased it, it felt like it may want to jump out into the ENVIRONMENT, TBD.

moped-etl/prefect/moped_test_create_flow.py
mddilley: I can come up with good arguments either way for how to handle the sleeps. Part of me thinks it is fine in bash because a task finishes when it finishes it. Part of me thinks that the delay is more hidden away that way 🤷 I know this isn't super helpful 😆

moped-etl/prefect/tasks/activity_log.py
```diff
@@ -45,17 +43,16 @@ def create_activity_log_lambda_config(
...
-    basename = slug["basename"]
     logger.info("Creating Activity Log deploy helper command")
 
-    aws_function_name = shared.create_activity_log_aws_name(basename, function_name)
+    aws_function_name = shared.create_activity_log_aws_name(slug["activity_log_slug"], FUNCTION_NAME)
 
-    helper_script_path = "/root/test_instance_deployment/atd-moped/.github/workflows"
-    deployment_path = f"/root/test_instance_deployment/atd-moped/moped-data-events/{function_name}"
+    helper_script_path = "/tmp/atd-moped/.github/workflows"
+    deployment_path = f"/tmp/atd-moped/moped-data-events/{FUNCTION_NAME}"
```
mddilley, line 51: Awesome - now we are working from the checked out branch 👍

moped-etl/prefect/tasks/activity_log.py
mddilley: That is a good catch!

moped-etl/prefect/tasks/api.py
```diff
@@ -174,7 +174,8 @@ def create_moped_api_deploy_command(slug, config_secret_arn, ready_for_api_deplo
     logger.info("Creating API Zappa deploy command")
 
     zappa_config = create_zappa_config(basename, config_secret_arn)
-    api_project_path = "/root/test_instance_deployment/atd-moped/moped-api"
+
+    api_project_path = "/tmp/atd-moped/moped-api"
```
mddilley, line 178: 🚀

moped-etl/prefect/tasks/api.py
```diff
@@ -212,7 +222,8 @@ def get_endpoint_from_deploy_output(output_list):
     match = re.search(r"(https://.*)", api_endpoint_item)
 
     if match == None:
-        return False
+        logger.info("No match found looking for API endpoint URL")
```
mddilley, line 225: I like this improved logging and error handling!

mddilley (approved):
Knocking down those `basename` vars one by one! The slug helper is getting really super-powered now! This is awesome, and these fixes really help me understand how the API deployment code was behind the feature branches. Thanks!

PR #761, opened 2022-08-17
Remove contractor and DO from signal table and search

chiaberry:
## Associated issues

companion PR to https://github.com/cityofaustin/atd-moped/pull/760

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #771, opened 2022-08-30
Drop contractor and purchase order columns

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9770

I updated the project list SQL view, dropped the contractor and purchase order columns from the moped_project table and r removed those columns from the permissions metadata. Also updated the seed data to no longer use those columns. 

## Testing
**URL to test:** 
need to run locally

**Steps to test:**

Purely database changes, the front end changes happened in an earlier PR here: https://github.com/cityofaustin/atd-moped/pull/761

---
#### Ship list
- [ ] Code reviewed 
- [ ] ~~Product manager approved~~
- [ ] ~~Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable~~


johnclary (approved):
Looks great! Thank you for knocking this out!

Maybe we should add comments to the [activity logs map](https://github.com/cityofaustin/atd-moped/blob/1111e50c055f307d2de7979ff6e86121e7082310/moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js#L120) that explain that we want to keep the field map definitions for these fields, because historical activities depend on them?

🚢

mateoclarke (approved):
🚢💨✨

PR #779, opened 2022-09-06
Updates toolbox script to generate Moped components geojson

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10189

I've overhauled this toolbox script to play with our current data model, and to make use of the `projects_list_view`. It generates `geojson` files which can be imported into AGOL.

Once we make the transition to using PostGIS column types, we should revisit this and, ideally, create a DB view specifically for this need.

Steps to test:
1. Activate a local python environment with `requests` installed
2. Start local Moped DB and editor
3. Create a project with a few different component types e.g. Buffered bike lane and signals
4. Run the script `python create_components_geojson.py -e local`
5. Locate the `points` and `lines` geojson files in the `_output/` directory.
6. You can upload these to [geojson.io](https://geojson.io/) to inspect the results.
7. Optionally, add `staging` or `prod` entries to `secrets.py` to process data against those environments.


chiaberry (approved):
tested locally and with data from prod. so many points!!

johnclary:
thank's y'all! this script isn't long for this world, but i'm merging it 🗺️

PR #783, opened 2022-09-13
Add Project Tags feature

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/10044

## Testing
**URL to test:** 
https://10044-tags--atd-moped-main.netlify.app/moped/projects/161

**Steps to test:**

Navigate to a project's summary page
Add Tags
Check that after adding a tag, the tag is no longer available as an option in the dropdown
Delete Tag

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
```diff
@@ -286,7 +287,10 @@ const ProjectSummary = ({ loading, error, data, refetch }) => {
               </ErrorBoundary>
             )}
           </Grid>
-          <Grid item xs={12}>
+          <Grid item xs={6}>
```
mddilley, line 290: This probably isn't a huge deal, but the Tags and Subprojects sections get pretty squished when they are next to each other on small screens. Maybe they could be set to 6 on medium or large and then 12 for small or xs to let them breathe some more?

moped-editor/src/queries/tags.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const ADD_PROJECT_TAGS = gql`
+  mutation AddProjectTags($objects: [moped_proj_tags_insert_input!]!) {
+    insert_moped_proj_tags(objects: $objects) {
+      returning {
+        id
+      }
+    }
+  }
+`;
```
mddilley, line 40: Just thinking through how the add/delete works here. 1. User adds a tag (`is_deleted` = false) 2. User deletes that tag (`is_deleted` = true) 3. Another user adds the same tag and a new `moped_proj_tags` record is created with `is_deleted` = false

Over time, `moped_proj_tags` records stack up in the table. Not saying this is how it should or shouldn't work - just making sure this is how it is intended to work. My brain is still waking up 🧠 😅 
Screen Shot 2022-09-13 at 1 20 38 PM

mddilley (approved):
Works great and looks great! Added a couple comments that don't affect how this works at all. 🚢

moped-database/migrations/1662751229997_create_table_public_moped_tags/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE "public"."moped_tags" (
+	"id" serial, 
+	"name" text NOT NULL, 
+	"type" text NOT NULL, 
+	"slug" text NOT NULL, 
+	"is_deleted" boolean NOT NULL DEFAULT false, 
+	PRIMARY KEY ("id") , 
+	UNIQUE ("name"), UNIQUE ("slug"));
```
frankhereford, line 8: unique slug - thank you for enforcing that 1:1 relationship between tag and slugs for URLs or what-have-you! 🐌

pull_request_template.md
```diff
@@ -1,7 +1,7 @@
 ## Associated issues
 
 ## Testing
-**URL to test:** <!-- https://moped-test.austinmobility.io/moped/ or Netlify -->
+**URL to test:** <!-- [branch-name].netlify.app/moped/ if test instance or deploy-preview-[pr-number]--atd-moped-main.netlify.app/moped/-->
```
frankhereford, line 4: 🙏

frankhereford (approved):
tag

This is really well done and polished work Chia. Thanks for answering my hair-brained question in the office regarding the `package-lock.json` file. 

I've tested it on the deployment you linked and read over the code.

moped-editor/src/queries/tags.js
johnclary: thanks — i think that's fine. since our feature is querying this table by `project_id`, and the table is going get rather larger, maybe we should index that `project_id` column?

johnclary (approved):
Looks great / works great! I requested one small change to track Hasura relationships but this is 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
chiaberry: good call -- I updated it. Thank you!

moped-database/metadata/tables.yaml
```diff
@@ -2455,6 +2455,74 @@
...
+        check: {}
+    - role: moped-editor
+      permission:
+        columns:
+          - id
+          - project_id
+          - tag_id
+          - is_deleted
+        filter: {}
+        check: {}
```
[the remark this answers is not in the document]
chiaberry: I updated the relationships

frankhereford:
@johnclary - to your question about indexing that column, I think that is a fine idea. My hot take is:

49.9% of me: I think it's a fair rule of thumb that whenever a column (ideally in the form of an integer) serves as a foreign key, adding that index is worthwhile.

50.1% of me: I also am open that tables get their primary key indexed and don't add other indexing until needed. Let performance profiling queries in dev and the slow query log spell out when and where you need to consider an index. It's reasonable to expect a small database to live entirely in memory and not need too much if any indexing beyond PK indexes. I think moped still resides comfortably at this scale.

johnclary:
> don't add other indexing until needed
👍 🙏 sounds good to me

moped-database/metadata/tables.yaml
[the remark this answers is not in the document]
johnclary: 🚢

johnclary:
i will update the QA test script as i begin testing later today. merge this!

PR #781, opened 2022-09-13
Bump moped graphql engine & Remove heroku workflows

frankhereford:
## Associated issues

This PR intends to close [10220](https://github.com/cityofaustin/atd-data-tech/issues/10220) & [10128](https://github.com/cityofaustin/atd-data-tech/issues/10128). 

## Considerations

Acceptance of this PR indicates an acceptance of bumping the production moped graphql-engine version to 2.11.2 during the next release party, this Thursday.

## Testing
* Spin up your local moped DB stack and see how it works
* Observe that there is no longer a github action for deployment of test to Heroku

---

#### Ship list

- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #786, opened 2022-09-14
Restore the Moped database migration action

johnclary:
## Associated issues
Oops, [we deleted our database CI](https://github.com/cityofaustin/atd-moped/pull/781/files#diff-7b8f8f0b9f093c1dd2b47f26b26ea1508ce1d97f8485ff5a8fc9e1d088e380eb).

## Testing
I don't believe it will be possible to test this until we merge this PR.

### DIFF
--- a/.github/workflows/atd_moped_database.yml
+++ b/.github/workflows/atd_moped_database.yml
@@ -1,0 +1,50 @@
+#
+# Builds Cognito Triggers for Moped Authentication
+#
+name: "Applies the migrations to the database"
+
+on:
+  push:
+    branches:
+      - main
+      - production
+    paths:
+      - "moped-database/**"
+      - ".github/workflows/atd_moped_database.yml"
+      - ".github/workflows/aws-moped-migrations-helper.sh"
+
+jobs:
+  apply:
+    name: Apply Migrations
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+          architecture: "x64"
+      # Get the code first
+      - name: "Checkout"
+        uses: actions/checkout@v2
+      # Then install the AWC CLI tools & boto3
+      - name: "Install AWS Cli & Hasura CLI"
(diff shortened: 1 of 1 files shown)


johnclary:
We're currently missing these migrations in main:
```
1662751229997  create_table_public_moped_tags                                                    Present        Not Present
1662752947833  create_table_public_moped_proj_tags                                               Present        Not Present
1662754037000  insert_moped_tags                                                                 Present        Not Present
1663016156653  pdd_new_milestones                                                                Present        Not Present
1663016172738  pdd_update_existing_milestones                                                    Present        Not Present
```

I think we have two options:
1. Merge this PR—the CI should apply all the missing migrations
2. Manually apply the missing migrations to staging before we merge this PR

.github/workflows/atd_moped_database.yml
```diff
@@ -0,0 +1,50 @@
+#
+# Builds Cognito Triggers for Moped Authentication
+#
```
chiaberry, line 3: can this be updated to reflect what the script does?

johnclary: yes thank you ✅

chiaberry (approved):

.github/workflows/atd_moped_database.yml
```diff
@@ -0,0 +1,50 @@
...
+# Applies database migrations to staging (main) and production
```
mddilley, line 2: 🙏

mddilley (approved):

frankhereford (approved):
Thank you so much for this PR and catching and providing a solution to the oversight in my PR from yesterday!

I like route 1 from John's question. I think that it has not attempted to run these migrations, and it should be able to catch itself up. If there is a problem, then doing step-by-step migrations may be needed, which is #2.

johnclary:
👍 here we go!

johnclary:
Added to the v1.8 RC [here](https://github.com/cityofaustin/atd-moped/pull/787).

PR #787, opened 2022-09-14
Adds DB migration action from main into v1.8 RC

johnclary:
Let's hold off until we know the staging database migrations are squared away.


johnclary:
@chiaberry FYI

PR #792, opened 2022-09-19
Transform geojson GIS data into native PostGIS geography data types

frankhereford:
### ✅ Hold off on merging this until after v1.11 is deployed

# Relationships

* ⚠️ This PR is the anticedent of #793. After this makes it way to production, intermediate steps must be taken including running a script to transfer data into the columns created herein.
* This PR intends to close [9930](https://github.com/cityofaustin/atd-data-tech/issues/9930).

# Todo
* [x] Update timestamps on migrations to make sure they don't overlap with executed ones

# Testing
* YATP: Yet another testing puzzle

### Demo & Review Path
* Observe the output of the script in a demo and read the code.

**or**

### Recreate Environment Path
* Spin up your local cluster
* Shut down your local graphql-engine
* Export a copy of production via the read replica using `pg_dump`
* Drop and recreate your local `moped` database using that export
* Spin back up your graphql-engine
* Have graphql-engine migrate from this checkout 
* Have graphql-engine apply the metadata from this checkout
* (Everything above is the steps for simulating a deployment to production for a migration, BTW)
* Test via graphql and via SQL

# Operations
* Transform geojson GIS data into native PostGIS geography data types
* Update graphql endpoint

# Considerations
* Dropping KV pairs out of top-level geojson object except for `geometry`, `type`, & `properties`, per the geojson spec. 
  * We do have some spurious fields named `id` but they don't appear to match the `properties["id"]` value.
* Genuine consideration is owed the best way to integrate this into the moped migration system
  * Past Frank asks: How will seed data end up with the right schema after building?
    * Now Frank: the team could consider updating the seed file with a snapshot of production from after this PR and transformation script are merged.
  * How best to handle the multi-step nature of this schema change?
    * 👍 Is it to have two PRs, allowing for the creation of fields by the first, the execution of the python script to transfer data, and a second PR to remove the source columns. This is what happens "naturally" as we create and delete columns, we just spread out the PRs by months or years.


frankhereford:
I had a real quick sync with @chiaberry about the idea of having this two-PR delivery of the GIS transformation to allow for an intermediate step of the actual transfer to take place once the this PR hits production. She agreed that it seemed to be a reasonable way to proceed.

moped-database/metadata/tables.yaml
```diff
@@ -1,16 +1,16 @@
 - table:
```
frankhereford, line 1: TODO: Figure out why this file re-scrambled? It should contain changes, but I was under the impression we shouldn't expect to see the `tables.yaml` jumble anymore.

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
```diff
@@ -0,0 +1,168 @@
...
+
+create table features (
+    id serial primary key, 
+    name character varying
+    );
+
+create table feature_signals (
+    signal_id integer default null,
+    knack_id character varying,
+    project_extent_id uuid,
```
frankhereford, line 97: Re this conversation: https://austininnovation.slack.com/archives/CNUEPKLB1/p1666042882917659

The values that come over in the `project_extent_id` field appear to be UUID fields. My understanding from John & Mike is that these IDs have been used as unique identifiers for drawable layers. Is this a suitable solution?

moped-database/metadata/tables.yaml
chiaberry: John's recent PR also had a scrambled tables.yaml :(

moped-toolbox/populate_geometry_transform/transpose_geometry.py
```diff
@@ -0,0 +1,281 @@
...
+            print("Component Subtype: ", record["component_subtype"])
+            print("Feature ID: ", record["feature_id"])
+            print("Default internal table: ", record["internal_table"])
+            print(str(record["feature_id"]) + ": " + str(feature["geometry"]["type"]))
+            pp.pprint(feature["properties"])
+            pp.pprint(feature["geometry"])
+
+        feature_id = None
+        if (record["component_name"] == "Project Extent - Generic" and
+                str(feature["geometry"]["type"]) == "Point"):
```
frankhereford, line 107: This traps the following features:

Feature 164 in project 159: Point feature when "Project Extent - Generic" calls for lines

moped-toolbox/populate_geometry_transform/transpose_geometry.py
```diff
@@ -0,0 +1,281 @@
...
+
+            ID_OF_INTERSECTION_TABLE_IN_FEATURE_LAYERS = 5
+            values = [ID_OF_INTERSECTION_TABLE_IN_FEATURE_LAYERS, record["project_component_id"]]
+
+            print(sql, values)
+            execute(sql, values, get_result=False)
+
+        elif ((record["component_name"] == "Sidewalk" and
+                record["component_subtype"] == 'With Curb and Gutter' and 
+                str(feature["geometry"]["type"]) == 'Point')
```
frankhereford, line 162: This traps feature 170 for project 163 when "Sidewalk / With Curb and Gutter" calls for lines but has a point.

johnclary: boo - i think this an artifact of a known bug we have that allows geometry mismatches following the right sequence of clicks. this is flagged as a test project and i'm going to delete it and its components. maybe we should take this opportunity to not migrate deleted components? i would be happy to see the cruft go away.

moped-toolbox/populate_geometry_transform/transpose_geometry.py
```diff
@@ -0,0 +1,281 @@
...
+                or
+                # the next three line are just being explicit about how i found record 297
+                (record["component_name"] == "Transit" and 
+                record["component_subtype"] == 'Transit/Bike Lane' and 
+                str(feature["geometry"]["type"]) == 'Point' and
+                record["feature_id"] == 297)):
```
frankhereford, line 168: This traps feature 297 in project 228 which is a point where the component_type calls for lines.

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
```diff
@@ -0,0 +1,201 @@
...
+      moped_components.component_id as component_archtype_id,
+      moped_proj_components.project_component_id as component_id,
+      moped_proj_components.is_deleted,
+      moped_project.project_name as project_name,
+      feature_layers.internal_table as table,
+      feature_layers.reference_layer_primary_key_column as original_fk,
+      moped_proj_components.name as component_name,
+      uniform_features.name as feature_name,
+      uniform_features.attributes as attributes,
+      uniform_features.geography as geography
```
frankhereford, line 194: What particular columns we should include in this view will likely change to reflect the needs and convenience of the work being done to the mapping front end systems.

moped-editor/src/views/projects/projectView/ProjectComponentsMapEdit.js
```diff
@@ -274,7 +274,7 @@ const ProjectComponentsMapEdit = ({
       }
     );
 
-    // identifiy IDs of existing project features
+    // identify IDs of existing project features
```
frankhereford, line 277: random spelling correction i found trying to find who knows what

moped-toolbox/populate_geometry_transform/transpose_geometry.py
frankhereford: @johnclary wrote:
> to not migrate deleted components

Hmm, i'm torn. As much as I'd love to see everything in a uniform state, I think deleting soft-deleted records does not align with the intent of having a soft delete system in place.

However, if it's an artifact of testing, and all of the records which can be reached via relations to the subject feature are deleted along with it, then I'm game. I guess put another way, I'm 100% down for deleting features by doing something like `delete from moped_project where ... cascade`, where the cascade reaches down and gets the features.

frankhereford:
@johnclary I was thinking about your comment that this work needs to hang out to mega-merge with the front-end stuff. I'm not entirely sure that's the case. These schema changes are non-destructive, and as long as we rerun the `transpose_geometry.py` script in conjunction with the launch-party containing the front-end work, we should be good.

I think it might also be easier on @mddilley's work if he can merge `main` and get these migrations. 

Totally game for whatever is the path of least resistance, of course!

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
frankhereford: We circled up to talk about the mapping systems today, and we observed that this view is really more of a example of what we can do, and not so much intended to be the prescribed form that the data should be made available via the graphql-engine. 

We agreed to spin-off a new issue to track the refinement of this view and creation of others as needed based on what @mddilley finds to be the best endpoint shape for his work on the frontend.

moped-database/metadata/tables.yaml
frankhereford: @johnclary It's jumbled again. I think it's the result of maybe this PR being old and maybe created prior to your work to get the `tables.yaml` file ironed out and "in order." Maybe I can do what you did somehow on this branch to get it back in-line with `main`? I'm curious to what you think is smart. I don't want to regress the ordered nature of this file.

johnclary:
> was thinking about your comment that this work needs to hang out to mega-merge with the front-end stuff

@frankhereford good call. I agree!

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
johnclary: 👍 will make that issue

moped-toolbox/populate_geometry_transform/transpose_geometry.py
johnclary: well said. my main concern is that more of these edge cases come about between now and when we run this script. i might take on some data cleanup in the meantime.

moped-toolbox/populate_geometry_transform/transpose_geometry.py
```diff
@@ -0,0 +1,281 @@
...
+            print("Component Name: ", record["component_name"])
+            print("Component Subtype: ", record["component_subtype"])
+            print("Feature ID: ", record["feature_id"])
+            print("Default internal table: ", record["internal_table"])
+            print(str(record["feature_id"]) + ": " + str(feature["geometry"]["type"]))
+            pp.pprint(feature["properties"])
+            pp.pprint(feature["geometry"])
+
+        feature_id = None
+        if (record["component_name"] == "Project Extent - Generic" and
```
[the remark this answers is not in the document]
johnclary: i'd like to change that :/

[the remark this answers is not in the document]
frankhereford: Me too! I think it is a change that is outside the scope of this issue, and may be a confusing tag-along, but I also think the "migration" to do it would be super quick. A snackoo if there ever were one. 🍫

moped-database/metadata/tables.yaml
johnclary: i'm starting to wonder if hasura is just capricious and we'll be chasing the ordered life nature for a long time. i did find [this issue](https://github.com/hasura/graphql-engine/issues/6348) which confirms that Hasura does intend to be consistent with metadata sorting but also me have flaws.

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
```diff
@@ -0,0 +1,197 @@
...
+update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_intersections'  ) where component_name = 'Lighting' and component_subtype = 'Intersection lighting';
+update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_intersections'  ) where component_name = 'Transit' and component_subtype = 'Bus stop';
+update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_street_segments') where component_name = 'Project Extent - Generic' and component_subtype is null;
+update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_intersections'  ) where component_name = 'Signage' and component_subtype = 'RRFB';
+update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_intersections'  ) where component_name = 'Intersection' and component_subtype = 'Roundabout';
```
johnclary, line 82: @frankhereford we have an additional component making it's way into the DB in v1.11:
https://github.com/cityofaustin/atd-moped/pull/844/files

This should cover it:

```sql
update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_street_segments'  ) where component_name = 'Intersection' and component_subtype = 'Improvement (linear)';

update moped_components set feature_layer_id = (select id from feature_layers where internal_table = 'feature_intersections'  ) where component_name = 'Intersection' and component_subtype = 'Improvement';
```

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
johnclary: i missed this — yes!

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
johnclary: Here's that issue https://github.com/cityofaustin/atd-data-tech/issues/10606

johnclary (changes requested):
As discussed with @frankhereford let's hold off on this merge until v1.11 ships. There are two new components to account for but otherwise this is bee's knees

moped-database/migrations/1665693611607_add_component_geometry_tables/up.sql
frankhereford: Thank you @johnclary - both for observing that these would be missing and then for even providing the `update` statements!

johnclary (approved):
🚢 🙌 🗺️

mddilley (approved):
Thanks for all the demos and updates as you've been working through this! I can't wait to get this sync'd up with the front end. This is a huge improvement that gives us a clear picture of what data comprises components. 🙌

PR #793, opened 2022-09-19
Remove GeoJSON fields previously used to store GIS data

frankhereford:
# Relationships

* ⚠️ This PR is a descendant PR of #792. It must proceed to production after #792 and with intermediate, manual steps taken. These include running a script to transfer GIS data from the columns removed in this PR.

# Todo
* [ ] Bump migration timestamps to something later than the antecedent PR's & and avoid out-of-order migrations.

# Operations

* Remove now unused GeoJSON fields previously used to store GIS data

# Consideration

* This PR is to be run **only** after all data contained in these columns has been moved, via script or other means, into native `geography` fields.


moped-toolbox/populate_geometry_transform/transpose_geometry.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python3
+
```
johnclary, line 2: i believe we'll want to merge #792/main into this branch, there are some shared files (like this one) which are more current on main. this is not currently reflected in this diff...

frankhereford: I agree John. I've just now merged `main` into this branch, which went in without conflict. ~470 commits from main made it into here.

Thanks for thinking of this!

johnclary:
@johnclary double-check that we have the v1.15 backup and that it won't be deleted

frankhereford:
@johnclary wrote:
> @johnclary double-check that we have the v1.15 backup and that it won't be deleted

We have an RDS snapshot of the moped database instance from that launch party. Since it was created by a human, as opposed to the daily, scheduled backup, it will persist until deleted. It can be used to create a new RDS instance from which we could query out the data, but you can't "apply the snapshot" to an existing database. Doing that sort of thing requires `pg_dump` and `psql`.

johnclary:
thank you @frankhereford!

Before we move forward:
1. I think we can actually drop the entire `moped_proj_features` table? We might want to hold off until #11313 just to be very clear that our codebase is not referring to it anymore...
2. This will break our seed data. I want to update and merge https://github.com/cityofaustin/atd-moped/pull/944 before this PR.


cc @mddilley

frankhereford:
@johnclary Just thinking out loud, but what do you think about planning on an RDS snapshot right before we let this loose onto production as well as part of a deployment agenda -- does that sound suitable to you? Really, I wouldn't mind doing one for every release with migrations - it's a heck of an undo button if something goes really haywire (I'm thinking `DELETE`s and `DROP`s).

johnclary:
@frankhereford that sounds perfect—i will update our agenda. Thank you!

johnclary (approved):
@frankhereford plz merge at your convenience—i updated our [release issue](https://github.com/cityofaustin/atd-data-tech/issues/11360) with the reminder to backup prod before this merges there.


🚢 🚢 🚢 🚢 🚢 🚢 

 🚢

frankhereford:
@johnclary I may have made a mistake here - I got these alerts that this failed in deploying the migrations, [here](https://github.com/cityofaustin/atd-moped/actions/runs/4129074997) and [here](https://github.com/cityofaustin/atd-moped/actions/runs/4129074995). I bet i have some out of order migrations or something along those lines. I will look into this tomorrow AM and triage what it broke. cc @cityofaustin/atd-devs

johnclary:
@frankhereford thanks for flagging this—please let me take this one! i'll make an issue and get it at least triaged tomorrow.

PR #804, opened 2022-09-28
Automatically show blank advanced search filter

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10289

## Testing
**URL to test:** 
https://deploy-preview-804--atd-moped-main.netlify.app/

**Steps to test:**
go to projects list view, click filter button, you should see an empty field. add a filter, search, delete a filter.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableFilters.js
+++ b/moped-editor/src/components/GridTable/GridTableFilters.js
@@ -416,6 +416,8 @@ const GridTableFilters = ({
   useEffect(() => {
     if (Object.keys(filterParameters).length === 0) {
       filterState.setFilterParameters(filterParameters);
+      // Add an empty filter so the user doesn't have to click the 'add filter' button
+      handleAddFilterButtonClick();
     }
   }, [filterParameters, filterState]);
 
--- a/moped-editor/src/components/GridTable/GridTableFilters.js
+++ b/moped-editor/src/components/GridTable/GridTableFilters.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { useLocation } from "react-router-dom";
 import { v4 as uuidv4 } from "uuid";
 import PropTypes from "prop-types";
@@ -152,6 +152,19 @@ const GridTableFilters = ({
     return { ...defaultNewFieldState, id: uuid };
   };
 
+  const generateEmptyFilter = useCallback(() => {
+    // Generate a random UUID string
+    const uuid = uuidv4();
+    // Clone state
+    const filtersNewState = {
+      ...filterParameters,
+    };
+    // Patch new state
+    filtersNewState[uuid] = generateEmptyField(uuid);
+    // Update new state
+    setFilterParameters(filtersNewState);
+  });
+
   /**
    * Returns true if Field has a lookup table associated with it and operator is case sensitive
    */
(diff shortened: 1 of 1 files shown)


moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -429,11 +434,6 @@ const GridTableFilters = ({
           <Icon fontSize={"small"}>close</Icon>
         </IconButton>
       </Grid>
-      {Object.keys(filterParameters).length === 0 && (
-        <Alert className={classes.filterAlert} severity="info">
-          You don't have any search filters, add one below.
-        </Alert>
-      )}
```
tillyw, line 436: this is no longer necessary! the user will never see this state :)

patrickm02L (approved):
Nice! 🤘
Screen Shot 2022-09-29 at 10 04 19 AM

johnclary (changes requested):
ah, this is much better! thank you @tillyw!

i'd like to request one small change—can you hide the delete filter trash can if there is only filter being displayed? the ability to delete the only filter then have it re-appear doesn't feel quite right.

moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -115,42 +115,54 @@ const GridTableFilters = ({
...
-    label: null,
-  };
-
   /**
    * Generates a copy of an empty field
    * @param uuid
    * @return {Object}
    */
-  const generateEmptyField = (uuid) => {
+  const generateEmptyField = useCallback((uuid) => {
```
mddilley, line 123: This `useCallback` works great and gets around the dependency array warnings that calling `generateEmptyField` inside of a `useEffect` probably caused. 

I think another way to get around this is pulling the definitions of `defaultNewFieldState` and `generateEmptyField` outside of the `GridTableFilters` component and into the top level of this file since they don't rely on any state. Just a different approach and no change needed here! There are many ways to handle this.

mddilley (approved):
Works great and is a really nice improvement that makes smooths out the filtering flow! ✨  I agree with John's feedback since the reset button will let users clear a single filter.

johnclary (changes requested):
@tillyw two more minor changes i'd like to request:
- If you'd add the `autoFocus` prop to this <TextField/> component, it will cause the filter field input to always autoFocus when it appears ✨ 
- Would you force the filter search value to be disabled until an operator is chosen? we do something similar to the operator input, but you can weirdly still type in a search value.

Screen Shot 2022-09-30 at 1 07 26 PM

sorry for the bit of scope creep 🙈

tillyw:
> 

@johnclary the last one was a bit tricky but I managed to do it using a useEffect!

moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -416,8 +424,26 @@ const GridTableFilters = ({
...
+      if (
+        filterParameters[filterKey].value === null ||
+        filterParameters[filterKey].value === ""
+      ) {
+        setFilterComplete(false);
+      } else {
+        setFilterComplete(true);
+      }
+    });
+  });
```
johnclary, line 446: @tillyw this looks good! you just need to add the dependency array at the end, `[filterParameters]`. otherwise i believe it just runs on every render.

moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -575,6 +599,7 @@ const GridTableFilters = ({
               <Hidden smDown>
                 <Grid item xs={12} md={1} style={{ textAlign: "center" }}>
                   <IconButton
+                    disabled={!filterParameters[filterId].value}
```
johnclary, line 602: There's a small bug here, because you can end up in this weird state where you cannot delete a filter without filling it out.

Screen Shot 2022-10-05 at 5 37 50 PM

I think the disabled condition we want to enforce is when there is currently only one filter visible. So maybe testing on `Object.keys(filterParameters).length` ?

johnclary (changes requested):
Looking great - so close!

chiaberry (approved):

johnclary (approved):
so good! thanks @tillyw! 🚢 🚢 🚢 🚢 🚢 🚢

PR #812, opened 2022-09-29
Add advanced search for project tags

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10288

This PR adds advanced search for project tags and patches a bug where autocomplete had stopped enabling with an option dropdown when a lookup table was set in the [ProjectsListViewFiltersConf](https://github.com/cityofaustin/atd-moped/blob/5b991e54b41d829b6d66291bc8c9e293eba05ed8/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js).

I'm confident that the bug patch won't lead to regressions, but please let me know if that is not the case. Also could use feedback on if this patch restores things to the way they were. 🙏 

Also, I'll add the tests that I mentioned in a separate PR since it requires adding a package and some boilerplate that will clutter this diff up.

## Testing
**URL to test:**
https://md-10228-adv-tag-search--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to https://md-10228-adv-tag-search--atd-moped-main.netlify.app/moped/projects
2. Use the advanced filter button to open the filters and select **Project tags** from the field dropdown
3. Select **is** from the operator dropdown and you should see the last field turn into a dropdown populated with existing tags. Repeat with the **is not** operator
4. Add and remove tags from projects and test:
   a. Using **is** to find a project with a single specific tag
   b. Using **is not** to filter out projects with a single specific tag
   c. Using  **contains** to find a project with a tag containing "Signals" after adding the "MAP Signals 2023" tag to a project along with others
   d. Using **begins with** to find a project with a tag starting with "MAP" after adding any tag since they all currently start with "MAP"
   e. d. Using **ends with** to find a project with a tag ending with "fiscal" after adding the "MAP TIA Fiscal" tag to a project along with others

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/GridTableFilters.js
+++ b/moped-editor/src/components/GridTable/GridTableFilters.js
@@ -159,7 +159,10 @@ const GridTableFilters = ({
     return (
       field.lookup_table &&
       !loading &&
-      ["_eq", "_neq", "_ilike"].includes(field.gqlOperator)
+      [
+        "string_equals_case_sensitive",
+        "string_does_not_equal_case_sensitive",
+      ].includes(field.operator)
     );
   };
 
--- a/moped-editor/src/components/GridTable/GridTableFilters.js
+++ b/moped-editor/src/components/GridTable/GridTableFilters.js
@@ -159,7 +159,7 @@ const GridTableFilters = ({
     return (
       field.lookup_table &&
       !loading &&
-      ["_eq", "_neq"].includes(field.gqlOperator)
+      ["_eq", "_neq", "_ilike"].includes(field.gqlOperator)
     );
   };
 
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -297,8 +297,6 @@ export const ProjectsListViewFiltersConf = {
         "string_ends_with_case_insensitive",
         "string_equals_case_sensitive",
         "string_does_not_equal_case_sensitive",
-        "string_is_null",
-        "string_is_not_null",
       ],
     },
   ],
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -285,7 +285,7 @@ export const ProjectsListViewFiltersConf = {
     {
       name: "project_tags",
       label: "Project tags",
-      placeholder: "Project tags",
+      placeholder: "Tags",
       type: "string",
       lookup: {
         table_name: "moped_tags",
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -54,7 +54,7 @@ export const ProjectsListViewFiltersConf = {
       defaultOperator: "string_equals_case_sensitive",
       lookup: {
         table_name: "moped_types",
-        field_name: "type_name"
+        field_name: "type_name",
       },
       operators: [
         "string_equals_case_sensitive",
@@ -116,7 +116,7 @@ export const ProjectsListViewFiltersConf = {
       type: "string",
       lookup: {
         table_name: "moped_entity",
-        field_name: "entity_name"
+        field_name: "entity_name",
       },
       operators: [
         "string_contains_case_insensitive",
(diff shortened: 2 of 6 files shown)


moped-editor/src/components/GridTable/GridTableFilters.js
```diff
@@ -159,7 +159,10 @@ const GridTableFilters = ({
     return (
       field.lookup_table &&
       !loading &&
-      ["_eq", "_neq"].includes(field.gqlOperator)
+      [
+        "string_equals_case_sensitive",
+        "string_does_not_equal_case_sensitive",
+      ].includes(field.operator)
```
mddilley, line 165: This change helps us rely on whichever GraphQL operator is set for the GridTable strings operators set [here](https://github.com/cityofaustin/atd-moped/blob/5b991e54b41d829b6d66291bc8c9e293eba05ed8/moped-editor/src/components/GridTable/GridTableFiltersCommonOperators.js#L152-L165)

moped-database/views/project_list_view.sql
```diff
@@ -113,6 +113,14 @@ AS WITH project_person_list_lookup AS (
...
         AND personnel.project_id = mp.project_id
       GROUP BY personnel.project_id) AS project_designer,
+    ( -- get me all of the tags added to a project
+    SELECT string_agg(tags.name, ', '::text) AS string_agg
+      FROM moped_proj_tags ptags
+        JOIN moped_tags tags ON ptags.tag_id = tags.id
+      WHERE 1 = 1
+        AND ptags.is_deleted = false
+        AND ptags.project_id = mp.project_id
+      GROUP BY tags.project_id) AS project_tags,
```
mddilley, line 123: Totally stole this from previous work in this view 😄 🙏 🙏 🙏

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -282,6 +282,23 @@ export const ProjectsListViewFiltersConf = {
         "string_is_not_null",
       ],
     },
+    {
+      name: "project_tags",
+      label: "Project tags",
+      placeholder: "Project tags",
```
johnclary, line 288: @mddilley would you change the label + placeholder to "Tags"? That will match the way we have them labeled on the project summary tab.

johnclary (approved):
Other than fixing the filter label, this works perfectly! I tested the other filters that have a `lookup` configured, and they are indeed back to working condition again. Very nice — thanks @mddilley!

🚢

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: Definitely - just pushed this update ✅

johnclary:
Just to flag that @tillyw's [PR 804](https://github.com/cityofaustin/atd-moped/pull/804/files) also touches `GridTableFilters.js`.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
chiaberry: I'm dense, where is the placeholder on the UI? I thought it was in the advanced search input, but I don't see any text in those inputs.

moped-database/views/project_list_view.sql
```diff
@@ -113,6 +113,14 @@ AS WITH project_person_list_lookup AS (
         AND personnel.is_deleted = false
         AND personnel.project_id = mp.project_id
       GROUP BY personnel.project_id) AS project_designer,
+    ( -- get me all of the tags added to a project
+    SELECT string_agg(tags.name, ', '::text) AS string_agg
+      FROM moped_proj_tags ptags
+        JOIN moped_tags tags ON ptags.tag_id = tags.id
+      WHERE 1 = 1
+        AND ptags.is_deleted = false
```
chiaberry, line 121: thank you!

chiaberry (approved):

roseeichelmann (approved):
Looks good! Just to be sure - If a project has multiple tags and you use "is not" to filter out one of the tags, the project should still come up because it also has other tags? This is the behavior I'm seeing which I think is what you want, so just double checking if I understand

mddilley:
@roseeichelmann - yep, that is how it currently works, but I am also curious about if that is what was originally intended since it caught me off guard as well. @johnclary @chiaberry - does that sound right or do we need a follow up issue?

johnclary:
Thanks for pointing that out @roseeichelmann and @mddilley! I agree that behavior is counterintuitive. Will flesh out a new issue [here](https://github.com/cityofaustin/atd-data-tech/issues/10490).

PR #814, opened 2022-10-04
Updates component geojson extract to include project tags

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10384

Adds `moped_proj_tags` to the components geojson extracts.

Steps to test:
1. Activate a local python environment with `requests` installed
2. Start local Moped DB and editor
3. Create a project with a few different component types e.g. Buffered bike lane and signals
4. Add Tags to the project.
5. Run the script `python create_components_geojson.py -e local`
6. Locate the `points` and `lines` geojson files in the `_output/` directory.
7. Search the output geojson to verify that the projects' tags are populated in the `moped_proj_tags` property.
8. You can upload these to [geojson.io](https://geojson.io/) to inspect the results.
9. Optionally, add `staging` or `prod` entries to `secrets.py` to process data against those environments.

### DIFF
--- a/moped-toolbox/component_geo_processing/settings.py
+++ b/moped-toolbox/component_geo_processing/settings.py
@@ -42,6 +42,11 @@ QUERY_TEMPLATE = """
     query ProjectComponentsQuery {
         moped_project(where: {is_deleted: {_eq: false}}) {
             project_id
+            moped_proj_tags(where: {is_deleted: {_eq: false}}) {
+                moped_tag {
+                    name
+                }
+            }
             moped_proj_components(where: {is_deleted: {_eq: false}}) {
                 moped_components {
                     component_name
--- a/moped-toolbox/component_geo_processing/create_components_geojson.py
+++ b/moped-toolbox/component_geo_processing/create_components_geojson.py
@@ -141,6 +141,14 @@ def add_project_properties(properties, project_data, project_list_keys):
         properties[key] = project_data[key]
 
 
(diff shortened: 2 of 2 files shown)


moped-toolbox/component_geo_processing/settings.py
```diff
@@ -42,6 +42,11 @@
     query ProjectComponentsQuery {
         moped_project(where: {is_deleted: {_eq: false}}) {
             project_id
+            moped_proj_tags(where: {is_deleted: {_eq: false}}) {
+                moped_tag {
+                    name
+                }
+            }
```
johnclary, line 49: tags are not available in the `project_list_view` yet - so i patched them into this part of the query.

it's a goofy script - but can be simplified later after we overhaul components schema

johnclary:
folks, there's no urgency here - i provided Andrew with the data he needs.

moped-toolbox/component_geo_processing/settings.py
chiaberry: Mike's PR adds tags to the project_list_view, but this works since we will overhauling things not too long from now

chiaberry (approved):

mddilley (approved):
Ran the script, got the geojson, plugged it into geojson.io, and, voila! 🪄 🐇 , I see project tags. 🚢

johnclary:
thanks y'all!

PR #815, opened 2022-10-05
Adds a delete confirmation modal to project tags and comments

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10292
fixes https://github.com/cityofaustin/atd-data-tech/issues/10318

## Testing
**URL to test:** 
https://deploy-preview-815--atd-moped-main.netlify.app/

**Steps to test:**
go to a project, add a comment, and click the delete icon. click cancel. click the delete icon again and click delete.
repeat the previous step on the project summary page for a project tag.

i created a reusable modal component that we should be able to implement in other places we want a delete confirmation modal. the only issue is that the delete icon on the project comments and the delete option on the chip on project tags act very differently, so the solution is not as elegant as i was hoping it would be, with a few conditionals and implementing the tag chip inside the modal rather than just rendering it as {children} (not ideal i think). on the other hand, maybe this is okay, but if anyone has any thoughts or suggestions i welcome them!

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/DeleteConfirmationModal.js
+++ b/moped-editor/src/views/projects/projectView/DeleteConfirmationModal.js
@@ -1,7 +1,6 @@
-import React, { useState } from "react";
+import React from "react";
 import {
   Button,
-  Chip,
   Dialog,
   DialogActions,
   DialogContent,
@@ -14,19 +13,16 @@ const useStyles = makeStyles((theme) => ({
   deleteButton: {
     color: theme.palette.error.main,
   },
-  chip: {
-    margin: theme.spacing(0.5),
-  },
 }));
 
-const DeleteConfirmationModal = ({ type, item, submitDelete, children }) => {
+const DeleteConfirmationModal = ({
+  type,
+  submitDelete,
+  deleteConfirmationOpen,
+  setDeleteConfirmationOpen,
+  children
+}) => {
   const classes = useStyles();
-
-  const [deleteConfirmationOpen, setDeleteConfirmationOpen] = useState(false);
-
-  const handleDeleteOpen = () => {
-    setDeleteConfirmationOpen(true);
-  };
 
   const handleDeleteClose = () => {
     setDeleteConfirmationOpen(false);
@@ -34,16 +30,7 @@ const DeleteConfirmationModal = ({ type, item, submitDelete, children }) => {
 
   return (
     <span>
-      {type === "comment" && (
-        <span onClick={() => handleDeleteOpen()}>{children}</span>
-      )}
-      {type === "tag" && (
-        <Chip
-          label={item.moped_tag.name}
-          onDelete={() => handleDeleteOpen()}
-          className={classes.chip}
-        />
-      )}
+      {children}
       <Dialog
         open={deleteConfirmationOpen}
         onClose={handleDeleteClose}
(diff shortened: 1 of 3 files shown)


chiaberry (commented):
I like the idea of using a reusable confirmation modal, but agree with you that the conditionals and rendering inside the modal isn't that idea. The modal is tasked with displaying the tag content and I think theres a way we can get away from that. 

I've been thinking about it -- what if the modal's open/close state is tracked in the TagsSection and ProjectComments components. Clicking on the delete button will then toggle the modal to be open (which can be passed in as a prop). The modal component would then only be the Dialog.

tillyw:
> I like the idea of using a reusable confirmation modal, but agree with you that the conditionals and rendering inside the modal isn't that idea. The modal is tasked with displaying the tag content and I think theres a way we can get away from that.
> 
> I've been thinking about it -- what if the modal's open/close state is tracked in the TagsSection and ProjectComments components. Clicking on the delete button will then toggle the modal to be open (which can be passed in as a prop). The modal component would then only be the Dialog.

@chiaberry yes we can do it that way! that was how i initially had it, then i reworked it to have the modal fully contained within the component (including the open/close state) which i thought would be better for implementing it in other places, but maybe that's not ideal in this situation. i can rework it!

tillyw:
@chiaberry i reworked the modal to pass the state in from the parent component. it's a bit more complicated on the parent side but the modal is a lot simpler, hopefully this is an improvement!

moped-editor/src/views/projects/projectView/DeleteConfirmationModal.js
```diff
@@ -0,0 +1,71 @@
...
+      >
+        <DialogContent>
+          <DialogContentText id="alert-dialog-description">
+            {`Are you sure you want to delete this ${type}?`}
+          </DialogContentText>
+        </DialogContent>
+        <DialogActions>
+          <Button
+            variant="outlined"
+            className={classes.deleteButton}
```
johnclary, line 48: i believerthe MUI-preferred way to do this is to instead define the color within our theme so you can use the `color` prop. using the `color` prop will keep the hover/press/border styles consistent with other button components.

you can update our theme by copying [from here](https://github.com/cityofaustin/atd-moped/pull/824/files#diff-44937b1f7ff295c1e359f2a53ba72aa26f164bc292c836acb7e3cc0775934825) - which are the `error` colors i pulled from the MUI docs.

moped-editor/src/views/projects/projectView/DeleteConfirmationModal.js
```diff
@@ -0,0 +1,71 @@
...
+            }}
+          >
+            <span>Delete</span>
+          </Button>
+          <Button
+            variant="outlined"
+            color="primary"
+            onClick={handleDeleteClose}
+            autoFocus
+          >
```
johnclary, line 62: i've changed my mind—would mind putting the "Delete" button to the right of the "Cancel" button? That will match the "Delete project" modal, which feels right to me.

would you also switch the "Delete" button's variant to `contained`, and leave the "Cancel" button without a variant?

johnclary (changes requested):
@tillyw I've requested a few minor changes but on the whole this is great.

i definitely prefer this approach of passing the event handlers as props to modal component!  🙌

roseeichelmann (approved):
This looks and works great! I also like this updated solution of passing the open/close state into the modal from the parents rather than having the modal rendering the chip. Nice job ✨

johnclary (approved):
thank you @tillyw! 🙌  merging this....

PR #822, opened 2022-10-07
Fix migration conflicts in local Hasura cluster 

mddilley:
## Associated issues
See https://austininnovation.slack.com/archives/CNUEPKLB1/p1665179295381869

TLDR:
Two migrations conflict when applying locally but applied with a regression in staging. Using the same fix as [the last time something like this happened](https://github.com/cityofaustin/atd-moped/pull/680).

**Order applied in staging**
1. `1664467952757_remove_deleted_contracts_from_project_view` patched soft deleted contractors from returning from the view
2. `1664402014381_add_project_tags_to_view` added project tags to view but also reintroduced soft deleted contractors to the returned data

**Order applied locally**
1. `1664402014381_add_project_tags_to_view` adds project tags
2. `1664467952757_remove_deleted_contracts_from_project_view` replaces the view with a version that doesn't have project tags and creates a mismatch between the view and its metadata which causes an error

**The fix:** 
1. Remove `1664467952757_remove_deleted_contracts_from_project_view` so local migration in chronological order don't error
2. Add `1665180093803_remove_deleted_contracts_again` to replace the view with a version that has both the project tags portion and also filters out soft deleted project contractors

## Testing
**URL to test:** 

https://md-patch-migrations--atd-moped-main.netlify.app/moped/ (can also be tested locally)

**Steps to test:**
1. Go through the test steps in https://github.com/cityofaustin/atd-moped/pull/812
2. Add project contractors to a project, delete them, and then make sure that the deleted contractors don't show up in the Project list view

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/views/project_list_view.sql
+++ b/moped-database/views/project_list_view.sql
@@ -120,7 +120,7 @@ AS WITH project_person_list_lookup AS (
       WHERE 1 = 1
         AND ptags.is_deleted = false
         AND ptags.project_id = mp.project_id
-      GROUP BY tags.project_id) AS project_tags,
+      GROUP BY ptags.project_id) AS project_tags,
     string_agg(contracts.contractor, ', ') AS contractors,
     string_agg(contracts.contract_number, ', ') AS contract_numbers
    FROM moped_project mp
--- a/moped-database/migrations/1665180093803_remove_deleted_contracts_again/down.sql
+++ b/moped-database/migrations/1665180093803_remove_deleted_contracts_again/down.sql
@@ -113,6 +113,14 @@ AS WITH project_person_list_lookup AS (
         AND personnel.is_deleted = false
         AND personnel.project_id = mp.project_id
       GROUP BY personnel.project_id) AS project_designer,
+    ( -- get me all of the tags added to a project
+    SELECT string_agg(tags.name, ', '::text) AS string_agg
+      FROM moped_proj_tags ptags
+        JOIN moped_tags tags ON ptags.tag_id = tags.id
+      WHERE 1 = 1
+        AND ptags.is_deleted = false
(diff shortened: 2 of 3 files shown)


moped-database/views/project_list_view.sql
```diff
@@ -120,7 +120,7 @@ AS WITH project_person_list_lookup AS (
       WHERE 1 = 1
         AND ptags.is_deleted = false
         AND ptags.project_id = mp.project_id
-      GROUP BY tags.project_id) AS project_tags,
+      GROUP BY ptags.project_id) AS project_tags,
```
mddilley, line 123: This was a type from my PR

johnclary (approved):
Here's what the migration status ends up looking like when we do these deletes (where `SOURCE` is local and `DATABASE`  is staging):

```
VERSION        NAME                                            SOURCE STATUS  DATABASE STATUS
1607388943390  -                                               Not Present    Present
1654646507541  -                                               Not Present    Present
1664467952757  -                                               Not Present    Present
```

Since the migrations no longer exist, i don't see any harm, although part of me wonders if it would be better to instead edit the problem migrations to be empty/comments. That way we'd leave breadcrumbs for ourselves? We've already got a pattern established here so I suppose it's best to just follow it.

thanks @mddilley for sorting this out!!

mddilley:
@johnclary I'm thinking that isn't the worst idea either. We could comment out the original migration and then add some notes about what happened. 

I'll bring it up in dev sync. I'm also thinking a GH action that is very similar to the [DB docs one](https://github.com/cityofaustin/atd-moped/blob/main/.github/workflows/atd_moped_db_docs.yml) (minus the docs bash script) that would run on newly opened PRs could help us catch these before merge.

PR #824, opened 2022-10-07
Establish baseline components code changes (#1)

mddilley:
### Note: 🚧  This PR is now being used to collect new features before releasing the updated moped components UI and database updates. This should only be merged once we test with the transform script for existing components and their features. 🚧 

Pre-merge steps:
- [x] ~Fix `Line 108:6:  React Hook useEffect has a missing dependency: 'errorMessageDispatch'. Either include it or remove the dependency array. If 'errorMessageDispatch' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps` build error. I will do this in one of the incoming PRs.~ Fixed in https://github.com/cityofaustin/atd-moped/pull/933
- [x] Test with output of the transform script
   - [x] Staging
   - [x] Prod
- [x] Pull main and resolve conflicts (make sure migrations don't break locally)
- [x] ~Update DB migration timestamps to avoid any conflicts there (should be safe since nothing else has changed this but we should be careful about this)~ Test applying migrations to local DB populated with staging and production snapshots
   - [x] staging
   - [x] production 

Post-merge steps:
- [ ] Run transform script on staging DB

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10400

This PR moves John's base for the project components updates to the project view folders. I kept the old code around since it will be needed to migrate existing components.

@johnclary - the only major changes to your code are:
- Wrapped `MapView` in a MUI full screen Dialog component
- Passing project name and project statuses props down from the `ProjectView` component for the map status badge (project ID comes from the URL)
- Splitting the `settings.js` into separate map settings and map style settings (we'll see how this works since they rely on a lot of common variables - we can always bring them back together)
- I couldn't get the "Close Map" button to go back to where the user came from and also fall back to the Summary tab - so it links back to the Summary tab every time for now. I can create an issue to come back to this, but it boils down how we use the `useRoutes` RR hook and me not finding a great way to inject the last route visited into the navigation state (if there is a last route -> go there, if not -> go to Summary tab). Found [this possible workaround](https://stackoverflow.com/a/71647428).

## Testing
**URL to test:** 

https://md-10400-baseline-components--atd-moped-main.netlify.app/moped/ (or test locally)

**Steps to test:**
1. Go to an existing project or create a new one
2. Go to the **Map** tab and a full screen map view should open
3. Check that the project name and project # match the project the map linked from
4. Click the **Close Map X** button in the top right corner and you should navigate back to the **Summary** view

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -0,0 +1,324 @@
+import { useState, useMemo } from "react";
```
johnclary, line 1: just want to mention that there was zero thought behind calling this component `TheMap` and i think `Map` would be better 🤪

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -0,0 +1,258 @@
...
+    const newComponentList = components.filter(
+      (comp) => comp._id !== clickedComponent._id
+    );
+    setComponents(newComponentList);
+    setClickedComponent(null);
+    setIsDeletingComponent(false);
+  };
+
+  return (
+    <Dialog fullScreen open={true}>
```
johnclary, line 153: very handy—works perfectly!

johnclary:
@mddilley i just revoked my approval so as not to forget that we don't actually want to merge this to main.....yet 😎

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
mddilley: Ok cool - thought is sounded pretty snazzy myself 😄

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
```diff
@@ -0,0 +1,66 @@
...
+this relies on nearby lines being avaialble in-memory, which 
+is not guaranteed. a reliable solution would be query the AGOL streets
+layer on-the-fly to grab street names - not sure if this is worth it TBH 
+*/
+export const getIntersectionLabel = (point, lines) => {
+  var radius = 10;
+  var options = {
+    steps: 10,
+    units: "meters",
+  };
```
frankhereford, line 26: I think you may want to use `radius` here instead of 10.

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+    return { type: "FeatureCollection", features };
+  }, [featureCollection, geomType]);
+
+/*
+Bit of a hack to generate intersection labels from nearby streets. 
+this relies on nearby lines being avaialble in-memory, which 
+is not guaranteed. a reliable solution would be query the AGOL streets
+layer on-the-fly to grab street names - not sure if this is worth it TBH 
+*/
```
frankhereford, line 20: Hmm, I'm torn about doing a GIS operation here in the map, but it is straightforward and certainly going to be valid as it's coming out of `turf`. I wonder what projection its using to allow it to use meters, probably web mercator (3857)?

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
```diff
@@ -0,0 +1,66 @@
...
+    const features = featureCollection.features.filter((feature) => {
+      const thisGeom = feature.geometry.type.toLowerCase();
+      return thisGeom.includes(geomType);
+    });
+    return { type: "FeatureCollection", features };
+  }, [featureCollection, geomType]);
+
+/*
+Bit of a hack to generate intersection labels from nearby streets. 
+this relies on nearby lines being avaialble in-memory, which
```
frankhereford, line 17: available

frankhereford (approved):
Talk about moving some code! This is great - thank you John & Mike!

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
mddilley: I think this is ok but I think that @johnclary can speak to this better than me

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
mddilley: Same here - I didn't look at this code that closely so I think @johnclary could respond better. The only constructive input that I have is that I also trust the output of `turf` 😅

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
mddilley: 🔍 Thanks Frank! I just fixed this in another PR after going through your feedback here. https://github.com/cityofaustin/atd-moped/commit/f0fe84003039b582f30236c6fede0064874013fe

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
johnclary: yes, thanks—i see three options (this current hack not being one of them):
1. remove this feature entirely
2. add the intersection labels to the underlying dataset w/ help from the GEO team
3. query an AGOL feature service on-the-fly to get the labels.
will make an issue 🤪

johnclary: https://github.com/cityofaustin/atd-data-tech/issues/10632

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
johnclary: it's confusing to see `10` twice. the `steps` prop refers to the number of vertices in the circle. i clearly copied from [the example](https://turfjs.org/docs/#circle) as evidenced by the use of `var`.

mddilley:
Note: this PR has received two approvals and has many branches merged into it now. Just noting in case anyone looks to review this one. It is now the destination for further work. 🧺

johnclary (approved):
🚢 !

PR #825, opened 2022-10-10
Change Funding Sources selects to autocompletes

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/10418

I also fixed a bug I had written where the tags were not filtering out correctly. 

## Testing
**URL to test:** 
https://deploy-preview-825--atd-moped-main.netlify.app/moped/projects/1887?tab=funding

**Steps to test:**

On funding tab
Add new funding source
In Source, try "Street Impact Fee" or other text in autocomplete. 
In Program "Signal" is a good thing to search as well

Test editing after submit. 

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #826, opened 2022-10-12
Change fund dropdown to autocomplete

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10418

## Testing
**URL to test:** 
https://deploy-preview-826--atd-moped-main.netlify.app/moped/projects/1815?tab=funding

**Steps to test:**

similiar to https://github.com/cityofaustin/atd-moped/pull/825
testing autocomplete in the funding table, but this time with the fund dropdown

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
@@ -332,7 +332,7 @@ const ProjectFundingTable = () => {
   );
 
   /**
-   * Component for autcomplete for Funds
+   * Autocomplete component for Funds
    * @param {*} props
    * @returns {React component}
    */
--- a/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
@@ -299,11 +299,11 @@ const ProjectFundingTable = () => {
     }, {});
   };
 
-/**
- * Component for autocomplete using a lookup table as options
- * @param {*} props
- * @returns {React component}
- */
+  /**
+   * Component for autocomplete using a lookup table as options
+   * @param {*} props
+   * @returns {React component}
+   */
   const LookupAutocompleteComponent = (props) => (
     <Autocomplete
       style={{ minWidth: "200px" }}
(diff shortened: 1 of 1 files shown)


johnclary (approved):
so, ideal world, i think `LookupAutocompleteComponent` and `FundAutocompleteComponent` should be the same component with different props, and i want to see them extracted from the main component.

i also think this table has presentation issues. the values are long and get wrapped in weird ways. i will make a backlog issue for enhancements. thank you @chiaberry! 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -333,40 +332,32 @@ const ProjectFundingTable = () => {
...
    */
-  const FundSelectComponent = (props) => (
-    <Select
-      id="moped_funds"
-      value={
-        props.value?.fund_id
-          ? `${props.value.fund_id} | ${props.value.fund_name}`
-          : ""
-      }
+  const FundAutocompleteComponent = (props) => (
```
mddilley, line 339: I can see where John is going, and I think the autocompletes could be a bit more abstracted by passing some functions as props that know how to get the options, selected option, and values for each type of data. That way the component doesn't have to be concerned about specific key names for options or values that it uses.

I ran into this the other day too and realized that our DB column names aren't always as predictable as they could be when it comes to record `id` and `name` which makes features like this tricky.

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -299,11 +296,11 @@ const ProjectFundingTable = () => {
...
- * Component for autocomplete using a lookup table as options
- * @param {*} props
- * @returns {React component}
- */
+  /**
+   * Component for autocomplete using a lookup table as options
+   * @param {*} props
+   * @returns {React component}
+   */
   const LookupAutocompleteComponent = (props) => (
```
mddilley, line 304: This could be refactored down the line but I agree that seeing these components move outside of the `ProjectFundingTable` component could be a good improvement. This is a totally a pattern that we have all used in this project, and I had a TIL moment that came from John and I chatting about anti-patterns the other day and me thinking through the pitfalls of components defined inside other components.

https://oozou.com/blog/6-react-anti-patterns-to-avoid-206

mddilley (approved):
Works just as described! This table is so powered up now! ⌨️ 🙌 🚢

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
chiaberry: ooh thanks for that blog post. I was wondering how the components defined inside other components render

frankhereford (commented):
I defer to y'all's better judgement on react performance, but I did check out the deploy preview - looks great. I couldn't get it to show me anything it shouldn't nor fail to show me something it should. Code looks great! 💯

PR #827, opened 2022-10-12
Increase number of default rows in dashboard 

chiaberry:
I was opening an issue regarding a bug i found in the team table, and spotted this snackoo. 

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10463

## Testing
**URL to test:** 

https://deploy-preview-827--atd-moped-main.netlify.app/moped/dashboard

**Steps to test:**

Check out your dashboard. If you have fewer than 15 projects, you wont see pagination. But once it goes over 15, pagination appears with the option to increase the count to 30 or 100. 

I arbitrarily chose the number 15 as the minimum count, on the signals dashboard thing we have it as 30. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/dashboard/DashboardView.js
+++ b/moped-editor/src/views/dashboard/DashboardView.js
@@ -311,11 +311,11 @@ const DashboardView = () => {
                       search: false,
                       toolbar: false,
                       tableLayout: "fixed",
-                      ...(selectedData.length < 15 + 1 && {
+                      ...(selectedData.length < 51 && {
                         paging: false,
                       }),
-                      pageSize: 15,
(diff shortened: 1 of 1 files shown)


johnclary (changes requested):
@chiaberry thank you! i think it should be very rare that the user has to encounter/deal with pagination.

we should provide enough rows for them to see what they need, and provide search/filter tools to enable the user to get every row they need on the one page.

johnclary (approved):
🙌 🚢

mddilley (approved):
Awesome to get this turned around so quickly - I like the updated default a lot! 🙌

PR #831, opened 2022-10-13
Re-implement geocoder and speed dial (#2)

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10406
https://github.com/cityofaustin/atd-data-tech/issues/10404

Notable changes:
- Reworks the speed dial styles to be set with `makeStyles` and without the CSS file imported before
- Prunes some redundant and unused speed dial styles
- Adds some constants for Mapbox-style pixel widths/heights that we often apply to map controls
- Removes some geocoder style overrides that served to make it the same width as the "Add component" form that sits below it in the current Moped map (happy to revert some of this if it makes sense to make it wider)

## Testing
**URL to test:** 

https://md-10406-geocoder-search--atd-moped-main.netlify.app/

**Steps to test:**
2. Use the geocoder search in the top left to search for "Austin City Hall" and select the result returned
3. In the bottom right corner, use the speed dial to switch from the **Streets** to the "Aerial" base map (Netlify deploy preview links are not on the approved NearMap API token referrer list so the tiles won't work, but you will see failed requests for them)
4. When **Streets** is selected, try different viewport width and see the height of the speed dial shift to keep the Mapbox copright text or mobile info icon in view. Super open to feedback on this since it maybe feels weird to have the controls shift, but this is what Mapbox does OOTB with their zoom controls so maybe it isn't that bad.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/components/Maps/GeocoderControl.js
```diff
@@ -25,8 +26,8 @@ export default function GeocoderControl(props) {
     () => {
       const ctrl = new MapboxGeocoder({
         ...props,
-        marker: false,
-        accessToken: props.mapboxAccessToken,
+        flyTo: true,
+        accessToken: mapParameters.mapboxAccessToken,
```
mddilley, line 30: Baking in some of these props since they are unlikely to change

moped-editor/src/views/projects/projectView/ProjectComponents/BaseMapSpeedDial.js
```diff
@@ -0,0 +1,163 @@
...
+   * @param basemapName
+   */
+  const onBasemapSelect = (basemapKey) => {
+    setIsSpeedDialOpen(false);
+    setBasemapKey(basemapKey);
+  };
+
+  /**
+   * Opens the speed dial menu
+   */
```
mddilley, line 84: Curious about comments like these - do we find them useful? I think the code tells the story, but I wanted to bring it up before removing them.

moped-editor/src/views/projects/projectView/ProjectComponents/EditModeDialog.js
```diff
@@ -31,7 +31,6 @@ const EditModeDialog = ({ showDialog, onClose, setIsEditingComponent }) => {
     <Dialog
       open={showDialog}
       onClose={onClose}
-      className={{ paper: classes.dialog }}
```
mddilley, line 34: Fixes a console warning about this classname not being defined

moped-editor/src/views/projects/projectView/ProjectComponents/BasemapSpeedDial.js
```diff
@@ -0,0 +1,163 @@
...
+        ? `${MAPBOX_PADDING_PIXELS}px`
+        : `${MAPBOX_PADDING_PIXELS * 3}px`,
+    // Mapbox copyright info collapses to a taller info icon at 990px and below
+    [theme.breakpoints.down(991)]: {
+      bottom: ({ basemapKey }) =>
+        basemapKey === "aerial"
+          ? `${MAPBOX_PADDING_PIXELS}px`
+          : `${MAPBOX_PADDING_PIXELS * 4}px`,
+    },
+  },
```
johnclary, line 60: i think this handles the rendering really well!

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -38,15 +42,19 @@ const useStyles = makeStyles((theme) => ({
...
+    minHeight: ({ appBarHeight }) => `calc(100vh - ${appBarHeight}px)`,
   },
   margin: {
     margin: theme.spacing(1),
   },
 }));
 
+/* per MUI suggestion - this empty toolbar pushes the list content below the main app toolbar  */
+const PlaceholderToolbar = () => <Toolbar />;
+
```
johnclary, line 54: nice touch :)

moped-editor/src/views/projects/projectView/ProjectComponents/utils.js
```diff
@@ -64,3 +65,26 @@ export function componentFormStateReducer(state, { key, value, action }) {
...
+  let currentToolbarMinHeight;
+  if (isDesktop) {
+    currentToolbarMinHeight = toolbar[toolbarDesktopQuery];
+  } else if (isLandscape) {
+    currentToolbarMinHeight = toolbar[toolbarLandscapeQuery];
+  } else {
+    currentToolbarMinHeight = toolbar;
+  }
+  return currentToolbarMinHeight.minHeight;
+}
```
johnclary, line 90: 😵 - this works really well.

makes me realize we'll want to be able to make the sidebar collapsable in the future. doesn't feel must-have. will make an issue.

johnclary (approved):
works perfectly! ran locally to test out the aerials. they look great. i think the shifting speedial position on smaller viewports is totally fine. nice work!

moped-editor/src/views/projects/projectView/ProjectComponents/BaseMapSpeedDial.js
chiaberry: i am a fan of removing redundant comments, but i dont think i docstring and comment enough so take my comment with a big grain of salt

roseeichelmann: personally i feel like its better to over comment than under comment/not comment at all. in the end its just making the code more readable and easily accessible to new eyes, so i feel like one can never comment too much. but i also had CS professors that would have us comment almost every line of code lol

roseeichelmann (approved):
i tested this with the netlify link and it looks/works great! im excited about the new and improved map component in moped, its so much less cramped. also i think it makes sense for the speedials to shift when changing the viewport width so looks good to me 🎆

moped-editor/src/views/projects/projectView/ProjectComponents/BaseMapSpeedDial.js
frankhereford: I think Rose's instinct is the right on for when the code is not lexical or difficult to read. Things like regex patterns and CIDR IP notations and like any `sed` command I've ever written - can't comment enough.

But .. for when the code is lexical and has full, verbose variables names and function names and stuff like that, then the effort that went into making those well-named symbols is actually more valuable than repeating yourself in the comments. ← This is where you want to be. Comments are for when you can't quite get there.

>There are only two hard things in Computer Science: cache invalidation and naming things.

>-- Phil Karlton

In this case, I got what i needed from `onBaseMapSelect` and would strike the comment, and would maybe consider changing the name of the parameter in the function def to basemapName, maybe?? (This change suggestion would not normally have been even close to enough of a thing to write up a comment for, BTW.)

moped-editor/src/views/projects/projectView/ProjectComponents/BasemapSpeedDial.js
```diff
@@ -0,0 +1,163 @@
...
+        <Typography>
+          <Icon
+            className={clsx(
+              classes.mapStyleToggleLabel,
+              classes.mapStyleToggleLabelIcon
+            )}
+          >
+            layers
+          </Icon>
+          <span className={classes.mapStyleToggleLabel}>Map</span>
```
frankhereford, line 106: Would you consider striking this label bit, the `Map` bit? If the stacked-layers icon were able to grow to be bigger in the speed dial, I think they would better convey what you adjust with it.

moped-editor/src/views/projects/projectView/ProjectComponents/BasemapSpeedDial.js
```diff
@@ -0,0 +1,163 @@
...
+   * Opens the speed dial menu
+   */
+  const onOpen = () => {
+    setIsSpeedDialOpen(true);
+  };
+
+  const onClose = () => setIsSpeedDialOpen(false);
+
+  return (
+    <SpeedDial
```
frankhereford, line 92: The speed dial opens `onMouseover`, and it opens/closes onClick. For those whose instinct is to click on it, you end up opening on mouseover and then immediately closing it onClick. What would you think about disabling the click handler entirely, provided you get a little grace to mouse from the speed dial to the basemap square to make your choice.

frankhereford (approved):
I looked at that height shift, and I think it looks natural. Thanks for pointing it out though.

I have a little thought on how to toggle the speed dial state, but don't let it hold you up - it's very minor.
⛴🚢

moped-editor/src/views/projects/projectView/ProjectComponents/BasemapSpeedDial.js
mddilley: I agree with this but it is the default Material UI behavior so I'll stick with it unless we get some user feedback about it. TIL because I always thought more than the appearance was customized when this component was originally written. https://mui.com/material-ui/react-speed-dial/

moped-editor/src/views/projects/projectView/ProjectComponents/BasemapSpeedDial.js
mddilley: I like this a lot. Going to push an update since I agree that the **Map** text doesn't add a lot of meaning here.

moped-editor/src/views/projects/projectView/ProjectComponents/BaseMapSpeedDial.js
mddilley: Thanks, y'all! 🙏 I removed the "Opens the speed dial menu" comment but kept the other one and filled in the docstring along with fixing a typo

PR #834, opened 2022-10-13
Fixes team member role editing

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10474

This PR adds a new table, `moped_proj_personnel_roles` which enables a normalized many-to-many relationship between project team members (`moped_proj_personnel`) and project roles (`moped_project_roles`).

Prior to this change, we actually stored duplicate personnel records, one per project role, and concatenated their details into a single table row in the editor. This PR includes a migration script that de-dupes the personnel records and creates their `moped_proj_personnel_roles`.

Once this has been merged, we'll need to create a separate migration that removes the depreacted `role_id` column from `moped_proj_personnel`.

A few other UI tweaks:
- show role descriptions in menu options
- autofocus team member input after adding a row
- prevent same team member from being added to the table twice

## Testing
**URL to test:**  Local

**Steps to test the editor UI:**
1. Add a team member to the project team with an **Inspector** role.
2. Verify their workgroup populates once the row is saved.
3. Edit the team member to remove their **Insepctor** role and add **Designer** and **Project Owner** roles. Verify the changes save correctly
4. Add another team member—make them an **Inspector**.
5. Take note of the names of your Inspector and Designer. Navigate to the Projects list, and verify these names populate in their respective columns on the team table. 
6.  Use the advanced search to search for your project based on team members you've added to your project.
7. Delete one of the team members from your project.
8. Verify the deleted team member is removed from the Team tab and the Projects list
9. Stop the Hasura cluster

**Steps to test the migration script:**
1. Start the hasura cluster
2. Activate a python env with `requests` installed
3. Navigate to `/moped-toolbox/migrate_personnel`
4. `python run.py -e local`
5. In the editor, check a few projects and observe that the team members look normal.
6. Edit these existing team members and confirm that they work normally.

Pre-merge todos:
- [ ] Verify that `not null` constraint migrations will succeed in staging and prod
- [ ] Update seed data

---
#### Ship list
- [x] Code reviewed 


moped-database/migrations/1665759117044_alter_table_public_moped_proj_personnel_alter_column_user_id/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."moped_proj_personnel" alter column "user_id" set not null;
+alter table "public"."moped_proj_personnel" alter column "project_id" set not null;
```
johnclary, line 2: we weren't enforcing it, but these columns should never be null. i need to double-check staging and prod to make sure these migrations will pass.

moped-editor/src/queries/project.js
```diff
@@ -192,19 +206,48 @@ export const TEAM_QUERY = gql`
...
+      affected_rows
+    }
+  }
+`;
+
+export const DELETE_PROJECT_PERSONNEL = gql`
+  mutation DeleteProjectPersonnel($id: Int!) {
+    update_moped_proj_personnel_by_pk(
+      pk_columns: { project_personnel_id: $id }
+      _set: { is_deleted: true }
```
johnclary, line 246: if we delete a team member (`moped_proj_personnel`), we do not delete their team member roles.

that seems like the right way to do this? if for some reason we wanted to revert a team member delete, its roles state would be intact.

moped-editor/src/queries/project.js
```diff
@@ -192,19 +206,48 @@ export const TEAM_QUERY = gql`
...
+    update_moped_proj_personnel_roles(
+      where: { id: { _in: $deleteIds } }
+      _set: { is_deleted: true }
     ) {
       affected_rows
     }
+    insert_moped_proj_personnel_roles(objects: $addRolesObjects) {
+      affected_rows
+    }
+  }
```
johnclary, line 239: this mutation does three things:
- updates one `update_moped_proj_personnel` - aka a team member
- deletes any team member roles specified in the array of `deleteIds`
- adds new team member roles

it _might_ be possible to this all with a nested mutation upsert, but i think we'd have to dynamically build the graphql query string.

moped-editor/src/queries/project.js
```diff
@@ -823,25 +866,25 @@ export const PROJECT_CLEAR_WORK_ASSIGNMENT_ID = gql`
...
-    _set: {interim_project_id: $interimProjectId}
-  ) {
-    interim_project_id
+  mutation UpdateProjectInterimId($projectId: Int!, $interimProjectId: Int) {
+    update_moped_project_by_pk(
+      pk_columns: { project_id: $projectId }
+      _set: { interim_project_id: $interimProjectId }
+    ) {
+      interim_project_id
+    }
```
johnclary, line 875: linting

johnclary:
Folks, this PR is ready for review, except I need to rebuild the seed data 😬. I'll chat with y'all next week about the best way to do this.

moped-editor/src/views/projects/projectView/ProjectTeamRoleMultiselect.js
```diff
@@ -71,15 +59,29 @@ const ProjectTeamRoleMultiselect = ({
...
+                <Checkbox checked={isChecked} color={"primary"} />
+                <ListItemText
+                  primary={<span>{project_role_name}</span>}
+                  secondary={
+                    <span style={{ whiteSpace: "normal" }}>
+                      {project_role_description}
+                    </span>
+                  }
+                />
+              </MenuItem>
```
johnclary, line 81: I'd like to know how folks feel about including the role description in the menu list item. We've gotten recent user feedback that it would be helpful to provide this kind of context, but this menu gets pretty cramped.

Screen Shot 2022-10-17 at 11 15 06 AM

moped-toolbox/migrate_personnel/run.py
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+"""
+
+
+def get_personnel_index(personnel):
+    index = {}
+    for p in personnel:
+        key = f"{p['project_id']}_{p['user_id']}"
+        index.setdefault(key, [])
```
chiaberry, line 79: TIL

moped-editor/src/views/projects/projectView/ProjectTeamRoleMultiselect.js
chiaberry: I initially like the descriptions in there, but dont know if once people get more familiar with the project roles, if it will feel like a pain to have to scroll down to find the role they want. Agree if it were only autocomplete.....

moped-database/metadata/tables.yaml
```diff
@@ -1,16 +1,16 @@
 - table:
-    schema: public
     name: moped_activity_log
+    schema: public
```
chiaberry, line 3: it bums me out when it changes the order thus making a lot of nonchanges

moped-database/metadata/tables.yaml
```diff
@@ -2172,18 +2180,77 @@
...
+          - project_role_id
+        filter: {}
+    - role: moped-editor
+      permission:
+        columns:
+          - is_deleted
+          - id
+          - project_personnel_id
+          - project_role_id
+        filter: {}
```
chiaberry, line 2233: are we going to skip setting permissions for moped-viewers since we aren't allowing anyone to be set as a viewer now?

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -321,14 +304,14 @@ const ProjectTeamTable = ({ projectId }) => {
...
-        data={makeListOfActivePersonnel(data.moped_proj_personnel)}
+        data={personnel}
         title={
           <Typography variant="h2" color="primary">
             Project team
           </Typography>
         }
         options={{
-          ...(data.moped_proj_personnel.length < PAGING_DEFAULT_COUNT + 1 && {
+          ...(data?.moped_proj_personnel?.length < PAGING_DEFAULT_COUNT + 1 && {
```
chiaberry, line 314: I don't think this length should be different from the `personnel` Array, but just in case, would you consider checking pagination on `personnel` instead?

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
```diff
@@ -350,96 +333,35 @@ const ProjectTeamTable = ({ projectId }) => {
...
-                    : false,
-                  notes: index === 0 ? newData.notes : "",
-                };
-              }
-            );
-
-            return upsertProjectPersonnel({
+          onRowUpdate: async (newData, oldData) => {
+            const { project_personnel_id } = oldData;
+            const payload = getEditPersonnelPayload({
```
chiaberry, line 345: i like this pattern of moving what happens out into its own separate function

chiaberry (approved):
Can't find a way to break it yet! left a note about pagination but otherwise 👨🏼‍🍳 💋

moped-editor/src/views/projects/projectView/ProjectTeamTable.js
johnclary: thanks for flagging this - i removed pagination entirely. this is the only content on the whole tab. if someone wants to add 100 team members, fine.

moped-database/metadata/tables.yaml
johnclary: no—thank you for catching this! i added select permissions for viewers.

moped-database/metadata/tables.yaml
johnclary: ugh, yes. i thought we were in a better place now. should i make a clean diff for y'all?

moped-toolbox/migrate_personnel/run.py
```diff
@@ -0,0 +1,151 @@
...
+        raise ValueError(data)
+
+
+PROJ_PERSONNEL_QUERY = """
+  query GetProjectPersonnel {
+    moped_proj_personnel(
+      where: {
+        _and: [
+          { is_deleted: { _eq: false } }
+          { _not: { moped_proj_personnel_roles: {} } }
```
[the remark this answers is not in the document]
johnclary: 👍 so as to take zero credit for it, [docs](https://hasura.io/docs/latest/queries/postgres/query-filters/#fetch-if-nested-objects-existdo-not-exist).

johnclary:
i just pushed a few small changes:
- nix pagination entirely
- prevent adding a team member to the table if they're already added
- Autofocus the user name picker when you add a new table row ✨ 
- Rename the `id` and `name` props of the user Autocomplete, b/c 1password was picking it up as a login field.

moped-toolbox/migrate_personnel/run.py
johnclary: was just looking for a js equivalent and i gotta say they are [pretty gross / clever](https://stackoverflow.com/questions/16107680/javascript-equivalent-of-pythons-dict-setdefault)

```javascript
d.key ??= value
```

one of those rare occasions where python has javascript beat wrt to the object API??

moped-editor/src/views/projects/projectView/ProjectTeamRoleMultiselect.js
johnclary: @amenity

johnclary:
I have merged main into this branch. I have also confirmed that there are no `moped_proj_personnel` records with a `null` `project_id` or `user_id`. 

I will now merge this branch into main.

moped-toolbox/migrate_personnel/secrets.py
```diff
@@ -0,0 +1,8 @@
+HASURA_AUTH = {
+    "hasura_graphql_endpoint": {
+        "local": "http://localhost:8080/v1/graphql",
+    },
+    "hasura_graphql_admin_secret": {
+        "local": "DuMmyApiKeyHFVOVto19otC1wX6sP2N0VSKrCD70L10B7Sm525ZR6L672i2F79M9!",
+    },
+}
```
frankhereford, line 8: I wrote up [some comments for Rose](https://github.com/cityofaustin/atd-vz-data/pull/1118#discussion_r1001949543) in a PR the other day that I think apply here as well. But otherwise, I think this is fine, as it was with hers, as this is in the toolbox, but in the future, a template of the secrets file may be safer.

frankhereford (approved):
I've looked at this quicky, so please take me with a grain of salt, but I think this looks good.

moped-toolbox/migrate_personnel/secrets.py
johnclary: thanks @frankhereford - i went ahead and ignored this file for all toolbox scripts 🔒 
https://github.com/cityofaustin/atd-moped/pull/834/commits/37ecba2c88307f947aad8b3501a0eb402ec00d97

johnclary:
sorry for all the noise—i noticed a console warning barking b/c of my mishandling of autocomplete options. [fixed that](https://github.com/cityofaustin/atd-moped/pull/834/commits/fe0d56de50c39333124804b5701fc60b237a47d9). moving on!

johnclary:
i ran the migration script against staging—it was successful. i added a reminder to the [v1.11 release issue](https://github.com/cityofaustin/atd-data-tech/issues/10499) to run this in production post-merge.

moped-editor/src/views/projects/projectView/ProjectTeamRoleMultiselect.js
patrickm02L: Oops, didn't see this earlier. The position title and description text work as visualized. 

My thought would have been to YOLO widen that drop down menu to the right for more real estate. Sure it would cover up the notes sections temporarily, but the "Notes" title would still be visible and would come back once finished selected.

PR #835, opened 2022-10-13
addresses project summary UI shortcomings

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10419

## Testing
**URL to test:** 

https://10419_address_ui_shortcomings_status--atd-moped-main.netlify.app/moped/

**Steps to test:**
go to a project summary page, clear all fields and add inputs to all the fields! add/edit a status update

---
#### Ship list
- [x] Code reviewed 


tillyw:
@johnclary @frankhereford  I have one last question! in order to change the 'project sponsor' blank/unselected state to '-' from 'none' I have to change the `entity_name` of the record with id `0` in `moped_entity`. i can write a script to include with this pr but just wanted to make sure that's not going to cause unwanted problems?

johnclary:
@tillyw i think the use of that `None` entity is a flaw in our original DB implementation. In cases where the project does not have a sponsor, we should just use a `null` value. let's sync on the best way to rectify this?

tillyw:
> @tillyw i think the use of that `None` entity is a flaw in our original DB implementation. In cases where the project does not have a sponsor, we should just use a `null` value. let's sync on the best way to rectify this?

thanks @johnclary as per our discussion i'm going to just stick with this approach and then we will address the db improvements in a separate issue. thanks!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -75,46 +76,45 @@ const ProjectSummaryProjectDescription = ({
...
+
+            <IconButton onClick={handleProjectDescriptionClose}>
+              <Icon>close</Icon>
+            </IconButton>
           </>
         )}
         {!editMode && (
           <ProjectSummaryLabel
-            text={description || "None"}
+            text={description || " - "}
```
frankhereford, line 117: Please ignore me if I'm an outlier with this opinion, but I feel like "None" is more informative than `-` and is clearer to the user.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectWebsite.js
```diff
@@ -113,15 +114,14 @@ const ProjectSummaryProjectWebsite = ({
...
-            className={classes.fieldLabelLink} // Override
             text={
               (website && website.length > 0 && (
                 <Link href={website} target={"_blank"}>
                   {website}
                   <OpenInNew className={classes.linkIcon} />
                 </Link>
               )) ||
-              "None"
+              ""
```
frankhereford, line 124: I find that I get displayed a `-` when I set the website to the empty string. Are we storing a website URL as the empty string when we mean `null`? Maybe I'm confused about where the `-` is coming from for this field.

frankhereford (approved):
I think it looks great! I bring up a few comments below about the data model for non-relationships and the use of `-` as a value label, but they are solidly outside the scope of this issue. This is ready to head out to sea -- ship it!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
roseeichelmann: personally i like " - ", it feels a lot cleaner to me from a design perspective and indicates that the field is blank/hasn't been filled in yet.

roseeichelmann (approved):
i love these changes, the summary page is looking and feeling so elegant! i have one super minor peeve about the status update and how the user/date that shows underneath draws my attention more than the status itself and feels more like a newline than a signature of the commenter, i wonder if anyone else feels this way? i thought indentation/italicizing like the screenshot below would help differentiate, maybe doesn't even need the dash, but once again this is just a minor design peeve. this looks great and is totally ready to ship, nice job!! ✨

Screenshot from 2022-10-21 12-31-24

Vs current:
Screenshot from 2022-10-21 12-31-34

moped-database/migrations/1666289149698_rename moped_entity none to dash/up.sql
```diff
@@ -0,0 +1 @@
+UPDATE "public"."moped_entity" SET entity_name = '-' WHERE entity_id = 0;
```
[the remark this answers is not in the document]
johnclary: yes—i agree. Tilly and I sync'd on it briefly. I will make a separate issue!

johnclary (changes requested):
This is looking really close @tillyw! A few requests:

Vertical spacing: i think we need clearer separation between the fields on the page. Can you knock a few points of vertical padding off of the gray input backround? anything to bring the input values closer to their labels. similarly, can you increase the vertical spacing between each row to be closer to what's in production today?

I agree with @roseeichelmann that the status update styles  need a little work, but let's punt that to a separate issue. I'd like to work in [humanized duration](https://evanhahn.github.io/HumanizeDuration.js/) along with it.

Make the status update input full-width, to make it match the rest of the form inputs on the page.

Remove the hover/pointer effect from the "Signal IDs" (knack data tracker sync) input. Users should click the URL. 

Move the "Signal IDs" / Data Tracker sync field to the bottom of the list, with only "ProjectSummaryWorkOrders"  below it.

Lastly, configure `autoFocus` for the sponsor, partners, and project type inputs. For sponsor and partners, you should be able to mirror the props [here](https://github.com/cityofaustin/atd-moped/blob/498d7db2c2ff76d0e079cf14a64bcacdf13a5664/moped-editor/src/views/projects/projectView/ProjectTeamTable.js#L209). Autofocusing project type might prove to be tricky, in which case we can shelve it.

tillyw:
thanks, all! @johnclary i'll get started on your requested changes and rerequest review. @roseeichelmann thanks for pointing out that design improvement! i agree and look forward to addressing that in a future issue. @frankhereford i also agree with you! look forward to addressing the database issues you raised in a future issue.

moped-database/migrations/1666289149698_rename moped_entity none to dash/up.sql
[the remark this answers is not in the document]
frankhereford: @johnclary I missed your comment about making the issue. I may have made a duplicate issue - https://github.com/cityofaustin/atd-data-tech/issues/10594. Please feel free to close it if it's not helpful. Thanks!

tillyw:
> 

I believe I addressed all your comments! rerequesting review.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -195,7 +196,7 @@ const ProjectSummaryProjectECapris = ({
                   </Link>
                 </>
               )) ||
-              "None"
+              ""
```
chiaberry, line 199: This component is also used on the funding tab to edit the ecapris ID, (https://10419_address_ui_shortcomings_status--atd-moped-main.netlify.app/moped/projects/228?tab=funding). I think you wanted to get rid of `None` everywhere, but just wanted to point this change out just i case it wasnt expected @johnclary

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectWebsite.js
chiaberry: I was curious too, but in the External Link component, if theres no text it renders as `-`

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
johnclary: ah hah—and now I'm remembering why we originally switched from " - " to "None". the eCapris ID field on the funding tab is very hard to edit now. @tillyw — this the absolute last thing — would you mind adding the same styles to the eCapris ID field on the funding tab, but don't make it full width? iirc we use `8em` as an input width in some places?

johnclary (changes requested):
This is so ready to ship but for the issue with the eCapris input on the funding tab :/

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -75,46 +76,45 @@ const ProjectSummaryProjectDescription = ({
...
    */
-  const handleProjectDescriptionChange = e => {
+  const handleProjectDescriptionChange = (e) => {
     setDescription(e.target.value);
   };
 
   return (
     <Grid item xs={12} className={classes.fieldGridItem}>
-      <Typography className={classes.fieldLabel}>Project description</Typography>
+      <Typography className={classes.fieldLabel}>Description</Typography>
```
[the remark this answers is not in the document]
johnclary: Yeah! Chia just stripped it out of the project list as well.

tillyw:
all this makes sense - thanks for catching that @chiaberry! will implement a fix shortly

johnclary (approved):
🙌 🏁 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #837, opened 2022-10-17
Add street labels over nearmap aerial imagery (#4)

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10405

## Testing
**URL to test:** 

This is best tested locally since the Netlify preview links are not on the approved list for the Moped NearMap token

**Steps to test:**
1. Go to an existing or new project's **Map** tab
2. Using the speed dial, select the aerial basemap
3. You should see the NearMap tiles pulling in with the Mapbox street labels presented on top
4. You can also try a viewport 990px or below and see that the speed dial moves above the Mapbox info icon that now it appears on the aerial view since we are using their street labels.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
@@ -1,4 +1,5 @@
 const MAPBOX_TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
+// This API key is managed by CTM. Contact help desk for maintenance and troubleshooting.
 const NEARMAP_KEY = process.env.REACT_APP_NEARMAP_TOKEN;
 
 // minimum map zoom to display selectable features, and
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
@@ -71,3 +71,95 @@ export const basemaps = {
     ],
   },
 };
+
+// Use a custom hook that manages basemap and the
+
+// The config from VZE
+export const LOCATION_MAP_CONFIG = {
+  mapStyle: "mapbox://styles/mapbox/satellite-streets-v11",
+  sources: {
+    aerials: {
+      id: "raster-tiles",
+      type: "raster",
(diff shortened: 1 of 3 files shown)


johnclary (approved):
@mddilley looks great!

i'm noting that the aerials render on top of the components layer, making that layer invisible. but we know that's a separate problem to tackle, see https://github.com/cityofaustin/atd-data-tech/issues/10407

this also has me thinking that it would be nice to alway have the street labels on top of everything else on the map, including project components. to do that, we would want to hide the street labels in our streets basemap, otherwise they would be duplicated. i'll file an issue away for that one.

🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
```diff
@@ -45,29 +46,101 @@ export const mapParameters = {
   mapboxAccessToken: MAPBOX_TOKEN,
 };
 
+/*
+ * This configuration sets the mapStyle, sources, and layers associated with basemaps
+ * mapStyle becomes the value of the mapStyle prop of the react-map-gl map
+ * https://visgl.github.io/react-map-gl/docs/api-reference/map#mapstyle
```
frankhereford, line 52: Do you want a period here after the first line? Even with that, the second line is unclear to me.

How's this: The `react-map-gl` map's prop `mapStyle` is used.

frankhereford: Also, do you want this comment in the `TheMap` component js file?

frankhereford (approved):
I left a comment that really is more of a question. If I'm confused about react language, please resolve the conversation! 

Tested using your instructions, and it worked like a charm. 🍀

moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
mddilley: I totally agree that the wording is kind of wonky so I updated it to explicitly say that the k/v pair is used to pass a value to  the `mapStyle` prop of the map. Thanks!

PR #839, opened 2022-10-20
Manage dynamic interactive layer IDs and layers (#5)

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10407

Changes:
- Updates from conditionally rendering Mapbox sources and layers with state conditionals to conditionally toggling [layer visibility ](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-background-visibility)
   - https://docs.mapbox.com/mapbox-gl-js/example/toggle-layers/
- Adds `isInteractive` boolean to the layers config so we can generate a list of [interactive layer IDs](https://visgl.github.io/react-map-gl/docs/api-reference/map#interactivelayerids) from it

## Testing
**URL to test:** 

https://md-10407-interactive-layers--atd-moped-main.netlify.app/ or local

**Steps to test:**
2. Click **New Component** and select the type called **Access Control - Driveway Closure**
3. Click **Continue**, zoom the map in until you see the dashed lines, and select a couple line segments
4. You should see the street labels above the selecting line segments like this example

Screen Shot 2022-10-21 at 1 38 49 PM

5. Change the base map to aerial and you should see the street labels remain above your selected segments

Screen Shot 2022-10-21 at 1 39 28 PM

6. Test out removing segments by clicking them too and then click **Save**
7. Click on a line segment of your saved component and you should see a pop up appear with details about the component and a link to edit or delete

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -216,11 +216,7 @@ export default function TheMap({
     projectPoints,
   };
 
-  const { interactiveLayerIds } = makeInteractiveIds({
-    isEditingComponent,
-    linkMode,
-    draftLayerId,
-  });
+  const interactiveLayerIds = makeInteractiveIds();
 
   return (
     <MapGL
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -14,8 +14,8 @@ import {
 import { getIntersectionLabel, useFeatureTypes } from "./utils";
 import { useFeatureService } from "./agolUtils";
 import {
-  useBasemapLayers,
-  useInteractiveIds,
+  BaseMapSourceAndLayers,
+  makeInteractiveIds,
   useComponentFeatureCollection,
   ProjectComponentsSourcesAndLayers,
 } from "./mapUtils";
@@ -216,9 +216,7 @@ export default function TheMap({
     projectPoints,
   };
 
-  const { BaseMapSourceAndLayers } = useBasemapLayers({ basemapKey });
-
-  const { interactiveLayerIds } = useInteractiveIds({
+  const { interactiveLayerIds } = makeInteractiveIds({
     isEditingComponent,
     linkMode,
     draftLayerId,
@@ -240,7 +238,7 @@ export default function TheMap({
     >
       <BasemapSpeedDial basemapKey={basemapKey} setBasemapKey={setBasemapKey} />
       <GeocoderControl position="top-left" marker={false} />
-      <BaseMapSourceAndLayers />
+      <BaseMapSourceAndLayers basemapKey={basemapKey} />
       <ProjectComponentsSourcesAndLayers
         data={data}
         isEditingComponent={isEditingComponent}
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -13,7 +13,12 @@ import {
 } from "./mapSettings";
 import { getIntersectionLabel, useFeatureTypes } from "./utils";
 import { useFeatureService } from "./agolUtils";
-import { useBasemapLayers, useMapLayers } from "./mapUtils";
+import {
+  useBasemapLayers,
+  useInteractiveIds,
+  useComponentFeatureCollection,
+  ProjectComponentsSourcesAndLayers,
+} from "./mapUtils";
 import "mapbox-gl/dist/mapbox-gl.css";
 
 // See https://github.com/visgl/react-map-gl/issues/1266#issuecomment-753686953
@@ -77,6 +82,9 @@ export default function TheMap({
 
   const draftComponentFeatures = useDraftComponentFeatures(draftComponent);
   const draftLayerId = `draft-component-${linkMode}`;
+
+  const componentFeatureCollection =
+    useComponentFeatureCollection(clickedComponent);
 
   // yeah, these props are mess :/
   const ctnLinesGeojson = useFeatureService({
@@ -210,14 +218,11 @@ export default function TheMap({
 
   const { BaseMapSourceAndLayers } = useBasemapLayers({ basemapKey });
 
-  const { interactiveLayerIds, ProjectComponentsSourcesAndLayers } =
-    useMapLayers({
-      data,
-      isEditingComponent,
-      linkMode,
-      draftLayerId,
-      clickedComponent,
-    });
+  const { interactiveLayerIds } = useInteractiveIds({
+    isEditingComponent,
+    linkMode,
+    draftLayerId,
+  });
 
   return (
     <MapGL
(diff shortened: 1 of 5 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -0,0 +1,256 @@
...
+  // This is a temporary to get data into the map sources
+  const {
+    ctnLinesGeojson,
+    ctnPointsGeojson,
+    projectLines,
+    projectPoints,
+    draftComponentFeatures,
+  } = data;
+
+  return (
```
mddilley, line 86: We could update this in the future to generate sources and layers from the `MAP_STYLES` config

moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js
```diff
@@ -69,13 +69,14 @@ export const basemaps = {
...
-        id: "simple-tiles",
+        id: "aerial-tiles",
         type: "raster",
-        source: "raster-tiles",
+        source: "aerial-tiles",
         minzoom: 0,
         maxzoom: 22,
       },
       streetLabels: {
+        id: "street-labels",
```
mddilley, line 79: Adding this id here so that we can target it with the [`beforeId`](https://visgl.github.io/react-map-gl/docs/api-reference/layer#beforeid) prop of other layers so that these street labels are always on top

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -208,20 +208,22 @@ export default function TheMap({
...
+  // This is a temporary to get data into the map sources
+  const data = {
+    draftComponentFeatures,
+    ctnLinesGeojson,
+    ctnPointsGeojson,
+    projectLines,
+    projectPoints,
+  };
+
+  const interactiveLayerIds = makeInteractiveIds();
```
johnclary, line 220: i was going to suggestion memoizing this, but, since it's completely static, you might just declare `interactiveLayerIds` in `mapUtils` and import it?

moped-editor/src/views/projects/projectView/ProjectComponents/mapStyleSettings.js
```diff
@@ -18,240 +18,278 @@ export const COLORS = {
...
-    _featureIdProp: "CTN_SEGMENT_ID",
-    type: "line",
-    paint: {
-      "line-width": 20,
-      "line-color": COLORS.white,
-      "line-opacity": 0,
-    },
-    layout: {
-      "line-cap": "round",
+    isInteractive: false,
```
johnclary, line 108: @mddilley in the previous version, we were using the project and ctn "underlays" as the interactive layers. these invisible layers just help with the mouse interaction by making the clickable area of each line wider. it feels like a bit of a hack, but i definitely prefer the click interaction when we use them.

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -0,0 +1,256 @@
...
+          ...basemaps.aerial.layers.streetLabels,
+          layout: {
+            ...basemaps.aerial.layers.streetLabels.layout,
+          },
+        }}
+      />
+    </>
+  );
+};
+
```
johnclary, line 59: Nice—so clean to have a basemap component like this.

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
johnclary: 👍 i think it's pretty great to have it all laid out visually like this!

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -233,101 +235,15 @@ export default function TheMap({
...
-        </Source>
-      )}
-      {basemapKey === "aerial" && (
-        <>
-          <Source {...basemaps[basemapKey].sources.aerials} />
-          <Layer {...basemaps[basemapKey].layers.aerials} />
-          {/* show street labels on top of other layers */}
-          <Layer {...basemaps[basemapKey].layers.streetLabels} />
-        </>
-      )}
```
johnclary, line 330: conditional rendering hurts my brain && this is a major improvement

johnclary (changes requested):
@mddilley this is looks super sharp! and it came with the treat of having street labels over the project features!!!!

i requested one small change to set the "underlay" layers to interactive.

🙌

moped-editor/src/views/projects/projectView/ProjectComponents/mapStyleSettings.js
mddilley: That makes perfect sense about why I felt like I made it harder to select segments somehow. 🙃 I'll gladly add that back in. 🙏

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
mddilley: Just made this change because I totally agree 👍

johnclary (approved):
🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -0,0 +1,251 @@
...
+      {/* Always show street labels so they can be the "target" of beforeId 
+      and always appear on top of everything else */}
+      <Layer
+        {...{
+          ...basemaps.aerial.layers.streetLabels,
+          layout: {
+            ...basemaps.aerial.layers.streetLabels.layout,
+          },
+        }}
+      />
```
frankhereford, line 50: I agree with John that it's much cleaner to have the labels in a component like this, but I am concerned about their legibility on the street layer. 

image

This white-with-black-stroke is on point for aerial imagery,  but I find it hard to read on the much lighter street layer. 

If there's agreement, would it be something that could be addressed via conditional styling of the labels or maybe a seperate layer defined for them and used conditionally?

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
frankhereford: My weakly held opinion here agrees with John. Until we need to export out this configuration as JSON or otherwise interact with this layout, this is super easy to read and work on.

frankhereford (approved):
I think this looks great, and it's ready to ship. ⛴

This is literally the most minor thing, please don't sweat this, but I wanted to point it out. When you click on a street segment toggle it, you can't click again to re-toggle _until you first move your mouse._ I bet a dime this is something inside the mapbox library itself, but it's not quite what one would expect in terms of behavior, I think. Verrrrry minor. 🐁

mddilley:
@frankhereford I think that the bug that you described with having to move the mouse before deselecting a segment is worth a bug issue. Thanks! https://github.com/cityofaustin/atd-data-tech/issues/10588 cc @johnclary

tillyw (approved):
tested and works as expected!

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
mddilley: @frankhereford I agree on this the more and more that I work on this. It makes my eyes feel unfocused 👓 so I'll create an issue to get the labels from the streets layer and show those conditionally. Thank you! 🙏

PR #841, opened 2022-10-24
Work around Mapbox's clipped geojson features (#6)

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10402

This PR updates the component draft mode to capture geometry from the same AGOL data that is populated in the map to show what lines and points are available to add to a component. Previously, we grabbed that geometry from a click event in the map library which led to the possibility of clipping geometries. Now, we click a feature, get the feature's unique ID, and then find the record within the features returned from AGOL to use its geometry and properties to eventually persist the component.

## Testing
**URL to test:** 

https://md-10402-clipped-geojson--atd-moped-main.netlify.app/moped/

**Steps to test:**

**This PR update doesn't change the way the UI works so this is mostly verifying that points and lines are still selectable.**

2. Create a new component with the type of **Access Control - Driveway Closure** and click **Continue**
3. Select some line segments and click **Save** to add the component
4. Repeat with a second component with type of **Bike box** but select some points this time and then **Save**
5. You should be able to select the components that you created from the sidebar menu or click on them on the map and then click the button in the popup to highlight a component
6. Check to make sure that you cannot click the greyed out components when highlighting a component

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -216,11 +216,7 @@ export default function TheMap({
     projectPoints,
   };
 
-  const { interactiveLayerIds } = makeInteractiveIds({
-    isEditingComponent,
-    linkMode,
-    draftLayerId,
-  });
+  const interactiveLayerIds = makeInteractiveIds();
 
   return (
     <MapGL
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -114,6 +114,7 @@ export default function TheMap({
   };
 
   const onClick = (e) => {
+    console.log(e.features);
     if (e.features.length === 0) {
       // clear clickedComponent if map is clicked elsewhere
       if (clickedComponent) {
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -70,6 +70,9 @@ export default function TheMap({
   setIsFetchingFeatures,
 }) {
   const [cursor, setCursor] = useState("grap");
+  // Move this state to the parent component?
+  // Move useAgolFeatures to the parent component and then move isFetchingFeatures to parent?
+  // Then setIsFetchingFeatures can be in the custom hook too
   const [bounds, setBounds] = useState();
   const [basemapKey, setBasemapKey] = useState("streets");
   const projectFeatures = useProjectFeatures(components);
@@ -80,16 +83,12 @@ export default function TheMap({
   const componentFeatureCollection =
     useComponentFeatureCollection(clickedComponent);
 
+  const currentZoom = mapRef?.current?.getZoom();
   const {
     ctnLinesGeojson,
     ctnPointsGeojson,
     findFeatureInAgolGeojsonFeatures,
-  } = useAgolFeatures({
-    linkMode,
-    setIsFetchingFeatures,
-    mapRef,
-    bounds,
-  });
+  } = useAgolFeatures(linkMode, setIsFetchingFeatures, currentZoom, bounds);
 
   const projectLines = useFeatureTypes(projectFeatures, "line");
   const projectPoints = useFeatureTypes(projectFeatures, "point");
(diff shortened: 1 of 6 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/mapStyleSettings.js
```diff
@@ -93,7 +93,7 @@ export const MAP_STYLES = {
   "project-lines-muted": {
     isInteractive: false,
     layerProps: {
-      id: "project-lines",
+      id: "project-lines-muted",
```
mddilley, line 96: Fixing two ID mismatches here that were causing multiple layers with the same ID and some issues with map click events

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -104,62 +108,73 @@ export const ProjectComponentsSourcesAndLayers = ({
...
               ...mapStyles["ctn-lines"].layerProps.layout,
-              visibility:
-                isEditingComponent && linkMode === "lines" ? "visible" : "none",
+              visibility: isEditingLines ? "visible" : "none",
             },
           }}
         />
       </Source>
 
       <Source
```
mddilley, line 117: The diff is a little deceptive here - I switched the order of two sources and layers (`project-lines` and `ctn-points`) so that points appear above lines

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -104,62 +108,73 @@ export const ProjectComponentsSourcesAndLayers = ({
...
+            },
+          }}
         />
         <Layer
           beforeId="street-labels"
-          {...mapStyles[
-            clickedComponent || isEditingComponent
-              ? "project-lines-muted"
-              : "project-lines"
-          ].layerProps}
```
mddilley, line 162: I didn't catch this conditional render in the last PR so I split this into two layers (`project-lines-muted` and `project-lines`) and toggle the visibility just like other layers.

moped-editor/src/views/projects/projectView/ProjectComponents/mapUtils.js
```diff
@@ -171,11 +186,22 @@ export const ProjectComponentsSourcesAndLayers = ({
       >
         <Layer
           beforeId="street-labels"
-          {...mapStyles[
-            clickedComponent || isEditingComponent
-              ? "project-points-muted"
-              : "project-points"
-          ].layerProps}
```
mddilley, line 178: Another conditional render that is now split into two layers with separated visibility logic

roseeichelmann (approved):
I tested on the netlify preview and it works just as described! Also this is a side note and has nothing to do with this PR but thought I would mention it here - the street labels on the Mapbox streets layer are kinda hard for me to read and I imagine may be hard for some of our users too. I wonder if the font style can be adjusted to be more vision-friendly or is it just an out of the box thing?

mddilley:
@roseeichelmann Thanks for the review and feedback, and the labels can definitely be updated. Glad to have more support to get https://github.com/cityofaustin/atd-data-tech/issues/10628 done before releasing this into the wild. 🙏 Frank found them hard to read as well.

frankhereford (approved):
This looks great. I tested using the instructions you provided and also went over the diffs. Thank you for the helpful comments. Everything is becoming more and more legible and organized the more you work on it - it's great.

moped-editor/src/views/projects/projectView/ProjectComponents/agolUtils.js
```diff
@@ -51,6 +52,67 @@ const featureReducer = (geojson, { features, featureIdProp }) => {
...
+        (feature) => feature.properties.CTN_SEGMENT_ID === clickedFeatureId
+      );
+    } else if (linkMode === "points") {
+      const clickedFeatureId = clickedFeature.properties.INTERSECTION_ID;
+
+      return ctnPointsGeojson.features.find(
+        (feature) => feature.properties.INTERSECTION_ID === clickedFeatureId
+      );
+    }
+  };
```
johnclary, line 107: 🙌 this looks good and works well. fwiw, i don't think we need to worry about our points layers being clipped, but if we for some reason moved our reference points into a to a multipoint geometry, it's possible clipping would occur. which is to say, i'm cool with playing it safe and handling clipping for points and lines.

@mddilley what do you think about nixing the hard-coded ID property - e.g. `clickedFeature.properties.CTN_SEGMENT_ID` in favor of looking that up from the configuration?

also, i kind of want to see this wrapped in `useCallback` or perhaps modified to accept the AGOL geojson layer as an argument and extracted completely from the hook? i don't feel strongly, and i liked seeing the feature service hooks combined into this wrapper, just not sure about cramming this new method.

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -179,12 +157,15 @@ export default function TheMap({
...
     const newFeature = {
-      geometry: clickedFeature.geometry,
+      geometry: clickedFeatureFromAgolGeojson.geometry,
       properties: {
-        ...clickedFeature.properties,
-        id: clickedFeature.id,
+        ...clickedFeatureFromAgolGeojson.properties,
+        id: clickedFeatureFromAgolGeojson.id,
+        // AGOL data doesn't include layer so we grab it from the clicked Mapbox feature
         _layerId: clickedFeature.layer.id,
```
johnclary, line 169: Nice! it will be interesting to see how these meta properties factor into the task of committing all this to the DB 🍿

johnclary (approved):
i shared some not-strongly-held opinions about the de-clipping hook, but this works great as is. 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/agolUtils.js
mddilley: Yes! I like these ideas a lot. Definitely could get that config in here - I'll update that along with moving that helper outside of the hook.

PR #850, opened 2022-10-27
Removes deprecated `role_id` column from `moped_proj_personnel`

johnclary:
## Associated issues
Resolves https://github.com/cityofaustin/atd-data-tech/issues/10586

With regard to the activity log—i initially thought the activity log would be impacted by this change, however we are not using "role_id" in the activity log, as Mike documented [here](https://github.com/cityofaustin/atd-data-tech/issues/9423). 

## Testing
**URL to test:** Local

**Steps to test:**
1. Verify that the Hasura cluster starts without error
2. Navigate to any of these project IDs: 176, 188, 182, 174
3. Verify that there are two team members, each with multiple roles.
4. Remove a role from one of these team members
5. Add a new team member.
6. Remove the new team member.
7. Navigate to the main projects list and verify that team members are listed in the "Team" column
8. Navigate to activity log for project 174. Verify that the previous activity renders correctly, notably that "John Clary added Renee O. to the project team".  

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -2077,88 +2074,81 @@
...
-          - date_added
           - added_by
+          - date_added
+          - is_deleted
           - notes
+          - project_id
+          - project_personnel_id
           - user_id
-          - role_id
-          - is_deleted
```
johnclary, line 2087: sorry :/ `role_id` is the only removal

moped-database/metadata/tables.yaml
```diff
@@ -3096,6 +3085,14 @@
...
     name: moped_project_roles
     schema: public
+  array_relationships:
+    - name: moped_proj_personnel_roles
+      using:
+        foreign_key_constraint_on:
+          column: project_role_id
+          table:
+            name: moped_proj_personnel_roles
+            schema: public
```
johnclary, line 3095: we weren't tracking this relationship and i tend to prefer that we just track everything, even when it's an unlikely use case.

moped-database/seeds/1602292389297_initial_seed_staging.sql
```diff
@@ -965,8 +896,7 @@ INSERT INTO public.moped_project_files (project_file_id, project_id, file_key, f
...
-INSERT INTO public.moped_project_files (project_file_id, project_id, file_key, file_name, file_description, file_size, file_permissions, file_metadata, api_response, created_by, create_date, is_scanned, is_deleted) VALUES (16, 157, 'private/project/157/11_03232021194501_0a7c422ceba66d29_whatabikelane.png', 'Existing connected lane', 'Looking west, southbound bike lane
-', 1235548, NULL, NULL, NULL, 11, '2021-03-23 19:46:07.093534+00', false, true);
+INSERT INTO public.moped_project_files (project_file_id, project_id, file_key, file_name, file_description, file_size, file_permissions, file_metadata, api_response, created_by, create_date, is_scanned, is_deleted) VALUES (16, 157, 'private/project/157/11_03232021194501_0a7c422ceba66d29_whatabikelane.png', 'Existing connected lane', 'Looking west, southbound bike lane', 1235548, NULL, NULL, NULL, 11, '2021-03-23 19:46:07.093534+00', false, true);
```
johnclary, line 899: linting

moped-database/migrations/1666899604790_alter_table_public_moped_proj_personnel_drop_column_role_id/up.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."moped_proj_personnel" drop column "role_id" cascade;
```
johnclary, line 1: This migration is a bit tricky to test, since in local dev the migration runs before our seeds are applied, but in reality it will be applied to a hydrated database.

I have tested this by manually running the migration against the previous seeds. I.e., by checking out main and starting the Hasura cluster.

I don't see any reason why this migration would fail in staging/prod, but if we wanted to be extra careful i believe we could do:
```
BEGIN;
alter table "public"."moped_proj_personnel" drop column "role_id" cascade;
ROLLBACK;
```

moped-database/metadata/tables.yaml
mddilley: I thought about this after you brought it up in dev sync the other day, and I'm fully on-board that we should expose all that we can out of Hasura. I can't think of a downside because if we don't need to query it - we'd just never write a query for it. 

Definitely could be overlooking something but I agree that we should build in flexibility for ad-hoc stuff like you mentioned instead of it requiring new work.

moped-database/migrations/1666899604790_alter_table_public_moped_proj_personnel_drop_column_role_id/up.sql
mddilley: I also don't see a reason why this would fail, but it could be a good use case for the using a test instance populated with staging data. But, I can't remember if that is an option at the moment. 

Do you remember @frankhereford? Definitely don't think it is worth extra effort right this moment if not.

mddilley (approved):
So nice to have this tidied up! ✨ All test steps worked when I ran locally. 🚢

moped-database/migrations/1666899604790_alter_table_public_moped_proj_personnel_drop_column_role_id/up.sql
frankhereford: I'm feeling pretty confident that it will not fail .. I think it will compile and execute, but what I'm wondering if the `cascade` would interact with any foreign key relationships and cause records in related tables to be deleted. To be honest, I really don't know how that would work, and I think Mike's inclination to test this statement with a database populated with production data is good idea. However, I 100% defer to John and his experiences while putting together this PR.

@johnclary I would be super happy to do that test if it'd be helpful -- I can reuse the `reload_db.sh` script from the recent geography work to spin one up.

moped-database/metadata/tables.yaml
frankhereford: I think I've been the most questioning of the hasura-esque attitude of relate-everything-and-expose-everything. 

It comes from my instinct that the graphql endpoint is a part of the frontend application. The frontend needs data in a certain form, and the `graphql-engine` can provide the data in that form. It's the frontend, not the backend (database) which governs the graphql endpoint's shape, if you will.

That said, I do agree that if your endpoint can answer all queries, then it by definition can answer the queries you may want in your application. A well normalized database with relations accurately represented in the `graphql-engine` metadata is really everything you could ask for.

A final thought which really sells me and makes me want to lean into relate-everything-and-expose-everything mentality:

* A fixed graphql endpoint shape allows the "backend team" to work on the database as a service and the "frontend team" to work on the front-end and not have to communicate or agree or talk even. The graphql endpoint's shape is the API they each agree on.
* A "dynamic" graphql endpoint based on the DB schema encourages (requires?) the front-end developer experience to include understanding and manipulating the database and it's middleware. This is anti-knowledge-silo and the makings for better developers and system administrators.

chiaberry (approved):
ran through the tests, everything worked as promised. also 🏆 for messing with the seed data

moped-database/metadata/tables.yaml
johnclary: > It's the frontend, not the backend (database) which governs the graphql endpoint's shape, if you will.
I think that's a very good point, and I think the fact that Hasura doesn't expose relationships by default hints that they have the same philosophy about shaping the API.

But the idea of a "fixed graphql endpoint" captures my sentiment so much better than i could articulate. it's nice to just know what you're gonna get when you ask Hasura for data. Particularly because we control access to the API through auth and permissions, it feels like the best approach to me.

johnclary: thank you both for your insight! 🙌

moped-database/migrations/1666899604790_alter_table_public_moped_proj_personnel_drop_column_role_id/up.sql
johnclary: Thanks, y'all. I will run a local "production" test just to be safe. That will definitely be worth the time vs having to fix this after merge.

Wrt to `cascade`, in the case of a dropped column, I don't _think_ it has any relation to the `on update` and `on delete` actions defined in the foreign key constraints, but it will be interesting to test on prod data.

Here's a bit of docs from the [pg tutorial](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-drop-column/):

> If the column that you want to remove is used in other database objects such as [views](https://www.postgresqltutorial.com/postgresql-views/), [triggers](https://www.postgresqltutorial.com/postgresql-triggers/), [stored procedures](https://www.postgresqltutorial.com/postgresql-stored-procedures/), etc., you cannot drop the column because other objects are depending on it. In this case, you need to add the CASCADE option to the DROP COLUMN clause to drop the column and all of its dependent objects.

johnclary:
Thanks for the reviews! As discussed, I will run a local test against a copy of staging and prod before merging.

frankhereford (approved):
I can't help but echo @chiaberry sentiment about the improvement to the seed data. That's a QOL improvement for us - thank you!

I have tested the feature as you described, and it's right on the money 🎯🔥.

PR #852, opened 2022-10-28
Re-implement map draw tools

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10403

Adds the draw tools to the components map refactor

Questions/concerns:

- there is some cross-activity when clicking a segment or point from CTN layers near a drawn line or point (capture below). The `clickBuffer` parameter of the draw tools could be updated to avoid this but I also don't want to make it super hard to click drawn features if we drop that too low.
clickLines

## Testing
**URL to test:** 

https://md-10403-draw-tools--atd-moped-main.netlify.app/moped/ or test locally

**Steps to test:**
1. Go to a new or existing project's **Map** tab and then click **New Component**
2. Choose component type **Access Control - Driveway Closure**
3. Zoom until you see the dashed lines of the CTN layer
4. Click the line drawing icon in the top right just above the trash can icon
5. You should see:
   a. the dashed lines of the CTN line layer disappear
   b. the cursor turn into a crosshair
6. Draw a line and complete the line by clicking twice in the last vertex. The vertices of the line should still be circled in green showing that the line is currently active in the draw tools
7. While still active, click the line and hold while dragging it and you should see the location of the drawn line update when you release your click
8. Click away from the line and the green circles should disappear and the dashed CTN line layer should reappear
9. Select a couple of segments from the CTN line layer and click **Save** - you will see the new component shown on the map in magenta
10. Repeat the above steps with the point component type called **Bike Box**
11. While drawing the point component, you should see the features of the first line component shown as muted with a gray color
12. After creating and saving a point type component, highlight each new component by clicking on the boxes in the left sidebar
13. You should see the highlighted component features highlighted in blue on the map and the other components muted with a gray color

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
@@ -1,7 +1,7 @@
 import React, { useRef } from "react";
 import ComponentsDrawControl from "src/components/Maps/ComponentsDrawControl";
+import { makeDrawnFeature } from "./featureUtils";
 import { cloneDeep } from "lodash";
-import { v4 as uuidv4 } from "uuid";
 
 const ComponentDrawTools = ({
   draftComponent,
@@ -22,13 +22,13 @@ const ComponentDrawTools = ({
         prevDraftComponent.features
       ).filter((feature) => !!feature.properties?.["DRAW_ID"]);
 
-      // Add a unique id to each drawn feature's properties
-      drawnFeatures.forEach(
-        (feature) => (feature.properties["DRAW_ID"] = uuidv4())
-      );
+      // Add properties needed to distinguish drawn features from other features
+      drawnFeatures.forEach((feature) => {
+        makeDrawnFeature(feature, linkMode);
+      });
 
       // We must override the features in the draw control's internal state with ones
-      // that have "DRAW_ID" properties so that we can find them later in onDelete
+      // that have our added properties so that we can find them later in onDelete
       drawControlsRef.current.set({
         type: "FeatureCollection",
         features: [...previouslyDrawnFeatures, ...drawnFeatures],
@@ -84,7 +84,6 @@ const ComponentDrawTools = ({
       setCursor("crosshair");
       setIsDrawing(true);
     } else {
-      setCursor("grab");
       setIsDrawing(false);
     }
   };
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
@@ -4,7 +4,6 @@ import { makeDrawnFeature } from "./featureUtils";
 import { cloneDeep } from "lodash";
 
 const ComponentDrawTools = ({
-  draftComponent,
   setDraftComponent,
   linkMode,
   setCursor,
@@ -92,11 +91,19 @@ const ComponentDrawTools = ({
     });
   };
 
-  const onModeChange = (e) => {
-    const { mode } = e;
-
+  const onModeChange = ({ mode }) => {
     if (mode === "draw_point" || mode === "draw_line_string") {
       setCursor("crosshair");
+      setIsDrawing(true);
+    } else {
+      setIsDrawing(false);
+    }
+  };
+
+  const onSelectionChange = ({ features: selectedFeaturesArray }) => {
+    const areFeaturesSelected = selectedFeaturesArray.length > 0;
+
+    if (areFeaturesSelected) {
       setIsDrawing(true);
     } else {
       setIsDrawing(false);
@@ -117,6 +124,7 @@ const ComponentDrawTools = ({
         onUpdate={onUpdate}
         linkMode={linkMode}
         onModeChange={onModeChange}
+        onSelectionChange={onSelectionChange}
         initializeExistingDrawFeatures={initializeExistingDrawFeatures}
       />
     )
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
@@ -1,14 +1,17 @@
-import React, { useRef, useState } from "react";
+import React, { useRef } from "react";
 import ComponentsDrawControl from "src/components/Maps/ComponentsDrawControl";
 
 const ComponentDrawTools = ({
   draftComponent,
   setDraftComponent,
   linkMode,
+  setCursor,
 }) => {
   const drawControlsRef = useRef();
-  const [isDrawing, setIsDrawing] = useState(false);
   const shouldShowDrawControls = linkMode === "points" || linkMode === "lines";
+
+  console.log({ draftComponent });
+  // TODO: Add/update/remove draftComponent.features
 
   const onCreate = () => {
     console.log("onCreate");
@@ -27,9 +30,9 @@ const ComponentDrawTools = ({
     const { mode } = e;
 
     if (mode === "draw_point" || mode === "draw_line_string") {
-      setIsDrawing(true);
+      setCursor("crosshair");
     } else {
-      setIsDrawing(false);
+      setCursor("grab");
     }
   };
 
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
@@ -13,16 +13,14 @@ const ComponentDrawTools = ({
   const shouldShowDrawControls = linkMode === "points" || linkMode === "lines";
 
   //   console.log({ draftComponent });
-  // TODO: Add/update/remove draftComponent.features
-
   const onCreate = ({ features: createdFeaturesArray }) => {
+    // TODO: Refactor this to not pass a function
     setDraftComponent((prevDraftComponent) => {
       const drawnFeatures = cloneDeep(createdFeaturesArray);
 
       const previouslyDrawnFeatures = cloneDeep(
         prevDraftComponent.features
       ).filter((feature) => !!feature.properties?.["DRAW_ID"]);
-      console.log({ previouslyDrawnFeatures });
 
       // Add a unique id to each drawn feature's properties
       drawnFeatures.forEach(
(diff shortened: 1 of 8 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -14,35 +15,25 @@ import {
...
 import "mapbox-gl/dist/mapbox-gl.css";
+import { isDrawnFeature, makeCapturedFromLayerFeature } from "./featureUtils";
 
 // See https://github.com/visgl/react-map-gl/issues/1266#issuecomment-753686953
 import mapboxgl from "mapbox-gl";
 mapboxgl.workerClass =
   // eslint-disable-next-line import/no-webpack-loader-syntax
   require("worker-loader!mapbox-gl/dist/mapbox-gl-csp-worker").default;
 
-const deDeupeProjectFeatures = (features) => {
```
mddilley, line 24: I removed this since the IDs checked here don't agree with the unique IDs of drawn features, and they were hidden as a result. I can spin off an issue to refactor this or include in this PR. 

Just realized that I forgot to circle back on this.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
```diff
@@ -0,0 +1,142 @@
...
+  setDraftComponent,
+  linkMode,
+  setCursor,
+  setIsDrawing,
+}) => {
+  const drawControlsRef = useRef();
+  const shouldShowDrawControls = linkMode === "points" || linkMode === "lines";
+
+  const onCreate = ({ features: createdFeaturesArray }) => {
+    setDraftComponent((prevDraftComponent) => {
```
mddilley, line 24: Not sure if I'm a huge fan of nesting all of this logic inside this state setter (same in `onUpdate` and `onDelete` below). I left them for now, but I'm going to think on it.

moped-editor/src/views/projects/projectView/ProjectComponents/featureUtils.js
```diff
@@ -0,0 +1,98 @@
+import { v4 as uuidv4 } from "uuid";
```
mddilley, line 1: Lots of util files stacking up - I might create a `utils` folder and then have named files inside like `features.js`, `agol.js`, etc. Any input on that would be great. 🙏

mddilley:
@patrickm02L I added the test deploy in the description above

frankhereford (approved):
This is more high quality work Mike - this whole interface is coming along at a shockingly fast clip! I went through the testing procedure that you laid out, and it's ✅ across the board. 

I have a couple of thoughts that I wanted to share, but I don't intend them as blocks on merging this PR. 

<hr>




image

When you click on a pink line in the above picture, you get the label shown above, and I think maybe it should only have one of the entries in the label.

<hr>

https://user-images.githubusercontent.com/3653206/199522404-df3e191d-d6d7-48f9-aa25-3cc25f4cb883.mp4



Another pair of things that I noticed belong on the ultimate bottom of the priority list, but I thought I might bring it up just in case it's made easy by the underlying mapbox library. In the following clip:


1) If it's easy to make it when you double-click on the first vertex you laid down, it would place the final vertex at exactly the same point as the first one. This allows you to close out a line (sorta faking a polygon) that actually encloses an area at any zoom level. I think I learned this behavior using GIS software, so it might be a similar experience for others.

(I recognize that this is a /linestring/ tool, so super all good to toss this idea in the bin!)

2) The vertices having a "selected" state indicates to me that they are editable, meaning they can have their location adjusted. I feel like I should be able to mouse over them and drag them. Any point or vertex of a line allllllways needs some sort of adjusting. If one thinks something is in the correct place, they just need to zoom in, and they will see it needs adjusting. 

Would you consider striking the green outline around selected vertices, or is there another need for the indication that is going over my head - quite possible!

moped-editor/src/views/projects/projectView/ProjectComponents/featureUtils.js
frankhereford: You have my support! Doubly so with your suggestion that the folder inherits the more generic name like `utils` or `utilities` and the files get more specific names indicating their purpose. 🎯

roseeichelmann (approved):
This drawing tool is so clean Mike, this is great work! I love that you can move the lines around while they are in edit mode,I bet that will be really handy. I went through all the steps and they all check out! 🥳

I do agree with Frank about how its unexpected that when you select a line that is overlapping with another line instead of just the label box for the selected line showing up, both labels show up like in the video below (okay i tried to add a screencast but it was so glitchy i dont think it was helpful lol). It makes it hard to know which label you want to click edit on and I think just having the one label box for the line you selected pop up would be less confusing?

Also it would be really cool to one day be able to edit the vertices themselves, like move them around as Frank said, because that's also what I would expect in a GIS tool lwith the vertices circled. This obviously wouldn't be addressed in this issue, but could be cool to explore in the future. Personally though I think the green circles should stay because it helps me understand that you have to double click on it to save the line. Without any indicator, it would be less obvious to me that I have to click the end of the line twice to save it.

mddilley:
@frankhereford @roseeichelmann Thanks for the thorough looks at this! I'm loving these ideas. ❤️  

Some thoughts:
1. The inability to close out a line string by connecting start to end point is a limitation of the [mode selected](https://github.com/mapbox/mapbox-gl-draw/blob/main/docs/API.md#modes) in the tools
2. Same with the inability to drag/edit vertices. We actually override the `direct_select` mode that allow this by default since it would take some custom dev to stop this mode from letting users do some unwanted things like splitting lines into further segments with stray clicks
3. I think both of these would be great to have and go into backlog issues, but I'll leave that up to @johnclary when he has the opportunity to look at this
4. Multiple entries in the pink feature click popup is a bug, and I need to fix that. 🐛 🙏 
5. Definitely open to different styling on active features - the green vertices are there to show which feature is active to drag around or delete and same with the wide outline on drawn points.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
johnclary: there's some repetition, but it's also very easy to track what's happening and i'm cool with optimizing later, especially if we're not super confident this is the final iteration of the draw behavior.

i'm not feeling very creating at the moment, but one way to improve readability a bit could be to wrap these handlers in `setDraftComponent` ?

```javascript
onCreate={(e) => setDraftComponent(onCreate(e)}
```

not sure that's actually an improvement 🤪

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
johnclary: 👍 this removes the ability to see overlapping features in the popups, but that functionality has not been well scoped and i'm happy to flesh out that behavior in [this issue](https://github.com/cityofaustin/atd-data-tech/issues/10715).

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -69,9 +60,11 @@ export default function TheMap({
   linkMode,
   setIsFetchingFeatures,
 }) {
-  const [cursor, setCursor] = useState("grap");
```
johnclary, line 72: 🙈

moped-editor/src/views/projects/projectView/ProjectComponents/agolUtils.js
```diff
@@ -48,13 +48,13 @@ on a previous state is a great use case for useReducer, because
 it allows us to work around an infinite recursion scenario in useEffect  */
 const featureReducer = (geojson, { features, featureIdProp }) => {
   const allFeatures = [...geojson.features, ...features];
-  const uniqueFeatures = deDeupeFeatures(allFeatures, featureIdProp);
```
johnclary, line 51: 🙈

moped-editor/src/views/projects/projectView/ProjectComponents/featureUtils.js
```diff
@@ -0,0 +1,98 @@
...
+ * Drawn feature object example
+ * {
+ *   "id": "9e0c688733283be4991cb9df39ae284d",
+ *   "geometry": { coordinates: [], type: "LineString" },
+ *   "type": "Feature",
+ *   "properties": {
+ *     "DRAW_ID": "88230580-099f-4c93-b8e5-ad3086caf0e6",
+ *     "_layerId": "drawnByUserLine",
+ * }
+ */
```
johnclary, line 71: these examples are super helpful

johnclary: i'm lazily not looking at Frank's branch, but i'm curious how this will shape into the new layer schema. i'm sure you're working through it!

johnclary (approved):
@mddilley this is working really well! and i agree Frank and Rose's feedback is spot on.

To echo some of their comments:
- I think we should add more distinct line and point styles when editing drawn features to make it clear which elements are drawn vs selected
- I'm wondering if there would be any harm in allowing the full sweet of line editing functions. I agree dragging vertices would be a big improvement.
- Ideally, we would disable the ability to click on selected features while drawing, because it's otherwise difficult to connect a drawn feature to a selected one. I'm curious how challenging you think that would be @mddilley.

In any case, I'm comfortable leaving those outside of this PR and will scope future enhancements in these two issues:
- https://github.com/cityofaustin/atd-data-tech/issues/10716
- https://github.com/cityofaustin/atd-data-tech/issues/10715

🚢 🚀

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentDrawTools.js
mddilley: Yep, I'm on the same page. I think that updating with some helpers is another way to handle this, but I think that I'll leave it this way for now. Punt! 🏈

moped-editor/src/views/projects/projectView/ProjectComponents/featureUtils.js
mddilley: For sure, it is definitely getting interesting! So far, I'm trying to focus on making the munging well-defined going in and out of the map. It will probably look similar to this in the end. 🤞

mddilley:
Thanks everyone for review! 🙏  I added some notes to https://github.com/cityofaustin/atd-data-tech/issues/10716 to capture some of the feedback here. Going to merge this one into the base feature branch.

PR #854, opened 2022-11-04
Creates project lead field in summary table

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10682

## Testing
**URL to test:** 
Please test locally

**Steps to test:**
go to a project summary page, there should be a new field called "Lead" above "Sponsor"
add or edit a a lead and save it
repeat this process with a project sponsor
that's it!

i decided to rename the ProjectSummaryProjectSponsor component to ProjectSummaryProjectEntity and add in an argument for which field it is displaying along with some logic for displaying and mutating the appropriate field so that it could be used for both fields. that way we didn't need to add another component!

---
#### Ship list
- [x] Code reviewed 


tillyw:
one question i had was whether it is worth trying to squash the two migrations that hasura generated or whether it's best to leave them separate!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectEntity.js
```diff
@@ -18,38 +21,59 @@ import { PROJECT_UPDATE_SPONSOR } from "../../../../queries/project";
...
+  const getOriginalSponsor = () => {
+    if (entityName === "Lead") {
+      return entityList.find(
+        (e) => e.entity_id === data?.moped_project?.[0]?.project_lead_id
+      );
+    }
+    if (entityName === "Sponsor") {
+      return entityList.find(
+        (e) => e.entity_id === data?.moped_project?.[0]?.project_sponsor
+      );
```
chiaberry, line 44: this isn't a comment on your implementation of this, because you didnt start the pattern that we have all over moped, but we could leverage the object relationships set up in hasura to get the entity name out of the query instead doing a lookup with in the component.

chiaberry (commented):
I found a bug, but it also exists in moped now -- I can add and edit a sponsor or lead, but I can't clear it (update the field with nothing)

chiaberry:
> I found a bug, but it also exists in moped now -- I can add and edit a sponsor or lead, but I can't clear it (update the field with nothing)

I am incorrect! This isnt a bug, I didn't realize that the way to clear a field is by selecting the `-` instead of clicking the X (the fainter one that has clear as a tooltip)
Screen Shot 2022-11-04 at 10 17 53 AM

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectEntity.js
tillyw: @chiaberry you're right, @johnclary mentioned using a hasura object relationship in the issue and i glossed over that part. i'm going to look into how to implement that now!

tillyw:
> > I found a bug, but it also exists in moped now -- I can add and edit a sponsor or lead, but I can't clear it (update the field with nothing)
> 
> I am incorrect! This isnt a bug, I didn't realize that the way to clear a field is by selecting the `-` instead of clicking the X (the fainter one that has clear as a tooltip) Screen Shot 2022-11-04 at 10 17 53 AM

i've noticed this too. it is weird behavior though and may be a result of the fact that we use a record to indicate no value. probably outside the scope of this issue but definitely worth looking into or addressing when we reconsider the database pattern!

tillyw:
ok, i reconfigured this to set and utilize object relationships as @chiaberry suggested! rerequesting your review

moped-editor/src/queries/project.js
```diff
@@ -741,6 +748,17 @@ export const PROJECT_UPDATE_SPONSOR = gql`
...
+  mutation ProjectUpdateLead($projectId: Int!, $entityId: Int!) {
+    update_moped_project(
+      where: { project_id: { _eq: $projectId } }
+      _set: { project_lead_id: $entityId }
+    ) {
+      affected_rows
+    }
+  }
+`;
+
```
johnclary, line 761: we don't do this consistently, but i think it's preferable to use the `update_<table>_by_pk` mutation when updating a single record. it takes away some risk of a accidentally updating multiple records.

the Hasura docs have examples of how to structure the variables the pk mutations, in this case `update_moped_project_by_pk`:

https://hasura.io/docs/latest/mutations/postgres/update/#auto-generated-update-mutation-schema

moped-database/migrations/1667514062398_set_fk_public_moped_project_project_lead_id/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."moped_project"
+  add constraint "moped_project_project_lead_id_fkey"
+  foreign key ("project_lead_id")
+  references "public"."moped_entity"
+  ("entity_id") on update cascade on delete set null;
```
johnclary, line 5: i would lean toward squashing this into one migration, since the feature requires both of them.

johnclary (changes requested):
@tillyw thanks for making such quick work of this! i'm glad to see this component re-used. Let me know if you have any questions about my feedback 🙌

moped-editor/src/queries/project.js
```diff
@@ -77,6 +78,12 @@ export const SUMMARY_QUERY = gql`
           type_id
         }
       }
+      moped_entity {
+        entity_name
+      }
+      moped_project_lead {
+        entity_name
+      }
```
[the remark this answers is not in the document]
tillyw: fixed, thanks!

moped-database/migrations/1667514062398_set_fk_public_moped_project_project_lead_id/up.sql
tillyw: done!

johnclary:
to close the loop on @chiaberry's comment about clearing out the input field, [here's](https://github.com/cityofaustin/atd-data-tech/issues/10594) the backlog issue related to it.

moped-editor/src/queries/project.js
tillyw: done! thanks for linking me to the docs :)

tillyw:
> one question i have is whether it is worth trying to squash the two migrations that hasura generated or whether it's best to leave them separate!



> @tillyw thanks for making such quick work of this! i'm glad to see this component re-used. Let me know if you have any questions about my feedback 🙌

thanks @johnclary ! resolved all your comments, rerequesting review now ^.^

tillyw:
just want to note if anyone is reviewing, the flow is down so you will have to test locally. but @frankhereford is working on a fix! thanks frank!!

johnclary (approved):
thanks for making those changes! i'd like to get one more dev review but i think this is ready to ship!

🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

chiaberry (approved):

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectEntity.js
```diff
@@ -106,7 +133,7 @@ const ProjectSummaryProjectSponsor = ({
         )}
         {!editMode && (
           <ProjectSummaryLabel
-            text={sponsor?.entity_name || "n/a"}
+            text={sponsor?.entity_name}
```
mddilley, line 136: I noticed a bug after this change. To replicate:

1. Go to a project that does not have a `Lead` set yet
2. Click the **Lead** field in the **Project Summary** form
3. Click the "X" icon on the far right of the field to cancel out
4. WSOD

This root of this bug is way beyond the scope of this issue because I noticed that in many instances we pass `text={<the value> || ""}` to `ProjectSummaryLabel` instead of handling nullish values in that component itself. For now, I think passing `text={sponsor?.entity_name || ""}` will fix this.

mddilley (changes requested):
This looks really amazing! One small bug fix (that is caused by existing code) and this is 🚢 🚢 🚢

mddilley (approved):
🚢 ✨✨✨

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectEntity.js
tillyw: thanks for catching this @mddilley! i just committed the fix you suggested. i think you're so right about reworking the `ProjectSummaryLabel` component to handle nullish values and removing this conditional from the parent components. i'll create a separate issue to do that.

PR #855, opened 2022-11-07
Add ability to search by project creator

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/6697

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1667853991173_add_created_by_sql_view/up.sql
+++ b/moped-database/migrations/1667853991173_add_created_by_sql_view/up.sql
@@ -33,7 +33,7 @@ AS WITH project_person_list_lookup AS (
       concat(users.first_name, ' ', users.last_name) AS added_by,
       project_id
       FROM moped_project
-        LEFT JOIN moped_users users ON users.user_id = added_by
+        LEFT JOIN moped_users users ON users.user_id = moped_project.added_by
     )
  SELECT
     mp.project_uuid,
@@ -163,7 +163,7 @@ AS WITH project_person_list_lookup AS (
      LEFT JOIN moped_proj_contract contracts ON (mp.project_id = contracts.project_id) AND contracts.is_deleted = false
      LEFT JOIN added_by_lookup ON mp.project_id = added_by_lookup.project_id
   GROUP BY mp.project_uuid, 
-    mp.project_id, 
+    mp.project_id,
     mp.project_name, 
     mp.project_description, 
     ppll.project_team_members, 
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -299,6 +299,19 @@ export const ProjectsListViewFiltersConf = {
         "string_does_not_equal_case_sensitive",
       ],
     },
+    {
+      name: "added_by",
+      label: "Added by",
+      placeholder: "Added by",
+      type: "string",
+      operators: [
+        "string_contains_case_insensitive",
+        "string_begins_with_case_insensitive",
+        "string_ends_with_case_insensitive",
+        "string_equals_case_sensitive",
+        "string_does_not_equal_case_sensitive",
+      ],
+    },
   ],
 
   operators: {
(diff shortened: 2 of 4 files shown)


patrickm02L (approved):
Good to go! 🤘
Vote for **Created By**


https://user-images.githubusercontent.com/5566727/200573162-261e13fc-dc38-445e-a3d5-39dbee593d74.mp4

moped-database/migrations/1667853991173_add_created_by_sql_view/up.sql
```diff
@@ -0,0 +1,188 @@
...
+        LEFT JOIN moped_types mt ON mpt.project_type_id = mt.type_id AND mpt.is_deleted = false
+    GROUP BY mpt.project_id
+  ),
+   added_by_lookup AS (
+    SELECT
+      concat(users.first_name, ' ', users.last_name) AS added_by,
+      project_id
+      FROM moped_project
+        LEFT JOIN moped_users users ON users.user_id = moped_project.added_by
+    )
```
johnclary, line 37: i'm not sure if this matters—this query is way beyond my SQL comfort zone—but because we're joining the one (a user) in a many-to-one relationship, i don't think we need a subquery. you can concat directly against moped_users.

```sql
     SELECT ...
     CONCAT(added_by_user.first_name, ' ', added_by_user.last_name) AS added_by_user
     ...
     FROM LEFT JOIN moped_users added_by_user ON mp.added_by = added_by_user.user_id
```

i would only be concerned about this if there's a performance hit for using a subquery

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -299,6 +299,19 @@ export const ProjectsListViewFiltersConf = {
...
+      placeholder: "Created by",
+      type: "string",
+      operators: [
+        "string_contains_case_insensitive",
+        "string_begins_with_case_insensitive",
+        "string_ends_with_case_insensitive",
+        "string_equals_case_sensitive",
+        "string_does_not_equal_case_sensitive",
+      ],
+    },
```
johnclary, line 314: this is out of scope, but it would be nice if we could use the `is` operator autocomplete to pick a user. let me know @chiaberry if you'd prefer to see that in a separate issue.

johnclary (approved):
🙌 this works perfectly! my feedback is optional.

> Should the created by column be added to the project list table? Does it feel weird to not see what you are searching on?

Yes, it should be added to the project list table. It can be hidden by default. Sorry that I missed that in the issue description. Lmk if you'd prefer to see that in a separate issue.

> Created By or is it Added By?

I prefer "Created By" - this feels like the most clear way to describe what it is, and it's how we refer to such fields in our Knack apps.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: Agreed, that would be nice to have for sure!

moped-database/migrations/1667853991173_add_created_by_sql_view/up.sql
chiaberry: yup, i overengineered this. ill update it

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
chiaberry: We should spin this into its own issue, its a little more involved to get the autocomplete to have the list of employee first name and last names

I noticed we also have the "is" option for Team, but there is no autocomplete and I can't figure out how to actually get a result using "is" and Team members

tillyw (approved):
tested and works as expected! i also prefer 'created by' and agree it should be added to the project list table but can be hidden by default. working from this branch for my 'search by lead' issue. should the package-lock.json file be uncommitted?

chiaberry:
I will go ahead and merge it so you can work from main -- the package-lock.json only has the version updated

PR #857, opened 2022-11-08
displays project lead in list view, searchable

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10683

## Testing
**URL to test:** 

https://10683_search_based_on_lead_workgroup--atd-moped-main.netlify.app/moped/

**Steps to test:**
- go to a project and add a 'lead'
- go to project list view, you should see the 'lead' to the left of 'sponsor'
- search by that 'lead' and you should see your record come up

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -109,6 +109,25 @@ export const ProjectsListViewFiltersConf = {
...
+        field_name: "entity_name",
+      },
+      operators: [
+        "string_contains_case_insensitive",
+        "string_begins_with_case_insensitive",
+        "string_ends_with_case_insensitive",
+        "string_equals_case_sensitive",
+        "string_does_not_equal_case_sensitive",
+        "string_is_null_special_case",
+        "string_is_not_null_special_case",
```
johnclary, line 128: this is clearly a pre-existing bug, but  the `string_is_null_special_case` ("is blank") and `string_is_not_null_special_case` ("is not blank") operators do not work correctly on this field, or on the Sponsor field. 

i'll make a separate issue.

johnclary (approved):
@tillyw thanks for adding this so quickly! i was happy to find the autocomplete advanced search working 🙌 

some of our lesser-used filter operators appear to have some problems, but that's outside the scope of this issue.

please hold off on merging until another dev takes a look


🚢 🚢 🚢 🚢

frankhereford (approved):
Tested via the instructions and reviewed your code. I started looking at how one needs to remove the null-indicating-dash when you click on the Project Lead field when setting it from its empty state, .... but I then realized that that is waaaay out of scope for what the issue driving this PR.

I think you nailed it - this looks great. Thank you @tillyw!

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -134,6 +134,18 @@ export const ProjectsListViewQueryConf = {
       width: "20%",
       filter: filterProjectTeamMembers,
     },
+    project_lead: {
+      label: "Project lead",
+      searchable: true,
```
chiaberry, line 139: since we are adding this to the search too, should we expand the prompt to include Project lead?
Screen Shot 2022-11-09 at 10 32 41 AM

tillyw: good point @chiaberry, i might have gone overboard with this pr because looking back i don't think adding this field to global search was part of the issue. i should either remove that or add it to this prompt as you suggested. @johnclary what are your thoughts?

johnclary: TIL we can currently search by sponsor and partners. I like this so i updated the placeholder text. Thank you both!

PR #859, opened 2022-11-09
Adds line breaks and auto-sizing to various columns on the project list

johnclary:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9853

This adjusts a number of columns on the project list so that they (1) have linebreaks between list items and (2) grow to the width of their content:

- project name
- team
- task orders
- contracts
- contract numbers
- tags
- partners
- funding

After user's react to this change, we may want to start truncating long strings in the cell render (task orders, team members with lots of roles....). My opinion is that if we're going to have lists of things in table cells, those things should never wrap. I did reach out to AMD and PDD for feedback on task orders specifically.

I will need to resolve conflicts once https://github.com/cityofaustin/atd-moped/pull/857 is merged.

## Testing
**URL to test:** [Deploy preview](https://deploy-preview-859--atd-moped-main.netlify.app/)

**Steps to test:**
1. Add multiple team members with multiple roles to a project
2. Add multiple task orders to a project
3. Add multiple contracts to a project
4. Add multiple tags to a project
5. In the project list view, observe that each of these columns display line-separated values and do not wrap
6. Remove the task orders and team members from the project
7. In the project list view, observe that the column widths for these fields minimal

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -429,18 +429,42 @@ const ProjectsListViewTable = ({ title, query, searchTerm, referenceData }) => {
       field: "contractors",
       hidden: hiddenColumns["contractors"],
       emptyValue: "-",
+      cellStyle: { whiteSpace: "noWrap" },
+      render: (entry) => {
+        return entry.contractors.split(",").map((contractor, i) => (
```
johnclary, line 434: The project list view sql puts us in this funny place where we're string-splitting our own data. The task orders field is an exception—we're leaving that column as an array and parsing the array in the client.

frankhereford: I agree this isn't great. I think the comma-joined value coming out of the DB right now is intended to be read-only content generally intended to be put in a table where it has to pack into a cell. I don't look forward to the day when we get a contractor named with a comma in it.

I think a solution is maybe going to be some form of list-like object that contains all the contractors and some metadata. Just for the sake of conversation, would adding this relationship in `graphql-engine` help?

image

The idea here is to make that graphql linkage from the view into the `moped_project_contract` table so you can ask for that contractor data as its own object or what have you.

Beyond just starting up your branch locally to get at the graphql-engine console, I have not tested that one iota, so please take this with a huge grain of salt.

frankhereford (approved):
# 📦 ✨ ⇨ 🚢 🌊 

Put this in a box and ship it! 

These columns are, at their core, lists, and this list format sure makes sense to me. 🙌 

Also, there's a ramble inspired by your comment re the `.split()` usage - nothing to developed though. Curious if it's interesting or if i'm framing this incompletely in my mind.

chiaberry (approved):

patrickm02L (approved):
Super clean! 🤘 
Screen Shot 2022-11-09 at 10 35 28 AM

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
mddilley: Agreed, y'all. Splitting feels a bit brittle for sure, but it works for now until we (hopefully not) see a bug surface. A spike into un-comma-ifying these lists would be good because I'm curious about it too but am finding it hard to grok as well.

While I'm thinking about it, computed fields could be handy for things like full name which would simplify this view in multiple places and also the codebase where we use a helper called `getUserFullName` currently. Just something to think about.

mddilley (approved):
Looks beautiful! ✨ I added some thoughts about the project list view too but this is 🚢

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
johnclary: Thanks, y'all. 

Totally agree @mddilley that we should be using generated columns things that get concatenated within their own table row, like full name.

@frankhereford manually adding relationships to the project list view is a really interesting idea. part of me thinks, well, if we're going to be adding graphql relationships to our list view, why not just use hasura entirely and ditch the list view? but there are very practical reasons to have the view, especially as users begin to look at project data in PowerBI. 

as another way to approach it, i wonder if it would be cleaner to create individual views for specific columns that clearly should be concatenated/rolled up on the database side, like project status, and creating manual relationships to the `moped_project` table for those edge cases.

given how complex search / filtering is currently, it's hard to imagine overhauling what we have in place, but it's fun to think about. i like @frankhereford's suggestion because that might be the lowest-lift way to move away from the brittle comma-splitting.

johnclary:
thank's y'all!!!!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
frankhereford: I think you're onto something -- why not have both? 

The view has the practicality that you mentioned, and with a view, the complexity of the joins and data roll-ups is sucked out of the app and into the DB layer. To me this feels right: postgres is a relational database package, and this task is a relational data task.  If Hasura were able to beat them at their game of doing that super-optimized data joining stuff internally to `graphql-engine`, then I would expect to see them switch to a NoSQL that is just all about speed like `memcached` or `reddis` or something, maybe.

The work the view does to flatten out that complex relationship of tables and records in our database ... For that effort,  I do recognize that we could implement all the joining logic with loops iterating over complex objects we could get back from a fully relationshiped-out `graphql-engine`. One thing for sure, we will always have to put in the work to define how things relate and how we want to mash them back together to get the output we want, somewhere. We can do it in JS in the app downstream of `graphql-engine` or we can do it in the DB, upstream of it. As time marches on, it looks like Hasura wants to see more of this data-mashing inside graphql-engine itself -- like with their remote data sources and other integration options.

All that to say, I think there are 2 places (pg view or JS app) where we can munge this data, and I think both places are 100% viable, and I think there are going to be perks to one over the other depending on the situation. I suggest we stay handy with both of the techniques and use them where we think they are best. All things equal, I like to push processing into the centralized server, but I can sure see the argument of pushing out to the clients too. I also trust PG about a billion times more in terms of performance and reliability than my own code, so that may be where I get that instinct. 

Anyway - 

The idea of specifying the linkage came from our recent conversations as a team regarding if we should fully embrace linking anything we can in graphql-engine, or if we should be more selective and deliberate in the graphql endpoint we create. I can't see any downside (performance or complexity) to creating the link in the graphql-engine config, so why not, right?

johnclary: Thanks @frankhereford  - I'm reading through this and nodding my head the whole way through. I agree we save ourselves a lot of future pain by munging/filtering/shaping our data as far upstream as possible. I think this will come to a head more as we re-evaluate the material-table component library and triage [some issues](https://github.com/cityofaustin/atd-data-tech/issues/10735) with the advanced search.

PR #863, opened 2022-11-09
Patches added_by backfill script

johnclary:
I uncovered a bug when i ran this against the staging database - i was not properly extracting a user ID from a `moped_user` object.

I ran this locally against the staging DB and it worked.


chiaberry (approved):

mddilley (approved):
Oh, yep! I see it now. Thanks for running this on both stages. We'll be all set for 🚀 tomorrow!

PR #870, opened 2022-11-11
Remove migration that does not match main

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/9724

This should fix our schema spy issue. 
I do not think it would affect the production database, because the removal of a migration file would not affect the migrations that have already been applied (or marked as ignore) in production. 

I am not sure if this would warrant a minor version bump. Production code is changing, but no changes to any functionality or appearance. 

## Testing
**URL to test:** 

**Steps to test:**

check out and run `./hasura-cluster restart` locally. 
Should be able to spin up the hasura cluster

If you check out the production branch, the hasura cluster will not spin up locally because of these migrations. 

---
#### Ship list
- [x] Code reviewed 


frankhereford (approved):
ezgif-5-1537b3d496

Thanks for the slack convo brainstorming on implications on this and making the fix! 🚢

johnclary (approved):
hooray! i changed the base to the v1.12.1 RC

mddilley (approved):
Tested this out and prod does not spin up but this branch does. Prod docs are coming back! 🙌 🎉

PR #877, opened 2022-11-15
Update components create, delete, and view to use new schema

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10409

## Testing
**URL to test:** 

https://md-10409-ui-new-comp-schema--atd-moped-main.netlify.app/moped/ or local

**Steps to test:**

**Normal mode**
1. Go to the link above
2. Go to the **Map** tab of a new or existing project
3. Create and save a line component selected and drawn lines
4. Create and save a point component selected and drawn lines
5. Close the **Map** view
6. Return to the **Map** tab, and you should see your previously created components and their features on the map
7. Click the magnifier icon in each component list item to zoom back and forth and make sure the features turn blue when selected
8. Select a component by clicking on its list item on the left side and click the **Delete** button
9. You should see the component details in a confirm modal. Delete the component.
10. You should see the component list item and its features on the map disappear from the UI

**Hard mode**
1. Fetch and checkout this branch locally
11. Follow [these steps](https://austininnovation.slack.com/archives/CNUEPKLB1/p1668447762094979) to populate your local database with either staging or production data (you might have to [update your Read Replica select permissions](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-MIQvl_rKnZ_-wHRdp4J/moped-data/moped-read-replica) first because there were recent changes to the schema)
12. To successfully apply the metadata in those steps, you might have to merge the up-to-date `main` branch into this one (there is a schema change in the staging DB that isn't in this branch or its feature branch base yet)
13. After getting your local DB loaded up with staging or production data, run [Frank's transpose geometries script](https://github.com/cityofaustin/atd-moped/blob/main/moped-toolbox/populate_geometry_transform/transpose_geometry.py). You will have to make sure that the [env vars](https://github.com/cityofaustin/atd-moped/blob/e0c307863716b21218b338055be90fe210aba014/moped-toolbox/populate_geometry_transform/transpose_geometry.py#L15-L18) point to your local database.
14. Go to the **Map** tab of an existing project with components (https://localhost:3000/moped/projects/1695 works if you used staging data) and see that existing components appear on the map
15. Follow the steps in normal mode to try creating and deleting components

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1668201314245_alter_table_public_feature_drawn_lines_alter_column_render_type/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."feature_drawn_lines" alter column "render_type" set default 'line';
+alter table "public"."feature_drawn_points" alter column "render_type" set default 'point';
+alter table "public"."feature_intersections" alter column "render_type" set default 'point';
+alter table "public"."feature_signals" alter column "render_type" set default 'point';
+alter table "public"."feature_street_segments" alter column "render_type" set default 'line';
```
mddilley, line 5: Looking for thoughts about these default values for `render_type` set in the DB instead of on the front end. These values aren't consistent in the current schema either. Some tables use `Point` and `Line` while others use `line` and `point`, and then the drawn features use `LineString` and `Point` too.

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -1,21 +1,25 @@
 import { useState, useMemo } from "react";
 import MapGL from "react-map-gl";
 import { cloneDeep } from "lodash";
-import FeaturePopup from "./FeaturePopup";
```
mddilley, line 4: Commenting out the feature popup for now since it is being updated in https://github.com/cityofaustin/atd-data-tech/issues/10715

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -26,14 +30,31 @@ mapboxgl.workerClass =
 // returns geojson of features across all components
 const useProjectFeatures = (components) =>
   useMemo(() => {
+    if (components.length === 0)
```
mddilley, line 33: This fallback value patches a Mapbox GL warning about sources not having valid GeoJSON - an empty feature collection makes it happy.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -85,12 +103,30 @@ export default function MapView({ projectName, projectStatuses }) {
...
+      variables: { projectId },
+      fetchPolicy: "no-cache",
+      onCompleted: () => {
+        setComponents(data.moped_proj_components);
+
+        // Create feature collections of all features in each component
+        const componentFeatureCollections = makeComponentFeatureCollectionsMap(
+          data.project_geography
+        );
+        setFeatureCollectionsByComponentId(componentFeatureCollections);
```
mddilley, line 120: Setting state here felt a little off to me at first, but it works well when `refetch()` is called. The alternative would be a useEffect that would watch `data` coming out of this `useQuery` or something like that.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -52,9 +67,12 @@ export default function MapView({ projectName, projectStatuses }) {
   const appBarHeight = useAppBarHeight();
   const classes = useStyles({ appBarHeight });
   const mapRef = useRef();
+  const { projectId } = useParams();
 
   /* holds this project's components */
   const [components, setComponents] = useState([]);
+  const [featureCollectionsByComponentId, setFeatureCollectionsByComponentId] =
```
mddilley, line 74: I'm open to other ideas here, but this piece of state made it easier to get a component's feature collection and make good use of the `project_geography` view.

Another direction would be making feature collections on the fly from component records in the `components` state array, but that would also require calling `.find` each time. IMO it will take projects with tons of components to find out what works best.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -108,8 +144,76 @@ export default function MapView({ projectName, projectStatuses }) {
   };
 
   const onSaveComponent = () => {
-    const newComponents = [...components, draftComponent];
-    setComponents(newComponents);
+    /* Start data preparation */
```
mddilley, line 147: I thought about extracting these data prep lines into a helper, but I kind of like seeing the data built up for insertion. Down to refactor if there are strong opinions about it.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatures.js
```diff
@@ -0,0 +1,188 @@
+import { getDrawId } from "./features";
+
+/*
+ * When we insert in the DB we need to translate the property names that either come from a CTN layer
+ * or the draw tools into columns names in the database.
+ * This is also used to track all layers of data present in the map.
+ */
+export const featureTableFieldMap = {
```
mddilley, line 8: Any commented out column names here with `NULL` as the value are columns that I think that we can remove from the DB tables. I noticed that we don't have any values populated there in the transpose script output.

@johnclary @frankhereford One question that I have here is about `symbol` for the `feature_street_segments` table. I saw `0` and `2` in that columns, but I'm not sure where that is coming from or what it represents.

frankhereford: @mddilley wrote: 
> One question that I have here is about symbol for the feature_street_segments table. I saw 0 and 2 in that columns, but I'm not sure where that is coming from or what it represents.

👋 Good question. TBH, I suspect that @johnclary or @chiaberry will have a better idea than I do. The various `feature_[name]` tables have the common field(s) inherited from `features` and then the columns that are unique to the feature table -- those were created as needed to handle whatever key-value pairs were in the geoJSON objects. I'm not sure where that data was originally sourced, I just took it at face value and did what I had to do to get it in the DB.

Long way to say - sorry I don't know, but I am hopeful others do. If no one knows, maybe dropping those columns is in order. 🤔

johnclary:
i had to spin this up real quick just to see it in action. looking forward to reviewing asap ✨ ✨ ✨

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatures.js
johnclary: yep, it's an artifact of just wholesale copying the geojson properties from whatever layer. i will give those columns a close look and come back with an opinion on what can be dropped.

 i do think we should minimize the columns down to what's essential + what we could reasonably need for a well-defined near-term enhancement, but i can also see us punting that cleanup until a few releases from now. i don't want us to get bogged down patching Frank's script.

moped-database/migrations/1668201314245_alter_table_public_feature_drawn_lines_alter_column_render_type/up.sql
frankhereford: These are naive questions -- not rhetorical at all. I'm not sure I don't think I have the whole history of this end-to-end in my head. 

Is the `render_type` column vestigial from the previous map invocation? Can you get the default dimensionality hint from `moped_components.line_representation`, or even better, we could use `ST_GeometryType(geography::geometry)`. For example, `select ST_GeometryType(geography::geometry) from feature_intersections limit 1;` returns `ST_MultiPoint`, which reveals the true type of geometry/geography that can be stored in the field. We could add the output of this function to the geography view, perhaps?

That final method of inspecting the geography field itself feels the most "normal" but I totally observe that it may be more trouble than it's worth, and it's not exactly hyper-readable either. 

I'm curious what you think -- but to your question -- I think what you've done in this migration is totally fine if it's working for you. And I agree that unification of them is 💯, which is why I started thinking about how we might be able to eliminate having to track it at all, so it can't go out of sync.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
frankhereford: 6 one way, half dozen the other. This doesn't feel wrong to me at all. The useEffect would be the way to go if there were other places in the app that manipulate the data coming out of the `useQuery()`, but I don't think that'll be the case here.

moped-editor/src/queries/components.js
```diff
@@ -10,10 +10,93 @@ export const GET_COMPONENTS_FORM_OPTIONS = gql`
...
+      feature_drawn_lines {
+        id
+        geometry: geography
+        source_layer
+      }
+      feature_drawn_points {
+        id
+        geometry: geography
+        source_layer
+      }
```
frankhereford, line 75: I see how the new `graphql-engine` relationships are created above linking `moped_proj_components` to these `feature_*` tables, which allows them to be drawn out here in the `moped_proj_components` portion of the GQL query. I also see the `project_geography` query, which feels like it might be pulling a very similar set of data. I think you have a comment below about using the view more, so there may be context I have not gotten to yet.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
frankhereford: It's got to be kept somewhere, right? I don't feel like there's anything wrong with this, and given that the geometry is not a fluid, changing thing on the database, I totally want to get-it-once and hang onto it. 

Now, I do think that now that you've nailed down the shape of the feature collection that is most useful to your app, I want to take a swing at making the view generate that exact shape so one could just get the data out of the graphql endpoint and start using it without having to transmogrify it. Nothing urgent, as your system here is primo -- it would be a practice exercise in SQL more than anything.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
frankhereford: My inclination is to not break things out into a helper until they are called multiple times and you're doing it to DRY up your code. This, as written, is the process for building up the data for insertion, and it's going to be a complex operation no matter where the code steps reside -- at least like this, you can see it all at once! This is actually a straight-forward, step-by-step set of instructions for the task, as written. I think you've nailed it.

(This is a weakly held assertion about utility / helper collections - it's not been too long since I wrote the VZ ETL and I pulled a ton of stuff into a helper when it was only called once -- because it gave a big boost to readability. Aways more than one way to do it!)

frankhereford (approved):
@mddilley, I saw your hard-mode instructions, and I couldn't resist. So glad I did -- it's awesome to browse real projects and look at real maps. Super coooooool!!! 🔥🗺️. I have tested it six ways to Sunday, and I can't seem to break it. 

# 🚢 🚢 ✨ 

Thanks for  your comments inlined into the diff portion of this PR. This branch's changes are pretty big, and it's hard to "think your way through it all" as a non-author. Your questions provide context hints which are massively helpful. I worry a little that my thoughts are going to be out in left field -- if so, forgive me!

Also, I wanted to note that my comment about using `ST_GeometryType` is very marginal and lightly held.  I think I may be confused too and wildly missing the point of your question. 🤦

Can't wait to hear what other folks say. Thank you for this awesome work!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatures.js
johnclary: i just noticed we have the old `project_extent_id` column in `feature_street_segment`. that column theoretically holds the same data as `ctn_segment_id`—we renamed `project_extent_id` to  `ctn_segment_id` at some point and never cleaned up after ourselves :/ 

i will make a backlog issue to handle that in the transpose script and drop `project_extent_id` - does that sound like the right way to deal with it?

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -108,8 +144,76 @@ export default function MapView({ projectName, projectStatuses }) {
   };
 
   const onSaveComponent = () => {
```
johnclary, line 146: as these callbacks continue to accumulate here in `index.js` - i'm wondering if/when we want to think about memoizing them with `useCallback`. my limited experience with that hook is that it can be a real chore to get the dependencies right.

i definitely think there are diminishing returns trying to optimize—whether it's the added complexity of `useCallback` or the dependency mess of extracting the function out of the component with a long list of params. 

taking a really quick look, this `onSaveComponent` actually might not be so bad if it were extracted? i trust your judgement on this @mddilley but wanted to throw that out there.

johnclary (approved):
boom! it's so fun to see this in action. i love the way you've organized the code and tackled the problem. this was a massive lift - thank you!

i offered a few comments and identified a bug that might not be endemic to this PR. i'm stamping with 🚢 🚢 🚢  and please let me know what makes sense to open as backlog issues.

🙌

moped-database/migrations/1668201314245_alter_table_public_feature_drawn_lines_alter_column_render_type/up.sql
mddilley: I'm looking at the code again, and I don't use `render_type` anywhere. I was going to use it to style points/lines using [Mapbox expressions](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/), but I ran into a wall there with `visibility` not supporting expressions. 

All of the style differentiation is handled by `line_representation` (which is captured from the `moped_components` table on create), and I agree that it would be great to move to something derived from the actual geometry - that is a great idea @frankhereford. 👍 

So, we could probably drop it, and this migration really doesn't serve any purpose.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
mddilley: Cool! I'm going to test this out before merging ✨

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
mddilley: Agreed, y'all! I was hoping that the code would guide us to some thoughts on what shape we could get out of the DB down the road. 💪

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatures.js
mddilley: That sounds perfect to me! `ctn_segment_id` is great since it is easier see that it connects to the source.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeData.js
```diff
@@ -0,0 +1,36 @@
...
+  const componentGeographyMap = {};
+
+  components.forEach((component) => {
+    const currentComponentId = component.component_id;
+
+    const currentFeature = {
+      type: "Feature",
+      properties: { ...component.attributes },
+      geometry: component.geometry,
+    };
```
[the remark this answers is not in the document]
mddilley: Cool, agreed! Glad that the code is clear enough to see things - that was my hope for sure! A backlog issue would be great. 🙏

moped-editor/src/queries/components.js
```diff
@@ -10,10 +10,93 @@ export const GET_COMPONENTS_FORM_OPTIONS = gql`
...
+      geometry: geography
+      component_id
+      attributes
+    }
+  }
+`;
+
+export const DELETE_PROJECT_COMPONENT = gql`
+  mutation DeleteMopedComponent($projectComponentId: Int!) {
+    update_moped_proj_components(
```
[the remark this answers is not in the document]
mddilley: Good call, I'll patch this right up. 🩹 And, yes, I think the same about features - they'll just be orphaned now instead of needing a soft delete column.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -108,8 +144,76 @@ export default function MapView({ projectName, projectStatuses }) {
...
+      feature_drawn_lines: { data: drawnLinesToInsert },
+      feature_drawn_points: { data: drawnPointsToInsert },
+    };
+    /* End data preparation */
+
+    addProjectComponent({ variables: { object: newComponentData } }).then(
+      () => {
+        refetchProjectComponents();
+      }
+    );
```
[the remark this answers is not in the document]
mddilley: Cool, I'll add a log here for now. It would be great to have some feedback like "Try again" or something. Good for a backlog issue, and we might be able to pull an idea from Rebecca's designs IIRC.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
```diff
@@ -41,8 +55,8 @@ export const makeCapturedFromLayerFeature = (
     properties: {
       ...featureFromAgolGeojson.properties,
       id: featureFromAgolGeojson.id,
-      // AGOL data doesn't include layer so we grab it from the clicked Mapbox feature
-      _layerId: clickedFeature.layer.id,
+      // AGOL data doesn't include source layer ID so we grab it from the clicked Mapbox feature
+      _layerId: clickedFeature.layer.source,
     },
   };
```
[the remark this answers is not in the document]
mddilley: Nice, thanks - I'll check this one out and see what is up. This seems like a new bug 🐛

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
mddilley: Cool, I'll give it a whirl to see and agree that it could be good to wait to optimize if it is too burdensome. 🙏

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
mddilley: I worked `useMemo` in, and I like it a lot since the previous state setter was only used once. TIL! 🙏

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeData.js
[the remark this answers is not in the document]
johnclary: 👍 will do

moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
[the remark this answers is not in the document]
mddilley: I think that I figured this one out. The CTN layers were rendered below the muted layer so I moved them up to the top of the stack but below the labels. I pushed this update, and hopefully no more bugs. 🤞

roseeichelmann (approved):
im not set up to run locally yet so i just tested in "normal mode", but all the steps work as expected! once again impressed by how much code you are pumping out, awesome work mike!

one thing i have noticed for me in the map view both in this PR and a previous PR i reviewed is that when i click to add a new component sometimes the gridded lines take forever to show up even with me zooming in and zooming out, so i cant add the component. i learned that a "hack" to this is clicking on the drawing tool and then clicking it again to go out of drawing mode and then the gridded lines show up. but it for moped users it may be frustrating for them if the grids never show up? unrelated to this PR but thought I would mention it and im curious if this happens to others! and once again this only happens sometimes, a lot of the time they show up pretty quick so im not sure if this is even worth delving into.

mddilley:
@roseeichelmann @tillyw Thank y'all for the reviews! 

Rose - I created a https://github.com/cityofaustin/atd-data-tech/issues/10865 for the bug with the street segments layer not showing since I'm pretty sure that I've experienced delay in loading that layer too. 🙏 

Tilly - I patched that bug in [this commit](https://github.com/cityofaustin/atd-moped/pull/877/commits/1252b8c359871b38894343d9b391b57fe5a76d9f) but didn't redeploy the test instance. Sorry about that! What happened was that the selectable layer was hiding underneath the blue and greyed out line layers so it couldn't be clicked.

PR #880, opened 2022-11-16
Set initial map extent from existing project components

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10426

This PR updates the components map to zoom to a bounding box that has all of a project's component features within. If there are no existing components, the map initializes to its default coordinates and zoom.

## Testing
**URL to test:** 

https://md-10426-zoom-on-load--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to https://md-10426-zoom-on-load--atd-moped-main.netlify.app/moped/projects/156?tab=summary
2. Go to the **Map** tab and you should see the map zoom to the existing components
3. You can also create a new project to see the default zoom, add some components, and then leave and revisit the **Map** tab to see the zoom on load

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
@@ -1,4 +1,4 @@
-import { useState, useRef } from "react";
+import { useState, useRef, useEffect } from "react";
 import { useMutation, useQuery } from "@apollo/client";
 import { useParams } from "react-router";
 import { makeStyles } from "@material-ui/core/styles";
@@ -121,6 +121,34 @@ export default function MapView({ projectName, projectStatuses }) {
       },
     }
   );
+
+  const [hasMapZoomedInitially, setHasMapZoomedInitially] = useState(false);
+
+  useEffect(() => {
+    if (!data || hasMapZoomedInitially) return;
+
+    if (data.project_geography.length === 0) {
+      setHasMapZoomedInitially(true);
+    }
+
+    const featureCollection = {
+      type: "FeatureCollection",
+      features: data.project_geography,
+    };
+
+    const bboxOfAllFeatures = bbox(featureCollection);
+    mapRef.current.fitBounds(bboxOfAllFeatures, {
+      maxZoom: 19,
+      padding: {
+        top: 75,
+        bottom: 75,
+        left: 75,
+        right: 75,
+      },
+    });
+
+    setHasMapZoomedInitially(true);
+  }, [data, hasMapZoomedInitially]);
 
   /* fits clickedComponent to map bounds - called from component list item secondary action */
   const onClickZoomToComponent = (component) => {
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
@@ -33,6 +33,7 @@ import {
 } from "./utils/makeFeatures";
 import { makeComponentFeatureCollectionsMap } from "./utils/makeData";
 import { getDrawId } from "./utils/features";
+import { fitBoundsOptions } from "./mapSettings";
 
 const drawerWidth = 350;
 
@@ -137,15 +138,7 @@ export default function MapView({ projectName, projectStatuses }) {
     };
 
     const bboxOfAllFeatures = bbox(featureCollection);
-    mapRef.current.fitBounds(bboxOfAllFeatures, {
-      maxZoom: 19,
-      padding: {
-        top: 75,
-        bottom: 75,
-        left: 75,
-        right: 75,
-      },
-    });
+    mapRef.current.fitBounds(bboxOfAllFeatures, fitBoundsOptions);
 
     setHasMapZoomedInitially(true);
   }, [data, hasMapZoomedInitially]);
@@ -159,16 +152,7 @@ export default function MapView({ projectName, projectStatuses }) {
     // close the map projectFeature map popup
     setClickedProjectFeature(null);
     // move the map
-    mapRef.current?.fitBounds(bbox(featureCollection), {
-      maxZoom: 19,
-      // accounting for fixed top bar
-      padding: {
-        top: 75,
-        bottom: 75,
-        left: 75,
-        right: 75,
-      },
-    });
+    mapRef.current?.fitBounds(bbox(featureCollection), fitBoundsOptions);
   };
 
   const onSaveComponent = () => {
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/index.js
@@ -19,7 +19,7 @@ import EditModeDialog from "./EditModeDialog";
 import ComponentMapToolbar from "./ComponentMapToolbar";
 import ComponentListItem from "./ComponentListItem";
 import DraftComponentListItem from "./DraftComponentListItem";
-import { useAppBarHeight } from "./utils/map";
+import { useAppBarHeight, useZoomToExistingComponents } from "./utils/map";
 import {
   ADD_PROJECT_COMPONENT,
   GET_PROJECT_COMPONENTS,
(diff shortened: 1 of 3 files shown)


frankhereford (approved):
Works like a charm 🍀. Thanks Mike!

You mention in your instructions about observing [the default viewport configuration](https://github.com/cityofaustin/atd-moped/blob/a783fc4544f85c2e98a6fdb09efad02fe3beb5b8/moped-editor/src/views/projects/projectView/ProjectComponents/mapSettings.js#L30-L34). 

I like the point in there, being mildly downtown, but you may want to make it `30.2747, -97.7406` which is the minimum precision needed to get the point inside the footprint of the state capitol building. 🏛️ (This change is very unimportant, and it's almost more of an easter egg to the GIS nerds out there who look for this kind of thing.)

That said, I do think the default zoom level of 15 is pretty tight in. I'd suggest coming off that until you can at least see the east and west bounds of the city, and maybe even the north and south ones? Perhaps put another way, what do you think about zooming it out so that a user can see that they are looking at a map of Austin without having to zoom out?

patrickm02L (approved):
Good to go! 🤘

https://user-images.githubusercontent.com/5566727/202506693-cdb81085-972b-4824-9003-382ca4a18991.mp4

chiaberry (approved):
It feels a little too close of a zoom for me, especially when adding just a point. its a smooth zoom though. 

but functinally and code wise, great!

tillyw (approved):
this looks great to me! i like the level of zoom per component and that it just fits whatever components you have. my only thought is that the default zoom (no components) shows a pretty small area and i wonder if it should be a little further out so the user can see more of the map initially. this is subjective and i think it's good to go as-is, just wanted to point that out!

Screen Shot 2022-11-17 at 2 47 45 PM

tillyw:
also! i don't know if this has been discussed and i think this would be a new issue but a 'zoom to extent' button that shows all components could be a useful addition in the future. for example, if you add a new component and want to see the extent of all existing components without reinitializing the map or refreshing the page. just an idea!

patrickm02L:
> It feels a little too close of a zoom for me, especially when adding just a point. its a smooth zoom though.

I agree with the sentiment now that I've had a minute to step away. The zoom takes up the whole window, I think you could step it back about 25% so there's a little more padding around what's drawn.

frankhereford:
I'll touch on this in dev sync, but I'm a little cautious feeling about wanting to prescribe the library's "zoom to a set of features" zoom level, because to do that, I imagine Mike would have to build up the collection of features an feed them through a GIS library that allows him to buffer by some amount. 

That alone is a lift, and then you have to dynamically compute the linear distance of that buffer, which has to scale with something like the `sqrt()` of the area or the longest X/Y dimension -- and even then, start to take in consideration the general aspect of the bounding box. That GIS work could be integrated into the DB via PostGIS, but still, it has to be worked out. Putting a GIS lib in the front-end app is even less appealing.

-----

Also, on my comment about zoom - that is about the initial zoom on a map that has zero features on it. When there is nothing on the map, I am suggesting that "Austin" be on the map.

johnclary:
> i think this would be a new issue but a 'zoom to extent' button that shows all components could be a useful addition in the future. for example

@tillyw agreed and i stashed an issue for it: https://github.com/cityofaustin/atd-data-tech/issues/10826

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -131,16 +134,7 @@ export default function MapView({ projectName, projectStatuses }) {
...
-      maxZoom: 19,
-      // accounting for fixed top bar
-      padding: {
-        top: 75,
-        bottom: 75,
-        left: 75,
-        right: 75,
-      },
-    });
+    mapRef.current?.fitBounds(bbox(featureCollection), fitBoundsOptions);
```
johnclary, line 137: thanks for applying the same options here as well!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
```diff
@@ -170,3 +172,32 @@ export const useSubcomponentOptions = (component) =>
...
+ * Use Mapbox fitBounds to zoom to existing project components feature collection
+ * @param {Object} mapRef - React ref that stores the Mapbox map instance (mapRef.current)
+ * @param {Object} data - Data returned from the moped_components query
+ * @param {Array} data.project_geography - Array of existing component features
+ */
+export const useZoomToExistingComponents = (mapRef, data) => {
+  const [hasMapZoomedInitially, setHasMapZoomedInitially] = useState(false);
+
+  useEffect(() => {
+    if (!data || hasMapZoomedInitially) return;
```
johnclary, line 186: you might want to add `mapRef` to this check and return if we don't have a `current` value? in my testing the value is never undefined but i would add it mostly b/c i don't fully understand refs.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
```diff
@@ -170,3 +172,32 @@ export const useSubcomponentOptions = (component) =>
...
+    const featureCollection = {
+      type: "FeatureCollection",
+      features: data.project_geography,
+    };
+
+    const bboxOfAllFeatures = bbox(featureCollection);
+    mapRef.current.fitBounds(bboxOfAllFeatures, fitBoundsOptions);
+
+    setHasMapZoomedInitially(true);
+  }, [data, hasMapZoomedInitially, mapRef]);
```
johnclary, line 202: this hook covers it and works perfectly!

johnclary (changes requested):
ah, this is such another nice improvement.

i agree with the gang that the zoom doesn't need to be so tight.

when the map is blank (when we pull from `initialViewState`), i think a zoom of `12` is ideal.

when the map loads with components, i think a max zoom of `15` would be better. 

after clicking around, i think we want to use a separate max zoom for zooming to specific components—i'm assuming that a user wants a higher level of detail to inspect a component and possibly start editing.

all that to say, would you mind adding yet another configuration to separate out the clicked component effect. so:

- initial zoom (no components): `12`
- initial maxZoom with components: `15`
- maxZoom on clicked component: `19`

and +1 if you want to through Frank's easter egg lat/lon into the `initialViewState`.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
mddilley: All good, I'll expand that config with keys for each case! 🙏

moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
johnclary: > the Mapbox instance is assigned to mapRef.current on first render since it is passed as its ref prop

thanks for that! sounds good, but does seems like an extreme edge case unless we were caching queries.

mddilley:
Thanks everyone for the feedback! Just updated the test instance with the last two commits, and that includes:
- John's suggestions for the zoom for no features, existing features, and clicked features
- Bumping up the padding around the zoomed features (this is most noticeable when zooming to features with the magnifier icon)
- Frank's suggestion to update the default coords of the map

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -131,16 +134,10 @@ export default function MapView({ projectName, projectStatuses }) {
...
-      padding: {
-        top: 75,
-        bottom: 75,
-        left: 75,
-        right: 75,
-      },
-    });
+    mapRef.current?.fitBounds(
+      bbox(featureCollection),
+      fitBoundsOptions.zoomToClickedComponent
```
johnclary, line 139: nice!

johnclary (approved):
thanks @mddilley!

🚢 🚢 🚢 🚢 🚢 🚢 🚢

roseeichelmann (approved):
Mike this is such a nice improvement! I like the updated zoom level, looks just right to me! 🚀🚀

PR #883, opened 2022-11-18
Use REST connector to update moped project activity log events

chiaberry:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/10600

evaluate REST connector to update activity log when changes are done to the moped_project table
Also https://github.com/cityofaustin/atd-data-tech/issues/10811, tracking changes to moped_proj_tags

This is the companion PR for changes to the moped-etl code, to get the test environment deployed with the new environment variable. https://github.com/cityofaustin/atd-moped/pull/887/

## Testing
**URL to test:** 
https://10600-rest-activity--atd-moped-main.netlify.app/moped/projects/227?tab=summary

**Steps to test:**
Update something from the moped project such as the description, or ecapris ID or website. 
Make changes to the project tags. 

Then check the activity log. You may need to refresh the activity log to see any changes. I see the same behavior in staging right now, updates dont immediately show on the activity log page. I think the query isn't being refetched between tab changes. Let me know if fixing that is a blocker for this PR to be approved. 

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -3533,45 +3536,70 @@
...
-            - parent_project_id
+            - project_priority
+            - interim_project_id
+            - ecapris_subproject_id
       retry_conf:
         interval_sec: 10
         num_retries: 0
         timeout_sec: 60
-      webhook_from_env: MOPED_API_EVENTS_URL
+      webhook_from_env: HASURA_ENDPOINT
```
chiaberry, line 3573: This needs to be updated in our test evironment, staging environment and prod environments

chiaberry: adding to staging and prod involves updating the task definitions in AWS, in test see this PR https://github.com/cityofaustin/atd-moped/pull/887/files

moped-database/metadata/tables.yaml
```diff
@@ -3546,41 +3596,58 @@
...
-            - added_by
-            - milestone_id
-            - project_id
-            - project_importance
-            - project_length
-            - project_order
             - project_sponsor
-            - timeline_id
             - task_order
+            - parent_project_id
```
chiaberry, line 3601: this is most all the columns on moped_project, except some of the ones we don't use (project_length) but most importantly it also omits "updated_at". Since we are updating updated_at with the REST connector, we don't need to trigger the REST connector again.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -74,33 +77,30 @@ const ProjectActivityLog = () => {
...
+    useActivityLogLookupTables();
 
   const { loading, error, data } = useQuery(PROJECT_ACTIVITY_LOG, {
     variables: { projectId },
-    onCompleted: data => getLookups(data, "activity_log_lookup_tables"),
+    onCompleted: (data) => getLookups(data, "activity_log_lookup_tables"),
   });
 
-  const [activityId, setActivityId] = useState(null);
+  // const [activityId, setActivityId] = useState(null);
```
chiaberry, line 88: i just commented out the details view in the activity log for now. If we think its not coming back at all in the near future, I can remove this code instead of just commenting out

moped-database/metadata/tables.yaml
```diff
@@ -3546,41 +3596,58 @@
...
-        - name: MOPED_API_EVENT_NAME
-          value: activity_log
+        - name: x-hasura-admin-secret
+          value_from_env: HASURA_GRAPHQL_ADMIN_SECRET
+      request_transform:
+        body:
+          action: transform
+          template: |-
+            {
+                            "query": "mutation InsertActivity($object: moped_activity_log_insert_input!, $project_id:Int!, $updated_at:timestamptz ) { insert_moped_activity_log_one(object: $object) { activity_id } update_moped_project_by_pk(pk_columns: {project_id: $project_id}, _set: {updated_at: $updated_at}) { updated_at }}",
```
chiaberry, line 3631: the mutation is all on one line, scroll far to the right to see the updated_at mutation

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -117,38 +117,54 @@ const ProjectActivityLog = () => {
...
 
-    outputList.forEach(event => {
-      // If this is the creation of a project
-      if (
-        event.record_type === "moped_project" &&
-        event.operation_type === "INSERT"
-      ) {
-        // Remove that object from the array and add it back on at the end
-        outputList.splice(outputList.indexOf(event), 1);
-        outputList.push(event);
```
chiaberry, line 140: this was code that went through all of the events and looked for the project creation event, then put it at the end of the array. 

I don't think its necessary, however I have found sometimes that seed data will show up in the activity log with project creation appearing after tags being added for example. I believe this would only happen in the seed data, since the events are sorted by creation date/time, and only when inserting seed data would all the events have the same time.

patrickm02L (commented):
Tested all the fields to see what would update in the Activity Log. 
_*= Not captured in Activity Log_

00:04 **Description** - Refresh to see update
*00:21 **Status update**
01:05 **Lead** - Refresh to see update
01:23 **Sponsor** - Refresh to see update
*01:41 **Partners** 
*02:20 **Project Type**
02:39 **Website** - No refresh
03:01 **Work assignment ID** - Refresh
03:22 **Interim MPD (Access) ID** - Refresh
03:33 **eCAPRIS Subproject ID** - Refresh
03:53 **Tag** - Works. Refresh
*04:12 **Subproject**
04:40 **Website** - Had to refresh this time.


https://user-images.githubusercontent.com/5566727/203649001-3caa19aa-8435-4c2f-bcb8-373429acba57.mp4

johnclary:
@chiaberry to get this running locally:
- i patched in the json blob for the JWT secret
- added an `ACTIVITY_LOG_API_SECRET` variable, which is the same as the `HASURA_GRAPHQL_ADMIN_SECRET` but is allowed by Hasura to be used in the events configuration. In `metadtaa/tables.yaml` I replaced `HASURA_GRAPHQL_ADMIN_SECRET` with `ACTIVITY_LOG_API_SECRET`

Here's the docker-compose:

```yaml
# Architecture Independent Docker Stack Components for Moped Development Environment

version: '3.7'
services:
    hasura:
        image: hasura/graphql-engine:v2.11.2
        restart: always
        depends_on:
            - moped-pgsql
        expose:
            - 8080
        ports:
            - 8080:8080
        environment:
            - HASURA_GRAPHQL_ENABLE_CONSOLE=false
            - HASURA_GRAPHQL_ENABLE_TELEMETRY=false
            - HASURA_GRAPHQL_DATABASE_URL=postgres://moped:moped@moped-pgsql:5432/moped
            - HASURA_GRAPHQL_CONSOLE_ASSETS_DIR=/srv/console-assets
            - MOPED_API_ACTIONS_URL=http://localhost:5000/actions
            - MOPED_API_EVENTS_URL=http://localhost:5000/events
            - HASURA_ENDPOINT=http://localhost:8080/v1/graphql
            - MOPED_API_APIKEY=DuMmyApiKeyHFVOVto19otC1wX6sP2N0VSKrCD70L10B7Sm525ZR6L672i2F79M9!
            - 'HASURA_GRAPHQL_JWT_SECRET={"type":"RS256","jwk_url": "https://cognito-idp.us-east-1.amazonaws.com/us-east-1_U2dzkxfTv/.well-known/jwks.json","claims_format": "stringified_json"}'
            - HASURA_GRAPHQL_ADMIN_SECRET=hasurapassword
            - ACTIVITY_LOG_API_SECRET=hasurapassword
        command: ["graphql-engine", "serve",  "--enable-console"]


# There is a volume which is common
volumes:
    moped-db-vol:
        driver: local
```

And here's a snippet from tables.yaml:

```yaml
...
  event_triggers:
    - name: activity_log_moped_project_tags
      definition:
        delete:
          columns: '*'
        enable_manual: false
        insert:
          columns: '*'
        update:
          columns:
            - is_deleted
            - tag_id
            - id
            - project_id
      retry_conf:
        interval_sec: 10
        num_retries: 0
        timeout_sec: 60
      webhook_from_env: HASURA_ENDPOINT
      headers:
        - name: x-hasura-admin-secret
          value_from_env: ACTIVITY_LOG_API_SECRET
      request_transform:
...
```

johnclary:
I created https://github.com/cityofaustin/atd-data-tech/issues/10896 to cover the work of converting the other activity log events to this new pattern.

moped-database/metadata/tables.yaml
```diff
@@ -3206,6 +3209,53 @@
...
+            - id
+            - project_id
+      retry_conf:
+        interval_sec: 10
+        num_retries: 0
+        timeout_sec: 60
+      webhook_from_env: HASURA_ENDPOINT
+      headers:
+        - name: x-hasura-admin-secret
+          value_from_env: ACTIVITY_LOG_API_SECRET
```
johnclary, line 3233: I don't think this is documented, but hasura will throw an error if you try to reference any  `HASURA_GRAPHQL_` env variables in the event trigger configuration.

Just a guess, but I suppose they're trying to prevent you from sending your admin secret off to some unknown endpoint. In our case, we're feeding the secret back to our own API, so I don't think this is a concern. It does create the awkward situation where we have the redundant var `ACTIVITY_LOG_API_SECRET` which is fact the `HASURA_GRAPHQL_ADMIN_SECRET`.

johnclary: as an alternative—we _could_ use the JWT token from a session variable. the problem there is that any events that are triggered outside of a session (e.g. a session-less python backfill script) will fail.

i think the simplest approach is to do what we see here 🤷‍♂️

moped-database/config/hasura.local.yaml
```diff
@@ -4,3 +4,4 @@ metadata_directory: metadata
 actions:
   kind: synchronous
   handler_webhook_baseurl: http://localhost:3000
+admin_secret: hasurapassword
```
chiaberry, line 7: the schemaspy github action uses this file, so this needs to be updated for the db docs to keep working

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -117,38 +116,70 @@ const ProjectActivityLog = () => {
...
       if (
-        event.record_type === "moped_project" &&
-        event.operation_type === "INSERT"
+        outputEvent.record_type === "moped_project" &&
+        outputEvent.operation_type === "INSERT"
       ) {
-        // Remove that object from the array and add it back on at the end
-        outputList.splice(outputList.indexOf(event), 1);
-        outputList.push(event);
+        createdEvent = outputEvent;
```
chiaberry, line 165: turned out this is needed, after testing locally with prod data I found that many projects have the status creation or if its a signal project, the proj_component event is listed as happening before project creation.

moped-database/config/hasura.local.yaml
johnclary: i believe folks will also need to run a one-time ` ./hasura-cluster setenv local` to make sure their local `config.yaml` gets updated with this variable. it was enlightening to debug this with you @chiaberry!

johnclary:
just an aside—now that’d i’ve had some experience watching the REST connector activity log overwhelm hasura when doing bulk write operations, i suspect that if our seeds grow larger (we only have 3 projects) we would want to disable the rest connectors before applying seeds. that's a problem for another day for sure.

johnclary (approved):
i re-tested your latest changes locally. this looks ready to ship to me! 🚀 🚢

moped-editor/src/utils/activityLogHelpers.js
```diff
@@ -76,6 +76,7 @@ export const buildLookupQuery = tableNames => {
 export const getActivityLogTableNames = (response, lookupDataKey) => {
   const lookupTableData = response?.[lookupDataKey];
 
+  // would this length ever be more than one??
```
mddilley, line 79: I looked into this a little, and, even in a brand new project, this length is 1 and then grows as more events are added that need lookup tables. Maybe there was a time when it could be 0 and that changed over time?

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
mddilley: Cool! I'm also curious about how much this detailed view is used (if at all). I think this will be a good way to see if it is missed. 👍

mddilley (approved):
So many nice improvements! ✨ I spun this up locally, and it was super fun to see those activities show up in the log. I really like how this work led to some rethinking about how activity is stored in the DB and shown in the UI too. And, if we ever need to use SQS in the future, maybe we could take John's lead and use Node for the script 😈

roseeichelmann (approved):
This looks great, most everything is updating on the activity log with a refresh. I'm getting the same responses as Patrick for the most part except the website did require a refresh. Also, I think in a a future issue it would be good to capture the updates to lead and sponsors as their corresponding label and not the ID #. 🚢🚢

frankhereford:
@chiaberry - I know this is after your merge, but I wanted to thank you for working all of this out and letting me pair with you today while you merged this and updated the task definitions in ECS. This is 🚀🔥.

PR #884, opened 2022-11-19
sets default project list sort and displays arrow

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10671

this fixes the bug where the column sorting icon on the projects list view only appeared on hover. it's also a step toward completing https://github.com/cityofaustin/atd-data-tech/issues/10538

in order to do this, i had to hardcode a default sorting order, so i used  the existing default (descending by date modified) set `ProjectsListViewQueryConf`. ultimately, we would populate the new default using url parameters from the previous session. i thought it made sense to always have a column sorted since this is what it will ultimately be like when the sorting is persistent. also, it used to be that if you click a column name a third time, the arrow disappears and the table is sorted back to default. i thought this was unnecessary and inconsistent with having a persistent icon, so i simplified the behavior.

https://deploy-preview-884--atd-moped-main.netlify.app/

## Testing
**URL to test:** 

https://deploy-preview-884--atd-moped-main.netlify.app/

**Steps to test:**
- go to the projects list page
- the columns should be sorted descending by date modified by default with an arrow indicating the order
- click the date modified column, the order should swap
- click another column, it should be sorted in the same order
- click it again, the order should swap

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -86,6 +86,9 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
    * This overrides MaterialTable props and determines how
    * the table is sorted when the page loads. Must remain consistent
    * with the sorting order passed in from ProjectsListViewQueryConf.
+   * @property {string} column - The column name in graphql to sort by
+   * @property {integer} columnId - The column id in graphql to sort by
+   * @property {string} order - Either "asc" or "desc" or ""
    */
   const defaultSortingProperties = {
     column: "updated_at",
@@ -96,10 +99,8 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
   /**
    * Stores the column name and the order to order by
    * @type {Object} sort
-   * @property {string} column - The column name in graphql to sort by
-   * @property {string} order - Either "asc" or "desc" or "" (default: "")
    * @function setSort - Sets the state of sort
-   * @default {{value: "", column: ""}}
+   * @default {defaultSortingProperties}
    */
   const [sort, setSort] = useState(defaultSortingProperties);
 
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -82,6 +82,18 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
   });
 
   /**
+   * The default sorting properties applied to the table.
+   * This overrides MaterialTable props and determines how
+   * the table is sorted when the page loads. Must remain consistent
+   * with the sorting order passed in from ProjectsListViewQueryConf.
+   */
+  const defaultSortingProperties = {
+    column: "updated_at",
+    columnId: 8,
+    order: "desc",
+  };
+
+  /**
    * Stores the column name and the order to order by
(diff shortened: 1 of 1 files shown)


tillyw:
@johnclary @chiaberry i have a question before i submit this for review. if we're going to have the sorting arrow appear when you load the table, we have to choose a column and direction to sort by default. from what i can tell, the default sorting method is 'last updated' in descending order, so recently edited projects appear first. should i make that the default sorting mechanism?

johnclary:
@tillyw yes please - last updated!

tillyw:
y'all will notice a few quirks with this issue that i tried for hours to fix and then paired with @johnclary on for over an hour before we came to the conclusion that they are a result of the way the data is queried and will require a major rework to fix. the main quirk is that when you click a new header, the table rerenders, which may be jarring to the user. this is how the table has already been functioning the abstract and query were designed, so that it doesn't flood the table with information if there are thousands of records and we only want to display the top 25 results (based on sorting) to the user. as of now there's no way to reorder the columns without retriggering the query and rerendering the entire table. @johnclary please correct me or elaborate if i'm not explaining this properly!

mddilley (approved):
Works great and looks great - it is really nice to have the arrow always present to see what sort is applied! ✨ Thank you for the write up about the table rerendering too - it will be good to keep thinking on how to improve that in the future.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -81,6 +81,18 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
+   * This overrides MaterialTable props and determines how
+   * the table is sorted when the page loads. Must remain consistent
+   * with the sorting order passed in from ProjectsListViewQueryConf.
+   */
+  const defaultSortingProperties = {
+    column: "updated_at",
+    columnId: 8,
+    order: "desc",
+  };
+
```
johnclary, line 95: 👍

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -178,9 +187,7 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
    * Query Management
    */
   // Manage the ORDER BY clause of our query
-  if (sort.column !== "" && sort.order !== "") {
-    query.setOrder(sort.column, sort.order);
-  }
+  query.setOrder(sort.column, sort.order);
```
johnclary, line 190: this setter runs every render, and is somehow not causing an infinite loop with the way it interacts with the query that `useQuery` is consuming. anyway, we didn't go to far into the code, but this hints at some underlying GQLAbstract behavior that needs a deeper look.

johnclary (approved):
thanks @tillyw! 🚢

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -89,10 +101,7 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
    * @function setSort - Sets the state of sort
    * @default {{value: "", column: ""}}
```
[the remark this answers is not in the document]
tillyw: thanks for catching this @roseeichelmann! i've updated the notes to reflect that this piece of state accepts the defaultSortingProperties object as its default and updated the comments for that object. merging away!

PR #890, opened 2022-11-29
sets project list view paging to 100

tillyw:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10885

## Testing
**URL to test:** 

https://deploy-preview-890--atd-moped-main.netlify.app/moped/

**Steps to test:**
- go to project list view, you should see 100 projects displayed sorted in descending order by date modified
- click the page forward button to see the rest of the records
- click a different column header, the table should rerender on page 1 with the top 100 records that match the column you clicked, in descending order
- go to the next page, you should see the remaining records according to the sort you applied
- reverse the sort, you should be taken back to the top 100 records (page 1) in ascending order according to the header you clicked

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -538,6 +538,13 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
     query.clearOrderBy();
     const columnName = columns[columnId]?.field;
 
+    // Resets pagination to 0 when user clicks a header to display most relevant results
+    setPagination({
+      limit: query.limit,
+      offset: query.offset,
+      page: 0,
+    });
+
     if (sort.column === columnName) {
       // If the current sortColumn is the same as the new
       // then invert values and repeat sort on column
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
@@ -241,6 +241,6 @@ export const ProjectsListViewQueryConf = {
   },
   or: null,
   and: null,
-  limit: 25,
+  limit: 100,
   offset: 0,
 };


tillyw:
while working on this issue, i noticed a bug in both production and staging where, if you reverse the sort on any other page besides page one, you see a completely different set of records on that page than you saw before. this was super confusing! at @johnclary's suggestion, i added a fix to this pr that sends you back to the first page when you click a table header so you see the most relevant records according to the sort you applied. you can then page over from there. hope this makes sense!

mddilley (approved):
Another user request ready for release! 🎉  Thanks for the explanation on the pagination update, and I like the solution. The quirks that are being surfaced in these list view PRs are going to be so valuable when we get to enhancing it down the line.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -538,6 +538,13 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
     query.clearOrderBy();
     const columnName = columns[columnId]?.field;
 
+    // Resets pagination to 0 when user clicks a header to display most relevant results
+    setPagination({
+      limit: query.limit,
+      offset: query.offset,
+      page: 0,
+    });
```
johnclary, line 546: i would think we need to also touch `query.offset` to reset that back to `0` as well. but that's not the case. i spent a little bit of time trying to understand why this works, and i couldn't grok it. all that to say - 👍 thanks for figuring out how to patch this!

chiaberry: this drove me nuts a little, so I did some digging. I think what is happening is the setPagination state update that happens when someone clicks next on the page at the bottom does update the query.offset in GQLAbstract to reflect the correct offset, which is then put into the query that is generated by GQLAbstract. 

Then the whole page refreshes which has the side effect of resetting the state to have the offset from the config, i.e. 0. But doesn't update the query itself. 

Now how come it works on page 3? I ask myself the same question and I think the pagination state is *actually* being kept track of in the TablePagination component thats in the GridTablePagination component. and now i will stop digging around in this.

roseeichelmann (approved):
this works just as i would expect it to tilly, going back to the first page makes sense when applying a new sort and this is the way it works in a lot of other apps i feel like. the pagination and sorting is so elegant 💅  maybe in the future there could be an indicator of what page you are on at the top of the table too so if people are on like page 4 and then re-sort they will know right away that it took them back to page 1 rather than having to scroll to the bottom to see that? just a thought!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
tillyw: thanks y'all for digging into this rabbit hole with me! :p it's been driving me crazy too. but i'm glad this fix is ready to ship. maybe we should revisit this as tech debt at some point?

tillyw:
> this works just as i would expect it to tilly, going back to the first page makes sense when applying a new sort and this is the way it works in a lot of other apps i feel like. the pagination and sorting is so elegant 💅 maybe in the future there could be an indicator of what page you are on at the top of the table too so if people are on like page 4 and then re-sort they will know right away that it took them back to page 1 rather than having to scroll to the bottom to see that? just a thought!

thanks, @roseeichelmann ! i had the same thought about putting an indicator at the top of the table so the user knows what page their on, especially after resorting the table. i'm going to ship this for now but this should be a quick addition - i'll look into adding this!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
johnclary: +1 thanks for that research @chiaberry!

yes, we'll definitely revisit. i haven't been sure how to write up issues related to this, but i'll capture it somewhere. i'm particularly excited about [this plugin](https://github.com/hasura/ra-data-hasura) that @mddilley found that looks to me like a replacement for the graphql abstract.

PR #891, opened 2022-11-30
adds missing columns to csv export

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10765

this pr adds all missing columns displayed in the project list view to the csv export. it also adds a few helper functions to properly display missing and more complex values that were displayed as `null` or `[Object Object]` in the csv. it does not strip out the html tags from the `project_note` column - please let me know if you think we should add a helper function to do so!

## Testing
**URL to test:** 

https://deploy-preview-891--atd-moped-main.netlify.app/

**Steps to test:**
1. go to project list view and click download to export a csv
2. open the csv
3. you should see a readable export of all columns
4. any missing values should be displayed as "-"
5. columns with multiple values should be displayed as comma separated lists

---
#### Ship list
- [ ] Code reviewed 


chiaberry (changes requested):
The PR introduced a regression where the team members are no longer displayed as a block on the project list view, making it really wide
Screen Shot 2022-12-07 at 9 48 29 AM

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
```diff
@@ -18,30 +23,83 @@ export const ProjectsListViewExportConf = {
   },
```
roseeichelmann, line 23: should the project description column also have "-" for empty rows?

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -1,6 +1,14 @@
+export const filterNullValues = (value) => {
+  if (!value) {
+    return "-";
```
johnclary, line 3: @tillyw would you remove the `-` and just return an empty string `""` ? i think that will be preferred for folks working with this data in a spreadsheet editor.

johnclary (changes requested):
✨ this is going to be a major improvement—thank you!

please fix the team rendering issue and nix those dashes.

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
johnclary: +1 - it would be good to add the null handler to `project_name` and `project_description` - our DB schema does not allow these fields to be blank, but the null handler would future-proof this export.

tillyw:
@chiaberry thank you for catching the regression, i've fixed it. @johnclary i replaced all instances of "-" with ""

johnclary (approved):
🚢 looks great!

mddilley (approved):
Test this out and it looks great - really nice improvement in the exported .csv! 🚀 🚀 🚀

PR #194, opened 2021-01-28
Moped Tables | Refine GridTable layout/design on Projects page #5028

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/5028

Preview:
https://deploy-preview-194--atd-moped-main.netlify.app/

- [x] Change page title from "Projects List" to "Projects"
- [x] Make [fixed tabs](https://material-ui.com/components/tabs/#fixed-tabs) for "Search" (currently "General Search") and "Advanced Search" (currently "Filter Search") inside the search box.
- [x] Move "Download" button (see #4785) into search box.

PR #893, opened 2022-12-02
Bump Hasura test instances to v2.15.1

johnclary:
## Associated issues
Our test instances are lagging behind local, staging, and prod. I'm also hoping this clears an issue where Hasura occasionally crashes after replacing metadata.

I also noticed that we don't define a health check, which I think will help the container service restart more quickly if it fails. We don't use autoscaling, but I think the health check is a precursor for auto scaling.

## Testing
**URL to test:** 
[This test instance](https://4838-jc-migrate-access-db--atd-moped-main.netlify.app/moped/projects) is running these changes.

**Steps to test:**
i need to learn from Chia + Frank on how to test this. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-etl/prefect/tasks/ecs.py
+++ b/moped-etl/prefect/tasks/ecs.py
@@ -352,7 +352,6 @@ def create_task_definition(slug, api_endpoint):
                     {"containerPort": 8080, "hostPort": 8080, "protocol": "tcp"}
                 ],
                 "essential": True,
-                "healthCheck": { "command": ["CMD-SHELL", "curl -f http://localhost:8080/healthz || exit 1"]},
                 "environment": [
                     {"name": "HTTP_PORT", "value": "8080"},
                     {"name": "HASURA_GRAPHQL_ENABLE_CONSOLE", "value": "true"},
--- a/moped-etl/prefect/tasks/ecs.py
+++ b/moped-etl/prefect/tasks/ecs.py
@@ -352,6 +352,7 @@ def create_task_definition(slug, api_endpoint):
                     {"containerPort": 8080, "hostPort": 8080, "protocol": "tcp"}
                 ],
                 "essential": True,
+                "healthCheck": { "command": ["CMD-SHELL", "curl -f http://localhost:8080/healthz || exit 1"]},
                 "environment": [
                     {"name": "HTTP_PORT", "value": "8080"},
                     {"name": "HASURA_GRAPHQL_ENABLE_CONSOLE", "value": "true"},
--- a/moped-etl/prefect/tasks/ecs.py
+++ b/moped-etl/prefect/tasks/ecs.py
@@ -344,7 +344,7 @@ def create_task_definition(slug, api_endpoint):
         containerDefinitions=[
             {
                 "name": "graphql-engine",
-                "image": "hasura/graphql-engine:v2.7.0",
+                "image": "hasura/graphql-engine:v2.15.1",
                 # "image": "mendhak/http-https-echo:latest",
                 "cpu": 256,
                 "memory": 512,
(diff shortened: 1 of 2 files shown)


moped-etl/prefect/tasks/ecs.py
```diff
@@ -344,14 +344,15 @@ def create_task_definition(slug, api_endpoint):
...
-                "image": "hasura/graphql-engine:v2.7.0",
+                "image": "hasura/graphql-engine:v2.15.1",
                 # "image": "mendhak/http-https-echo:latest",
                 "cpu": 256,
                 "memory": 512,
                 "portMappings": [
                     {"containerPort": 8080, "hostPort": 8080, "protocol": "tcp"}
                 ],
                 "essential": True,
+                "healthCheck": { "command": ["CMD-SHELL", "curl -f http://localhost:8080/healthz || exit 1"]},
```
johnclary, line 355: the `-f` flag:

```
       -f, --fail
              (HTTP) Fail silently (no output at all) on server errors.
              This is mostly done to enable scripts etc to better deal
              with failed attempts. In normal cases when an HTTP server
              fails to deliver a document, it returns an HTML document
              stating so (which often also describes why and more). This
              flag will prevent curl from outputting that and return
              error 22.

              This method is not fail-safe and there are occasions where
              non-successful response codes will slip through,
              especially when authentication is involved (response codes
              401 and 407).

              See also --fail-with-body.
```

my new gripe is that it's really hard to quickly find docs for specific flags for command line tools. `curl` has soo many flags, and searching the `man` page for `-f` is not useful.  [here's](https://stackoverflow.com/questions/19198721/is-there-a-way-to-look-for-a-flag-in-a-man-page) a not helpful post about it.

chiaberry (commented):
I'll take this for a spin tomorrow, I want to just click approve but should run it just to be sure I suppose

moped-etl/prefect/tasks/ecs.py
```diff
@@ -344,14 +345,24 @@ def create_task_definition(slug, api_endpoint):
...
-                "image": "hasura/graphql-engine:v2.7.0",
+                "image": "hasura/graphql-engine:v2.15.1",
                 # "image": "mendhak/http-https-echo:latest",
                 "cpu": 256,
                 "memory": 512,
                 "portMappings": [
                     {"containerPort": 8080, "hostPort": 8080, "protocol": "tcp"}
                 ],
                 "essential": True,
+                # see: https://github.com/hasura/graphql-engine/issues/1532#issuecomment-1161637925
```
johnclary, line 356: tldr: the hasura image does not have `curl` installed, so we have to hack the health check a bit or use a custom image.

johnclary:
folks, i've only tested this by manually configuring an ECS task definition - [Revision #194](https://us-east-1.console.aws.amazon.com/ecs/home?region=us-east-1#/taskDefinitions/moped-test-graphql-endpoint/194).

i updated the docker-compose so that you can see how this works locally, but the AWS ECS config is a bit different and we'll need to test this the changes made through `ecs.py`.

```shell
❯ docker ps
CONTAINER ID   IMAGE                          COMMAND                  CREATED          STATUS                    PORTS                    NAMES
05cbb264672c   hasura/graphql-engine:latest   "graphql-engine serv…"   14 seconds ago   Up 10 seconds (healthy)   0.0.0.0:8080->8080/tcp   moped-database_hasura_1
d640fd9d44d3   frankinaustin/postgis:12-3.2   "docker-entrypoint.s…"   15 seconds ago   Up 14 seconds             0.0.0.0:5432->5432/tcp   moped-database_moped-pgsql_1
```

chiaberry (approved):
🥌

mddilley (approved):
Going to have to spend more time looking at the health check magic but going to ✅ so we can get test instances caught up to staging 🚀

PR #894, opened 2022-12-06
Add a "contract_amount" column to the contracts table

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10728
and
https://github.com/cityofaustin/atd-data-tech/issues/10727

Moves the work assignment ID field from summary to contracts table. 
Adds contract amount to contracts table. 

## Testing
**URL to test:** 
https://10727-move-work-assignment--atd-moped-main.netlify.app/moped/ 

**Steps to test:**
Navigate to a project's funding tab. Add / edit / delete work assignment IDs and contract amounts. 

---
#### Ship list
- [x] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -142,11 +142,9 @@
   delete_permissions:
     - role: moped-admin
       permission:
-        backend_only: false
```
chiaberry, line 145: i have no idea why these changed

chiaberry: oh maybe its because we updated the hasura version?

moped-editor/src/views/projects/projectView/ProjectContractsTable.js
```diff
@@ -90,6 +92,17 @@ const ProjectContractsTable = () => {
       title: "Description",
       field: "description",
     },
+    {
+      title: "Work Assignment ID",
```
chiaberry, line 96: not sure if this should be Work assignment ID to match the sentence casing, and also how it was on the project summary page. 

When I checked the project summary page, I noticed "eCAPRIS Subproject ID" isnt sentence case because of the S

moped-database/metadata/tables.yaml
mddilley: I think that you've got it. Seems harmless and TIL - https://hasura.io/docs/latest/auth/authorization/permission-rules/#backend-only-permissions

moped-editor/src/views/projects/projectView/ProjectContractsTable.js
mddilley: That is a good question, I'm thinking sentence case and maybe patch the eCAPRIS one? Not a super strong opinion from me but that is a good detail to catch for consistency!

moped-editor/src/views/projects/projectView/ProjectContractsTable.js
```diff
@@ -23,6 +23,8 @@ import MaterialTable, {
 import typography from "../../../theme/typography";
 
 import { PAGING_DEFAULT_COUNT } from "../../../constants/tables";
+import { currencyFormatter } from "../../../utils/numberFormatters";
+import FundingAmountIntegerField from "./FundingAmountIntegerField";
```
mddilley, line 27: Maybe this could be renamed to DollarIntegerField (or something to that effect) now that it is being used outside of funding? Either way, just glad to see this component get reused 🚀

mddilley (approved):
Tested and it works amazingly! 🙌 I bet it was fun 🪓 ing some of this code, and glad to see an existing component taking on some more work. 🚢

johnclary (changes requested):
@chiaberry thank you for knocking this out so quickly! there's a lil bug with the type handling of `contract_amount` otherwise this looks great.

also, if you don't mind, it would be nice to update our project seeds with values for these new columns. [here](https://github.com/cityofaustin/atd-moped/blob/main/moped-database/README_seed_data.md) are the docs for that task.

i will open a separate issue to delete the `work_assignmet_id` column from `moped_project` - that can wait until the next release.

moped-editor/src/views/projects/projectView/ProjectContractsTable.js
```diff
@@ -198,6 +216,9 @@ const ProjectContractsTable = () => {
 
             // Remove unneeded variable
             delete updateContractData.__typename;
+            updateContractData.contract_amount = Number(
+              newData.contract_amount
+            );
```
chiaberry, line 221: this is what the funding amount table has that the contract amount didnt -- what changes the "" to a 0 so it works in the mutation that is looking for an Int!

johnclary: ah hah. ok, so if we're defaulting the amount to $0 in the UI, i think you should also set the DM column default to `0`.

chiaberry: updated!

johnclary (commented):
🚢

johnclary (approved):

PR #910, opened 2022-12-16
Activity Log enhancements 

chiaberry:
## Associated issues

~~This is my first stab at the way I want to refactor the activity log, by splitting the different activity updates into their own components. That way it will be easier to debug (i hope) and fine tune behaviors based on which table the update is coming from.~~

* I removed all the code for the getLookups hooks, its not working in production. Also I think instead of doing a lot of work to check which lookup tables need to be queried, let's just query the lookup tables. Eventually I'd like to investigate moving them into Context app wide, since I think we look them up in a few places. 

~~While this PR is ready to review, the whole activity log is still a work in progress, I am going to chip away at all the tables and make new components for them as I update their event hooks to be the rest connector ones.~~

Thanks to feedback and help from John, I've restructured it again. Instead of each table/activity type getting its own component, there is one activity component and multiple functions to format the information that should be rendered. 

Also (thanks again to John), I dropped the "from" part of the activity log text to only say "[thing] was updated to [update]" instead of including. 

Stylistically and textwise it needs improving but this is a first pass to clean things up. 

## Testing
**URL to test:** 
https://deploy-preview-910--atd-moped-main.netlify.app/moped/projects/1905?tab=activity_log

**Steps to test:**

Check the activity log, since this is a preview app it points to staging, so there is a mix of legacy (pre REST connector) activity log events and also the new ones (moped project and moped tags). Updating tags and updating the project summary fields will add new activity entries. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -259,131 +266,109 @@ const ProjectActivityLog = () => {
...
-                                    <span className={classes.boldText}>
-                                      {getCreationLabel(
-                                        change,
-                                        userList,
-                                        phaseList
-                                      )}
-                                    </span>
-                                  </Grid>
+                        {["moped_project", "moped_proj_tags"].includes(change.record_type) ? (
+                          selectEntryComponent(change)
```
chiaberry, line 273: this is temporary, the array of table names. Once everything is updated and I remove the old way of showing events, we won't have to use this wonky thing.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -182,10 +178,24 @@ const ProjectActivityLog = () => {
...
   }
 
+  const selectEntryComponent = (change) => {
+    switch (change.record_type) {
+      case "moped_project":
+        return <ProjectActivityEntry change={change} />;
+      case "moped_proj_tags":
+        return <TagsActivityEntry change={change} tagList={tagList} />;
+      default:
+        return <div>There is no matching component</div>;
```
chiaberry, line 193: we shouldn't have this happen, this is more of a message to me as im working on the components.

moped-editor/src/views/projects/projectView/ActivityLogComponents/ProjectActivityEntry.js
```diff
@@ -0,0 +1,149 @@
...
+    );
+  } else {
+    return (
+      <Typography variant="body2" className={classes.entryText}>
+        Changed {entryMap.fields[change.description[0].field].label} from "
+        <span className={classes.strikeText}>
+          {change.description[0].old[change.description[0].field]}
+        </span>
+        " to "{change.description[0].new[change.description[0].field]}";
+      </Typography>
```
frankhereford, line 102: My inclination is to suggest disregarding this message, because design-by-committee in a PR review is a special kind of uncool, but ... 

image 
👆An example of string-to-a-string type change

Even when the strings are a much more manageable length, the use of a sentence fragment in the `Change` column coupled with the use of strikethrough text decoration on the outgoing value is confusing and at the very least, hard to read. Maybe something like this is clearer? I'm not sure and far from convinced this is best, but it's an idea..

IMG_2225

In terms of maybe an example style to borrow from, the code blocks in this page seem like they could be good textual idea containers. 

http://web.simmons.edu/~grabiner/comm244/weekfour/code-test.html

But,  if a sentence-based description of the change is the way we want to go, I think we should consider forming the sentences just as simply as we can. Like:

The [field name] was changed from "[previous value]" to "[new value]."
ex: The project description was changed from "herd cats" to "runnin' down a dream."

moped-editor/src/views/projects/projectView/ActivityLogComponents/ProjectActivityEntry.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+};
+
+const ProjectActivityEntry = ({ change }) => {
+  const classes = useStyles();
+
+  let changeData = {};
+  // the legacy way has the change data in event/data.
+  // In the new schema, event is undefined
+  const legacyVersion = !!change.record_data.event;
```
frankhereford, line 113: Do you envision the larger arc of work here including a script which will transform all the old activity log entries into the same form as the new ones so that everything ends up going through the non-legacy pipeline? 

It may be totally out of scope of the issue you're working on, but if you had an idea of if that'll be included at some point, it may influence choices you make as you structure the moving parts of this system. I'm trying to think of ways to maybe minimize the task of refactoring this code to shed the legacy rendering support at some point down the road when we no longer need it.

moped-editor/src/views/projects/projectView/ActivityLogComponents/TagsActivityEntry.js
```diff
@@ -0,0 +1,41 @@
...
+}));
+
+const getEntryText = (change, tagList) => {
+  // Adding a new tag
+  if (change.description.length === 0) {
+    return `Project tagged with "${
+      tagList[change.record_data.data.new.tag_id]
+    }" `;
+  }
+  // Soft deleting a tag is the only update a user can do (is_deleted is set to true) can do
```
frankhereford, line 21: Extra `can do` in your comment, and maybe it could read:

A user can only perform one type of update. They can soft-delete a tag which sets `is_deleted` to true.

moped-editor/src/views/projects/projectView/ActivityLogComponents/ProjectActivityEntry.js
```diff
@@ -0,0 +1,149 @@
...
+  return (
+    <Box display="flex" p={0}>
+      <Box p={0}>
+        <span class="material-symbols-outlined">summarize</span>
+      </Box>
+      <Box p={0} flexGrow={1}>
+        {getEntryText(change, legacyVersion, classes)}
+      </Box>
+    </Box>
+  );
```
frankhereford, line 146: Is the idea that you can't get here? Of if we do, we've managed to pull up a project which is no longer active?

I guess I'm not sure what the meaning of `summarize` is, but I'm not sure what I need to do to a project to reach this branch.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -182,10 +178,24 @@ const ProjectActivityLog = () => {
...
+    }
+  };
```
frankhereford, line 195: Seeing this, I'm 1,000% sold behind on your idea of how to structure this into various components. 🎯 Good call!

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -1,166 +1,4 @@
...
-        icon: "",
-        label: "updated timestamp",
-        data_type: "timestamp with time zone",
-      },
-      // deprecated column, but keeping because historical activities depend on it
-      contractor: {
-        icon: "",
-        label: "contractor",
-        data_type: "text",
-      },
```
frankhereford, line 125: 😢 I hope we can find a way to zap this historical stuff out of the codebase. I hope we don't get into a situation where we have reserved words that we can't use in the database as a column because we used it once in the past. Not the end of the world but ... here's an idea we might discuss as a team.

I think it would be neat to add a field to the activity log table named `overloaded_message`, and have it default to `null`. For any record that does get it populated, such as we might on activity log entries that concern depreciated columns and features, we could see that field being not-null in this activity log rendering system and print out whatever message is in it, just raw without formatting. This would allow us to drop support for the special names in the data fields of the activity log.

frankhereford (approved):
I think this looks great. Zapping all that bold is already a win. If you end up collecting a bunch of PRs into this, please feel free to re-request a review anytime, if that would be helpful. 

I can't wait to see what you make of this - it's going to be a huge upgrade! 🪵🔥

patrickm02L:
Can you change the calendar icon to: 
https://mui.com/material-ui/material-icons/?query=calendar&theme=Outlined&selected=CalendarTodayOutlined

Screen Shot 2022-12-21 at 10 01 39 AM

moped-editor/src/views/projects/projectView/ActivityLogComponents/TagsActivityEntry.js
chiaberry: thanks :)

chiaberry:
@patrickm02L I will use that icon when I get to the timeline tabs, thank you for linking to the right one. I haven't touched all the entries yet, so you will still some of the filled in icons here.

roseeichelmann:
When I change the project sponsor on a project & then head to the activity log and refresh the page i get a WSOD. anyone else? its also happening with the interim MPD

chiaberry:
@roseeichelmann totally a bug. thanks for catching it. ill fix it now

roseeichelmann (approved):
Looks good! That bug is fixed now. However when I add project types the activity log just says "+ Added". Idk if that is out of scope for this issue though so I'm going to approve bc everything else works as expected!

johnclary (changes requested):
@chiaberry i'm really liking this approach of isolating the change event handling for specific record types!

i want to propose a slightly different approach that i hope will help to further clean up the code.

rather than have table-specific components, i'd prefer to see table-specific data munging helpers, and do all of the data-munging up-front, inside of a hook that does something similar to what to what `getDiffs` is doing. the goal would be to have an array of change event objects, each with all of the values that we can plug directly into the render components. 

something like:

```javascript
[
    {
        date: '2022-12-18T12:10:11',
        user: '<moped_user object>',
        message: "Project tagged with 'MAP Vision Zero - Ped Safety 2023',
    },
    ...
]
```

with the data structured like that, i think we could get down to a handful of generic components, and really clean up the render code in `ProjectActivityLog`.

i'm requesting changes here, but i know you intend to keep chipping away at this code in a separate PR. what you have works great and is safe to ship IMO.

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -0,0 +1,66 @@
...
+    if (legacyVersion) {
+      changeText = `Changed ${
+        entryMap.fields[change.description[0].field].label
+      } to "${entityList[change.description[0].new]}"`;
+    } else {
+      changeText = `Changed ${
+        entryMap.fields[change.description[0].field].label
+      }
+        " to "
+        ${entityList[change.description[0].new[change.description[0].field]]}"`;
```
roseeichelmann, line 51: i think you can remove the quotes around " to " and add a quote in front of `${entityList[change.description[0].new[change.description[0].field]]}"`;`

moped-database/metadata/tables.yaml
```diff
@@ -3201,6 +3202,13 @@
         query_params: {}
         template_engine: Kriti
         version: 2
+      cleanup_config:
+        batch_size: 10000
+        clean_invocation_logs: false
+        clear_older_than: 168
+        paused: true
+        schedule: 0 0 * * *
+        timeout: 60
```
chiaberry, line 3211: hasura did this automatically

moped-database/metadata/tables.yaml
```diff
@@ -3188,7 +3189,7 @@
                   "record_type":  {{ $body.table.name }},
                   "activity_id": {{ $body.id }},
                   "record_project_id": {{ $body.event.data.new.project_id }},
-                  "record_data": {{ $body.event }},
+                  "record_data": {"event" : {{ $body.event }}},
```
chiaberry, line 3192: updating this so the new activities logged match the old (legacy) ones

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -0,0 +1,59 @@
...
+    // Changing a field, but need to use lookup table to display
+    changeDescription = `Changed ${entryMap.fields[changedField].label} to `;
+    changeValue = entityList[changeData.new[changedField]];
+  } else {
+
+    // If the update is an object, show just the field name that was updated. 
+    if (typeof changeData.new[changedField] === "object") {
+      changeDescription = `Changed ${entryMap.fields[changedField].label}`;
+      changeValue = "";
+    }
```
chiaberry, line 46: I noticed this when testing with prod data. There was a task order that was added back in June and `changeData.new[changedField]` is a jsonb object. Instead of trying to pick apart the pertinent information for a deprecated field, I'm just showing this. We can circle back and add more details later, if users are requesting it.

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -0,0 +1,59 @@
+import BeenhereOutlinedIcon from "@material-ui/icons/BeenhereOutlined";
+import { ProjectActivityLogTableMaps } from "../../views/projects/projectView/ProjectActivityLogTableMaps";
+
+export const formatProjectActivity = (change, entityList) => {
+  const entryMap = ProjectActivityLogTableMaps["moped_project"];
```
johnclary, line 5: the casing on `ProjectActivityLogTableMaps ` is throwing me off because i'm expecting this to be a class. exporting a name like `TABLE_MAPS` would be better? i see the pattern throughout this feature so nbd to leave this 🧹 for a future PR.

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -0,0 +1,59 @@
...
+  let changeDescription = "Project updated";
+  let changeIcon = <span className="material-symbols-outlined">summarize</span>;
+  let changeValue = "";
```
johnclary, line 8: since these are the "must have" props in your function return (and we're going to repeat the pattern a whole bunch of times), it might self-document better to setup the object up-front:

```javascript
const entry = { 
  changeDescription: "Project updated",
  changeIcon: <span className="material-symbols-outlined">summarize</span>,
  changeValue: "",
}
```

moped-editor/src/utils/activityLogHelpers.js
```diff
@@ -1,156 +1,19 @@
...
+    <span className="material-symbols-outlined">summarize</span>
+  );
+
+  switch (change.record_type) {
+    case "moped_project":
+      return formatProjectActivity(change, lookupData.entityList);
+    case "moped_proj_tags":
+      return formatTagsActivity(change, lookupData.tagList);
+    default:
+      return { changeIcon, changeDescription, changeValue };
```
johnclary, line 17: i guess it'd be nice to  separate the default/fallback behavior into a `defaultFormatActivity` function?

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+  boldText: {
+    fontWeight: 600,
+  }
+}));
+
+const ProjectActivityEntry = ({ changeIcon, changeDescription, changeValue }) => {
+  const classes = useStyles();
+
+  console.log(changeValue)
```
johnclary, line 16: 🧹

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <Box display="flex" p={0}>
+      <Box p={0}>{changeIcon}</Box>
+      <Box p={0} flexGrow={1}>
+        <Typography variant="body2" className={classes.entryText}>
+          {changeDescription}
+          {changeValue && changeValue.length > 0 && <span className={classes.boldText}>{changeValue}</span>}
```
johnclary, line 24: is there a case where there is a  `changeValue` but it's length is not > 0?

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -71,64 +69,50 @@ const useStyles = makeStyles((theme) => ({
...
-   * Opens the details dialog
-   * @param {string} activityId - The activity uuid
-   */
-  // const handleDetailsOpen = activityId => {
-  //   setActivityId(activityId);
-  // };
+/**
+ * Memoized object containing lookup tables
+ */
+const useLookupTables = (data) =>
```
johnclary, line 75: 🙌

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -71,64 +69,50 @@ const useStyles = makeStyles((theme) => ({
...
-   * @returns {Array}
-   */
-  const getDiffs = (eventList) => {
+/**
+ * if event is from new schema, finds the updated difference and includes in the object
+ * Makes sure the project creation event is the last in the returned Array
+ * @param {Array} eventList - The data object as provided by apollo
+ * @returns {Array}
+ */
+const usePrepareActivityData = (activityData) =>
```
johnclary, line 108: 🔥

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -71,64 +69,50 @@ const useStyles = makeStyles((theme) => ({
...
 
-  /**
-   * if event is from new schema, finds the updated difference and includes in the object
-   * Makes sure the project creation event is the last in the returned Array
-   * @param {Array} eventList - The data object as provided by apollo
-   * @returns {Array}
-   */
-  const getDiffs = (eventList) => {
+/**
+ * if event is from new schema, finds the updated difference and includes in the object
```
johnclary, line 103: ~if event is from new schema~ - ?

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -173,24 +157,44 @@ const ProjectActivityLog = () => {
...
-      phaseList[`${phase.phase_id}`] = phase.phase_name;
-    });
-  }
+  /**
+   * Returns True if the field should be a generic type (i.e., maps, objects)
+   * @param {string} field - The field name (column name)
+   * @return {boolean} - True if the field is contained in the ProjectActivityLogGenericDescriptions object
+   */
+  const isFieldGeneric = (field) =>
+    field in ProjectActivityLogGenericDescriptions;
```
johnclary, line 189: nbd but you could boot this and `getTotalItems` outside of this component

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -259,131 +265,115 @@ const ProjectActivityLog = () => {
...
               </TableBody>
             </Table>
           </TableContainer>
         )}
-        {/*        {!!activityId && (
-          <ProjectActivityLogDialog
-            activity_id={activityId}
-            handleClose={handleDetailsClose}
-          />
-        )}*/}
```
johnclary, line 386: there's a lot of great code slashing in this render! looking forward to getting all of the tables up to this new standard so we can continue to pare it down 🚀

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
johnclary: definitely an improvement 🙏

johnclary (changes requested):
@chiaberry this looks fantastic!

would you mind changing the `changValue` of `(blank)` to be `(none)` instead? i think that is slightly more clear to users?

along those lines—let me know if you'd prefer a separate issue for this—but we need to add that same special handling for project sponsor and project lead. here's what they look like if `null`.

 
Screen Shot 2023-01-03 at 6 02 54 PM


Screen Shot 2023-01-03 at 6 40 46 PM

that's it—otherwise this is ready to gooo 🚀

chiaberry:
Screen Shot 2023-01-03 at 7 56 59 PM
@johnclary What do you want it to say if the project lead or sponsor was set to null

project lead was reset?
project lead was removed?

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
chiaberry: This has been around forever, I'll change the name in a future PR

moped-editor/src/utils/activityLogHelpers.js
chiaberry: Definitely, if we want to be more detailed than everything else being the generic "Project was updated"

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -2,163 +2,71 @@ export const ProjectActivityLogTableMaps = {
...
+      project_lead_id: {
+        label: "project lead",
+        lookup: "moped_entity",
+      },
+      capitally_funded: {
+        label: "capitally funded"
+      },
+      is_retired: {
+        label: "is retired"
+      }
```
chiaberry, line 69: I was testing other projects using the prod data and found a few more deprecated fields that have activity log entries

johnclary: interesting—so this is a lingering bug that exists in prod today? do you have any thoughts on how we could reliably avoid potential WSOD if other things fall through the cracks? shall i make a separate issue?

chiaberry: It appears to be a regression I was going to introduce with this PR. Going to make sure I didnt lose any of the other fields when moving the table maps back to this file

johnclary (approved):
thanks @chiaberry!

🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
johnclary: - https://github.com/cityofaustin/atd-data-tech/issues/11105

PR #911, opened 2022-12-20
disable comment save button if empty

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10996

## Testing
**URL to test:** 

https://deploy-preview-911--atd-moped-main.netlify.app/

**Steps to test:**
- go to a project comments tab
- the 'save' button should be disabled until you've input text
- delete the text, the 'save' button should be disabled again

---
#### Ship list
- [x] Code reviewed 


tillyw:
I used a simple regex to remove the html tags from the comment input to test for content. Please let me know if there's a preferred way to do this!

moped-editor/src/views/projects/projectView/CommentInputQuill.js
```diff
@@ -68,6 +68,8 @@ const CommentInputQuill = ({
               </div>
             )}
             <ProjectSaveButton
+              // disable save button if no text after removing html tags
+              disabled={!noteText.replace(/(<([^>]+)>)/gi, "")}
```
frankhereford, line 72: This is such an edge case, but the string `<>` would get past this regex and come out the otherside still as `<>`. I also think the parentheses are unneeded since you are not capturing the matched values. I think you could go with `/<[^>]*>/g`. I read it as: "This matches the situation where there is content in between a less-than and greater-than symbol, and when that content is anything that does not include a greater-than symbol, including the null set." What do you think? That also drops the `i` flag off there since there isn't any case-sensitive text being considered.


image

frankhereford (approved):
Hey Tilly -- This works like a charm. To your question about using a regex -- I think that is a great way to handle checking the contents of the note. I do think the regex can be simplified a little - what do you think about that?

roseeichelmann (approved):
This works really nice Tilly, I think using a regex is a great solution ✨

moped-editor/src/views/projects/projectView/CommentInputQuill.js
mddilley, line 72: I took a brief look at the [Quill docs](https://github.com/zenoamaro/react-quill#methods), and, since we have a ref for the autofocus behavior already, there is the option to call the `getEditor()` method and then `getLength()` to check the editor content length. It [always returns at least 1](https://github.com/quilljs/quill/issues/1105) though so that would have to be considered. 🤷

mddilley (approved):
Works great, and I'm glad that improvements like this are making there way in! 🛥️✨ Agreed with Frank that the regex could be adjusted a bit, and I offered up another alternative too if you want to go down that road.

moped-editor/src/views/projects/projectView/CommentInputQuill.js
johnclary: interesting find @mddilley! @tillyw i defer to you if you want to make use of the `getLength()` method or use Frank's modified regex

tillyw: thanks, y'all! I'm going to go with @frankhereford's modified regex for simplicity, but @mddilley thanks for pointing out those methods of the Quill component!

PR #915, opened 2022-12-22
Re-implement project summary map

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10995
https://github.com/cityofaustin/atd-data-tech/issues/10689

This PR updates the project summary map to show features that come from the `project_geography` view. I also updated the map to use some of the same config and patterns as the components map. I removed the `ErrorBoundary` component from the summary view since it isn't a pattern that we've used elsewhere in the app. Definitely up for debate because having those can be handy, but I think that if we remove it then I should remove the package for it as well.

## Testing
**URL to test:** 

local or https://md-10995-summary-map--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Create a new project and, when you reach the summary page, notice that the fallback component with the map and scooter are showing instead of the summary map'
2. Click the "Add Components" button in the fallback and use the map to add some components with features of point and line type
3. Go back to the summary page, and you should see all of the component features that you added and the map should zoom to them just like in the components map
4. Try out the speed dial selector in the bottom right of the summary map and switch between street and aerial

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -117,6 +117,13 @@ export const SUMMARY_QUERY = gql`
       project_id
       user_id
     }
+    project_geography(
+      where: { project_id: { _eq: $projectId }, is_deleted: { _eq: false } }
+    ) {
+      geometry: geography
+      component_id
+      attributes
+    }
   }
 `;
 
--- a/moped-editor/src/queries/project.js
+++ b/moped-editor/src/queries/project.js
@@ -121,7 +121,6 @@ export const SUMMARY_QUERY = gql`
       where: { project_id: { _eq: $projectId }, is_deleted: { _eq: false } }
     ) {
       geometry: geography
-      component_id
       attributes
     }
   }
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
@@ -2,6 +2,18 @@ import { useMemo } from "react";
 import { featureTableFieldMap } from "./makeFeatures";
 
 /**
+ * Take a project_geography record and return a valid GeoJSON feature
+ * TODO: We should look into returning GeoJSON features directly from the DB
+ * @param {Object} record - project_geography record from GET_PROJECT_COMPONENTS query
+ * @returns {Object} - GeoJSON feature
+ */
+export const makeFeatureFromProjectGeographyRecord = (record) => ({
+  type: "Feature",
+  properties: { ...record.attributes },
+  geometry: record.geometry,
+});
+
+/**
  * Create an object map of component feature collections by component id
  * Ex. { 1: { type: "FeatureCollection", features: [...] } }
  * @param {Object} data - data returned by GET_PROJECT_COMPONENTS query
@@ -15,12 +27,7 @@ export const useComponentFeatureCollectionsMap = (data) =>
 
     data.project_geography.forEach((component) => {
       const currentComponentId = component.component_id;
-
-      const currentFeature = {
-        type: "Feature",
-        properties: { ...component.attributes },
-        geometry: component.geometry,
-      };
+      const currentFeature = makeFeatureFromProjectGeographyRecord(component);
 
       if (!componentGeographyMap[currentComponentId]) {
         componentGeographyMap[currentComponentId] = {
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
@@ -116,7 +116,6 @@ const ProjectSummary = ({ loading, error, data, refetch }) => {
   const { projectId } = useParams();
   const classes = useStyles();
 
-  const [mapError, setMapError] = useState(false);
   const [snackbarState, setSnackbarState] = useState(false);
 
   /**
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
@@ -8,11 +8,6 @@ import { createProjectFeatureCollection } from "src/utils/projectComponentHelper
 import { Grid, CardContent, CircularProgress } from "@material-ui/core";
 import ApolloErrorHandler from "../../../../components/ApolloErrorHandler";
 
-/*
-  Error Handler and Fallback Component
-*/
-import ProjectSummaryMapFallback from "./ProjectSummaryMapFallback";
-import { ErrorBoundary } from "react-error-boundary";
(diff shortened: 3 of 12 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryMap.js
```diff
@@ -1,107 +1,80 @@
...
-} from "../../../../utils/mapHelpers";
+  mapParameters,
+  initialViewState,
+} from "../ProjectComponents/mapSettings";
+import { makeFeatureFromProjectGeographyRecord } from "../ProjectComponents/utils/makeFeatureCollections";
+import { useZoomToExistingComponents } from "../ProjectComponents/utils/map";
+import "mapbox-gl/dist/mapbox-gl.css";
+
+/**
+ * Use a callback ref to get the map instance and store it in state so we can watch it with useEffect
```
mddilley, line 18: This solves a bug that I was facing with nothing in this component triggering the `useEffect` in `useZoomToExistingComponents`. In a nutshell, this callback gives us both the normal React ref behavior and also the ref value in state so that we can watch it for updates.

It revolves around how React refs hold the same value throughout a component's lifecycle but they do not trigger a re-render when that value is updated. Here is [a thread](https://twitter.com/teh_builder/status/1093497029693763584) that I always find my way back to on this. This type of stuff is a head-scratcher for me every time.

Also, if you are wondering why it works in the component map, it is because there is an async call for `data` there so that is what triggers the zoom to features. `data` has already been fetched here and is passed as a prop so it doesn't change/trigger `useEffect`.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryMap.js
```diff
@@ -1,107 +1,80 @@
...
-        ref={mapRef}
-        maxZoom={20}
-        style={{ width: "100%", height: "60vh" }}
-        /* Access Key */
-        mapboxAccessToken={MAPBOX_TOKEN}
-        onRender={fitMapToFeatureCollectionOnRender}
-        /* Get the IDs from the layerConfigs object to set as interactive in the summary map */
-        /* If specified: Pointer event callbacks will only query the features under the pointer of these layers.
-              The getCursor callback will receive isHovering: true when hover over features of these layers */
-        interactiveLayerIds={getSummaryMapInteractiveIds(
```
mddilley, line 85: We don't have any hover behavior scoped in this map or the component map yet so I'm removing the hover and interactive layer IDs needed to make layers hover-able.

roseeichelmann (approved):
This works great! That aerial map is so cool to see. One thing I noticed was that when I spaced the components farther apart the summary map auto zoom looked a little far away. But the auto zoom when you click on the map looks good. Definitely not something that's blocking this PR for me though!

Screenshot from 2022-12-27 12-15-22
Screenshot from 2022-12-27 12-15-34

moped-editor/src/queries/project.js
```diff
@@ -117,6 +117,12 @@ export const SUMMARY_QUERY = gql`
       project_id
       user_id
     }
+    project_geography(
+      where: { project_id: { _eq: $projectId }, is_deleted: { _eq: false } }
+    ) {
+      geometry: geography
+      attributes
+    }
```
johnclary, line 125: 🚀

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
```diff
@@ -1,6 +1,18 @@
...
+ * Take a project_geography record and return a valid GeoJSON feature
+ * TODO: We should look into returning GeoJSON features directly from the DB
+ * @param {Object} record - project_geography record from GET_PROJECT_COMPONENTS query
+ * @returns {Object} - GeoJSON feature
+ */
+export const makeFeatureFromProjectGeographyRecord = (record) => ({
+  type: "Feature",
+  properties: { ...record.attributes },
+  geometry: record.geometry,
+});
```
johnclary, line 14: 👍

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummary.js
```diff
@@ -3,16 +3,10 @@ import { useParams } from "react-router-dom";
...
-import { createProjectFeatureCollection } from "src/utils/projectComponentHelpers";
 
 import { Grid, CardContent, CircularProgress } from "@material-ui/core";
 import ApolloErrorHandler from "../../../../components/ApolloErrorHandler";
 
-/*
-  Error Handler and Fallback Component
-*/
-import ProjectSummaryMapFallback from "./ProjectSummaryMapFallback";
-import { ErrorBoundary } from "react-error-boundary";
```
johnclary, line 15: yep i agree and am happy to see this dependency go away

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryMap.js
johnclary: very cool! so glad to see this resolved!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryMap.js
```diff
@@ -1,107 +1,81 @@
...
+            setBasemapKey={setBasemapKey}
+          />
+          <BaseMapSourceAndLayers basemapKey={basemapKey} />
+          <ProjectSummaryMapSourcesAndLayers
+            projectFeatureCollection={projectFeatureCollection}
+          />
+        </MapGL>
+      ) : (
+        <ProjectSummaryMapFallback />
+      )}
```
johnclary, line 78: ✨

johnclary (approved):
woohoo—it's so fun to see this map come back to life!

i agree with @roseeichelmann that the map zoom doesn't feel quite right when features are spread out — i believe that's the effect of the `fitBoundsOptions.padding` being too large for the small viewport of the summary map. i'll make a separate issue for this since it looks like a non-trivial fix. and in any case it will be good to see how the summary map looks with more real-life projects before making a decision on how best to refine that zoom/padding.

🙌  🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

johnclary:
oh btw @mddilley -  if you want to grab an extra story point, we have https://github.com/cityofaustin/atd-data-tech/issues/10689 floating out there to add "cooperative gestures" to this map. i think that's as easy as a `MapGL` prop addition?

mddilley:
@johnclary Thanks for the review - I'll fully remove the error boundary dependency in this PR and squeeze https://github.com/cityofaustin/atd-data-tech/issues/10689 in here too since it is a one-liner. 👍

johnclary:
🚀 🚀 🚀 🚀 🚀 🚀

PR #917, opened 2022-12-27
Drop partner name moped proj partners 10607

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10607

## Testing
**URL to test:** 

https://drop_partner_name_moped_proj_partners_10607--atd-moped-main.netlify.app/moped/

**Steps to test:**

go to a project, add or edit the project partners and save. the ui should reflect the changes.

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -920,10 +920,8 @@ export const ProjectActivityLogCreateDescriptions = {
   },
   moped_proj_partners: {
     label: (record, userList) => {
-      return (
-        fieldFormat(record.record_data.event.data.new.partner_name, true) +
```
tillyw, line 924: i wanted to change this to display ``proj_partner_id`` instead of ``partner_name`` in the activity log but there is no way locally to test whether that would work, so we went with this safe option that should not break anything. i don't know that this is relevant because it will probably be overhauled with the activity log overhaul, but i just wanted to point it out! this was a learning experience for me better understanding how the activity log works. @chiaberry

johnclary: @chiaberry what do you think about this? since we don't have a way to test, it might be best to wait to merge until we migrate this table to a REST connector?

roseeichelmann (approved):
Works like a charm ✨

mddilley (approved):
Yayyy! More normalization 🎉 Good to go pending the resolution of the convo around the partner name in the activity log. 🚀

frankhereford (approved):
perfect

Thanks Tilly! 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
chiaberry: I think we can merge this now, if we are okay with just seeing the generic "New partner added" in the meantime

johnclary:
thanks @chiaberry let's hold on this until we release v1.14. @tillyw looks like we now have some merge conflicts that need your attention 👀

chiaberry:
Now that the moped_proj_partners are in the new activity log, this can be revisited to fix merge conflicts and then is it good to go?

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
tillyw: i went ahead and deleted this whole entry since i don't think this is necessary anymore

tillyw:
> Now that the moped_proj_partners are in the new activity log, this can be revisited to fix merge conflicts and then is it good to go?

i fixed the merge conflicts and deleted the `moped_proj_partners` entry from `ProjectActivityLogTableMaps`. i think this is good to merge! @chiaberry will wait for you to confirm!

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -250,6 +250,7 @@ export const ProjectActivityLogTableMaps = {
         label: "partner ID",
         type: "int4",
       },
+      // deprecated column, but keeping because historical activities depend on it
```
tillyw, line 253: can we actually go ahead and get rid of this too now? this is the only instance of `partner_name` left in the repo (aside from migrations)

moped-toolbox/knack_migration/1_prepare_projects.py
```diff
@@ -402,7 +402,6 @@ def set_sponsor_and_partners(projects, entities_index):
         if partner:
             entity = entities_index[partner]
             new_partner = {}
-            new_partner["partner_name"] = entity["entity_name"]
```
tillyw, line 405: figured this could go too - @chiaberry

moped-toolbox/random_project_seeds/index.js
```diff
@@ -74,7 +74,6 @@ const randomProjContract = () => ({
 });
 
 const randomProjPartner = () => ({
-  partner_name: randomString(),
```
tillyw, line 77: and this!

moped-toolbox/knack_migration/1_prepare_projects.py
chiaberry: will we ever use these toolbox scripts again to import projects from knack @johnclary? If we might someday, then yes, move this. But if we won't, then I guess its more like a historical record? and then it doesnt matter to me if it stays or goes.

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
chiaberry: We should leave this for the time being, because previous activity log entries may have a reference to this field.

moped-toolbox/knack_migration/1_prepare_projects.py
johnclary: i very much hope we never need to run these scripts again, so let's leave those old toolbox script frozen in time. i'll pull this commit outta here so we can get this merged.

moped-toolbox/random_project_seeds/index.js
johnclary: thanks for catching this one!

PR #922, opened 2023-01-02
Signal component autocomplete

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/10511

## Testing
**URL to test:** 

local or https://md-10511-signal-autocomplete--atd-moped-main.netlify.app/

**Steps to test:**

**Create a signal project**
1. Create a new project by switching on the **Use signal asset** toggle and then choosing a signal
2. When the **Project Summary** view loads after the redirect, you should see the signal on the project summary map
3. Go to the **Map** tab and you should see the signal on the map and one list item on the left with the details of the signal

**Add a signal component to an existing project**
1. Add the following types of components to the project that you just created and check the form and UI flow for the corresponding bullet points:
   - Signal - PHB
      - Should show the signal autocomplete dropdown
      - **New component** modal button should say "Save"
      - After the form is complete and the button is clicked, the map should update with a new list item and a point on the map to show where the signal is
 
   - Signal - Traffic
      - Should show the signal autocomplete dropdown
      - **New component** modal button should say "Save"
      - After the form is complete and the button is clicked, the map should update with a new list item and a point on the map to show where the signal is
   - Signal - RRFB
      - Should **NOT** show the signal autocomplete dropdown
      - **New component** modal button should say "Continue"
      - After the form is complete and the button is clicked, the map UI should appear and you can select from the CTN intersection layer and/or draw points and then save the component
   - Signal - School Zone Beacon
      - Should **NOT** show the signal autocomplete dropdown
      - **New component** modal button should say "Continue"
      - After the form is complete and the button is clicked, the map UI should appear and you can select from the CTN intersection layer and/or draw points and then save the component

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/utils/signalComponentHelpers.js
```diff
@@ -19,21 +19,17 @@ export const getSignalOptionSelected = (option, value) =>
...
  */
-export const getSignalOptionLabel = option =>
+export const getSignalOptionLabel = (option) =>
   // this label formatting mirrors the Data Tracker formatting
   `${option.properties.signal_id}: ${option.properties.location_name}`;
 
 /**
  * Imitate a "drawn point" feature from a traffic signal geojson feature. Sets required
- * fields so that featureCollection can be used in the DB mutation on submit
+ * fields so that a Knack feature can be inserted into the feature_signals table
```
mddilley, line 28: Lots of these changes are making the signals records that we fetch from the Socrata dataset (and originate from Data Tracker) into the format expected by the new `feature_signals` table instead storing the signal as a GeoJSON feature collection like we currently do. The existing helpers were super handy and were straightforward to update.

moped-editor/src/utils/signalComponentHelpers.js
```diff
@@ -42,72 +38,25 @@ export const signalToFeatureCollection = signal => {
...
-      geometry: signal.geometry,
-      id: featureUUID,
+      knack_id: signal.properties.id,
+      location_name: signal.properties.location_name,
+      signal_type: signal.properties.signal_type,
+      signal_id: signal.properties.signal_id,
+      // I was considering changing this to be INTERSECTIONID, but since the sourceLayer is being saved
+      // as "drawnByUser" and not "ATD_ADMIN.CTN_Intersections", I've left it as PROJECT_EXTENT_ID
+      project_extent_id: featureUUID,
+      source_layer: "drawnByUser",
```
mddilley, line 55: I left these comments so that we can rethink the source layer name for the signal features. 

It doesn't seem as important as it once was since we can now get the name of the table that a feature is stored in from the component record itself.

moped-editor/src/utils/signalComponentHelpers.js
```diff
@@ -42,72 +38,25 @@ export const signalToFeatureCollection = signal => {
...
     };
-    featureCollection.features.push(feature);
-  }
-  return featureCollection;
-};
-
-export const useInitialSignalComponentValue = (
```
mddilley, line 63: React Hook Form takes over the state management for the signal autocomplete now so this helper is no longer needed.

moped-editor/src/utils/signalComponentHelpers.js
```diff
@@ -42,72 +38,25 @@ export const signalToFeatureCollection = signal => {
...
-    setSignal(editFeatureCollection.features[0]);
-    // only fire on init
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
-};
 
-/*
-/ Hook which updates the component editor state based on the selected signal
-*/
-export const useSignalChangeEffect = (
```
mddilley, line 98: Same, React Hook Form takes care of this now.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -141,7 +167,7 @@ const ComponentForm = ({
             startIcon={<CheckCircle />}
             type="submit"
           >
-            {formButtonText}
+            {isSignalComponent ? "Save Signal Component" : formButtonText}
```
mddilley, line 170: I don't feel strongly about the "Save Signal Component" text here. Maybe there is a better way to put this? I want it to communicate to users that they will be saving a signal component immediately when they click this button. 

Normally, the button says "Continue" because the next step is to define the geometry of the component with the map UI. We don't have to do that since there is already a geometry associated with the choice in the signal autocomplete dropdown.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -85,6 +89,9 @@ const ComponentForm = ({
     subcomponentOptions
   );
 
+  const isEditingExistingComponent = initialFormValues !== null;
+  const isSignalComponent = internalTable === "feature_signals";
```
mddilley, line 93: Another payoff from being able to inspect the table that a component should insert features into! 🙌

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -43,6 +43,15 @@ export default function ComponentListItem({
...
     editDispatch({ type: "start_edit", payload: component });
 
+  const isSignalComponent =
+    component?.moped_components?.feature_layer?.internal_table ===
+    "feature_signals";
+  const componentName = component?.moped_components?.component_name;
+  const signalLocationName = component?.feature_signals?.[0]?.location_name;
+  const listItemPrimaryText = isSignalComponent
+    ? `${componentName} -${signalLocationName}`
+    : componentName;
```
mddilley, line 53: Thought about wrapping `ComponentListItem` with another component like `SignalComponentListItem` and moving this logic there, but just decided to ship this. If anyone feels strongly about refactoring, I'll update for sure. 🙏

moped-editor/src/utils/signalComponentHelpers.js
johnclary: i think we should drop these two columns unless you have a need for them in your code @mddilley. or i can earmark it as something to do after we run the transpose geometry script for the last time.

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
```diff
@@ -1,69 +1,62 @@
...
-  classes,
-  setSelectedComponentSubtype,
-  setEditFeatureCollection,
-  editFeatureCollection,
-}) => {
-  const [signal, setSignal] = useState(null);
-  const { features, loading, error } = useSocrataGeojson(SOCRATA_ENDPOINT);
+const SignalComponentAutocomplete = React.forwardRef(
+  ({ classes, autocompleteProps, onChange, value }, ref) => {
+    const { features, loading, error } = useSocrataGeojson(SOCRATA_ENDPOINT);
```
johnclary, line 22: Because we've done away with the "Subtype" field that used to exist on this form, it's actually possible to select the "Signal - PHB" component but choose a traffic signal (and vice versa). That's an oversight on my part and not a big deal—but i will make a backlog issue to filter the autocomplete options by signal type.

Screen Shot 2023-01-04 at 2 13 03 PM

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
johnclary: thanks for catching this! let's go with plain ole' "Save" since that's the text we show to users after they're done mapping.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
johnclary: AMD will be very happy added the location name here - nice! i think this is fine for now. we're going to have to revisit the list item styles as we add additional component properties, e.g. [component phasing](https://github.com/cityofaustin/atd-data-tech/issues/10680).

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -86,15 +95,17 @@ export default function ComponentListItem({
...
+                  <Button
+                    fullWidth
+                    size="small"
+                    color="primary"
+                    startIcon={<EditOutlined />}
+                    onClick={onStartEditingComponent}
+                  >
+                    Edit
+                  </Button>
+                )
```
johnclary, line 108: - https://github.com/cityofaustin/atd-data-tech/issues/11114

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
```diff
@@ -1,69 +1,62 @@
...
-const SignalComponentAutocomplete = ({
-  classes,
-  setSelectedComponentSubtype,
-  setEditFeatureCollection,
-  editFeatureCollection,
-}) => {
-  const [signal, setSignal] = useState(null);
-  const { features, loading, error } = useSocrataGeojson(SOCRATA_ENDPOINT);
+const SignalComponentAutocomplete = React.forwardRef(
+  ({ classes, autocompleteProps, onChange, value }, ref) => {
```
johnclary, line 21: i happened to be eyeballing this in my editor and noticed you're not using `autocompleteProps` so it can be removed 🔍

johnclary (approved):
dang, @mddilley, this is killer! it is such a relief to see this data flowing into a normalized table instead of the jsonbleh. my feedback is minor and this is 🚢 🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
johnclary: - https://github.com/cityofaustin/atd-data-tech/issues/11115

moped-editor/src/utils/signalComponentHelpers.js
mddilley: 😎 Thanks, sounds great to me - I'll remove them in this PR and go through the test steps to make sure it doesn't break anything. I think that it will be good to take care of it now.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
mddilley: Perfect - I'll update that text to keep it simple. 🙏

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
mddilley: Awesome - I initially filtered this way but then was confused about the lack of school zone beacon and rrfd types. Forgot to follow up on that when we synced on what should/should not use the signal dropdown. Long way to say that it should be a small effort to put that back in. 👍

moped-editor/src/views/projects/projectView/SignalComponentAutocomplete.js
mddilley: Just removed this 🙏 🕵️

mddilley:
Just made some changes based on John's feedback above, the diff doubled in size but, fear not, it is mostly a couple migrations to update the `uniform_features` view and remove the unused `project_extent_id` and `source_layer` columns from the `feature_signals` table

roseeichelmann (approved):
Tested all the steps and it works like a charm, more great work from you Mike 🚀🚀🚀

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 !

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -141,7 +166,7 @@ const ComponentForm = ({
             startIcon={<CheckCircle />}
             type="submit"
           >
-            {formButtonText}
+            {isSignalComponent ? "Save" : formButtonText}
```
frankhereford, line 169: In the first two sets of bullet points, I think the intent is to have this button's content text say `Save Signal Component`, but I'm not sure that it's getting set. I do see that it's set to say "Continue" on line 91 of `src/views/projects/projectView/ProjectComponents/CreateComponentModal.js`. Here is what I see in this case:

image

frankhereford (approved):
signal

This is yet another fantastic PR Mike, really great. I observed two ultra-minor things, one of which might be worthy of a low priority spin off issue, and other may be a prop that needs to be defined. Neither should even come close to holding this back - this is so great. Thank you!

moped-editor/src/views/projects/projectView/ProjectComponents/CreateComponentModal.js
```diff
@@ -51,9 +58,20 @@ const CreateComponentModal = ({ showDialog, setLinkMode, createDispatch }) => {
...
-    createDispatch({ type: "close_create_dialog" });
+    // Signal components get their geometry from the Knack signal dataset so we save them
+    // immediately. All other components are saved after the user selects or draws their geometry.
+    if (isSavingSignalFeature) {
+      const newComponentWithSignalFeature = {
+        ...newComponent,
+        features: [formData.signal],
+      };
+
+      onSaveDraftSignalComponent(newComponentWithSignalFeature);
```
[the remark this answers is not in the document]
mddilley: Thanks for the screencap, Frank. Perfectly captures how people could get lost irt the change they just made. I think this is worth a backlog issue for sure. What do you think - @johnclary?

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
mddilley: This is actually supposed to say "Save". I updated from some of John's feedback, and I forgot to update the test instructions. Gotta get better at sweeping through test instructions after updates like this. 😅 Thanks for looking out for this! 🙏

PR #926, opened 2023-01-03
relates project notes to current phase and displays badge

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10679
fixes https://github.com/cityofaustin/atd-data-tech/issues/11082

this pr stores the current phase of a project in the `moped_proj_notes` and table and displays it next to the associated status update or project comment. next pr will allow users to edit the associated phase!

## Testing
**URL to test:** 

[10679_relate_project_notes_to_current_phase--atd-moped-main.netlify.app/moped/](https://10679_relate_project_notes_to_current_phase--atd-moped-main.netlify.app/moped/)

**Steps to test:**
- create a new project
- add a status update named 'potential' (for easier tracking during testing)
- the new status update should appear with the 'potential' badge next to it (the modal will close automatically)
- add a comment with the same name
- the new comment should appear with the 'potential' badge next to it
- change the current phase under the 'timeline' tab
- add a new status update with the name of that phase
- check to make sure the badge reflects the update
- repeat the last pairs of steps with a comment
- delete a comment and/or status update, the appropriate ones should be deleted and the rest should persist

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
+++ b/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
@@ -1,3 +1,4 @@
+-- create foreign key relationship between moped_proj_notes and moped_phases on phase_id
 alter table "public"."moped_proj_notes"
   add constraint "moped_proj_notes_phase_id_fkey"
   foreign key ("phase_id")
--- a/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
+++ b/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
@@ -3,4 +3,4 @@ alter table "public"."moped_proj_notes"
   add constraint "moped_proj_notes_phase_id_fkey"
   foreign key ("phase_id")
   references "public"."moped_phases"
-  ("phase_id") on update cascade;
+  ("phase_id") on update cascade on delete set null;
--- a/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
+++ b/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
@@ -1,0 +1,5 @@
+alter table "public"."moped_proj_notes"
+  add constraint "moped_proj_notes_phase_id_fkey"
+  foreign key ("phase_id")
+  references "public"."moped_phases"
+  ("phase_id") on update cascade on delete cascade;
--- a/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
+++ b/moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
@@ -3,4 +3,4 @@ alter table "public"."moped_proj_notes"
   add constraint "moped_proj_notes_phase_id_fkey"
   foreign key ("phase_id")
   references "public"."moped_phases"
-  ("phase_id") on update cascade on delete cascade;
+  ("phase_id") on update cascade;
--- a/moped-editor/src/queries/comments.js
+++ b/moped-editor/src/queries/comments.js
@@ -14,6 +14,9 @@ export const COMMENTS_QUERY = gql`
       project_note_id
       project_note_type
       is_deleted
+      moped_phases {
+        phase_name
+      }
     }
   }
 `;
--- a/moped-editor/src/queries/comments.js
+++ b/moped-editor/src/queries/comments.js
@@ -15,6 +15,7 @@ export const COMMENTS_QUERY = gql`
       project_note_type
       is_deleted
       moped_phases {
+        phase_key
         phase_name
       }
     }
--- a/moped-editor/src/queries/comments.js
+++ b/moped-editor/src/queries/comments.js
@@ -14,7 +14,7 @@ export const COMMENTS_QUERY = gql`
       project_note_id
       project_note_type
       is_deleted
-      moped_phases {
+      moped_phase {
         phase_key
         phase_name
       }
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -370,8 +370,8 @@ const ProjectComments = (props) => {
                                   component={"span"}
                                 >
                                   <ProjectStatusBadge
-                                    phaseKey={mopedProjNotes[i].moped_phases.phase_key}
-                                    phaseName={mopedProjNotes[i].moped_phases.phase_name}
+                                    phaseKey={mopedProjNotes[i]?.moped_phases.phase_key}
+                                    phaseName={mopedProjNotes[i]?.moped_phases.phase_name}
                                     comment
                                   />
                                 </Typography>
--- a/moped-editor/src/views/projects/projectView/ProjectComments.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComments.js
@@ -123,6 +123,7 @@ const ProjectComments = (props) => {
   });
 
   const mopedProjNotes = data?.moped_proj_notes;
+  console.log(mopedProjNotes);
 
   const [addNewComment] = useMutation(ADD_PROJECT_COMMENT, {
     onCompleted() {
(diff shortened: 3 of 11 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryStatusUpdate.js
```diff
@@ -36,6 +36,7 @@ const ProjectSummaryStatusUpdate = ({ projectId, data, refetch, classes }) => {
         <DashboardStatusModal
           projectId={projectId}
           projectName={projectName}
+          data={data}
```
johnclary, line 39: since the only property you care about in `data` is the current phase ID, it would be better to require that prop explicitly, eg `currentPhaseId` instead of the generic `data` blob.

also, the `DasboardView` also uses this component and needs to be updated in the same way.

johnclary (changes requested):
it's looking good! let me know if you have any questions about this feedback 👇

tillyw:
thanks, @johnclary! will address these comments now, removing review requests from other others in the meantime.

tillyw:
@johnclary updated to reflect your feedback and fix a bug! rerequesting review from everyone here. thanks!

frankhereford:
This is squarely outside of the scope of this, but making a note of something I've observed while testing:

Screenshot 2023-01-04 at 5 22 02 PM

Screenshot 2023-01-04 at 5 22 14 PM

@johnclary An issue for consideration - should we unify the color of the font in the Quill WYSIWYG editor between when one is creating a new comment vs. when one is editing one?

moped-editor/src/views/projects/projectView/ProjectComments.js
```diff
@@ -91,6 +92,11 @@ const projectNoteTypes = ["", "Internal Note", "Status Update"];
 
 const ProjectComments = (props) => {
   const isStatusEditModal = props.modal;
+  // use currentPhaseId if passed down from ProjectSummaryStatusUpdate component,
+  // otherwise use data passed from ProjectView
+  const currentPhaseId =
+    props.currentPhaseId ??
+    props.data?.moped_project[0]?.moped_proj_phases[0]?.moped_phase.phase_id;
```
johnclary, line 99: ahh—i see now that `props.data` is being passed through the `ProjectView` component and can't easily be avoided. thanks for adding that `currentPhaseId` for `ProjectSummaryStatusUpdate` and adding the comment!

johnclary (approved):
@tillyw thanks for making such quick work of this. i'm really excited about this feature! 🚢 🚢 🚢

johnclary:
> should we unify the color of the font in the Quill WYSIWYG editor between when a user is creating a new comment vs. when a user is editing a comment?

@frankhereford thanks for flagging that — will file a backlog issue 🙏

moped-editor/src/views/projects/projectView/ProjectComments.js
tillyw: yeah, i went back and forth on this multiple times - this was how i originally did it before i submitted my initial pr. thanks for helping me pick an approach!!

moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
```diff
@@ -0,0 +1,6 @@
+-- create foreign key relationship between moped_proj_notes and moped_phases on phase_id
+alter table "public"."moped_proj_notes"
+  add constraint "moped_proj_notes_phase_id_fkey"
+  foreign key ("phase_id")
+  references "public"."moped_phases"
+  ("phase_id") on update cascade on delete cascade;
```
frankhereford, line 6: I'm wondering about the `on update cascade` and `on delete cascade` usage here, and I am not sure that want that behavior. 

I believe, as written, this says:

> When the `phase_id` of a `moped_phases` record is changed, go into `moped_proj_notes` and find all the records that point to the old ID and update them to point to the new one. 

> When a record in `moped_phases` is deleted, go into `moped_proj_notes` and delete any records that point to the being-deleted `moped_phases` record.

✅ The first behavior, the `on update` one, seems pretty great to me. I can't imagine a world when we'd be changing `phase_id`s, but it's very technically correct. 

🤔 The second behavior gives me more pause. I think we'd want the query to fail if we tried to remove a referred-to phase so we could be very intentional in how we want to handle those soon-to-be orphaned `moped_proj_notes` records. I'm pretty sure we wouldn't want to just delete them.

See next comment, for some more details, outside of this general observation.

moped-editor/src/queries/comments.js
```diff
@@ -14,6 +14,10 @@ export const COMMENTS_QUERY = gql`
       project_note_id
       project_note_type
       is_deleted
+      moped_phase {
+        phase_key
+        phase_name
+      }
```
mddilley, line 20: 🥳

moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
frankhereford: In practice, right now, today, and probably not forever, we're safe-ish, because we have /another/ constraint named `moped_proj_phases_moped_phases_phase_id_fk` on the `moped_proj_phases` table which maintains the relationship between the value in the `phase_id` column in `moped_proj_phases` and the FK `moped_phases(phase_id)`. 

This constraint does not have `on update` or `on delete` behavior defined, so it's this constraint that prohibits the deletion of a referred to `moped_phase` record which would cascade on to the `moped_proj_notes` table as directed by the constraint in this PR.

frankhereford: My recommendation is to remove the `on delete cascade` directive so we don't cascade into deleting users comments because we remove a phase without us first hand-crafting a resolution.

I think a low priority spin-off issue would be helpful to review these constraints and make sure we've set in directives that align to what we want to delete or not and when and also to ensure that we don't have conflicting or interacting constraints that prevent each other from doing what we told them to do. The constraint in this PR enters us into that state now but ...

The whole thing is largely moot because we almost universally soft-delete records and in this case, the records that have updates and deletes cascade from them are records that are "part of the app" and not user entered data, and as a result, are very stable. 

A final thought - this is very hard for me to get all loaded into my mind and to keep straight. It's complicated! Please forgive me for what I get wrong, but FWIW, these observations are borne out of testing in my local DB.

moped-editor/src/views/projects/projectView/ProjectComments.js
mddilley: +1 the comments and variable names make this super readable 🙏

mddilley (approved):
Test steps worked great - this is such a cool feature that really levels up comments with the context of what phase it was at the time! 🚀 🚀 🚀

frankhereford (approved):
Edit: 1st, I have to agree that this is a killer feature. It's so excellent in how well it conveys the state of the project when a comment was made, and it's just wonderfully clear and digestible. The design and implementation are top shelf. 

I tested this, including making comments when there is not an active phase, and it did just what I wanted it to. 💯

I tried really hard to write a concise explanation of my thoughts on the `on update` and `on delete` directives in the FK relationship. 😣 Sorry y'all -  It ended up kinda long, but we have a bit of a tangle here. 

It's a bit of a tangle that we should work out, but I don't think it needs to block this from shipping, except for I do think we should consider the `on delete cascade` again just in case we forget about circling back on it. My reasoning is below, and even as is, by happenstance, this PR does not put us in a danger zone, only danger-zone adjacent.

moped-database/migrations/1672851579949_set_fk_public_moped_proj_notes_phase_id/up.sql
tillyw: thanks, @frankhereford ! i fully agree and will push an update that remove `on delete cascade` from the migration. for context, i was copying the behavior of the other fkey relationship in this table, `project_id`. that may be the desired behavior in that case, but i wanted to point that out in case we want to reevaluate that.

Screen Shot 2023-01-04 at 7 51 26 PM

tillyw:
thanks y'all for the quick review! i'll leave it up to @johnclary's discretion as to whether to merge this tomorrow :)

johnclary:
thanks @frankhereford for catching that and @tillyw for patching it right quick! totally agree we do not want to cascade delete from lookup tables. i've been opting for `set null` on the delete rules for optional foreign keys like this one, so i went ahead and patched that in.

i agree it'd be good to go through all our fk constraints and make sure we're being consistent. we've also got [#10525](https://github.com/cityofaustin/atd-data-tech/issues/10525) which would add another layer of protection by making sure users cannot edit lookup tables (like `moped_phases`) through the API.

PR #932, opened 2023-01-05
Disable "Continue" button when the create component form isn't complete

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11124

This PR disables the **Continue** (or **Save** when creating a signal component) button in the modals that appears when creating a new component or when editing an existing component's attributes.

Button is disabled when:
- form does not have required fields complete
   - for non-signal components: the component type field is required
   - for signal components: the component type and signal fields are required
- the edit attributes modal is open but no edit has been made yet

## Testing
**URL to test:** 

local for now

**Steps to test:**
2. Click **New Component** and notice that the **Continue** button in the modal is greyed out and you can't click it
3. Add a Description and notice that the button is still greyed out
4. Choose a non-signal **Component Type** and the **Continue** button will turn blue and active
5. Finish creating the component and save it
6. Now create another component but this time select **Signal - PHB** from the type dropdown
7. Notice that the **Save** button is greyed out
8. Select a signal from the **Signal** dropdown and see the **Save** button turn blue and active and then click it
9. Last, edit the attributes of the non-signal component that you created and notice that the **Save** button is greyed out
10. Update the description and see that the button turns blue and active
11. Save the update

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -3,7 +3,7 @@ import { useQuery } from "@apollo/client";
 import { Controller, useForm } from "react-hook-form";
 import { Button, Grid, TextField } from "@material-ui/core";
 import { CheckCircle } from "@material-ui/icons";
-import { Autocomplete } from "@material-ui/lab";
+import { ControlledAutocomplete } from "./utils/form";
 import { GET_COMPONENTS_FORM_OPTIONS } from "src/queries/components";
 import SignalComponentAutocomplete from "../SignalComponentAutocomplete";
 import {
@@ -20,46 +20,6 @@ const defaultFormValues = {
   signal: null,
 };
 
-const ControlledAutocomplete = ({
-  id,
-  options,
-  renderOption,
-  name,
-  control,
-  label,
-  autoFocus = false,
-  multiple = false,
-  disabled,
-}) => (
-  <Controller
-    id={id}
-    name={name}
-    control={control}
-    render={({ onChange, value, ref }) => (
-      <Autocomplete
-        options={options}
-        multiple={multiple}
-        getOptionLabel={(option) => option?.label || ""}
-        getOptionSelected={(option, value) => option?.value === value?.value}
-        renderOption={renderOption}
-        value={value}
-        disabled={disabled}
-        renderInput={(params) => (
-          <TextField
-            {...params}
-            inputRef={ref}
-            size="small"
-            label={label}
-            variant="outlined"
-            autoFocus={autoFocus}
-          />
-        )}
-        onChange={(_event, option) => onChange(option)}
-      />
-    )}
-  />
-);
-
 const ComponentForm = ({
   formButtonText,
   onSave,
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -20,45 +22,16 @@ const defaultFormValues = {
   signal: null,
 };
 
-const ControlledAutocomplete = ({
```
mddilley, line 23: Moved this to the form helper file

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -68,8 +41,17 @@ const ComponentForm = ({
...
+  const {
+    register,
+    handleSubmit,
+    control,
+    watch,
+    setValue,
+    formState: { isValid },
+  } = useForm({
     defaultValues: defaultFormValues,
+    mode: "onChange",
```
mddilley, line 53: This mode causes validation with the yup schema each time a form value changes. There is a warning in the [docs](https://react-hook-form.com/api/useform) that I quoted below, but we can take another look if/when this forms get larger or more complex. It is key in keeping the button disabled until valid inputs are entered for now.

> Validation will trigger on the change event with each input, and lead to multiple re-renders. Warning: this often comes with a significant impact on performance.

johnclary: TIL `onChange` mode.  Thanks for the heads-up. I did not notice any sluggishness when testing our humble form.

johnclary (approved):
@mddilley this looks great and works great!

one interesting observation: when editing, you cannot save w/o editing the form in some way (by changing subcomponents are description).  

i guess this is happening because isValid is initializing as `false` until some form interaction fires the `onChange` validation? in any case, this is totally fine, and has the same effect of making use of RFH's `isDirty` [formState](https://react-hook-form.com/api/useform/formstate).

Screen Shot 2023-01-10 at 11 26 13 AM

🚢 🚢 🚢

johnclary:
one last comment, I probably should have scoped in some visual error validation, like we have on the staff form. nbd—i will throw that in a separate backlog issue.

Screen Shot 2023-01-10 at 11 40 48 AM

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -12,6 +13,7 @@ import {
   useSubcomponentOptions,
   useInitialValuesOnAttributesEdit,
 } from "./utils/form";
+import * as yup from "yup";
```
frankhereford, line 16: TIL `yup`. Reminds me of `zod` used for typescript aware validation. Another one is `joi`, I think. Who is naming these libraries? They are a lifesaver though!

frankhereford (approved):
Works great - this is sharp. I agree with John that seeing the validation done with a validation library - what more could one ask. Thanks for laying down more great patterns!

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
mddilley: So true! Like there is a rule that they must be 3 letters long. `zod` sounds so cool and also reminds that I should rewatch the old Superman movies soon. 😄 
image

frankhereford: This is the funniest comment I have read all week. So good!

PR #933, opened 2023-01-09
Edit signal components

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11114

This updates the edit component attributes form to be able to update the signal, subcomponents, and description of an existing signal component.

## Testing
**URL to test:** 

local or https://md-11114-edit-signal-component--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Create a new component and select either **Signal - PHB** for **Signal - Traffic**
2. Select a signal, add one or more subcomponents, and add a description
3. Click **Save**
4. Click your new component's list item on the left and then click **Edit**
5. When the edit modal appears, click **Attributes**
6. Update the signal, subcomponents, and description
7. Try updating each of those individually
8. Try updating different combinations
9. You should see your component update each time

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/EditModeDialog.js
```diff
@@ -53,15 +61,17 @@ const EditModeDialog = ({ showDialog, editDispatch, onEditFeatures }) => {
...
-            <Button
-              variant="outlined"
-              color="primary"
-              className={classes.margin}
-              startIcon={<TimelineIcon />}
-              onClick={onEditFeatures}
-            >
-              Map
-            </Button>
+            {!isSignalComponent && (
```
johnclary, line 64: @mddilley this works just fine—but i'm curious what you think it'd look like to skip this dialog entirely (or immediately fire off `onEditAttributes`) rather than having the user click the "Attributes" button

moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
```diff
@@ -74,12 +76,29 @@ export const useInitialValuesOnAttributesEdit = (
...
+    if (!isSignalComponent) return;
+    if (!areSignalOptionsLoaded) return;
+
+    const componentSignal = initialFormValues.component?.feature_signals?.[0];
+    const knackFormatSignalOption =
+      featureSignalsRecordToKnackSignalRecord(componentSignal);
+
+    setValue("signal", knackFormatSignalOption);
+  }, [initialFormValues, areSignalOptionsLoaded, setValue]);
+
```
johnclary, line 101: this util is working well, but if we look to add many more fields to this form we might want to revisit this to see if we can cut down on the useEffects. i've never understood RFH's preferred way of handling this very common edit situation where your form defaults may not be ready on first render, but i _think_ an alternative is to memoize the the object you pass to RFH's `defaultValues` - and the default form values will update until the form is dirtied. that might offer a way to combine some of the up-front data munging and not have to call `setValue` a whole bunch. it's not important right now; just had to kick in my two cents.

moped-editor/src/queries/components.js
```diff
@@ -130,6 +133,48 @@ export const UPDATE_COMPONENT_ATTRIBUTES = gql`
   }
 `;
 
+// This mutation updates component subcomponents in same way as UPDATE_COMPONENT_ATTRIBUTES.
+// It also updates all component feature_signals to is_deleted = true and then inserts the new signal.
+export const UPDATE_SIGNAL_COMPONENT = gql`
```
johnclary, line 138: this autocomplete feature really does throw in so many complications - thank you for sorting through them!

johnclary (approved):
🙌 @mddilley this is fantastic!

one minor issue—i'm getting the dreaded MUI warning about an unmatched input value:

```
The value provided to Autocomplete is invalid.
None of the options match with ....
```

i debugged your `getSignalOptionSelected` helper, and noticed that there's a type issue between the options features and the input value. `option.properties.signal_id` is a string, while `value.properties.signal_id` is a number.

i _think_ Socrata is to blame for that—it appears to be stringifying the signal IDs. i have no opinion about the best way to deal with that in your code, but it'd be good to get that equality matching squared.

i'm going to stamp `approved` because this is otherwise so ready to 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
mddilley: @johnclary - glad that you caught this. I almost broke down and refactored this last night. Those `useEffects` that fire on the `clickedComponent` state updates cause a brief blip of the old values in the form before it closes and some re-renders. You are spot on on a better way to do this, and I think the only field that needs some special care would be the signal field that needs to wait on the async call.

Glad that spotted this too! It felt real messy when I dove back in. 🙏

moped-editor/src/views/projects/projectView/ProjectComponents/EditModeDialog.js
mddilley: Oooo yeah, that would be a good one because it will feel faster and look less awkward for sure. 👍

roseeichelmann: i like this idea!

johnclary:
> when i test these deploy previews and click on a project i cant see any components on the map that were already there (not created by me). is that to be expected?

thanks for mentioning that. yes—we need to [update our local env seed](https://github.com/cityofaustin/atd-moped/blob/main/moped-database/README_seed_data.md) data to match this new database schema. i might open a PR for that while testing https://github.com/cityofaustin/atd-moped/pull/824

moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
johnclary: cool cool—will make a backlog issue 🚢

frankhereford (approved):
I tested this with your instructions, and I couldn't find a way to break it. This works great, and your hard work is paying out in spades with this wonderfully intuitive and accessible map.

mddilley:
Thanks y'all for the reviews - just pushed two updates to: 
1. Fix the MUI warning that John described above
2. Update the UI to immediately show the modal for editing attributes when you click the **Edit** button from the list item of a signal component

PR #935, opened 2023-01-11
Drops moped_proj_notes.added_by column

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11182

## Testing
**URL to test:**  Local

**Steps to test:**
1. Start the Hasura cluster and editor

3. Create, edit, and delete comments through the three places this is possible:

- Project comments tab
- Project summary status update
- Status update modal from the Dashboard page

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -2505,21 +2505,20 @@
     name: moped_proj_notes
     schema: public
   object_relationships:
+    - name: moped_phase
+      using:
+        foreign_key_constraint_on: phase_id
     - name: moped_project
       using:
         foreign_key_constraint_on: project_id
     - name: moped_user
       using:
         foreign_key_constraint_on: added_by_user_id
-    - name: moped_phase
-      using:
-        foreign_key_constraint_on: phase_id
   insert_permissions:
     - role: moped-admin
       permission:
         check: {}
         columns:
-          - added_by
           - added_by_user_id
           - date_created
           - is_deleted
@@ -2532,7 +2531,6 @@
       permission:
         check: {}
         columns:
-          - added_by
           - added_by_user_id
           - date_created
           - is_deleted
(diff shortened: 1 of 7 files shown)


moped-database/metadata/tables.yaml
```diff
@@ -2641,6 +2633,13 @@
           value_from_env: MOPED_API_APIKEY
         - name: MOPED_API_EVENT_NAME
           value: activity_log
+      cleanup_config:
+        batch_size: 10000
+        clean_invocation_logs: false
+        clear_older_than: 168
+        paused: true
+        schedule: 0 0 * * *
+        timeout: 60
```
johnclary, line 2642: Hasura did this. it's a new cloud-only feature.

chiaberry (changes requested):
I can't edit or delete any of comments, which I have tracked down to 
` const editableComment =
                      userSessionData.user_id === item.added_by_user_id;` coming back as False

the COMMENTS_QUERY needs to return the added_by_user_id and then its all good

frankhereford (approved):
@chiaberry's comment is right on the money. I observed the same behavior, and when I added `added_by_user_id` to `COMMENT_QUERY` in `moped-editor/src/queries/comments.js#19, it worked like a charm. 

Given that change is in place, I was able to make, edit and delete comments. Thanks for this!

johnclary:
thanks @frankhereford and @chiaberry! i completely missed this :/

taking a closer look, i see we're defaulting the `added_by_user_id` column to `1` in the DB. i believe that value would just fail in production, but in any case i don't think we want a default value for this column. i'm going to spend a little more time thinking about how this—it won't make tomorrow's release.

chiaberry (approved):

johnclary:
turns out i've already broken comment editing in main (and prod?) thanks @chiaberry for patching this up!

PR #938, opened 2023-01-17
Updates project list to be compatible with new components schema

johnclary:
## Associated issues
- 
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/queries/signals.js
```diff
@@ -4,20 +4,9 @@ export const SIGNAL_PROJECTS_QUERY = gql`
...
-          }
-          {
-            moped_proj_components: {
-              moped_components: { component_name: { _ilike: "signal" } }
-              is_deleted: { _eq: false }
-            }
-          }
+        _and: [
+          { is_deleted: { _eq: false } }
+          { moped_proj_components: { feature_signals: {} } }
```
johnclary, line 9: this is a slight change in behavior—the project must have a signal component _feature_, whereas before the project merely needed a signal component. it should have an affect from the user perspective and seemed more explicit?

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -318,7 +318,11 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
       field: "project_team_members",
       hidden: hiddenColumns["project_team_members"],
       cellStyle: { whiteSpace: "noWrap" },
-      render: (entry) => renderProjectTeamMembers(entry.project_team_members, "projectsListView"),
+      render: (entry) =>
+        renderProjectTeamMembers(
+          entry.project_team_members,
+          "projectsListView"
+        ),
```
johnclary, line 325: linting

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -375,21 +379,13 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
-              signalIds.push({
-                signal_id: signal,
-                knack_id: projectFeature.properties.id,
-              });
-            }
-          });
-          return <RenderSignalLink signals={signalIds} />;
+          const signals = entry.project_feature.filter(
+            (signal) => signal.signal_id && signal.knack_id
+          );
```
johnclary, line 387: these two columns (`signal_id` and `knack_id`) are currently nullable in the DB. we should change that after we run the transpose geom script, right?

moped-database/views/project_list_view.sql
```diff
@@ -52,16 +52,15 @@ AS WITH project_person_list_lookup AS (
...
-          ON (features.project_component_id = components.project_component_id)
+        LEFT JOIN uniform_features feature
+          ON (feature.component_id = components.project_component_id)
         WHERE TRUE
           AND components.is_deleted = false
           AND components.project_id = mp.project_id
-        ),
-      '{}'::json) as project_feature, -- close out that coalesce; if null, give us a empty json object
+          AND feature.table = 'feature_signals'
+        ) as project_feature,
```
johnclary, line 63: this column used to return an empty object `{}` when there weren't any signal components. it just returns `null` now.

moped-editor/src/queries/signals.js
mddilley: This makes sense and looks more straightforward to me! I did think of one thing while poking around. Is a signal project still considered a signal project if a user deletes the signal component that was originally added to the project? 

I keep getting hung up on how `is_deleted` in the `features` table factors into all of this. I hope that makes sense.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
mddilley: I agree on that. A feature in the `feature_signals` table should always have those two columns populated.

mddilley (approved):
Looks great, John! Thanks for updating these views to keep the signal features alive and well. I raised one question that I think is more of an edge case but I may be underestimating it. Either way, I'm slapping a ✅ on this. 🚀

moped-editor/src/queries/signals.js
johnclary: oof—yes, you're totally right that the this query needs to consider `is_deleted`. i think hasura should be able to handle this but i'm going to have to wrap my brain around it first. thanks for flagging this!

frankhereford (approved):
excellent

This is great John, thank you!

I tested using your instructions and looked through your code. 

I'm confused about what was going on in the `_or: []` part of the graphql query in `moped-editor/src/queries/signals.js` as it looks like we were `or`ing two identical conditions, but that is outgoing code, and what you've replaced it with makes sense. 👍

I also agree with the idea you proposed about changing the schema to reflect our expectation of always having the `signal_id` and `knack_id`. That seems like a route to catch errors earlier and more reliably, should they come up. 👍

johnclary:
> it looks like we were oring two identical conditions

indeed—we were.

thanks @frankhereford!

i opened [#11223](https://github.com/cityofaustin/atd-data-tech/issues/11223) for to add those `not null` constraints.

mddilley (approved):
Just tested with the updates and with the base feature branch pulled in, and I'm throwing another ✅ on here 🚀

johnclary:
OK! thanks @mddilley for nudging me to do some extra testing of this with the signal component edit code installed. it's time to 🚢

PR #941, opened 2023-01-18
Add new geospatial views to view tracking

johnclary:


moped-database/views/project_geography.sql
```diff
@@ -0,0 +1,23 @@
+-- latest version: 1666978291816_add_component_geometry_tables
```
johnclary, line 1: do y'all think it would be helpful to keep this here? if someone forgets to update it things can get confusing quickly

frankhereford: I'm split. Yes, it's monstrously helpful because you know where to look, but it has the drawback of requiring humans to keep it up to date. Humans forget this sort of thing ... In my dream world, it would be populated by CI, which would just be too cool, but I can't see us justifying the time to make that work at this point. 

For the time being, I'd leave it in there, because at the _very least_ it's a hint that lets you limit the search space if you're using your IDE, grep or what-have-you trying to find migrations by giving you the maximally oldest migration you have to look through. I hold this opinion so lightly, it's a feather; I could have my mind changed by anything.

mddilley: Well put, Frank, and I agree. Some breadcrumb that might be up to date is better than none at all!

PR #948, opened 2023-01-20
adds public process status to activity log

tillyw:
this pr adds 'public process status' to the activity log. merging this pr will update the [10833_public_process_field_ui](https://github.com/cityofaustin/atd-moped/tree/10833_public_process_field_ui) branch, which will then be ready to merge into main. i'm keeping the standard moped project update icon for this update in the activity log since this is somewhat intangible.

## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/11255

## Testing
**URL to test:** 
https://11255_add_public_process_status_to_activity_log--atd-moped-main.netlify.app/moped/

**Steps to test:**
- go to a project summary page and add a public process status
- go to the activity log and refresh to see the update
- remove the public process status and note the change in the activity log

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -2,7 +2,7 @@
   "name": "atd-moped-editor",
   "author": "ATD Data & Technology Services",
   "license": "CC0-1.0",
-  "version": "1.15.0",
+  "version": "1.16.0",
   "homepage": "/moped",
   "private": false,
   "repository": {
--- a/moped-editor/src/utils/activityLogHelpers.js
+++ b/moped-editor/src/utils/activityLogHelpers.js
@@ -11,13 +11,18 @@ export const formatActivityLogEntry = (change, lookupData) => {
 
   switch (change.record_type) {
     case "moped_project":
+      // if the user is changing the public process status id field,
+      // provide the publicProcessStatustList lookup object
       if (change?.description[0]?.field === "public_process_status_id") {
         return formatProjectActivity(
           change,
           lookupData.publicProcessStatustList
         );
       }
-      return formatProjectActivity(change, lookupData.entityList);
+      // otherwise, provide the entityList lookup object
+      else {
+        return formatProjectActivity(change, lookupData.entityList);
+      }
     case "moped_proj_tags":
       return formatTagsActivity(change, lookupData.tagList);
     case "moped_proj_funding":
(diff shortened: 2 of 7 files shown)


moped-editor/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "atd-moped-editor",
   "author": "ATD Data & Technology Services",
   "license": "CC0-1.0",
-  "version": "1.15.0",
+  "version": "1.16.0",
```
tillyw, line 5: i'm not sure why this is here because both branches are supposedly up to date with main but hopefully this is okay. if not i'll uncommit this.

moped-editor/src/utils/activityLogHelpers.js
```diff
@@ -11,7 +11,18 @@ export const formatActivityLogEntry = (change, lookupData) => {
...
+      if (change?.description[0]?.field === "public_process_status_id") {
+        return formatProjectActivity(
+          change,
+          lookupData.publicProcessStatustList
+        );
+      }
+      // otherwise, provide the entityList lookup object
+      else {
+        return formatProjectActivity(change, lookupData.entityList);
+      }
```
chiaberry, line 25: i was on the fence about having this logic be here or in the function but I can't come up with a good reason for it not being here

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -83,6 +83,7 @@ const useLookupTables = (data) =>
     lookupData.fundingSources = {};
     lookupData.fundingPrograms = {};
     lookupData.fundingStatus = {};
+    lookupData.publicProcessStatustList = {};
```
chiaberry, line 86: theres an extra t in here 
Statust List

chiaberry (approved):
remove the extra t from StatustList (its in a few places) and 🚢

moped-editor/src/utils/activityLogHelpers.js
tillyw: @chiaberry i went through the same thought process! i think ultimately maybe we'd want to name the `lookupData` objects the same thing as the table names and then we would just be able to interpolate that into the logic of the component? that would be a departure from our naming schema though. it didn't feel necessary to do that for this one case but i imagine there will be more and i could tackle it then

PR #949, opened 2023-01-23
Refactor activity log text formatting

chiaberry:
## Associated issues

No specific issue per se, but looking at activity log issues like the milestones one (https://github.com/cityofaustin/atd-data-tech/issues/11092), the current way in main to represent activity log updates, i.e., changeDescription and changeValue, was limiting how updates could be formatted. 

This refactoring doesn't tie us down to only two strings and opens the door for more flexibility. 

also this https://github.com/cityofaustin/atd-data-tech/issues/11233 (ecapris is stored as a number, and you cant find the lengths of numbers)

## Testing
**URL to test:** 

Test locally until its parent PR (milesontes) is approved and merged. 

**Steps to test:**
 This updates all of the existing activity log formatters
* funding
* milestones
* partners
* project
* tags

Add / edit / delete records on the funding table, milestones. 
update a project's information on the summary page (be sure to test ecapris and interim project id) as well as tags, project partners and public process status. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
@@ -10,61 +10,83 @@ export const formatFundingActivity = (
   const entryMap = ProjectActivityLogTableMaps["moped_proj_funding"];
 
   const changeIcon = <MonetizationOnOutlinedIcon />;
-  const changeText = [];
 
   // add a new funding source
   if (change.description.length === 0) {
     // if the added record has a funding source, use that as the change value
     if (change.record_data.event.data.new.funding_source_id) {
-      changeText.push({ text: "Added ", style: null });
-      changeText.push({
-        text: fundingSources[
-          change.record_data.event.data.new.funding_source_id
+      return {
+        changeIcon,
+        changeText: [
+          { text: "Added ", style: null },
+          {
+            text: fundingSources[
+              change.record_data.event.data.new.funding_source_id
+            ],
+            style: "boldText",
+          },
+          { text: " as a new funding source.", style: null },
         ],
-        style: "boldText",
-      });
-      changeText.push({ text: " as a new funding source.", style: null });
+      };
       // if not, then check if theres a funding program
     } else if (change.record_data.event.data.new.funding_program_id) {
-      changeText.push({ text: "Added ", style: null });
-      changeText.push({
-        text: fundingPrograms[
-          change.record_data.event.data.new.funding_program_id
+      return {
+        changeIcon,
+        changeText: [
+          { text: "Added ", style: null },
+          {
+            text: fundingPrograms[
+              change.record_data.event.data.new.funding_program_id
+            ],
+            style: "boldText",
+          },
+          { text: " as a new funding source.", style: null },
         ],
-        style: "boldText",
-      });
-      changeText.push({ text: " as a new funding source.", style: null });
+      };
     } else {
-      changeText.push({ text: "Added a new funding source", style: null });
+      return {
+        changeIcon,
+        changeText: [{ text: "Added a new funding source", style: null }],
+      };
     }
-
-    return { changeIcon, changeText };
   }
 
   // delete an existing record
   if (change.description[0].field === "is_deleted") {
-    // if the added record has a funding source, use that as the change value
+    // if the deleted record has a funding source, use that as the change value
     if (change.record_data.event.data.new.funding_source_id) {
-      changeText.push({ text: "Deleted a funding source: ", style: null });
-      changeText.push({
-        text: fundingSources[
-          change.record_data.event.data.new.funding_source_id
+      return {
+        changeIcon,
+        changeText: [
+          { text: "Deleted a funding source: ", style: null },
+          {
+            text: fundingSources[
+              change.record_data.event.data.new.funding_source_id
+            ],
+            style: "boldText",
+          },
         ],
-        style: "boldText",
-      });
+      };
       // if not, then check if theres a funding program
     } else if (change.record_data.event.data.new.funding_program_id) {
-      changeText.push({ text: "Deleted a funding source: ", style: null });
-      changeText.push({
-        text: fundingPrograms[
-          change.record_data.event.data.new.funding_program_id
+      return {
+        changeIcon,
+        changeText: [
+          { text: "Deleted a funding source: ", style: null },
+          {
+            text: fundingPrograms[
+              change.record_data.event.data.new.funding_program_id
+            ],
+            style: "boldText",
+          },
         ],
-        style: "boldText",
-      });
+      };
     } else {
-          changeText.push({ text: "Deleted a funding source", style: null });
+      return {
+        changeIcon,
+        changeText: [{ text: "Deleted a funding source.", style: null }],
+      };
     }
-    return { changeIcon, changeText };
   }
 
   // Multiple fields in the moped_proj_funding table can be updated at once
@@ -86,14 +108,17 @@ export const formatFundingActivity = (
     }
   });
 
-  changeText.push({
-    text: "Edited a funding source by updating the ",
-    style: null,
-  });
-  changeText.push({
-    text: changes.join(", "),
-    style: "boldText",
-  });
-
-  return { changeIcon, changeText };
+  return {
+    changeIcon,
+    changeText: [
+      {
+        text: "Edited a funding source by updating the ",
+        style: null,
+      },
+      {
+        text: changes.join(", "),
+        style: "boldText",
+      },
+    ],
+  };
 };
--- a/moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
@@ -14,7 +14,6 @@ export const formatFundingActivity = (
 
   // add a new funding source
   if (change.description.length === 0) {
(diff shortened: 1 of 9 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -14,9 +14,27 @@ const ProjectActivityEntry = ({
   changeIcon,
   changeDescription,
   changeValue,
+  changeText,
 }) => {
   const classes = useStyles();
 
+  if (changeText) {
```
chiaberry, line 21: temporary until I finish refactoring

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -10,22 +10,27 @@ const useStyles = makeStyles((theme) => ({
...
-  changeIcon,
-  changeDescription,
-  changeValue,
-}) => {
+/**
+ * One row/entry in the project activity log
+ * @param {changeIcon} Material UI icon component
+ * @param {changeText} - Array of objects of the shape:
+ *  {text: String to display, style: String name of style or null}
+ */
```
chiaberry, line 18: would love feedback on how to make this better

moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
```diff
@@ -10,46 +10,61 @@ export const formatFundingActivity = (
...
-    // if there isnt a funding source or program added, then default to ""
-    return { changeIcon, changeDescription, changeValue };
+
+    return { changeIcon, changeText };
   }
 
   // delete an existing record
   if (change.description[0].field === "is_deleted") {
-    changeDescription = "Deleted a funding source";
     // if the added record has a funding source, use that as the change value
```
roseeichelmann, line 46: should this comment say deleted record instead of added 🤔

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
roseeichelmann: i really like this pattern!

moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
```diff
@@ -10,46 +10,61 @@ export const formatFundingActivity = (
...
-      changeValue =
-        fundingSources[change.record_data.event.data.new.funding_source_id];
+      changeText.push({ text: "Added ", style: null });
+      changeText.push({
+        text: fundingSources[
+          change.record_data.event.data.new.funding_source_id
+        ],
+        style: "boldText",
+      });
+      changeText.push({ text: " as a new funding source.", style: null });
```
johnclary, line 26: instead of doing all this pushing, you could `return` from the if/else blocks throughout?

```javascript
  if (change.record_data.event.data.new.funding_source_id) {
      return {
        changeIcon,
        changeText: [
          { text: "Added ", style: null },
          {
            text: fundingSources[
              change.record_data.event.data.new.funding_source_id
            ],
            style: "boldText",
          },
          { text: " as a new funding source.", style: null },
        ],
      };
```

johnclary (approved):
this looks excellent @chiaberry! 

would you mind removing the quotes from these three fields ?

Screen Shot 2023-01-24 at 1 14 49 PM
 
besides that, i think you could DRY this up a little but but it's otherwise good to 🚢

roseeichelmann (approved):
Wowww the activity page is looking so good ✨✨ love this refactor. My only comment is that maybe on editing the project name the activity log could say "project name" instead of just "name"? looks lovely thooo 😍

Screenshot from 2023-01-24 12-41-18

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
tillyw: this make sense to me! so every item in the array can have its own styling

moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
chiaberry: you are very right thank you!

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -10,22 +10,27 @@ const useStyles = makeStyles((theme) => ({
...
       <Box p={0} flexGrow={1}>
         <Typography variant="body2" className={classes.entryText}>
-          {changeDescription}
-          {!!changeValue && (
-            <span className={classes.boldText}>{changeValue}</span>
-          )}
+          {
+            // maps through the array of objects and applies specified style to the text
+            changeText.map((changeObject) => (
+              <span className={classes[changeObject.style]}>
```
tillyw, line 29: would it be worth adding a conditional to only add a className if the style prop exists?

moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
```diff
@@ -10,46 +10,83 @@ export const formatFundingActivity = (
...
   if (change.description.length === 0) {
-    changeDescription = "Added a new funding source";
     // if the added record has a funding source, use that as the change value
     if (change.record_data.event.data.new.funding_source_id) {
-      changeValue =
-        fundingSources[change.record_data.event.data.new.funding_source_id];
+      return {
+        changeIcon,
+        changeText: [
+          { text: "Added ", style: null },
```
tillyw, line 21: related to my comment about only adding a className when rendering if this style field exists, would it be worth leaving this key value pair out of the object if there is no necessary styling? just a thought!

tillyw (approved):
this is a great solution for making the activity log updates more modular and flexible. i just left a couple of comments but this is ready to merge!

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
tillyw, line 29: i'm getting a warning in the console saying each of these spans should have a unique key

chiaberry: good question, i thought if the classname was `null` it would ignore it but i am going to make sure its null and not `undefined`

moped-editor/src/utils/activityLogFormatters/mopedFundingActivity.js
chiaberry: I did this so that changeObject.style would be null and not undefined, but I didn't double check to confirm thats what was happening

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
chiaberry: thank you for checking :)

moped-editor/src/utils/activityLogFormatters/mopedTagsActivity.js
```diff
@@ -2,18 +2,30 @@ import LocalOfferOutlinedIcon from "@material-ui/icons/LocalOfferOutlined";
...
+      ],
+    };
   } else {
     // Soft deleting a tag is the only update a user can do (is_deleted is set to true)
-    changeDescription = "Project tag deleted "
-    changeValue = tagList[change.record_data.event.data.new.tag_id]
+    return {
+      changeIcon,
+      changeText: [
+        { text: "Project tag deleted  ", style: null },
```
chiaberry, line 23: this could now be "Project tag TAGNAME was deleted" 

as long as we are okay with passive voice

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -29,40 +38,83 @@ export const formatProjectActivity = (change, lookupList) => {
...
     }
 
     // the update can be rendered as a string
-    changeDescription = `
-          Changed ${entryMap.fields[changedField].label}
-          to `;
-    changeValue =
-      changeData.new[changedField].length > 0
+    const changeValue =
+      String(changeData.new[changedField]).length > 0
```
mddilley, line 101: Can `changeData.new[changedField]` ever return as `null`? Just wondering because apparently `String(null)` returns `'null'` (as a string) which would have a length greater than 0 here. Not sure of the result of that here. Totally a TIL for me. 🤯

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
mddilley: Same for me! I like this way of representing the different pieces of the activity log text.

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
mddilley: I did some digging in the React docs, and they say that using the index as keys is okay as a last resort. The bugs that stem from index as keys only apply to items that change order in the UI. https://reactjs.org/docs/lists-and-keys.html#keys

mddilley (approved):
Very exciting changes coming to the activity log - looks beautiful. 🙌 All the test steps check out, and I added a comment about using index as keys after looking around a bit. 🚢 🚢 🚢

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
chiaberry: good catch! I'll add a conditional to prevent nulls and undefineds slipping through

PR #939, opened 2023-01-17
 Make Moped <> Data Tracker sync compatible with new components schema

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11216

## Testing
**URL to test:** local

**Steps to test:**
1. To test locally, you'll need to uncomment [L122](https://github.com/cityofaustin/atd-moped/blob/6213fdb27f4808da26a9e9147ba4447355535313/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js#L122) :(
2. Create a project with a signal component
3. On the project's summary tab, click the "Synchronize with Data Tracker" link to push this moped project to the Data Tracker test environment. 
4. Keep an eye on the top right-hand corner of the page to see the green "Success" toast 🥪 
5. Now move your eyes back to where you clicked the "Synchronize with Data Tracker" link, and you should instead see a hyperlinked signal ID. Clicking that ID will take you to a knack app.
6. If you'd like to be thorough, you can login to the test Knack app [here](https://builder.knack.com/atd/test-25-feb-2022-dt-austin-transportation-data-tracker/records/objects/object_201?builder_table_sort=field_3858|desc) using the admin builder creds in our password store. You should see your project in the first row.

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #952, opened 2023-01-24
Moped projects being duplicated in the Data Tracker

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11224

The project summary UI wasn't always updating after a project was sync'd with Knack successfully and opened the door to duplicate syncs after the first one succeeded. It seems that, in the original chaining, sometimes the code would move forward to the `refetch` call before the mutation to update the project completed. The `refetch` would return stale data and miss the update, and the UI would not reflect the accurate Knack sync state.

This updates the promise chaining to refetch data and show the snack bar only after the Moped record successfully updates and after the Knack update request succeeds. I know, it is a mouthful. 😅 

## Testing
**URL to test:** 

Local only 💻 

**Steps to test:**
1. Use the same test steps from https://github.com/cityofaustin/atd-moped/pull/939
2. Sync multiple projects and make sure the Project Summary page updates as shown below each and every time the **Success** snackbar appears. Goal is to try to break it. I've tried about 5 or so.

Screenshot 2023-01-24 at 12 36 54 PM

---
#### Ship list
- [x] Code reviewed 
- ~[ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable~

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
@@ -191,16 +191,17 @@ const ProjectSummaryKnackDataTrackerSync = ({
             project_id: project.project_id,
             knack_id: knackRecord.record.id,
           },
-        });
-      })
-      .then(() => refetch()) // ask the application to update its status from our graphql endpoint
-      .then(() => {
-        // End of the chain; advise the user of success
-        snackbarHandle(
-          true,
-          "Success: Project data synchronized with Data Tracker",
-          "success"
-        );
+        })
+          // ask the application to update its status from our graphql endpoint
+          .then(() => refetch())
+          .then(() =>
+            // End of the chain; advise the user of success
+            snackbarHandle(
+              true,
+              "Success: Project data synchronized with Data Tracker",
+              "success"
+            )
+          );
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryKnackDataTrackerSync.js
```diff
@@ -191,16 +191,17 @@ const ProjectSummaryKnackDataTrackerSync = ({
...
-      })
-      .then(() => refetch()) // ask the application to update its status from our graphql endpoint
-      .then(() => {
-        // End of the chain; advise the user of success
-        snackbarHandle(
-          true,
-          "Success: Project data synchronized with Data Tracker",
-          "success"
-        );
+        })
```
mddilley, line 194: Kind of hard to see in the diff, but these `then` calls were moved to chain from `mutateProjectKnackId` (the mutation to update the Moped project record) instead of the `fetch` that starts all of the updates

johnclary (approved):
i tested this a few times, including with low-bandwidth throttling enabled. it works 💯  ✨ 

🙌 🚢

chiaberry (approved):
yay!

PR #954, opened 2023-01-25
refines project components activity log events

tillyw:
this pr refines components activity log events using REST connectors.

note: reflecting component feature changes in the activity log is outside the scope of this issue, that will be addressed in the future.

## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/10806

## Testing
**URL to test:** 

https://deploy-preview-954--atd-moped-main.netlify.app/

**Steps to test:**
- add, edit and remove a component while noting the activity log updates
- the changes should be reflected in the activity log

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/activityLogFormatters/mopedComponentsActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedComponentsActivity.js
@@ -1,0 +1,60 @@
+import RoomOutlinedIcon from "@material-ui/icons/RoomOutlined";
+import { ProjectActivityLogTableMaps } from "../../views/projects/projectView/ProjectActivityLogTableMaps";
+
+export const formatComponentsActivity = (change, componentList) => {
+  const entryMap = ProjectActivityLogTableMaps["moped_proj_components"];
+
+  const changeIcon = <RoomOutlinedIcon />;
+  const component =
+    componentList[change.record_data.event.data.new.component_id];
+  const componentText = {
+    text: component,
+    style: "boldText",
+  };
+
+  // add a new component
+  if (change.description.length === 0) {
+    return {
+      changeIcon,
+      changeText: [{ text: "Added a component: ", style: null }, componentText],
+    };
+  }
+
+  // delete an existing component
+  if (change.description[0].field === "is_deleted") {
+    return {
+      changeIcon,
+      changeText: [
+        { text: "Removed a component: ", style: null },
+        componentText,
+      ],
+    };
+  }
+
+  // Multiple fields in the moped_proj_components table can be updated at once
+  // We list the fields changed in the activity log, this gathers the fields changed
+  const newRecord = change.record_data.event.data.new;
+  const oldRecord = change.record_data.event.data.old;
+
+  let changes = [];
+
+  // loop through fields to check for differences, push label onto changes Array
+  Object.keys(newRecord).forEach((field) => {
+    if (newRecord[field] !== oldRecord[field]) {
+      changes.push(entryMap.fields[field]?.label);
+    }
+  });
+
+  return {
+    changeIcon,
+    changeText: [
+      { text: "Edited ", style: null },
+      componentText,
+      { text: " by updating the ", style: null },
+      {
+        text: changes.join(", "),
+        style: "boldText",
+      },
+    ],
+  };
+};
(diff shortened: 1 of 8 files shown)


moped-editor/src/utils/activityLogFormatters/mopedMilestonesActivity.js
```diff
@@ -16,7 +16,7 @@ export const formatMilestonesActivity = (change, milestoneList) => {
           text: milestoneList[change.record_data.event.data.new.milestone_id],
           style: "boldText",
         },
-        { text: " as a new milestone.", style: null },
+        { text: " as a new milestone", style: null },
```
mddilley, line 19: 🔍 🙌

moped-editor/src/utils/activityLogFormatters/mopedComponentsActivity.js
```diff
@@ -0,0 +1,60 @@
...
+  const componentText = {
+    text: component,
+    style: "boldText",
+  };
+
+  // add a new component
+  if (change.description.length === 0) {
+    return {
+      changeIcon,
+      changeText: [{ text: "Added a component: ", style: null }, componentText],
```
mddilley, line 19: Noticed a `:` here and in the remove message that I didn't see anywhere else in the log. Gotta admit that I didn't look at every change type though!

mddilley (dismissed):
Looks awesome! I added, edited, and removed and saw it all logged in the feed. 🗺️  🚀 

I think this is definitely outside of the scope of this issue, but it would be cool to log if the geometry of a component updated. Not sure what that would look like and maybe there is a backlog issue for it.

tillyw:
> Member

thanks, @mddilley! i had the same thought and had a conversation with @johnclary before i submitted this for review and he said it's out of the scope of the issue but definitely something we should revisit! i do wonder if there is an issue for it

chiaberry (approved):
great! i agree that future enhancements will be good, but this is exactly what the issue requested

roseeichelmann (changes requested):
when i choose a component that has a "x - x" pattern in the name, the full name doesn't show in the activity log. for example if i choose "Crosswalk - Creative", only "Crosswalk" shows. 
image

should this be adjusted to capture the full name?

tillyw:
> when i choose a component that has a "x - x" pattern in the name, the full name doesn't show in the activity log. for example if i choose "Crosswalk - Creative", only "Crosswalk" shows. image
> 
> should this be adjusted to capture the full name?

thanks for catching this @roseeichelmann! I think this is a bug - I neglected to take into account a naming pattern for subcomponents. i will need to address this but i likely won't get to it until next week so i'm going to put this pr back into 'in progress' for now

mddilley:
> should this be adjusted to capture the full name?

@roseeichelmann @tillyw I should have called this out in my review because I thought the same thing. I found https://github.com/cityofaustin/atd-data-tech/issues/10807, and I think that covers this? Need to check with @johnclary

tillyw:
actually on further investigation the field we're not displaying is `component_subtype` which is part of the `moped_components` table - i just need to make sure to populate that when it exists! this isn't related to the `moped_subcomponents` table

tillyw:
@roseeichelmann @mddilley this was going to drive me crazy so i fixed it!
Screen Shot 2023-01-26 at 5 37 18 PM
mike you still raise a good point about subcomponents - those don't display either. thanks for pulling up that issue - i'll be happy to add that next week.

roseeichelmann (approved):
you zapped that 🐛 so quick, looks beaut ✨

johnclary (approved):
Yep, we have follow up issues for subcomponents and map features:
- https://github.com/cityofaustin/atd-data-tech/issues/10809
- https://github.com/cityofaustin/atd-data-tech/issues/10807

It's going to be tricky to render meaningful information about these nested / many-many tables in the activity log. We'll have to query the related records out of the project and use them as a lookup. A fun challenge 🍿 

this looks fantastic @tillyw — thanks for getting the component subtype added in there 💯 ✨ 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🙌

tillyw:
thanks, all for the thorough reviews! and thanks, @johnclary for pointing me to those follow-up issues - that does sound like a fun challenge indeed 🤓

PR #958, opened 2023-01-26
Adds council districts layer to Moped DB

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/10891

This PR gets council districts into our database. We'll definitely need to optimize our queries and make use of indexes if we want to [add council district to the project list](https://github.com/cityofaustin/atd-data-tech/issues/10889). But that's a problem for another day, I hope.

## Testing
**URL to test:** local

**Steps to test:**
1. Spin up the hasura cluster

2. Follow the steps in [the README](https://github.com/cityofaustin/atd-moped/blob/ab72ceb780a3830c467431879d5e962c8b72a36a/moped-toolbox/council_districts_load/README.md) to run both scripts. I've left a copy of `data/council_districts.geojson` in the repo. It seems like we might as well save it, right?

3. You can inspect `data/council_districts.geojson` via https://geojson.io/

4. Open the hasura console and observe the 10 records in the `layer_council_district` table.

5. Create a new project with components spread out across the city.

6. Run this query in your SQL client—it will return **one row per council district** that a project intersects with.

```sql
WITH projects AS (
	SELECT
		plv.*,
		project_geography.feature_id,
		project_geography.geography
	FROM
		project_geography
	LEFT JOIN project_list_view plv ON plv.project_id = project_geography.project_id
)
SELECT
	*
FROM
	projects,
	layer_council_district
WHERE
	ST_Intersects(projects.geography, layer_council_district.geography);
```

7. You'll need to eyeball these results against a [council district map](https://maps.austintexas.gov/GIS/CouncilDistrictMap/)

---
#### Ship list
- [ ] Code reviewed 


moped-toolbox/council_districts_load/insert_council_districts.py
```diff
@@ -0,0 +1,124 @@
+"""Downloads council districts geojson from AGOL and inserts it into Moped"""
+import argparse
+import requests
+from secrets import HASURA
+
+
+DISTRICTS_URL = "https://services.arcgis.com/0L95CJ0VTaxqcmED/ArcGIS/rest/services/BOUNDARIES_single_member_districts/FeatureServer/0/query?where=1%3D1&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&resultType=none&distance=0.0&units=esriSRUnit_Meter&relationParam=&returnGeodetic=false&outFields=*&returnGeometry=true&returnCentroid=false&featureEncoding=esriDefault&multipatchOption=xyFootprint&maxAllowableOffset=&geometryPrecision=8&outSR=4326&defaultSR=&datumTransformation=&applyVCSProjection=false&returnIdsOnly=false&returnUniqueIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&returnQueryGeometry=false&returnDistinctValues=false&cacheHint=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&having=&resultOffset=&resultRecordCount=&returnZ=false&returnM=false&returnExceededLimitFeatures=true&quantizationParameters=&sqlFormat=none&f=pgeojson&token="
```
johnclary, line 7: Esri adds in all these default params when you build a query from their UI. the params of note:

- geometryPrecision=8: cap coordinates at 8 decimal places
- outSR=4326: use WGS84 coordinates
- f=pgeojson: return features as geojson

moped-toolbox/council_districts_load/insert_council_districts.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def make_multipoly(features):
+    for feature in features:
+        if feature["geometry"]["type"] == "Polygon":
+            feature["geometry"]["type"] = "MultiPolygon"
+            feature["geometry"]["coordinates"] = [feature["geometry"]["coordinates"]]
+            print(
+                f"Converted district {feature['properties']['COUNCIL_DISTRICT']} to MuliPolygon"
+            )
```
johnclary, line 63: the Esri feature type for this layer is `esriGeometryPolygon`, which is agnostic to multi vs single-part shapes. when the AGOL server converts the features to geojson it returns either `MultiPolygon` or `Polygon` as needed.

we're using the `MultiPolygon` type in the DB, so so the hack to convert poly to a multipoly is to simply wrap the poly's coordinates in `[]`.

moped-toolbox/council_districts_load/insert_council_districts.py
```diff
@@ -0,0 +1,124 @@
...
+    for f in district_2s:
+        if f["geometry"]["type"] != "MultiPolygon":
+            raise Exception("Feature must be multipolygon for this hack to work")
+        new_district_2_coordinates = new_district_2_coordinates + f["geometry"]["coordinates"]
+    # update the first district 2 with the geometry of all district 2s
+    new_district_2 = district_2s[0]
+    new_district_2["geometry"]["coordinates"] = new_district_2_coordinates
+    # add this back into our feature array
+    other_districts.append(new_district_2)
+    return other_districts
```
johnclary, line 79: as complained about [here](https://austininnovation.slack.com/archives/C015VKQSA23/p1674750864318159), the layer has two district 2 polygons 😑

johnclary: @frankhereford i'd love to hear your thoughts on this.

i also wonder if we should accept the AGOL features as-is and just have two district 2s? part of me thinks it's best to keep the data as close to the authoritative layer as possible. it really won't affect any of our querying. my biggest concern is that we'd have to drop the `unique` constraint from the `council_district` column and that certainly invites future trouble.

moped-database/migrations/1674748665079_create_layer_council_district/up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE layer_council_district (
+    id serial PRIMARY KEY,
+    council_district integer UNIQUE NOT NULL,
+    geography geography ('MULTIPOLYGON', 4326) NOT NULL
+);
```
johnclary, line 5: i am very open to ideas about the naming convention here! my thinking was that `layer_council_district` makes clear this is a reference layer, in contrast to, e.g., `feature_signals`.

moped-toolbox/council_districts_load/secrets_template.py
```diff
@@ -0,0 +1,8 @@
+HASURA_AUTH = {
```
frankhereford, line 1: I believe this line needs to read `HASURA = {`

moped-database/migrations/1674748665079_create_layer_council_district/up.sql
mddilley: I like this naming - it is easy to spot when scanning through the tables as well. 👍

moped-toolbox/council_districts_load/insert_council_districts.py
```diff
@@ -0,0 +1,124 @@
...
+    features_de_duped = merge_district_2(features)
+    # parse out the only columns of interest and make lower case (the agol layer has additional metadata)
+    inserts = [
+        {
+            "council_district": f["properties"]["COUNCIL_DISTRICT"],
+            "geography": f["geometry"],
+        }
+        for f in features_de_duped
+    ]
+    # test to make sure we now have exactly the districts we're expecting
```
mddilley, line 97: 🚀

moped-toolbox/council_districts_load/secrets_template.py
mddilley, line 1: Looks like the script is expecting this to be `HASURA` and not `HASURA_AUTH`

mddilley (commented):
One small issue with the secrets file, but all looks good. Just saw @frankhereford's comment in Slack so I'm ready to approve when that gets sorted out. Thanks for getting this work in! 🗺️

moped-toolbox/council_districts_load/insert_council_districts.py
```diff
@@ -0,0 +1,124 @@
...
+
+def main(env):
+    print(f"Env: {env}")
+    print("Downloading features from AGOL...")
+    features = get_council_districts_geojson()
+    # ensure all features are of same type (AGOL returns a mix of multi/not multi polygons)
+    make_multipoly(features)
+    # combine bizarre multiple district twos into one multipoly
+    features_de_duped = merge_district_2(features)
+    # parse out the only columns of interest and make lower case (the agol layer has additional metadata)
```
frankhereford, line 89: We spoke about this briefly in slack: https://austininnovation.slack.com/archives/CNUEPKLB1/p1674757493224919

moped-toolbox/council_districts_load/insert_council_districts.py
frankhereford: My heart wants to go with whatever clean-up needing to be done be done in this script, and then this script can become an infrequently run ETL to keep the data up to date. I can totally see a world where this becomes a spin-off issue to avoid blocking getting something out the door that is pressing.

frankhereford: Also, I can see that districts 1, 2, & 5 are interesting in that they have wholly disjointed, disconnected areas (think an "island off the coast" of the main district area, which may explain why they are malformed, but certainly does not excuse it.

frankhereford (commented):
Let me know if I can help with any of this. You're doing something here that I have never quite gotten right (nor attempted in a very long time) -- that is to consume directly from AGOL and come up with data that is workable in a non-ESRI format. I have a very similar task on my plate with consuming the VZ polygons from AGOL or `maint` and putting them in the VZ DB. @patrickm02L - we should track this task and use as much of it as we can.

Your work is an awesome starting off point on that - thank you!.

johnclary:
@mddilley @frankhereford thanks for your feedback. i rewrote this in javascript to make use of `@terraformer`.  Esri's built-in support for geojson is unequivocally busted, so this new approach downloads the esri-style JSON and converts it to geojson with a battle-tested tool.

i re-requested reviews 🙏

frankhereford (approved):
This is a succinct, elegant solution to a nasty problem! This is really great work John.

image

Those districts have holes in them as good as any Utah Teapot. Anchors up, this is ready to ship.

mddilley (approved):
Tested this out locally and all looks good to me! Thanks for getting to the bottom of all of this and for adding another Node script into the toolbox. ✨ 🚀

johnclary:
thank you both for your feedback and thorough testing!!

moped-database/migrations/1674748665079_create_layer_council_district/up.sql
johnclary: ugh, just realized we left this unfinished conversation. i see both sides. a table is a table that may/may not have a geometry column. i do definitely agree that this is clearly a use case for separate db schema.

in any case, it will still be some time before this table is referenced in application code, so we can definitely change it.

johnclary:
✅ i imported the district data into the staging DB. we'll take care of prod during https://github.com/cityofaustin/atd-data-tech/issues/11240

PR #959, opened 2023-01-26
Files activity log events

tillyw:
## Associated issues
pending issue @johnclary 

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -3973,31 +3973,47 @@
         insert:
           columns: '*'
         update:
-          columns:
-            - is_deleted
-            - is_scanned
-            - created_by
-            - file_size
-            - file_type
-            - project_file_id
-            - project_id
-            - api_response
-            - file_metadata
-            - file_permissions
-            - file_description
-            - file_key
-            - file_name
-            - create_date
+          columns: '*'
       retry_conf:
         interval_sec: 10
         num_retries: 0
         timeout_sec: 60
-      webhook_from_env: MOPED_API_EVENTS_URL
+      webhook_from_env: HASURA_ENDPOINT
       headers:
-        - name: MOPED_API_APIKEY
-          value_from_env: MOPED_API_APIKEY
-        - name: MOPED_API_EVENT_NAME
-          value: activity_log
+        - name: x-hasura-admin-secret
+          value_from_env: ACTIVITY_LOG_API_SECRET
+      request_transform:
+        body:
+          action: transform
+          template: |-
+            {
+              "query": "mutation InsertActivity($object: moped_activity_log_insert_input!, $project_id:Int!, $updated_at:timestamptz ) { insert_moped_activity_log_one(object: $object) { activity_id } update_moped_project_by_pk(pk_columns: {project_id: $project_id}, _set: {updated_at: $updated_at}) { updated_at }}",
+              "variables": {
+                "object": {
+                  "record_id": {{ $body.event.data.new.project_id }},
+                  "record_type":  {{ $body.table.name }},
+                  "activity_id": {{ $body.id }},
+                  "record_project_id": {{ $body.event.data.new.project_id }},
+                  "record_data": {"event": {{ $body.event }}},
+                  "description": [{"newSchema": "true"}],
+                  "operation_type": {{ $body.event.op }},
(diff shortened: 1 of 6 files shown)


moped-database/metadata/tables.yaml
```diff
@@ -4072,31 +4072,47 @@
...
-            - project_file_id
-            - project_id
-            - api_response
-            - file_metadata
-            - file_permissions
-            - file_description
-            - file_key
-            - file_name
-            - create_date
+          columns: '*'
```
frankhereford, line 4075: Is this a new `graphql-engine` syntax? If so, I'm really happy to see it! `*` can expand with new columns without manually having to specify them, which I think is a good fit for a lot of what we do and also very within the spirit of Hasura's `graphql-engine` philosophy.

frankhereford (approved):
This is great work Tilly! I've tested the new activity formatter using your instructions, and gone over this code, it's 🚢🚢🚢.

I'm behind in keeping up with all the developments that have been going on with the activity log, and I think it's really top-shelf, the whole thing. Thank you so much! 💯

tillyw:
> This is great work Tilly! I've tested the new activity formatter using your instructions, and gone over this code, it's 🚢🚢🚢.
> 
> I'm behind in keeping up with all the developments that have been going on with the activity log, and I think it's really top-shelf, the whole thing. Thank you so much! 💯

thanks @frankhereford !

moped-database/metadata/tables.yaml
tillyw: agreed! it's super useful and looks so much cleaner

moped-database/metadata/tables.yaml
```diff
@@ -4072,31 +4072,47 @@
...
+          value_from_env: ACTIVITY_LOG_API_SECRET
+      request_transform:
+        body:
+          action: transform
+          template: |-
+            {
+              "query": "mutation InsertActivity($object: moped_activity_log_insert_input!, $project_id:Int!, $updated_at:timestamptz ) { insert_moped_activity_log_one(object: $object) { activity_id } update_moped_project_by_pk(pk_columns: {project_id: $project_id}, _set: {updated_at: $updated_at}) { updated_at }}",
+              "variables": {
+                "object": {
+                  "record_id": {{ $body.event.data.new.project_id }},
```
johnclary, line 4092: would you set this to `project_file_id` instead of `project_id`? this column is meant to capture the unique ID of the record itself, and `record_project_id` forms the link between this record and it's parent project.

Chia went through and updated most of our other triggers to follow this pattern, though it was kinda buried in [this PR](https://github.com/cityofaustin/atd-moped/pull/965).

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -321,7 +321,8 @@ const ProjectActivityLog = () => {
                           "moped_proj_notes",
                           "moped_proj_components",
                           "moped_project_types",
-                          "moped_proj_personnel"
+                          "moped_project_files",
+                          "moped_proj_personnel",
                         ].includes(change.record_type) ? (
```
johnclary, line 326: i believe we can now do away with this special case handling—all the event triggers have been migrated to this new pattern 🎉 

this is fine as-is: we will clean this up in https://github.com/cityofaustin/atd-data-tech/issues/10897

johnclary (changes requested):
@tillyw this looks great + works great 🙌  just one lil change to the event mutation, please.

moped-database/metadata/tables.yaml
tillyw: thanks for pointing this out @johnclary! updated!

johnclary (approved):
🚢 🚢 🚢 🚢  🎉 


thanks @tillyw!!

johnclary:
out of curiosity, did y'all find that these activity log events take a while (15 seconds?) to show up in the feed? these were noticeably slower for me than other events. i poked around in the code to try and understand the DB operations for files, and i don't see anything particularly different.

frankhereford:
@johnclary wrote:

> did y'all find that these activity log events take a while (15 seconds?)

Yes, 100%. It puzzled me because I figured that getting the λ out of the loop would hasten it, but maybe `graphql-engine` queues it and has a lazy processor. 🤷🏼

PR #957, opened 2023-01-26
Use CTN intersection points instead of COA intersection points

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11306

This PR updates the feature service endpoint for the point component data from https://services.arcgis.com/0L95CJ0VTaxqcmED/ArcGIS/rest/services/LOCATION_coa_intersection_points/FeatureServer/0  to https://services.arcgis.com/0L95CJ0VTaxqcmED/arcgis/rest/services/CTN_Intersections_MOPED_FS/FeatureServer/0.

## Testing
**URL to test:** 

local or https://deploy-preview-957--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. In a new or existing project, create a component that uses point geometries (like Bike Box)
2. When creating, select and deselect points a lot and then save
3. Edit the component and select and deselect a lot
4. For extra close 🔍, run locally and compare the data captured in the `feature_intersections` table to the data in the same table of the staging read replica - the same columns should populate

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager approved
~- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable~

PR #960, opened 2023-01-26
Patch 1.15.1

mddilley:
This is a companion PR to https://github.com/cityofaustin/atd-moped/pull/957, which we're shipping as a patch release to minimize the fallout from https://github.com/cityofaustin/atd-data-tech/issues/11306.

#### Ship list
- [ ] Bump `main` back to `1.16.0` after merging this


chiaberry:
🩹

johnclary (approved):
🚢 🪡

PR #961, opened 2023-01-27
Use relative timestamps in activity log

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11356
- https://github.com/cityofaustin/atd-data-tech/issues/11357

## Testing
**URL to test:** local, but [netlify preview](https://deploy-preview-961--atd-moped-main.netlify.app/moped/projects) will be best

**Steps to test:**
1. Take a look at some existing projects and try to to find some activity events that match these criteria. You'll need to make your own edits to see "Just now" :)
 - less than 1 minute: `Just now`
 - less than 1 hour: `X minute(s) ago`
 - less than 1 day: `x hours ago`
 - less than 30 days: `x days ago`
 - else `m/dd/yyyy` - 1/23/2023

2. Hover over some of the relative timestamps in the activity log. A tooltip will open and show the full timestamp

Screen Shot 2023-01-27 at 3 47 05 PM

3. Run the tests. In the prompt that follows, you can use the `p` option to run tests by pattern then type `date` to select only our `dateAndTime` tests.

```shell
$ npm run test
```

You should see this test pass:

```shell
  formatRelativeDate()
    ✓ formats millesconds ago to a human description relative to now
```

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/utils/dateAndTime.js
```diff
@@ -57,11 +57,58 @@ export const makeHourAndMinutesFromTimeStampTZ = timeStampTZString => {
  * @param {String} timeStampTZString - Timestamptz string in ISO format
  * @return {String} Time formatted as HH:MM:SS XM
  */
-export const makeFullTimeFromTimeStampTZ = timeStampTZString => {
+export const makeFullTimeFromTimeStampTZ = (timeStampTZString) => {
```
johnclary, line 60: linting above  ☝️

moped-editor/src/utils/dateAndTime.js
```diff
@@ -57,11 +57,58 @@ export const makeHourAndMinutesFromTimeStampTZ = timeStampTZString => {
...
+};
+
+/**
+ * Format millseconds into a relative description
+ * @param {integer} millsElapsed - the elapsed time in milliseconds to format
+ * @param {string} unitName - the name of the units to convert to - the string must
+ * be defined in the MILLS_CONVERSION objeect
+ * @returns {string} in format  `<value> <unit-name> ago`
+ */
+const formatRelativeTime = (millsElapsed, unitName) => {
```
johnclary, line 82: i had fun with this coding challenge, but i am very open to using a library instead. [this one](https://github.com/EvanHahn/HumanizeDuration.js) works well.

moped-editor/src/utils/dateAndTime.js
```diff
@@ -57,11 +57,58 @@ export const makeHourAndMinutesFromTimeStampTZ = timeStampTZString => {
...
+export const formatRelativeDate = (timeStampTZString) => {
+  const targetDate = new Date(timeStampTZString);
+  const millsElapsed = new Date().getTime() - targetDate.getTime();
+  if (millsElapsed > MILLS_CONVERSION.day * 30) {
+    return targetDate.toLocaleDateString("en-US");
+  } else if (millsElapsed < MILLS_CONVERSION.minute) {
+    return "Just now";
+  } else if (millsElapsed < MILLS_CONVERSION.hour - MILLS_CONVERSION.minute) {
+    return formatRelativeTime(millsElapsed, "minute");
+  } else if (millsElapsed < MILLS_CONVERSION.day - MILLS_CONVERSION.hour) {
```
johnclary, line 109: we subtract one notch from the next smaller unit of measure so that later on so that 59 minutes and 30 seconds return "1 hour" instead of "60 minutes".

moped-editor/src/utils/test/dateAndTime.test.js
```diff
@@ -54,3 +55,39 @@ describe("makeFullTimeFromTimeStampTZ()", () => {
     expect(formattedTime).toBe("11:14:45 AM");
```
johnclary, line 55: these time-specific tests fail for me when i run them from Eastern Time :/  

looks like you can [set the timezone in a Jest config](https://stackoverflow.com/a/56482581/2103784) but i did wander down this path

johnclary:
👋 no urgency here—this was just something i noodled on after hours and it doesn't need to ship this week.

moped-editor/src/utils/dateAndTime.js
```diff
@@ -57,11 +57,58 @@ export const makeHourAndMinutesFromTimeStampTZ = timeStampTZString => {
...
     hour: "numeric",
     minute: "numeric",
     second: "numeric",
   });
 };
+
+const MILLS_CONVERSION = {
+  minute: 60_000,
+  hour: 3_600_000,
+  day: 86_400_000,
```
chiaberry, line 72: TIL _

chiaberry (approved):
Cool to see the relative timestamps and I really thankful for the tooltip with the exact date since I kind of obsess over dates and like also knowing what day of the month things happened. 

I am not super into the date column being to the right and I think its because I care about when something happened when looking at the activity log, and the date column being to right makes it harder to line up when something happened. Also makes the whole thing not left justified. But I don't feel super strongly about this and it could just be change is hard. 

I was going to say I'd rather see the user name to the right than the date, but users may care more about who made a change than when it was made.

johnclary:
thanks @chiaberry! i think i'm with you that the order of the columns was better with the date on the left, esp b/c of the alignment.

i also wonder if there's a way to make it more obvious to users that they can hover over the date to see the tooltip. like maybe it should be hyperlinked or have an <info> icon next to it?

patrickm02L (approved):
This works overall. Two things stuck out to me...

The hover text is on the small side, I could see some of our older guests having a rough time reading it, and might bump it up to the same text size as the Date. 

Also, I prefer having a single word "Now" rather than "Just Now". 

Screen Shot 2023-02-03 at 11 10 18 AM

moped-editor/src/utils/dateAndTime.js
mddilley: +1

moped-editor/src/utils/test/dateAndTime.test.js
mddilley: Glad that you tripped over this. Didn't realize that would be the case and thanks for pointing to the Jest config 🙏

moped-editor/src/utils/dateAndTime.js
mddilley: I think this is great without the library - especially with test coverage! 🙌

mddilley (approved):
This is such a great feature ⏳, and thanks for polishing up some of the details as well. The simplified tab label and removal of the h2 really clean this up. Looks amazing! ✨ 🚀 👏

moped-editor/src/utils/dateAndTime.js
```diff
@@ -57,11 +57,58 @@ export const makeHourAndMinutesFromTimeStampTZ = timeStampTZString => {
...
+};
+
+/**
+ * Format millseconds into a relative description
+ * @param {integer} millsElapsed - the elapsed time in milliseconds to format
+ * @param {string} unitName - the name of the units to convert to - the string must
+ * be defined in the MILLS_CONVERSION objeect
```
mddilley, line 79: Forgot this in the review - sp on object

roseeichelmann (approved):
i love these changes, this looks so great john! super clean. i agree about it not being obvious that you can hover over the timestamp to get the tooltip. an icon would be useful for this otherwise i would have never thought to hover and i think for our users it may be especially helpful for them to find exact dates easily. i personally dont mind the date column being right-aligned. 🚢🚢🚢

moped-editor/src/utils/test/dateAndTime.test.js
```diff
@@ -54,3 +55,39 @@ describe("makeFullTimeFromTimeStampTZ()", () => {
...
+describe("formatRelativeDate()", () => {
+  it("formats millesconds ago to a human description relative to now", () => {
+    const nowMills = new Date().getTime();
+    const twoSecondsAgoMills = nowMills - 1000 * 2;
+    const oneMinute20sAgoMills = nowMills - 1000 * 70;
+    const twoMinutesAgoMills = nowMills - 1000 * 60 * 2;
+    const fiftyNineMinutesAgo = nowMills - 1000 * 60 * 59;
+    const twoHoursAgoMills = nowMills - 1000 * 60 * 60 * 2;
+    const twoDaysAgoMills = nowMills - 1000 * 60 * 60 * 2 * 24;
+    const twoMonthsAgoMills = nowMills - 1000 * 60 * 60 * 2 * 24 * 60;
```
frankhereford, line 68: I always enjoy seeing patterns like this; these are magic numbers that I can get behind. ✨ I like to turn off linting rules for chunks like this and line them all up -- it is totally useless but the columns look so organized 🤩

moped-editor/src/utils/dateAndTime.js
frankhereford: I agree wholeheartedly with Mike - this works great. A dependency we didn't we didn't pick up is 👍. 🚀

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -137,7 +137,7 @@ const TABS = [
   { label: "Comments", Component: ProjectComments, param: "comments" },
   { label: "Files", Component: ProjectFiles, param: "files" },
   {
-    label: "Activity Log",
```
frankhereford, line 140: Less is more - I like this.

frankhereford (approved):
🚢🛳️ This is super well done both in terms of the UX and the implementation. Embedding the `onHover` tooltip with the actual timestamp is just so nice. Also, the tests -- very cool how that testing system presents each test and the results of the evaluation. I like that tooling a bunch - that is TIL for me 💯.

tillyw (approved):
tested in the deploy preview and ran the test scripts - all went perfectly! this looks great and i love the relative timestamps. i agree that an icon or indicator might be helpful to point people to the tooltip - maybe just an information icon with a tooltip next to the date header would suffice?

patrickm02L:
>  what do y'all think about rendering the full date when hovering over a row?

The full date timestamp in light grey underneath the min/hour/day time representation works for me.

However, rather than revealing the timestamps on hover, my druthers would be to see all the full dates rendered on the page all at once. When looking at the updates that were made on the same day, I found myself wanting to compare their time stamps to each other.

chiaberry:
Druthers! I think it looks great as is, I still like the timestamp to the left but its just a personal preference. I vote to get this in production and we can see what the MUG thinks. 

Dreaming about UX interviews and research!

johnclary:
> I vote to get this in production and we can see what the MUG thinks.

agreed—than's y'all! 🚢

PR #964, opened 2023-02-01
Update event triggers for moped personnel and personnel roles

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/10810

## Testing
**URL to test:** 
https://deploy-preview-964--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

1. Add team members to a team
2. Remove a team member
3. Swap out one team member for a different name
4. Do any other edits you can to a team member

Changing roles are tracked in the activity log, but aren't shown in the UI. You can check that the mutations are working locally by checking http://localhost:9695/console/events/data/activity_log_moped_proj_personnel_roles/processed or querying the db. 

I also have a note about a bug I saw in staging where if you save empty spaces as a project description, you can't edit the description again (ex: https://moped.austinmobility.io/moped/projects/1790). Try saving a description as "      " and you should see the "-" render so you can still edit the field. 

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -3150,6 +3171,52 @@
...
+          template: |-
+            {
+                            "query": "mutation InsertActivity($object: moped_activity_log_insert_input!) { insert_moped_activity_log_one(object: $object) { activity_id } }",
+                            "variables": {
+                                "object": {
+                                    "record_id": {{ $body.event.data.new.project_personnel_id }},
+                                    "record_type":  {{ $body.table.name }},
+                                    "activity_id": {{ $body.id }},
+                                    "record_data": { "event": {{ $body.event }}},
+                                    "record_project_id": null,
```
chiaberry, line 3204: `moped_proj_personnel_roles` does not have a column for the project_id, so this is null. This also means that while we are tracking these activities, they will not show up in our activity log query since we query on project_id

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -114,7 +114,7 @@ const ProjectSummaryProjectDescription = ({
         )}
         {!editMode && (
           <ProjectSummaryLabel
-            text={description || " - "}
+            text={description.trim().length > 0 ? description : " - "}
```
chiaberry, line 117: This is to fix a bug I found where if you save empty spaces as a project description, you can't edit the description again (ex: https://moped.austinmobility.io/moped/projects/1790) `Boolean("  ")` is true.

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -95,7 +95,6 @@ export const formatProjectActivity = (change, lookupList) => {
         ],
       };
     }
-    console.log(changeData.new[changedField])
```
chiaberry, line 98: 😬

moped-database/metadata/tables.yaml
johnclary: this is great—thanks for getting the tracking in place.

i also want to point out that, because we don't have the project ID here, we cannot update the project's modified date through this mutation. that's frustrating because i think user's should be able to expect that any project edit will update the project's modified date. if we wanted to get clever, we _could_ mutate the `moped_proj_personnel` record from here, which would then fire off another event which would mutate the parent project's modified date. that would require us adding some mutable field on `moped_proj_personnel` like an `updated_at` timestamp. 

just throwing it out there—i don't necessarily think this is a good idea. i believe we're also going to encounter this problem when we set up tracking for project features.

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
```diff
@@ -26,7 +26,11 @@ const ProjectActivityEntry = ({ changeIcon, changeText }) => {
     <Box display="flex" p={0}>
       <Box p={0}>{changeIcon}</Box>
       <Box p={0} flexGrow={1}>
-        <Typography variant="body2" className={classes.entryText}>
+        <Typography
+          variant="body2"
+          className={classes.entryText}
+          component="span"
```
johnclary, line 32: this fixes an invalid DOM nesting error

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
johnclary: thank you for fixing this!

we have [a snackoo](https://github.com/cityofaustin/atd-data-tech/issues/10339) to prevent users from saving an empty description in the first place. we require description on project create but not project edit.

johnclary (approved):
looks great and works great! since you're out-of-electricity i made a tiny patch to rework the edit event text.

🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

mddilley (approved):
Tried all things personnel, and it works like a dream! 🙌 Some cool bug fixes in here too. 👋 🐛 Seeing rich events like "Replaced team member `team member` with `team member`" really makes this activity log look so pro!!! 🚢

moped-editor/src/views/projects/projectView/ProjectActivityEntry.js
chiaberry: thank you!!!

PR #970, opened 2023-02-08
Zoom to new components after create or edit

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11241

This PR updates the map to pan and zoom to the features of signal and non-signal components when saving on create or edit. I used the same padding that was used for the zoom that happens when you click on a list item's magnifier icon. That felt right, but open to any feedback on all of this. 🙏 

## Testing
**URL to test:** 

https://deploy-preview-970--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Add a line or point component (that isn't a signal component) to a project
2. Select or draw some points or lines, drag the map away from the features, and then save
3. The map should zoom to the features in your newly created component
4. Create a signal component (PHB or Traffic) and save
5. The map should zoom to the location of the signal that you selected
6. Edit the non-signal component and save
7. The map should zoom to the updated features of the component
8. Edit the signal component by selecting a signal far from the original one
9. The map should zoom to the updated signal feature

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
@@ -14,6 +14,8 @@ import {
   UPDATE_SIGNAL_COMPONENT,
 } from "src/queries/components";
 import { knackSignalRecordToFeatureSignalsRecord } from "src/utils/signalComponentHelpers";
+import { zoomMapToFeatureCollection } from "./utils/map";
+import { fitBoundsOptions } from "./mapSettings";
 
 const useStyles = makeStyles((theme) => ({
   dialogTitle: {
@@ -25,12 +27,13 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const EditComponentModal = ({
+const EditAttributesModal = ({
   showDialog,
   editDispatch,
   componentToEdit,
   refetchProjectComponents,
   setClickedComponent,
+  mapRef,
 }) => {
   const classes = useStyles();
 
@@ -68,6 +71,7 @@ const EditComponentModal = ({
       const signalFromForm = formData.signal;
       const featureSignalRecord =
         knackSignalRecordToFeatureSignalsRecord(signalFromForm);
+
       const signalToInsert = {
         ...featureSignalRecord,
         component_id: projectComponentId,
@@ -89,7 +93,15 @@ const EditComponentModal = ({
           signals: [signalToInsert],
         },
       })
-        .then(() => onComponentSaveSuccess(updatedClickedComponentState))
+        .then(() => {
+          onComponentSaveSuccess(updatedClickedComponentState);
+          // Zoom to the new or existing signal
+          zoomMapToFeatureCollection(
+            mapRef,
+            { type: "FeatureCollection", features: [signalFromForm] },
+            fitBoundsOptions.zoomToClickedComponent
+          );
+        })
         .catch((error) => {
           console.log(error);
         });
@@ -142,4 +154,4 @@ const EditComponentModal = ({
   );
 };
 
-export default EditComponentModal;
+export default EditAttributesModal;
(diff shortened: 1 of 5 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -25,12 +27,13 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const EditComponentModal = ({
+const EditAttributesModal = ({
```
johnclary, line 30: 👍 always love to see variable names iterated on

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -89,7 +93,15 @@ const EditComponentModal = ({
...
-        .then(() => onComponentSaveSuccess(updatedClickedComponentState))
+        .then(() => {
+          onComponentSaveSuccess(updatedClickedComponentState);
+          // Zoom to the new or existing signal
+          zoomMapToFeatureCollection(
+            mapRef,
+            { type: "FeatureCollection", features: [signalFromForm] },
+            fitBoundsOptions.zoomToClickedComponent
+          );
+        })
```
johnclary, line 104: i was going to ask if there was a race condition here with the `onComponentSaveSuccess` refetch, but you're zooming straight to the map data we have in memory, which isn't changing. this is snappy!

johnclary (approved):
dang, i really like this! it makes that create/edit save feel super snappy, and it's such a huge improvement for signal components.

and it's awesome the way you were able to stitch this into the existing code with a new util — so good!

🚢 🚢 🚢 🚢 🚢

frankhereford (approved):
This is premium polish on something that is already so strong. It's so good. This is ship-ready as is!

One thing that I'd like to offer for consideration. What would you think about zooming to a feature when a user clicks edit from the left pane and then selects 'Map' as what they want to target? The idea being that when someone is going to edit an existing feature, they will require that feature to be in the viewport to interact with it. I've tried to capture this, albeit it poorly, in this video.

https://user-images.githubusercontent.com/3653206/218806088-7294aebf-1072-4806-9822-aa415ce0c198.mp4

mddilley:
@frankhereford I like that idea to zoom when editing starts! It sounds like a small lift that would fit in this PR.

@johnclary - what do you think? I could get this in today and so it would make it into tomorrow's release.

johnclary:
@frankhereford thanks for that suggestion. i lean toward wanting to preserve the map extent when someone clicks the edit button, because they may have moved the map to a specific location where they want to edit, e.g. if you were going to edit a very long bike lane. there's kind of a similar idea with clicking on the component in the sidebar, which highlights it but does not zoom to it, vs explicitly using that mag button to zoom in to a component—i guess the philosophy is to default to the user knowing what they want the map state to be. 

for the same reason, i actually had minor reservations about adding in this feature—but seeing it in action i think 99% of the time the user will be ok with that re-focus on save.

johnclary:
along those lines—two navigation improvements could be:

1. Showing the zoom-to mag icon on the component list item when editing (currently that gets hidden)
2. Adding a "Home" button the top-right corner of the map that will re-fit the entire project's components into the map extent.

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
mddilley: 🙏 Laziness wins 😆

mddilley:
@johnclary @frankhereford Thanks for the reviews and the thoughts on this. I added John's feedback to https://github.com/cityofaustin/atd-data-tech/issues/10826 so it can potentially fall under the work there when scoped. 🙏 Going to merge now

PR #973, opened 2023-02-09
Drops `moped_proj_features.feature` JSON column again

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11426

This is a revised version of #793 - which drops the unused `moped_proj_features.feature` column and updates Hasura's metadata. The previous migration did not actually fail, but left the metadata in an inconsistent state because the `feature` column was still referenced by permissions objects. 

To keep things tidy, I reverted the original migration (timestamp: `1665694663101`) from staging and reset its timestamp to now. So, right now the `feature` column does exist in the staging database, but it is completely empty b/c it was dropped via #793. Merging this PR will drop it once and for all and keep the metadata in sync.

@frankhereford i am sorry for urging that yolomerge with #793—it slipped my mind that we hadn’t actually tested this migration since you originally wrote it. that was my bad!

## Testing

You would need to test this on a copy of staging. You can login to the staging console to peep the current state of the db, or use `hasura migrate status` to inspect the state of the staging DB's migrations.

---
#### Ship list
- [x] Code reviewed 


moped-database/seeds/1602292389297_initial_seed_staging.sql
```diff
@@ -133,48 +133,6 @@ INSERT INTO public.moped_proj_contract (id, contractor, contract_number, descrip
...
 
 
---
--- Data for Name: moped_proj_entities; Type: TABLE DATA; Schema: public; Owner: moped
---
-
-
-
---
--- Data for Name: moped_proj_features; Type: TABLE DATA; Schema: public; Owner: moped
```
johnclary, line 143: we are totally done with this table!

frankhereford:
@johnclary wrote:

> urging that yolomerge

I ❤️ yolo merge - super no worries, not even in the slightest!! Thank you a ton for this PR, and I can't wait to review it this morning! Thank you for keeping everything moving forward and on the rails John - I mega-appreciate it!

frankhereford (approved):
I tested this out by pulling production and running the migrations contained herein locally. Worked real fine! Thank you for putting this together and updating it John!

second_time_around

Also, FWIW, again, super no worries on the yolo merge. If we don't break `main` every now and then, we're not doing it right!

mddilley (commented):
Tested this out with a staging dump so I think we are all covered - thanks for patching this up! Glad we have some nice docs on loading up the DB with staging and production now. They are super handy! 🧰

PR #982, opened 2023-02-16
Show signal ID in component list item

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11496

This PR adds the signal ID the component list item text. I also reworked the text styles, so that component name and subtype are always the primary text. We're going to keep refining these styles as we add more component attributes, e.g. [component phasing](https://github.com/cityofaustin/atd-data-tech/issues/11429), so we'll revisit this again soon.

## Testing
**URL to test:** [Netlify](https://deploy-preview-982--atd-moped-main.netlify.app/)

**Steps to test:**
1. Create a traffic and phb signal component.
2. Observe the signal ID and location name in the secondary list item text (muted gray)
3. Create other components.
4. Observe that the component name and subtype are now shown in the primary list item text.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -19,6 +20,31 @@ const useStyles = makeStyles((theme) => ({
...
+    listItemText.primary = componentSubtype
+      ? `${componentName} - ${componentSubtype}`
+      : componentName;
+    if (isSignalComponent(component)) {
+      const signalLocationName = component?.feature_signals?.[0]?.location_name;
+      const signalId = component?.feature_signals?.[0]?.signal_id;
+      listItemText.secondary = `${signalId}: ${signalLocationName}`;
+    }
+    return listItemText;
+  }, [component]);
```
johnclary, line 46: should i move this to a separate util module?

johnclary:
Y'all—no worries if we don't have time to squeeze this in to v1.17 today. It's one of those that was easier to just do instead of making the design specs.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -65,10 +84,7 @@ export default function ComponentListItem({
       }}
     >
       <ListItem dense button onClick={onListItemClick}>
-        <ListItemText
-          primary={listItemPrimaryText}
-          secondary={component.moped_components?.component_subtype}
-        />
+        <ListItemText {...listItemText} />
```
mddilley, line 87: What do you think about keeping the `ListItem`s the same height regardless of whether there is secondary text or not? I don't see a way to do this without CSS so this could probably go into consideration with the style refinements that you mentioned.

mddilley (approved):
I really like this update a lot, and I'm super excited to see you pack away some of the list item text handling into a custom hook. That really declutters this code. 🙏 💯 🚀

johnclary:
Thanks @mddilley for your feedback. I'm going to leave this as-is for now, bearing in mind that you're also re-working component list items in https://github.com/cityofaustin/atd-data-tech/issues/9764.

chiaberry (approved):
I like this enhancement, and the style refinements to primary text. And I appreciate every time you use useMemo, helps me get a better sense of everywhere we could be using it.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
mddilley: Moved towards this in https://github.com/cityofaustin/atd-moped/pull/982

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
mddilley: I moved these helpers to a separate util file in https://github.com/cityofaustin/atd-moped/pull/982

PR #988, opened 2023-02-28
Script to add moped user ID to old activity log entries

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/11513

Before the activity log refactor, we used to associate changes with a users's Cognito ID. However when we deactivate a user, we delete their cognito ID, so the activity log ends up saying "Unknown User"

Screenshot 2023-02-27 at 5 20 59 PM

This PR backfills all of the entries with a user's ID, which doesn't get deleted even when a user is deactivated. When I was testing against prod data locally, I found 4 cognito IDs that are not in the prod database. I went through the logs and tracked who the IDs belong to so we can backfill those in prod. 

Any missing IDs in staging will just stay as "Unknown User". 

While I was testing I found a WSOD bug on one of the projects.  (https://mobility.austin.gov/moped/projects/216?tab=activity_log) I patched that in this PR. 

## Testing
**URL to test:** 

https://deploy-preview-988--atd-moped-main.netlify.app/moped/projects/1657?tab=activity_log 
This will show Unknown User for everything, since I've removed references to the cognito ID in the code, but haven't run the script against staging. 

**Steps to test:**

Running the script locally on the local db won't find any entries to update, since locally we use the user id and not the cognito ID. You can test by downloading a copy of the prod data and running the `run.py` script in `/moped-toolbox/backfill_user_id/`

I tested it locally that way and the way you know it really worked is activities from last year render with the users name and not Unknown. 

Screenshot 2023-02-28 at 9 26 08 AM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/config/.gitignore
+++ b/moped-database/config/.gitignore
@@ -1,6 +1,5 @@
 *.*
 !.gitignore
-!primary_key_map.json
 !hasura.local.yaml
 !hasura_cluster_template.json
 !README.md
--- a/moped-editor/src/utils/activityLogFormatters/mopedProjectTypesActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedProjectTypesActivity.js
@@ -1,6 +1,7 @@
 export const formatProjectTypesActivity = (change, projectTypeList) => {
-
-  const changeIcon = <span className="material-symbols-outlined">summarize</span>
+  const changeIcon = (
+    <span className="material-symbols-outlined">summarize</span>
+  );
   const projectType =
     projectTypeList[change.record_data.event.data.new.project_type_id];
   const displayText = {
@@ -31,4 +32,10 @@ export const formatProjectTypesActivity = (change, projectTypeList) => {
       ],
     };
   }
+
+  // Fallback text for other updates. Catches old updates before database refactoring. (status_id etc)
+  return {
+    changeIcon,
+    changeText: [{ text: "Updated ", style: null }, displayText],
+  };
 };
--- a/moped-toolbox/backfill_user_id/run.py
+++ b/moped-toolbox/backfill_user_id/run.py
@@ -36,8 +36,14 @@ mutation UpdateActivityUserId($activity_id: uuid!, $updated_by_user_id: Int!) {
     updated_by_user_id
   }
 }
-
 """
+
+prod_cognito_ids = {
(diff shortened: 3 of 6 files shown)


moped-database/config/.gitignore
```diff
@@ -1,6 +1,5 @@
 *.*
 !.gitignore
-!primary_key_map.json
```
chiaberry, line 3: missed this when deleting the primary_key_map in an earlier PR

moped-editor/src/utils/activityLogFormatters/mopedProjectTypesActivity.js
```diff
@@ -31,4 +32,10 @@ export const formatProjectTypesActivity = (change, projectTypeList) => {
       ],
     };
   }
+
+  // Fallback text for other updates. Catches old updates before database refactoring. (status_id etc)
+  return {
+    changeIcon,
+    changeText: [{ text: "Updated ", style: null }, displayText],
+  };
```
chiaberry, line 40: prevents the WSOD

moped-toolbox/backfill_user_id/run.py
```diff
@@ -0,0 +1,175 @@
...
+
+    for activity in activities:
+        activity_id = activity["activity_id"]
+        updated_by_cognito_id = activity["updated_by"]
+        try:
+            user_id = cognito_lookup[updated_by_cognito_id]
+            ready.append({"activity_id": activity_id, "updated_by_user_id": user_id})
+        except KeyError:
+            # if the cognito id is not in our pool, and we are not in prod, add to unable_to_process
+            if env != "prod":
```
chiaberry, line 114: change this to local if you are testing locally with prod data

mddilley (approved):
Looks great! I tested with a prod dump locally, and I saw that there were many rows without `updated_by_user` populated before running the script, and then 0 after running the script. I clicked through many projects and each one had an activity full of usernames. The WSOD bug is gone on the project that you linked as well. 🙌 🎉 🚀

chiaberry:
I'm going to merge this so the WSOD bug is fixed in the next release, but will hold off running the script until I get another pair of eyes on it.

johnclary:
thanks @chiaberry!!!

PR #992, opened 2023-03-08
Store ecapris subproject ID as text instead of number

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11653

As it turns out, eCapris Subproject IDs often end with a trailing zero, e.g. `12440.010`. It's taken a while for this problem to fully surface, because:
1. Postgres actually lets you store [decimals with a trailing zero](https://www.migops.com/blog/handling-trailing-zeros-with-numeric-datatype-in-postgresql/).
2. The [Socrata data](https://data.austintexas.gov/Transportation-and-Mobility/Transportation-Capital-Project-Financial-Codes/jega-nqf6) we use to connect projects to funding numbers is sourced from a database view which also had the trailing zeroes truncated. (That has since been corrected effective March 8 2023).

This PR:
1. Changes the `ecapris_subproject_Id` column from `numeric` to `text`. Postgres kindly handles the re-casting automatically. 
2. Fixes existing records in the DB which are missing their trailing zero.
3. Drops the `capitally_funded` column, which we were not actually using.
4. Updates the editor code to properly handle `ecapris_subproject_id` as text.

## Testing
**URL to test:** Local - ideally with a copy of production.

**Steps to test:**
1. Create a new project
2. From the summary tab, set the **eCAPRIS Subproject ID** field to `12440.010`
3. Refresh the page. Observe that the subproject ID correctly renders with the trailing zero.
4. Try editing the subproject ID to have four digits after the decimal and/or a letter. You should get an error:
```
Invalid eCapris value: 12440.0100 must not contain letters and have exactly three digits after the decimal place. E.g., 12680.010.
```
5. Navigate to the "Funding" tab. Add a new funding source by clicking **Add Funding Source** and then select the**From eCapris** dropdown item.
6. In the modal that appears, you will see the funding source for **Pleasant Valley at Lighthouse PHB**. Add it to the project.

You did it!

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -3662,7 +3662,6 @@
       permission:
         check: {}
         columns:
-          - capitally_funded
```
johnclary, line 3665: i decided to remove this column from the DB. we're not using anywhere and can generate it on the fly (if eCapris project id is not null then it is capitally funded).

chiaberry: 🙏🏼

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -182,7 +180,7 @@ const ProjectSummaryProjectECapris = ({
         {!editMode && (
           <ProjectSummaryLabel
             text={
-              (isValidNumber(eCapris) && (
```
chiaberry, line 185: was this check before to remove invalid ecapris values from being shown? and now all the ecapris are valid?

johnclary: i'm really not sure! maybe it was preventing `undefined` from being rendered?

johnclary: to answer your question—all ecapris values were always valid in the sense that nothing would break if they were rendered. we actually have a handful of ids that are not "valid" according to the new regex implemented here. so those will still be rendered but the user would need to edit the value in order to save it again.

mddilley (approved):
Tested this out locally, the steps work perfectly, and I got some affirmation for the end of the day! 🙌 ❤️
> You did it!
 
Thanks for untangling this and giving us a TIL in Slack along the way. 🙏 🚢

chiaberry (approved):

PR #994, opened 2023-03-14
Drop unused moped_project columns and financial tables

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/10954
- https://github.com/cityofaustin/atd-data-tech/issues/11721

## Testing
**URL to test:** Local

**Steps to test:**
First, you're testing to make sure the app behaves normally.

1. Start the hasura cluster
2. Verify the project list loads
3. Verify the project summary and all other tabs load normally
4. Try updating a few fields on the project summary tab - make sure your changes are landing in the activity log
5. Create a new project and verify that it loads - double check its activity log

There is also a migration which copies the old `work_assignment_id` column into new `moped_proj_contract` records. To test this, you'll need to start the cluster with a copy of production. After applying migrations, you can:

1. Connect to the Moped DB with a SQL client and run

```sql
select * from moped_proj_contract order by id desc;
```

2. Observe that the first 33 records have null values except for the `work_assignment_id` column. Those are the newly created contract records.

3. These inserts will **not** show up in the activity log, because the event mutation is failing because of an error with `updated_by_user_id` . You can inspect the failing mutations by inspecting the `request` and `response` columns returned from:

```sql
select * from hdb_catalog.event_invocation_logs where trigger_name = 'activity_log_moped_proj_contract' order by created_at desc;
```

Here's an example error:

```json
{
  "data": {
    "body": "{\"errors\":[{\"extensions\":{\"code\":\"constraint-violation\",\"path\":\"$.selectionSet.insert_moped_activity_log_one.args.object\"},\"message\":\"Foreign key violation. insert or update on table \\\"moped_activity_log\\\" violates foreign key constraint \\\"moped_activity_log_updated_by_user_id_fkey\\\"\"}]}",
    "headers": [
      { "name": "Transfer-Encoding", "value": "chunked" },
      { "name": "Date", "value": "Tue, 14 Mar 2023 16:05:30 GMT" },
      { "name": "Server", "value": "Warp/3.3.23" },
      { "name": "Content-Type", "value": "application/json; charset=utf-8" }
    ],
    "status": 200
  },
  "type": "webhook_response",
  "version": "2"
}
```

As you can see, we [use a fallback user ID](https://github.com/cityofaustin/atd-moped/blob/abd890d1f8f7af826d8124c19e01f487bb1cf610/moped-database/metadata/tables.yaml#L1903) value of `1` - this value works locally (it's our default JD user), but we do not have a user ID value of `1` in production. 

I'm going to think about that problem but I think it's ok for now that these activity events will be missing from the log. Right?

---
#### Ship list
- [ ] Code reviewed 


chiaberry (approved):
Thanks for all the cleanup. I only ever saw the error messages in the hasura console, so its nice to know how to see them with sql / dbeaver. 

Since we don't have a user with the ID of 1 in production, should the fallback user id be something else? Or should we add a user with ID 1 in production....

frankhereford (approved):
Thanks for cleaning up these crufty columns, John. I've been through your code and tried to break the app, but it seems solid as a rock. This is ship-shape! 🚀

I've been thinking about Chia's question and your response about user ID #1 being a catch all. 

I think defining a user role to act as a catch-all for system events makes sense. It lets the database constraint work as-is, and would keep us from having to write special cases in the editor to display something when there isn't a human to attribute an action to.

Normally, I'm hesitant to want to imbue a particular ID with special meaning or to see the [magic number](https://github.com/cityofaustin/atd-moped/blob/abd890d1f8f7af826d8124c19e01f487bb1cf610/moped-database/metadata/tables.yaml#L1903) in the `tables.yaml` file, so the database nerd in me is wanting to see `null` used to represent an action without a user in the `updated_by_user_id` column. 

But in this case, it makes sense. The "moped administrator" is a user-like entity, and as such, I'm totally 👍 on seeing them in the users table.

As the tiniest aside, the `down.sql` migration for the `drop_fiscal_year_tables` is missing it's EOF linefeed, but hasura doesn't mind, and it's totally not a big deal if it stays.

johnclary:
Hmm, that's a great point. We do already allow `null` values in that column, and that would make for a pretty clear indicator that backend process was the initiator of the change event. I believe we would need to address the way we render a `null` user ID in the activity log UI, but that would be a very quick fix.

johnclary:
I'm going to go ahead and merge because we have so many metadata edits here and resolving conflicts is not fun. Reviews are still welcome and I'm happy to open another PR with any revisions.

PR #1001, opened 2023-03-21
update moped-api packages

chiaberry:
## Associated issues

dependabot PRs: https://github.com/cityofaustin/atd-moped/pull/980 & https://github.com/cityofaustin/atd-moped/pull/969

After some trial and error! https://github.com/cityofaustin/atd-moped/actions/runs/4483671158
It builds and the healthcheck works. https://moped-api-test.austinmobility.io/

## Testing

I tested the `/users/` endpoint which GETs all the users
https://moped-api-test.austinmobility.io/users/
You need to include the Authorization header, log into moped staging and get your Bearer token from the network tab as the value. 

A GET request to https://moped-api-test.austinmobility.io/auth/current_user with the same header as above returns your information

I'm wondering if this needs more testing before its merged to main, testing more than just GET requests. The moped-api-test deployment uses the staging cognito pool and dynamodb table. 

---
#### Ship list
- [ ] Code reviewed 


johnclary (approved):
thanks for figuring this out 😵 

i am comfortable merging this so long as the health checks work.

with future package updates in mind, should i make a separate issue to get these requirements files pruned down to only the requirements that we need to explicitly pin? these files are the output of `pip freeze`, which is akin to committing a `package-lock.json` file into `package.json`. 

after a quick scan of all the imports, it seems like these are  the packages of concern:

```
zappa
flask_cognito
flask
flask_cors
werkzeug
cryptography
```

mddilley (approved):
I just tested getting all of the users from the Cognito user pool from this test endpoint, and it worked perfectly! If that works, then I don't see why anything else wouldn't so I agree that this is ready to go.

I hadn't thought about the `moped-api-test` deployment being a usable remnant from the old test instance until now. It is the perfect tool for this kind of work, and I could see this taking a lot longer if we relied solely on the Prefect flow for tasks like this. 🙌

PR #1006, opened 2023-03-30
Add new potential and construction subphases

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/11859

## Testing

Database changes, so test locally

**Steps to test:**

spin up database and editor, there are three new subphases with ids 25, 26, 27. 

checking the https://localhost:3000/moped/dev/lookups lookup page you can see one new one in potential phase and two new ones in the construction phase

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1680199250609_add_new_subphases/down.sql
+++ b/moped-database/migrations/1680199250609_add_new_subphases/down.sql
@@ -1,3 +1,3 @@
-DELETE FROM moped_subphases WHERE subphase_name = 'Requested by partner';
-DELETE FROM moped_subphases WHERE subphase_name = 'Fiber/Communications';
-DELETE FROM moped_subphases WHERE subphase_name = 'Detection/CCTV';
+DELETE FROM moped_subphases WHERE subphase_name = 'Requested by partner' and subphase_id = 25;
+DELETE FROM moped_subphases WHERE subphase_name = 'Detection/CCTV' and subphase_id = 26;
+DELETE FROM moped_subphases WHERE subphase_name = 'Fiber/Communications' and subphase_id = 27;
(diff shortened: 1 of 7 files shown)


chiaberry:
Total problem on my end -- I forgot to update the tables.yaml file to remove the permissions for columns that don't exist. 

I also went ahead and removed the permissions to update or insert into the moped_subphases table, further chipping away at https://github.com/cityofaustin/atd-data-tech/issues/10525

moped-database/migrations/1680199250609_add_new_subphases/down.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM moped_subphases WHERE subphase_name = 'Requested by partner';
+DELETE FROM moped_subphases WHERE subphase_name = 'Fiber/Communications';
+DELETE FROM moped_subphases WHERE subphase_name = 'Detection/CCTV';
```
mddilley, line 3: These might need an additional condition for their `related_phase_id` since subphase name isn't unique from row to row. Like how `Study in progress` is the value of two rows in this table.

Probably never going to run this down migrations but just in case.

mddilley (approved):
This worked for me perfectly when I tested it out locally. I see the new subphases, and I'm happy to see these unused columns ✌️out. 🚢  🙌  🚀

roseeichelmann (approved):
test steps worked! 🚢🚢

tillyw (approved):
works perfectly! I see the new subphases in the lookup table and in the dropdown when I select the corresponding phase.

moped-editor/src/queries/project.js
```diff
@@ -250,7 +250,7 @@ export const TIMELINE_QUERY = gql`
       phase_id
       phase_name
       phase_order
-      moped_subphases(order_by: { subphase_order: asc }) {
```
johnclary, line 253: @chiaberry would you sort this query and the ones below by `subphase_name` instead? that way the dropdown menu's will render the options consistently.

moped-database/migrations/1680199250609_add_new_subphases/down.sql
johnclary: good catch @mddilley. agreed we will never run this migration but it's probably best to have it right.

johnclary (changes requested):
looking very good! just a few tiny changes. 

it's awesome to see this feature come together! let's hold off on merging until AMD has a chance to review.

johnclary (approved):
🚢

PR #1020, opened 2023-04-17
Fix file upload ampersand bug

chiaberry:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/FileUpload/FileUpload.js
+++ b/moped-editor/src/components/FileUpload/FileUpload.js
@@ -83,9 +83,12 @@ const FileUpload = props => {
    * @return {Promise<boolean>}
    */
   const handleBeforeAdd = item => {
+    // strip out & because of file upload bug
+    // https://github.com/cityofaustin/atd-data-tech/issues/11900#issuecomment-1505701452
+    const filteredFilename = item.filename.replace("&", "")
     return fetch(
       withQuery(`${config.env.APP_API_ENDPOINT}/files/request-signature`, {
-        file: item.filename,
+        file: filteredFilename,
         ...(props?.projectId ? { project_id: props.projectId } : {}),
         ...(props?.uploadType ? { type: props.uploadType } : {}),
       }),
@@ -110,7 +113,7 @@ const FileUpload = props => {
                 setErrors([]);
                 if (data?.credentials) {
                   const newFileSignatureState = { ...fileSignatures };
-                  newFileSignatureState[item.filename] = data.credentials;
+                  newFileSignatureState[filteredFilename] = data.credentials;
                   setFileSignatures(newFileSignatureState);
                 }
               })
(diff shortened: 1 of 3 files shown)


moped-editor/src/components/FileUpload/FileUpload.js
```diff
@@ -252,7 +248,7 @@ const FileUpload = props => {
   };
 
   return (
-    <Grid fullWidth>
+    <Grid fullwidth>
```
chiaberry, line 251: fixing a warning in the console

moped-editor/src/components/FileUpload/FileUpload.js
```diff
@@ -274,13 +270,7 @@ const FileUpload = props => {
         }}
         beforeAddFile={handleBeforeAdd}
         onprocessfile={(error, file) => handleFileAdded(error, file)}
-        onupdatefiles={setFiles}
-      >
-        {/* Update current files  */}
-        {files.map(file => (
-          <File key={file} src={file} origin="local" />
```
chiaberry, line 281: this is showing an error in the console when running locally, 
```
react_devtools_backend.js:2655 Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check your code at FileUpload.js:283.
```

files is an array of filepond file objects, not an array of our files. when I commented out this section, I saw no difference in the UI. I think this is supposed to show the files that are staged for upload as you add them to the dialog. But since we only allow users to upload one file at a time, the upload status that shows up represents the file that is added. 

tldr: this isnt doing anything but throwing an error so i removed it.

moped-editor/src/components/FileUpload/FileUpload.js
```diff
@@ -83,9 +74,12 @@ const FileUpload = props => {
    * @return {Promise<boolean>}
    */
   const handleBeforeAdd = item => {
+    // strip out & because of file upload bug
+    // https://github.com/cityofaustin/atd-data-tech/issues/11900#issuecomment-1505701452
+    const filteredFilename = item.filename.replace("&", "")
```
johnclary, line 79: @chiaberry could you replace `&` with `_` instead? that leaves the filename a bit more readable.

johnclary (approved):
looks great—works great! thank you for cleaning up this code!!

one small change to add replace with `_`.

mddilley (approved):
Uploaded a file with an & in the filename, it worked in the UI, it uploaded, and it downloaded without & in the filename. Thanks for cleaning up this code while you were at it! ✨🙏

tillyw:
i didn't get to this before it was merged but this looked good - thanks for cleaning up that code as well!

PR #1025, opened 2023-04-19
Fixes component feature editing bug

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/12059

## Testing
**URL to test:** [Netlify preview](https://deploy-preview-1025--atd-moped-main.netlify.app/)

**Steps to test:**
1. Create a bike lane component by selecting street segments.
2. Edit that component by adding additional segments and drawing lines. Confirm that you can save your edits. ✅ 
3. Repeat with a crosswalk (or other point-type component) and drawing points as well.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
@@ -30,13 +30,11 @@ export const getAllComponentFeatures = (component) => {
   });
   // Flatten array of arrays containing features from each feature table
   const allComponentFeaturesFlat = allComponentFeatures.flat();
-
   // Make features valid GeoJSON by adding type and properties attributes
   return allComponentFeaturesFlat.map((feature) => {
-    feature.properties = {
-      // we need this property to tie individual map features back to their parent component
-      project_component_id: component.project_component_id,
-    };
+    feature.properties ??= {};
+    // we need this property to tie individual map features back to their parent component
+    feature.properties.project_component_id = component.project_component_id;
     feature.type = "Feature";
     return feature;
   });


moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
```diff
@@ -30,13 +30,11 @@ export const getAllComponentFeatures = (component) => {
...
   // Flatten array of arrays containing features from each feature table
   const allComponentFeaturesFlat = allComponentFeatures.flat();
-
   // Make features valid GeoJSON by adding type and properties attributes
   return allComponentFeaturesFlat.map((feature) => {
-    feature.properties = {
-      // we need this property to tie individual map features back to their parent component
-      project_component_id: component.project_component_id,
-    };
+    feature.properties ??= {};
```
johnclary, line 35: Makes sure not to overwrite existing properties, which will be present on edited features.

I'm still not totally sure how this working before—it obviously was!—but here's the bit of code throwing me off:
https://github.com/cityofaustin/atd-moped/blob/724bb200f672e57f60c07c2f6ca4ffad3a01710b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js#L85-L89

I'd like to spend more time tracing the data flow, but this patch definitely brings us back to the right state so I'm going to leave it.

chiaberry (approved):
edited and saved my bike lane and crosswalks! 🥇

mddilley (approved):
Tested this out with the deploy preview, and it works great! Thanks for fixing this up, and thanks @tillyw for catching this! 🙏🚢

roseeichelmann (approved):
test steps worked for me! 🚀

johnclary:
> I think this has to do with the code before this improvement not mapping through the features. It just returned them as-is so we didn't have to worry about making sure the existing properties came through.
> 
> That line above is puzzling, and it appears to discount any properties coming through when displaying component features on the map which could be improved. From what I remember, we don't rely on feature properties for anything. 

thanks @mddilley - that helps and makes sense wrt to why the previous code was working normally. i also think it makes total sense that we weren't surfacing some of these properties into the map feature since we didn't need them. we can extend this down the road so i think we're good to go 😅

PR #1028, opened 2023-04-27
Restores localstorage column config

johnclary:
## Associated issues
Restores the local storage of project list column visibility which we lost during the MUI 5 upgrade. 

## Testing
**URL to test:** Local or Netlify

**Steps to test:**
1. On the project list, use the column visibility toggle to hide some columns.
2. Navigate away from the project list, then return to it.
3. Observe that your column visibility settings are persisted.

---
#### Ship list
- [ ] Code reviewed 


frankhereford (approved):
Code looks good after a real quick read, and this was tested locally, successfully. 🚀

chiaberry (approved):
my changes stick around! 

You can also delete the ProjectsListViewTableToolbar file since we don't need it anymore

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -88,6 +81,17 @@ const DEFAULT_HIDDEN_COLS = {
...
 
+/**
+ * Keeps localStorage column config in sync with UI interactions
+ * @param {Object} column - the MT column config with the `field` prop - aka the column name
+ * @param {Bool} hidden - the hidden state of the column
+ */
+const handleColumnChange = ({ field }, hidden) => {
+  let storedConfig = JSON.parse(localStorage.getItem("mopedColumnConfig"));
+  storedConfig = { ...storedConfig, [field]: hidden };
+  localStorage.setItem("mopedColumnConfig", JSON.stringify(storedConfig));
```
johnclary, line 92: at first glance it looked to me like this would break if the user did not already have `mopedColumnConfig` in local storage, but `getItem` will return `null` if the item is not found, and `JSON.parse(null)` returns `null`, and `{...null}` returns `{}`. ✨

johnclary:
thanks y'all!

PR #978, opened 2023-02-13
Normalize subcomponents schema

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11452

This PR adds a many <> many table between `moped_components` and `moped_subcomponents`. We will be adding many more subcomponents in the near future, so it seemed worthwhile to get this taken care of.

## Testing
**URL to test:** Local

**Steps to test:**
1. Create new components with subcomponents. Signals of subtype PHB, Traffic, and RRFB  should have three subcomponent options: Audible push button, Bicycle signal, Leading pedestrian interval. Bike lane components (except **Ramp to Sidewalk** and **Advisory**) have Raised pavement markers.

2. Edit components by adding and removing subcomponents.

3. If you're feeling thorough, you can run this against a copy of production. I have done this and the data looks goood.

4. You can peep the components data dictionary (https://localhost:3000/moped/dev/lookups) —it should match staging:

Component ID | Component name | Component subtype | Geometry type | Subcomponents
-- | -- | -- | -- | --
5 | Bike Box |   | Point |  
7 | Bike Lane | Buffered | Line | Raised pavement markers
8 | Bike Lane | Climbing | Line | Raised pavement markers
9 | Bike Lane | Colored | Line | Raised pavement markers
10 | Bike Lane | Contraflow | Line | Raised pavement markers
11 | Bike Lane | Protected One-Way | Line | Raised pavement markers
12 | Bike Lane | Protected Two-Way | Line | Raised pavement markers
21 | Bike Lane | Turn Lane | Line | Raised pavement markers
22 | Bike Lane | Wide Curb Lane | Line | Raised pavement markers
6 | Bike Lane |   | Line | Raised pavement markers
20 | Bike Parking | Corral | Point |  
19 | Bike Parking |   | Point |  
16 | Signal | PHB | Point | Audible push button, Bicycle signal, Leading pedestrian interval
17 | Signal | RRFB | Point | Audible push button, Bicycle signal, Leading pedestrian interval
18 | Signal | Traffic | Point | Audible push button, Bicycle signal, Leading pedestrian interval

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1032, opened 2023-05-01
Fixes subcomponent editing

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11968

This change unbreaks [a previous patch](https://github.com/cityofaustin/atd-moped/pull/978/files#diff-be66300e240d2abb87784e0db4a1a85c1c9040251d43a82a17500c9091b4f63c) which is preventing users from selecting/editing subcomponents.

This feels a wee bit hacky, since it only works because we do not allow users to edit their component's type/subtype. If that were not the case this would be a lot harder!

## Testing
**URL to test:** [Netlify](https://deploy-preview-1032--atd-moped-main.netlify.app/)

**Steps to test:**
1. Start creating a new bike lane component and select the "Raised pavement markers" subcomponent
2. Before you save, switch the component type to Traffic signal.
3. Observe that the subcomponent input is cleared ("Raised pavement markers" is removed)
4. Now select the "Bicycle signal" subcomponent and save.
5. Edit the component again to remove the bicycle signal subcomponent and add an Audible push button subcomponent.
6. Refresh your page and open the component editor one more time to confirm your subcomponent selection is persisted/correct

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -53,7 +53,7 @@ const validationSchema = yup.object().shape({
 /**
  * Return a Date object from a string date
  * @param {string} value - the string formatted date
- * @returns 
+ * @returns
  */
 const parseDate = (value) => {
   if (value) {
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
works on the netlify deploy, tried the same steps on staging and no dice. 

✅

mddilley (approved):
Tested on staging and the preview, and I see it working now! 🚀 Thanks for adding the comment about this working by not being able to update the component type too. I could see that regressing without the heads up. 🙏

johnclary:
thanks y'all!

johnclary:
thanks y'all!

PR #1034, opened 2023-05-02
Validate project website and store null when a user removes a project website

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12158
https://github.com/cityofaustin/atd-data-tech/issues/12159

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/test/urls.js
+++ b/moped-editor/src/utils/test/urls.js
@@ -1,7 +1,7 @@
 import { isValidUrl } from "../urls";
 
 describe("isValidUrl()", () => {
-  it("returns a full name string from a user object", () => {
+  it("validates a url using Yup", () => {
     const httpAddress = "http://www.austintexas.gov";
     const httpsAddress = "https://www.austintexas.gov";
     const wwwOnlyAddress = "www.austintexas.gov";
--- a/moped-editor/src/utils/test/urls.js
+++ b/moped-editor/src/utils/test/urls.js
@@ -1,0 +1,15 @@
+import { isValidUrl } from "../urls";
+
+describe("isValidUrl()", () => {
+  it("returns a full name string from a user object", () => {
+    const httpAddress = "http://www.austintexas.gov";
+    const httpsAddress = "https://www.austintexas.gov";
+    const wwwOnlyAddress = "www.austintexas.gov";
+    const justLetters = "fjdksaljfdlksa";
+
+    expect(isValidUrl(httpAddress)).toBe(true);
+    expect(isValidUrl(httpsAddress)).toBe(true);
+    expect(isValidUrl(wwwOnlyAddress)).toBe(false);
+    expect(isValidUrl(justLetters)).toBe(false);
+  });
+});
--- a/moped-editor/src/utils/urls.js
+++ b/moped-editor/src/utils/urls.js
@@ -2,6 +2,7 @@ import * as yup from "yup";
 
 /**
  * Validate a url using yup
+ * @see https://github.com/jquense/yup/blob/1086aa93fdd08a554936a409c30788058dbc7c32/src/string.ts#L24-L26
  * @param {string} url - url to validate
  * @returns {Boolean} is the url valid?
  */
--- a/moped-editor/src/utils/urls.js
+++ b/moped-editor/src/utils/urls.js
@@ -1,5 +1,12 @@
 import * as yup from "yup";
 
+/**
+ * Validate a non-empty string url using yup
+ * @param {string} url - url to validate
+ * @returns {Boolean} is the url valid?
+ */
 export const isValidUrl = (url) => {
+  if (url === "") return false;
+
   return yup.string().url().isValidSync(url);
 };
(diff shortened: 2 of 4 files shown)


frankhereford (approved):
As is, this is already way over the finish line. This is great. Here's an idea, and pleeeeease disregard it if it's not helpful.

What would you think of prepending `https://` to a URL if it doesn't match `/^http/i`? The idea here being to give people a way to enter a site without having to get the right protocol part of the URL, sort of how people say URLs?

As an aside, I like how this is strict about the URL having to be a for a HTTP/HTTPS resource. I can't sneak `irc://irc.fef.net` in there or anything like that.

moped-editor/src/utils/urls.js
```diff
@@ -0,0 +1,11 @@
+import * as yup from "yup";
+
+/**
+ * Validate a url using yup
+ * @see https://github.com/jquense/yup/blob/1086aa93fdd08a554936a409c30788058dbc7c32/src/string.ts#L24-L26
```
roseeichelmann, line 5: thank you for including the repo link/the line to the specific logic in this comment, it makes it super easy to go and checkout how yup works!

roseeichelmann (approved):
the code looks good and it works great! im glad you pulled in the yup package to use. this is one of those smaller changes that actually makes such a huge difference/improvement in both the UI and the back-end with the handling of null values instead of empty strings. i would support a popup for an example of a valid url for users to reference if they arent sure whats going wrong, but definitely something that could be addressed in a diff issue if its decided that would be helpful.

chiaberry (approved):
thanks for adding the validation and the test!

I think either adding helper text or Frank's idea to prepend https:// to the urls would be helpful, I can foresee someone putting in www.someurl.com and wondering why the input is still erroring.

mddilley:
Thank y'all for the suggestions! I agree that some kind of prompt is helpful. I know that the last thing I ever want on the user end is seeing red with no indication of what I'm doing wrong. 🙏

moped-editor/src/utils/test/urls.js
```diff
@@ -0,0 +1,15 @@
...
+    const httpsAddress = "https://www.austintexas.gov";
+    const wwwOnlyAddress = "www.austintexas.gov";
+    const justLetters = "fjdksaljfdlksa";
+
+    expect(isValidUrl(httpAddress)).toBe(true);
+    expect(isValidUrl(httpsAddress)).toBe(true);
+    expect(isValidUrl(wwwOnlyAddress)).toBe(false);
+    expect(isValidUrl(justLetters)).toBe(false);
+  });
+});
```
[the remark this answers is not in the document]
johnclary: +1 🚀 🙌

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectWebsite.js
```diff
@@ -97,9 +97,15 @@ const ProjectSummaryProjectWebsite = ({
               id="moped-project-website"
               label={null}
               onChange={handleProjectWebsiteChange}
-              value={website} />
+              value={website}
+              error={!isWebsiteValid}
```
johnclary, line 101: would you also add the `helperText` prop here? it can say "Website is not a valid URL" when there's an error.

johnclary (changes requested):
@mddilley this works perfectly. 

bit of scope creep, but i think it would be more intuitive to handle a user input that may not have `http(s)://`.  i'm wondering what you think about adding some special parsing to get this behavior:

1. If the URL is valid or `null`, save it
2. If the URL can be made valid by prefixing the input with `"https://"` - do that and save it.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectWebsite.js
mddilley: Perfect - I will add that! 🙏

mddilley:
@johnclary That is a really great idea to try to make it valid, and the discussion here was pushing in that direction too. I don't think that is a large lift at all and will surely make this more user-friendly. I'll make that update - thanks!

mddilley:
This is ready to review again. 🙏 The update makes entries like `austintexas.gov` valid as well fyi.

moped-editor/src/utils/urls.js
```diff
@@ -0,0 +1,26 @@
...
+ * Attempt to make a url valid by prepending https:// to it
+ * @param {string} url - url to try to make valid
+ * @returns {string} - valid url or original url
+ */
+export const makeUrlValid = (url) => {
+  if (url === null || url === "") return url;
+
+  if (isValidUrl(`https://${url}`)) {
+    return `https://${url}`;
+  }
```
chiaberry, line 23: i like how simple this is! i pictured a much more convoluted solution

johnclary (approved):
nice!!! love it. and thanks for adding more tests 💯 🚢 🚢 🚢 🚢 🚢

PR #1035, opened 2023-05-02
New subproject action buttons are hidden and fix console error on signal component edit

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11275

This PR fixes:
- an issue where the action buttons shown when adding a new subproject are out of view and require side scrolling to see when the browser is at smaller widths
- a tiny bug where the map complains about trying to zoom to an invalid location when a signal project is edited without updating the signal location

## Testing
**URL to test:** 

https://deploy-preview-1035--atd-moped-main.netlify.app/moped/

**Steps to test:**

**New subproject side scroll fix**
1. Go to an existing project summary view and collapse the browser width until just before the breakpoint where the summary map stacks below the summary text fields
2. Try to add a new subproject and see that the action buttons (✔️ and ✖️) are visible on the screen
3. Try adding a second subproject and you should still see the action buttons

**Signal component zoom error**
1. Open the dev tools console
5. Create a new **Signal - Traffic** component
6. Edit this component and update anything except for the **Signal** field
7. You shouldn't see an error in the console
8. Repeat this in staging
9. You should see `Error: Invalid LngLat latitude value: must be between -90 and 90` displayed in the console because the map is trying to zoom to its current lat./long.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -102,12 +102,21 @@ const EditAttributesModal = ({
...
-            { type: "FeatureCollection", features: [signalFromForm] },
-            fitBoundsOptions.zoomToClickedComponent
-          );
+          const [existingLongitude, existingLatitude] =
+            featureSignalRecord.geography.coordinates[0];
+          const [newLongitude, newLatitude] =
+            signalFromForm.geometry.coordinates;
+          const hasLocationChanged =
+            existingLongitude !== newLongitude ||
+            existingLatitude !== newLatitude;
```
mddilley, line 111: Maybe this belongs in a helper?

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
```diff
@@ -51,7 +54,7 @@ const SubprojectsTable = ({ projectId = null }) => {
     {
       title: "Name",
       field: "project_name",
-      width: "45%",
+      width: "35%",
```
mddilley, line 57: This is the only change in the file. The rest is prettier formatting. ✨

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
frankhereford: Do you think they will / are being used in other places? Take this with a huge grain of salt, but I'd wait until they are unless they are distracting or otherwise untidy in here. I think they look good, but you have a ton more experience in this code, to say the least, so please go with what you think is best.

frankhereford (approved):
The code looks great and the app tests like a dream. This is right on, thank you Mike!

chiaberry (approved):
thanks for fixing the action buttons and that tricky console error

johnclary (approved):
Yay for seeing the subproject action buttons and squashing a console error! 

🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
mddilley: I like that take so I'm going to leave it here - thanks, Frank!

mddilley:
Thanks y'all - merging 🚀

PR #1038, opened 2023-05-04
Upgrades production cognito lambda packages

johnclary:
This bumps our production cognito hook requirements. Not in the diff, but we are working around a known Lambda env limitation (discussed [here](https://github.com/pyca/cryptography/issues/6391)) by installing the `manylinux2014_x86_64` packages with  `--only-binary=:all:`.

See here: https://github.com/cityofaustin/atd-moped/blob/60a015765b6cdccdefb74354a102181540134c51/.github/workflows/aws-moped-cognito-helper.sh#L26

I did make some other changes to the github action and helper while debugging this, starting with [this commit.](https://github.com/cityofaustin/atd-moped/commit/991d73f8005024a2833ee5bcd96ca45d4bf10f20)

This is deployed on staging so you can test staging.

### DIFF
--- a/.github/workflows/atd_moped_cognito_hooks_build.yml
+++ b/.github/workflows/atd_moped_cognito_hooks_build.yml
@@ -21,6 +21,7 @@ jobs:
     steps:
       - uses: actions/setup-python@v4
         with:
+          # this python version must be compatible with aws lambda!
           python-version: "3.8"
           architecture: "x64"
       # Get the code first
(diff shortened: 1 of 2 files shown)


.github/workflows/atd_moped_cognito_hooks_build.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     steps:
       - uses: actions/setup-python@v4
         with:
+          # this python version must be compatible with aws lambda!
```
johnclary, line 24: i tried bumping to 3.11 and ran into new issues. i think we could safely go up to 3.10 but i did not test it.

chiaberry (approved):
I am able to log into staging instances so 👍🏼

mddilley (approved):
Thanks for de🐛ing and getting to the bottom of this fix. I was able to log in to staging as well. 🚢 ✨

PR #1042, opened 2023-05-08
Fix bug where Mapbox fitBounds fails with small screen widths

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12213

Mapbox's fitBounds method can fail with certain combinations of its padding parameter and browser screen widths. You can check out https://stackoverflow.com/a/76011418 for more background. This adds some handling of that failure and some logging in case we need a better fix down the line.

## Testing
**URL to test:** 

https://deploy-preview-1042--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to https://deploy-preview-1042--atd-moped-main.netlify.app/moped/projects/1933
2. Go the summary page, open dev tools, and decrease your screen width until you see a console message that looks like `fitBounds failed, trying again without padding`. I'm getting 512px or smaller on my screen
3. Go to the Map tab and do the same and you should see the message at 750px or smaller
---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
@@ -101,7 +101,20 @@ export const zoomMapToFeatureCollection = (
   if (!mapRef?.current) return;
 
   const bboxOfFeatureCollection = bbox(featureCollection);
-  mapRef.current.fitBounds(bboxOfFeatureCollection, fitBoundsOptions);
+
+  /**
+   * There is a known issue with fitBounds throwing an error with padding on smaller screens
+   * So, we try to fitBounds with padding, and if that fails, we try again without padding
+   * @see https://stackoverflow.com/a/76011418
+   * */
+  try {
+    return mapRef.current.fitBounds(bboxOfFeatureCollection, fitBoundsOptions);
+  } catch {
+    console.log(
+      "fitBounds failed, trying again without padding",
+      fitBoundsOptions
+    );
+  }
 };
 
 /**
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
```diff
@@ -101,7 +101,26 @@ export const zoomMapToFeatureCollection = (
...
+    console.log("fitBounds failed, trying again without padding", {
+      fitBoundsOptions,
+      screenWidth: window.innerWidth,
+    });
+  }
+
+  try {
+    mapRef.current.fitBounds(bboxOfFeatureCollection);
+  } catch {
+    console.log("Unable to fitBounds to feature collection");
```
johnclary, line 122: since this would be unusual, might be good to include the bboxFc or the entire `featureCollection` in the log message

johnclary (approved):
thanks @mddilley for identifying this tricky bug and patching it so quickly! i was able to test this successfully with the expected logging/fallback behavior. 

this is ready to ship to main, and i'd also like to push a `v1.22.2` patch to production since this is blocking at least one user's work. would you mind opening a PR against main with this change and version bump as well?

moped-editor/src/views/projects/projectView/ProjectComponents/utils/map.js
mddilley: Good call - I'll add the feature collection 🙏

chiaberry (approved):
🚢

PR #1045, opened 2023-05-09
Data Grid Demo

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/12166

No need to approve this PR, as there are no plans to merge it. 

## Observations

https://mui.com/x/react-data-grid/getting-started/

The very basics of DataGrid is similar to Material Table, you pass in a data array (here called rows) and define the columns.  

Out of the box DataGrid doesn't provide an easy way to add new rows to the data displayed in the table. I gave it a shot with the milestones table, but ran into the same issues with the row ids as here https://www.reddit.com/r/learnreactjs/comments/wmpvvn/cant_figure_out_how_to_add_row_to_mui_datagrid/

So this netlify deployment only has DataGrid as the projects list view table. 

"On the DataGrid, pagination is enabled by default and can't be disabled". We use our own pagination on material table because we handle the filtering and querying of data via the graphql query. We could query all of the projects and use the GridTable filtering, but we would need to figure out what to do with search. 

For this demo, I removed the limit on the query and am just getting all of the projects. 

I enabled the toolbar where you can hide and show columns, filters and change the density. you can also export or print the data. Print isnt formatted very well. GridTable pro gives the option to export as excel as well. 

You can also hide columns and filter by hovering over a column header and clicking the three dots. 

Hover over the Name column header or the Sponsor column header to see how the description works. 

## Testing

https://deploy-preview-1045--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -275,309 +277,537 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
+          to={`/moped/projects/${entry.row.project_id}`}
           state={{
             filters: Object.keys(filters).length
               ? btoa(JSON.stringify(filters))
               : false,
           }}
           className={classes.colorPrimary}
         >
-          {entry.project_name}
+          {entry.value}
```
chiaberry, line 601: entry.value is whatever the field is, but i could have also entry.row.project_name here.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -275,309 +277,537 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
-          </span>
-        ));
+      renderCell: (entry) => {
+        return entry.row.project_partner
+          ? entry.row.project_partner.split(",").map((partner) => (
+              <span key={partner} style={{ display: "block" }}>
+                {partner}
+              </span>
+            ))
+          : "-";
```
chiaberry, line 651: material table has a property called "emptyValue" that we defined, but DataGrid does not

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -607,64 +837,91 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
-                  data={data["project_list_view"]}
-                  columns={columns}
-                  title=""
-                  options={{
-                    search: false,
-                    rowStyle: {
-                      fontFamily: typography.fontFamily,
-                      fontSize: "14px",
+                <DataGrid
+                  getRowHeight={() => "auto"}
```
chiaberry, line 841: if we dont do this, the rows are all the same height. but then we have to add the padding like below

patrickm02L:
I know this is a proof of concept, but I could see adding this functionality to Moped. 

Here's my take on the UX/UI.

## Toolbar

Recommend rethinking the placement and aesthetics of the toolbar. Right now, it feels crowded and a bit disjointed from the rest of the design language of the index UI. Also, I could not get Print to work under the Export menu. 

Screen Shot 2023-05-10 at 8 36 30 AM

## Dropdown Menu

The individual menus on the columns add value. Recommend replacing icons with outlined version. "sort by ASC" and "sort by DESC" could be spelled out "ascending" and "descending"

Screen Shot 2023-05-10 at 8 45 56 AM

 The filter selection threw me off. It opens under the first column of the table. I expected it to open underneath the column from which I was accessing the dropdown menu. 


Screen Shot 2023-05-10 at 8 47 31 AM

mddilley (commented):
Thanks you so much for getting this demo out, Chia! Agree with a lot of what Patrick and John mentioned. Also fully support paying for more features instead of trying to build and maintain them over time. Having a button like "unsort" in the single column dropdown OOTB is a great example of a feature that we would have to build out otherwise. I think that this update will provide a lot of features that we haven't thought of yet and that the user group will find useful.

- hide all/show all is in the **Columns** dropdown is so handy! I really like being able to clear all columns and add from there
- being able to select a single row and highlight it is cool even though I couldn't figure out how to unselect it
- i like how the code config for the columns is similar enough to Material Table that it allowed the custom cells like badges to pop right in
- you mentioned that this uses flex the other day, and I hope that will help us adjust and keep the UI consistent across views with less effort. Make quirky stuff like this [recent subprojects controls snackoo](https://github.com/cityofaustin/atd-moped/pull/1035) easier to manage.
- I like where John is going with this grid making it easier to add new columns and not forget to set the filter config needed. I found [this reddit thread](https://www.reddit.com/r/reactjs/comments/wry9jv/comment/ikvsu0k/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) that mentions generating column definitions dynamically. Might be a cool path to check out but it would still require handling custom cell behavior like rendering badges so idk

As far as query all vs. query remote data, I found these examples in the past for using remote data/query building kind of like the GraphQL abstract does with DataGrid and material table. Not trying to argue for either approach - just trying to stir up some ideas. I also think benchmarking is a great idea - maybe it just works as is!
- [Remote data with DataGrid demo/code](https://devexpress.github.io/devextreme-reactive/react/grid/demos/featured/remote-data/)
- [Remote data example with material table](https://medium.com/@harshverma04111989/material-table-with-graphql-remote-data-approach-f05298e1d670)

johnclary:
this was a great success and i am closing it. See also https://github.com/cityofaustin/atd-moped/pull/1054

PR #1050, opened 2023-05-16
Show summary map if a project's subprojects are mapped

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12262

This PR updates the summary map to show subproject components even if the current project has no components of its own. I updated the summary map to use the same hooks and layers that the component map uses. I also updated the project summary query to fetch the components for the current project _and_ its subprojects so that we can refetch that top-level data when adding or removing subprojects in the `SubprojectTable` component. That way the summary map updates without a page refresh.

@johnclary and everyone else - I couldn't decide on whether the initial map zoom should include subprojects. The components map in the **Map** tab only zooms to components in the current project - not related projects. So, that is what I did here. Happy to update if needed!

## Testing
**URL to test:** 

https://deploy-preview-1050--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Create a new project and notice that the summary map does not appear since there are no components or subprojects associated with it
2. In the summary view, add a subproject, and you should see the summary map appear with the subproject's components displayed
3. Remove the subproject and the map should go back to the fallback view with the "Add Components" button
4. Add some components through the **Map** tab and you can see them on the summary map along with any subproject components

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/queries/components.js
+++ b/moped-editor/src/queries/components.js
@@ -41,7 +41,7 @@ export const ADD_PROJECT_COMPONENT = gql`
   }
 `;
 
-export const GET_PROJECT_COMPONENTS = gql`
+export const PROJECT_COMPONENT_FIELDS = gql`
   fragment projectComponentFields on moped_proj_components {
     project_component_id
     component_id
@@ -109,6 +109,10 @@ export const GET_PROJECT_COMPONENTS = gql`
       component_id
     }
   }
+`;
+
+export const GET_PROJECT_COMPONENTS = gql`
+  ${PROJECT_COMPONENT_FIELDS}
   query GetProjectComponents($projectId: Int!, $parentProjectId: Int = 0) {
     moped_proj_components(
       where: { project_id: { _eq: $projectId }, is_deleted: { _eq: false } }
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
@@ -151,6 +151,7 @@ const SubprojectsTable = ({ projectId = null, refetchSummaryData }) => {
           actionsColumnIndex: -1,
           tableLayout: "fixed",
           addRowPosition: "first",
+          idSynonym: "project_id",
         }}
         localization={{
           header: {
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
@@ -30,7 +30,7 @@ import {
 } from "../../../../queries/subprojects";
 import typography from "../../../../theme/typography";
 
-const SubprojectsTable = ({ projectId = null }) => {
+const SubprojectsTable = ({ projectId = null, refetchSummaryData }) => {
   const addActionRef = React.useRef();
 
   const { loading, error, data, refetch } = useQuery(SUBPROJECT_QUERY, {
@@ -177,6 +177,7 @@ const SubprojectsTable = ({ projectId = null }) => {
             })
               .then(() => {
                 refetch();
+                refetchSummaryData();
               })
               .catch((error) => console.error(error));
           },
@@ -189,6 +190,7 @@ const SubprojectsTable = ({ projectId = null }) => {
             })
               .then(() => {
                 refetch();
+                refetchSummaryData();
               })
               .catch((error) => console.error(error));
           },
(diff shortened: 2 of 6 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
```diff
@@ -177,6 +177,7 @@ const SubprojectsTable = ({ projectId = null }) => {
             })
               .then(() => {
                 refetch();
+                refetchSummaryData(); // Refresh subprojects in summary map
```
mddilley, line 180: We refetch summary data so that the data going to the summary map updates and the map updates.

moped-editor/src/queries/components.js
```diff
@@ -109,6 +109,10 @@ export const GET_PROJECT_COMPONENTS = gql`
       component_id
     }
   }
+`;
+
+export const GET_PROJECT_COMPONENTS = gql`
+  ${PROJECT_COMPONENT_FIELDS}
```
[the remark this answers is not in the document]
johnclary: cool to see a fragment in the wild!

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
johnclary: this works great! ✨

johnclary (changes requested):
@mddilley thanks for knocking this out—users are really going to appreciate it.

it looks great. i'm happy to leave the map zoom as-is. we might revisit that, but not today.

bit of a tangent, but would you mind adding the option `idSynonym: "project_id"` to the subprojects [table settings](https://github.com/cityofaustin/atd-moped/blob/e70d9fcb8be1fc3a6bef9cc7d420547cf056ad5d/moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js#L153)? i'm seeing the unique ID warning in the console.

mddilley:
@johnclary Thanks for pointing that out - just patched up that warning! 🩹

frankhereford (approved):
Mike, this is great. I'm all-in on approving here, but I do want to contribute to the question and discussion around the initial zoom bounding box for the summary map. 

I created a project with no GIS data, and when I added a subject, I did get a map, but I got this:

image

This is unclear to me if I'm missing the data on it, or if it's like super ultra micro because i'm looking at the whole city, or as it turns out to be the case, the subproject that I missed just happens to fall outside of the default extent. 

👇 Frank Opinion Time

My gut is telling me to expect the map to show me data 100% of the time that it flips from the off-state to the on-state, otherwise, it would not have flipped on. 

Because of this, and due to the summary page holding the broadest, most general information including subprojects, the map should be equally inclusive and have the subprojects shown on it. It should zoom to the minimum bounding box that contains all projects and subprojects, and the user should feel free to manipulate the viewport if they need more specific information. A map that has just set its state always should have set its most important state -- namely the viewport -- as part of that.

If you want to differentiate subprojects from the main project's components, then symbolism could be used, where you have a main color and a subproject color.  I feel less strongly on this than the viewport, but it might be interesting to consider. 

Let me know if this rambling doesn't make any sense or I can clarify. This is great work, and I think it's a super quality addition!

mddilley:
@frankhereford Thanks for the thoughts on the the map zoom. That is also unexpected that a subproject would be off-screen if the map is shown. @johnclary mentioned revisiting the initial zoom of the map down the road so I'll leave it up to him if we want to fix this now or spin an issue off. 

As for the symbolism/colors, the main project's components appear blue while the subprojects appear green like they do in the edit map. Open to any ideas around that!

chiaberry (approved):
I noticed the other day that the summary page map wasn't rendering even though there was a project and didn't know why, I'm glad you added this feature. 

Agree with Frank that it was a little weird to not see my subproject when its not in downtown Austin, but at least seeing the map render is better than seeing nothing.

roseeichelmann (approved):
This works great Mike, test steps work just as expected! I do agree about w/ Frank and Chia about the map extent showing the subprojects too. 
I also noticed some interesting behavior while testing this that is unrelated to this PR I think but captured it in this video below. When I edit a component and add a phase I can't save the completion date? Obvi to be addressed in another issue though.

[simplescreenrecorder-2023-05-22_13.45.56.webm](https://github.com/cityofaustin/atd-moped/assets/67130963/1bd419ad-49d8-4050-86d7-27856e3bd80e)

johnclary (approved):

mddilley:
thanks for taking another look @johnclary - i'll try to remember to re-request review next time 😅🙏. Thanks everyone - merging!

johnclary:
I created https://github.com/cityofaustin/atd-data-tech/issues/12356 to address the date editing issue.

PR #1057, opened 2023-05-24
[Enhancement] Ability to advanced search by interim_project_id 

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/12340

## Testing
**URL to test:** 

Test locally

**Steps to test:**

You can advanced search/filter by access ID, our seed data has one project with the interim project id of 1. he column is available on the project list view, you can togggle it hidden or shown. 

Add / edit interim project ids from a project's summary page, filter using that on the projects list page. 

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
chiaberry, line 1: i added the project_interim_id to the permissions and it decided to change all the " to ' while it was at it

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -238,6 +238,9 @@ export const ProjectsListViewQueryConf = {
     added_by: {
       type: "string",
     },
+    interim_project_id: {
+      type: "string",
```
mddilley, line 242: This works just like it is, but I noticed that `project_id` is another field that is set to type = number in `ProjectsListViewFiltersConf.js` but is set to `Int` in this file. But, I think this is a Grid Table vs. Material Table config thing so it shouldn't matter.

mddilley (approved):
Test this locally, compared the up and down migrations, and everything looks solid! Thanks for adding this to the list view. 🚢 🚀 🙌

johnclary (approved):
this works perfectly! i'm happy to see another must-have feature done for the migration. we're getting close for real!!!

PR #1065, opened 2023-06-14
Upgrade to react-hook-form v7

mddilley:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11612
- https://github.com/cityofaustin/atd-data-tech/issues/12518

This PR gets us to RHF v7 and makes some small adjustments to some of the data handling. I used [this migration guide](https://legacy.react-hook-form.com/migrate-v6-to-v7/) and the [current docs](https://react-hook-form.com/docs).

A second PR will handle the refactor of the form mentioned in the issue. **Edit: The form flashing old values is not fixed yet in this PR.**

## Testing
**URL to test:** 

https://deploy-preview-1065--atd-moped-main.netlify.app/moped/

**Steps to test:**

**Component attributes form**
1. Create a new project component (non-signal) with all of fields filled out
2. Edit the component attributes after it is created
3. Clear some attributes and save
4. Repeat these same steps but use a signal component like `Signal - PHB` this time.

**Staff form**
1. Go to https://deploy-preview-1065--atd-moped-main.netlify.app/moped/staff/new and fill out the form to create a new user in the staging user pool. For the password, try something not complex like `weak`. For the email, use your email to create something like `mike.dilley+rhftest@austintexas.gov`. 
2. The form should highlight the password field in red and not submit. Create a complex password with 1Password and submit the form again. It should redirect you to the Staff list view.
3. Find your new user and edit details about them. I added test to all of the text fields and updated the role. You can try different combinations. The password field is not required to edit a user, but you can try updating that too.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -48,10 +48,14 @@ const validationSchema = yup.object().shape({
   completionDate: yup.date().nullable().optional(),
   description: yup.string(),
   // Signal field is required if the selected component inserts into the feature_signals table
-  signal: yup.object().when("component", {
-    is: (val) => val?.data?.feature_layer?.internal_table === "feature_signals",
-    then: yup.object().required(),
-  }),
+  signal: yup
+    .object()
+    .nullable()
```
mddilley, line 53: The form validation Yup resolver is now complaining about `null` for this field when a non-signal component was being edited.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -180,11 +184,9 @@ const ComponentForm = ({
               id="signal"
               name="signal"
               control={control}
-              render={({ onChange, value, ref }) => (
+              render={({ field }) => (
                 <SignalComponentAutocomplete
-                  onChange={onChange}
-                  value={value}
-                  ref={ref}
+                  {...field}
```
mddilley, line 189: `onChange`, `value`, and `ref` are now packed into `field` which can be spread out into the component props.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -219,10 +221,9 @@ const ComponentForm = ({
         </Grid>
         <Grid item xs={12}>
           <TextField
-            inputRef={register}
+            {...register("description")}
```
mddilley, line 224: Similarly, `register` is no longer assigned as the ref and is invoked to return `ref` and other props that are spread into these components.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -300,16 +300,16 @@ const ComponentForm = ({
...
                 name="completionDate"
                 control={control}
-                render={({ onChange, value, ref }) => {
+                render={({ field }) => {
                   return (
                     <MobileDatePicker
-                      inputRef={ref}
-                      value={parseDate(value)}
+                      {...field}
+                      value={parseDate(field.value)}
```
mddilley, line 307: Since we have custom handling of the date, we can spread `field` which contains the `value` prop and then immediately override it with the custom date handling.

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -53,8 +53,8 @@ const EditAttributesModal = ({
     const { subcomponents, phase, subphase, tags } = formData;
     const completionDate = !!phase ? formData.completionDate : null;
     const description =
-      formData.description.length > 0 ? formData.description : null;
-    const srtsId = formData.srtsId.length > 0 ? formData.srtsId : null;
+      formData.description?.length > 0 ? formData.description : null;
+    const srtsId = formData.srtsId?.length > 0 ? formData.srtsId : null;
```
mddilley, line 57: The form is now returning `null` when these are blank so we need to check for this. Seems like these length checks can be removed in the form refactor since it will pass along `null` to the DB without special handling.

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -87,10 +87,9 @@ const StaffForm = ({
   const {
     register,
     handleSubmit,
-    errors,
     control,
     watch,
-    formState: { isSubmitting },
+    formState: { isSubmitting, errors },
```
mddilley, line 92: `errors` comes out of `formState` now

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -301,8 +293,15 @@ const StaffForm = ({
           <FormControl variant="standard" component="fieldset">
             <FormLabel id="roles-label">Role</FormLabel>
             <Controller
-              as={
```
mddilley, line 304: `Controller`'s `as` prop is removed in v7 so we have to refactor with the `render` prop like the other instances of `Controller` in this codebase use.

moped-editor/src/views/account/AccountView/ProfileDetails.js
mddilley, line 1: Forgot to note that I removed this file because this component isn't rendered anywhere. 🧹 It was made so that users could update their own password.

moped-editor/src/views/staff/EditStaffView.js
```diff
@@ -59,13 +59,33 @@ const EditStaffView = () => {
...
+    const payload = {
+      first_name,
+      last_name,
+      title,
+      email,
+      password,
+      roles,
+      workgroup_id,
+    };
+
```
johnclary, line 81: is this restructuring doing something important or were you just debugging here?

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -260,20 +253,19 @@ const StaffForm = ({
...
+                    {...field}
+                    variant="outlined"
+                    id="workgroup_id"
                     labelId="workgroup-label"
                     label="Workgroup"
                     disabled={!isUserActive}
                     onChange={(e) => {
-                      onChange(e.target.value ?? "");
+                      field.onChange(e.target.value ?? "");
                     }}
```
johnclary, line 268: @mddilley i noticed that the "workgroup" error message doesn't render if you leave the field blank, and i'm pretty sure that's a bug i introduced a few releases ago. if you look  down below in L286 the render logic is checking `errors.workgroup` when it should now be checking `errors.workgroup_id`.

johnclary (approved):
@mddilley this looks great! i ran through the tests locally and on main and everything checks out. it's awesome to see us take this step forward: thanks for making it look easy!

i flagged a bug that predates this PR and i'm happy to open a separate issue if it's not as simple as i hoped.

🚢 
🚢 
🚢 
🚢

moped-editor/src/views/staff/EditStaffView.js
mddilley: The form submission broke since I was lazily passing the whole user payload through the form which included the `moped_workgroup` data that the backend no longer needs. I'll take a closer look at this loose end when I check the workgroup error mentioned below. I think that we can trim down the data requested to populate this form. 🙏

moped-editor/src/views/staff/StaffForm.js
mddilley: I'll make an attempt to fix this in this PR. It doesn't seem like a big lift. 👍

moped-editor/src/views/staff/EditStaffView.js
johnclary: ah, of course, you're being explicit about what values to handle—i just misinterpreted. this looks fine to me!

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -87,17 +87,15 @@ const StaffForm = ({
...
     watch,
-    formState: { isSubmitting },
+    formState: { isSubmitting, errors },
     reset,
   } = useForm({
     defaultValues: {
       ...initialFormValues,
       // Roles are stored as an array in the DB but we need to feed the form a string
       roles: findHighestRole(initialFormValues.roles),
-      workgroup_id: initialFormValues?.moped_workgroup?.workgroup_id ?? "",
```
mddilley, line 100: Workgroup was recently normalized so we no longer need to get the nested id here with special handling. We get it from the `workgroup_id` column of the `moped_users` table. That comes in `initialFormValues` that is spread a few lines above this.

moped-editor/src/views/staff/NewStaffView.js
```diff
@@ -40,7 +40,12 @@ const validationSchema = yup.object().shape({
   first_name: yup.string().required(),
   last_name: yup.string().required(),
   title: yup.string().required(),
-  workgroup_id: yup.number().integer().required(),
+  workgroup_id: yup
+    .number()
+    .integer()
+    .required()
+    // see https://github.com/jquense/yup/issues/211
+    .typeError("workgroup is a required field"),
```
mddilley, line 48: The Yup error message that comes through here by default is not user friendly (complains about `NaN` and not about the required part) so I set a message that matches up with the other fields.

moped-editor/src/queries/staff.js
```diff
@@ -28,10 +28,7 @@ export const GET_USER = gql`
       staff_uuid
       title
       user_id
-      moped_workgroup {
-        workgroup_id
-        workgroup_name
-      }
+      workgroup_id
```
mddilley, line 31: We can request the `workgroup_id` here so that we don't have to get it from the `moped_workgroup` part of the return object which required special handling to set it in the form that I mentioned above.

moped-editor/src/views/staff/EditStaffView.js
mddilley: Cool! I kept it around for future eyes. 👀

moped-editor/src/views/staff/StaffForm.js
mddilley: @johnclary I fixed this up - wasn't bad at all. I added some more comments about the changes that I pushed today.

roseeichelmann (approved):
reviewed the code and test steps checked out! you made this upgrade look easy 🚢🚢

frankhereford (changes requested):
Hi Mike - first off, this is awesome - a refactor like this is always a big task!!

I wanted to suggest one thing, and to ask if another is an example of the behavior you are trying to eliminate.

### Password form
When the weak password warning is shown, it reads `min length is 8, value does not match regex '^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$'`. Do you think we should offer a message that's more like:

Your password must meeting the following criteria:
 * At least 8 characters
 * Include at least 
   * One capital letter
   * One lower-case letter
   * One numeral
   *  One of the following special characters: #?!@$%^&*-


Also, I think the regex is requiring a 12 character password but the form handler is asking for 8? `aA!1aA!1` isn't valid but `aA!1aA!1aA!1` is. 

### Flashing-old-value issue:

I was able to get it to do this. This isn't what you want, right?


https://github.com/cityofaustin/atd-moped/assets/3653206/cd81ac30-c351-4b8f-b0c6-2bf6465a3c1a

mddilley:
@frankhereford Thanks for the feedback. 🙏 The password feedback could definitely be improved, and there is an issue for it https://github.com/cityofaustin/atd-data-tech/issues/10085 but it hasn't been prioritized. I'll add a note about converting the API validation errors into human-readable errors in there.

And, yes, it is expected that the old values flashing bug is still in this PR. I split up the react hook form v7 upgrade and the form refactor into two PRs. The next PR is https://github.com/cityofaustin/atd-moped/pull/1068.

frankhereford (approved):
@mddilley Thank you for clarifying that the behavior is what is expected and also on the issue regarding the password feedback. I should have done more digging into those issues, and your ability to keep track of so many things at once is /amazing!/ 

This is fantastic work - I can't wait to see it 🚢🚢🚢!

chiaberry (approved):

PR #1068, opened 2023-06-23
Edit component attributes form briefly shows previous values on save

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11612

This PR updates the component attributes edit form to fix an issue where the previous values briefly flashed in the UI when saving. I included a diagram of the refactor but it comes down to updating the UI through the new edits through the same channel that the initial data comes through. This PR also pulls out all of the data prep for the form fields out of a custom hook and into individual helpers.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Through a project's Map tab, create a signal component (Signal - PHB or Signal - Traffic) and another non-signal component with all parts of the component form filled in
2. Edit these attributes and make sure the previous values don't show as the modal is closing
3. Try creating other components with different combinations of attributes filled in the form
4. Edit the components to make sure:
   - the form will not let you submit without selecting a component type
   - the form will not let you submit without selecting a signal type if you choose a signal component
   - all of the fields that you added when creating a component are shown in the form when you edit a component
   - the submit button is disabled until you make a change from the initial value of a field

change

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/DateFieldEditComponent.js
+++ b/moped-editor/src/views/projects/projectView/DateFieldEditComponent.js
@@ -9,22 +9,25 @@ import { parseISO, format } from "date-fns";
  * @constructor
  */
 
-const DateFieldEditComponent = ({ onChange, value, ...props }) => {
-  const handleDateChange = (date) => {
-    const newDate = date ? format(date, "yyyy-MM-dd") : null;
-    onChange(newDate);
-  };
+const DateFieldEditComponent = React.forwardRef(
+  ({ onChange, value, ...props }, ref) => {
+    const handleDateChange = (date) => {
+      const newDate = date ? format(date, "yyyy-MM-dd") : null;
+      onChange(newDate);
+    };
 
-  return (
-    <MobileDatePicker
-      format="MM/dd/yyyy"
-      value={value ? parseISO(value) : null}
-      onChange={handleDateChange}
-      InputProps={{ style: { minWidth: "100px" } }}
-      slotProps={{ actionBar: { actions: ["accept", "cancel", "clear"] } }}
-      {...props}
-    />
-  );
-};
+    return (
+      <MobileDatePicker
+        ref={ref}
+        format="MM/dd/yyyy"
+        value={value ? parseISO(value) : null}
+        onChange={handleDateChange}
+        InputProps={{ style: { minWidth: "100px" } }}
+        slotProps={{ actionBar: { actions: ["accept", "cancel", "clear"] } }}
+        {...props}
+      />
+    );
+  }
+);
 
 export default DateFieldEditComponent;
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -92,7 +92,7 @@ const ComponentForm = ({
 
   const componentOptions = useComponentOptions(optionsData);
   const phaseOptions = usePhaseOptions(optionsData);
-  const { component, phase } = watch();
+  const { component, phase } = watch(["component", "phase"]);
   const subphaseOptions = useSubphaseOptions(phase?.data.moped_subphases);
   const internalTable = component?.data?.feature_layer?.internal_table;
   const isSignalComponent = internalTable === "feature_signals";
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import { useQuery } from "@apollo/client";
 import { Controller, useForm } from "react-hook-form";
 import { yupResolver } from "@hookform/resolvers/yup";
@@ -72,7 +72,7 @@ const ComponentForm = ({
     control,
     watch,
     setValue,
-    formState: { isValid },
+    formState: { isDirty, isValid },
   } = useForm({
     defaultValues: initialFormValues ? initialFormValues : defaultFormValues,
     mode: "onChange",
@@ -109,7 +109,6 @@ const ComponentForm = ({
     dependentFieldName: "subphase",
     valueToSet: defaultFormValues.subphase,
     setValue,
-    valuePath: "value",
   });
 
   useResetDependentFieldOnParentChange({
@@ -117,7 +116,6 @@ const ComponentForm = ({
     dependentFieldName: "signal",
     valueToSet: defaultFormValues.signal,
     setValue,
-    valuePath: "value",
   });
 
   // reset subcomponent selections when component to ensure only allowed subcomponents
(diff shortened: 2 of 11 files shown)


moped-editor/src/views/projects/projectView/DateFieldEditComponent.js
```diff
@@ -9,22 +9,25 @@ import { parseISO, format } from "date-fns";
  * @constructor
  */
 
-const DateFieldEditComponent = ({ onChange, value, ...props }) => {
-  const handleDateChange = (date) => {
-    const newDate = date ? format(date, "yyyy-MM-dd") : null;
-    onChange(newDate);
-  };
+const DateFieldEditComponent = React.forwardRef(
```
mddilley, line 12: Forwarding the ref here cleared up a console warning.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -72,9 +72,9 @@ const ComponentForm = ({
     control,
     watch,
     setValue,
-    formState: { isValid },
+    formState: { isDirty, isValid },
```
mddilley, line 75: We couldn't use `isDirty` reliably before because the custom hook that set initial values was making the form dirty. Using `isDirty` (has a field changed from its default or initial value) along with `isValid` helps us keep the submit button inactive unless an actual change has been made.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -93,54 +93,40 @@ const ComponentForm = ({
   const { component, phase } = watch();
   const subphaseOptions = useSubphaseOptions(phase?.data.moped_subphases);
   const internalTable = component?.data?.feature_layer?.internal_table;
-  const [areSignalOptionsLoaded, setAreSignalOptionsLoaded] = useState(false);
-  const onOptionsLoaded = () => setAreSignalOptionsLoaded(true);
```
mddilley, line 97: We used to track when async signal options returned from Knack because the initial values wouldn't set correctly until after the options results returned. This is no longer needed as we prep the current signal value needed for the form in the parent component and pass it as a prop.

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -28,23 +35,16 @@ const EditAttributesModal = ({
...
-  const onComponentSaveSuccess = (updatedClickedComponentState) => {
-    // Update component list item and clicked component state to keep UI up to date
+  const onSaveSuccess = () => {
     refetchProjectComponents().then(() => {
       onClose();
-      // Update clickedComponent with the attributes that were just edited
-      setClickedComponent((prevComponent) => ({
-        ...prevComponent,
-        ...updatedClickedComponentState,
-      }));
```
mddilley, line 47: This set state call is not used anymore and is what caused the momentary flash of old values. We now:
1. Submit the edit mutation
2. Refetch the data about the updated component
3. Close the modal

This removes the effort of needing to make the form data into the shape returned by Hasura which is stored in `clickedComponent` state. Now, the updated data returns and then populates `clickedComponent` state instead of being set by the code here.

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -171,16 +144,21 @@ const EditAttributesModal = ({
...
+        description: componentToEdit.description,
+        subcomponents: makeSubcomponentsFormFieldValues(
+          componentToEdit.moped_proj_components_subcomponents
+        ),
+        signal: makeSignalFormFieldValue(componentToEdit),
+        phase: makePhaseFormFieldValue(componentToEdit.moped_phase),
+        subphase: makeSubphaseFormFieldValue(componentToEdit.moped_subphase),
+        completionDate: componentToEdit.completion_date,
+        srtsId: componentToEdit.srts_id,
+        tags: makeTagFormFieldValues(componentToEdit.moped_proj_component_tags),
```
mddilley, line 159: These helpers are all extracted from the previous `useInitialValuesOnAttributesEdit` custom hook removed later in this diff. Now, `EditAttributesModal` is in charge of making the shape of data needed for the form. The form is no longer responsible for doing all of this work.

roseeichelmann (approved):
Wow this is a lot of code mike, awesome work! Its so nice to have this detail fixed. It happens in VZE a lot and has always bothered me, maybe this can be inspiration to address it there too :p 
All the test steps worked perfectly!
I like how when you create a signal component it says "Required" under the Signal field and if you try to click continue while leaving it blank you get a popup saying "Please fill out this field." Maybe a similar thing could happen when the Component field is left blank. Currently if the Component field is left blank and you fill out another field the continue button turns blue and you can click on it but then nothing else happens and there is no indicator of why it didnt go through. Not related to this PR though, so rockets all the way 🚀🚀

chiaberry (approved):
Mike this is such a great refactor! And thank you for adding the diagrams to help us understand all the changes you made. You keep raising the bar :)

moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
```diff
@@ -22,6 +23,16 @@ export const makeComponentLabel = ({ component_name, component_subtype }) => {
...
 
+/**
+ * Create component tag label out of the type and name
+ * @param {string} type The type of the component tag
+ * @param {string} name The name of the component tag
+ * @returns {string}
+ */
+export const makeComponentTagLabel = ({ type, name }) => {
+  return `${type} - ${name}`;
+};
```
johnclary, line 34: 🙌

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -93,54 +95,40 @@ const ComponentForm = ({
   const { component, phase } = watch();
```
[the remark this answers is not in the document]
mddilley: Coming right up! Thanks for pointing this out. Totally missed it in [the docs](https://react-hook-form.com/docs/useform/watch). 🙏

johnclary (approved):
@mddilley dang, this is such an improvement! i love how much simpler things become with the new field-specific hook. I tested this a lot, and everything works perfectly 🙌 

Wrt @roseeichelmann's comment:

> Currently if the Component field is left blank and you fill out another field the continue button turns blue and you can click on it but then nothing else happens and there is no indicator of why it didnt go through.

Thanks for flagging that. It looks like we might just need to pass a `required` prop and `helperText` to that component. I will open a separate issue for that.

I also think we should add a loading state to the form to handle when the form is submitting. I'll open a separate issue for that as well.

awesome work, Mike! 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

mddilley:
Agreed about the loading state! 🍄 I'll merge this down so it gets in the release. 🚀 Also, thanks @roseeichelmann for catching the component field loose end.

PR #1074, opened 2023-07-11
Fixes component form so that signal autocomplete input renders

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/12810

Well, this one slipped through release testing :/ It stems from a minor change that Mike made at the end of the PR review process that I should have caught in testing. Ivonne brought it to my attention so I think we'll want to bring this forward as a patch release as well. 

## Testing
Local or Netlify (https://deploy-preview-1074--atd-moped-main.netlify.app/moped/projects)

**Steps to test:**
1. Create a new component. Select the  "Signal - Traffic" or "Signal - PHB" component
2. Observe that the autocomplete input appears and enables you to search/select a traffic signal or PHB.
3. Change the component type to something else. The autocomplete input goes away.
4. Save your signal component. Edit it. Observe that the autocomplete input is visible in the form.

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/auth/user.js
```diff
@@ -109,7 +109,6 @@ export const initializeUserDBObject = (userObject) => {
     }).then((res) => {
       // Then we parse the response
       res.json().then((resData) => {
-        // If we have any user data,  use it
```
johnclary, line 112: i'm v confused about the commit history in this PR. my local `git log` looks normal 🤷‍♂️ 

i will try rebasing after y'all review.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -92,7 +92,7 @@ const ComponentForm = ({
 
   const componentOptions = useComponentOptions(optionsData);
   const phaseOptions = usePhaseOptions(optionsData);
-  const { component, phase } = watch(["component", "phase"]);
+  const [component, phase] = watch(["component", "phase"]);
```
johnclary, line 95: this was causing `component` to always be `undefined`, and so `isSignalComponent` always `false`.

moped-editor/src/auth/user.js
mddilley: definitely a puzzler! Ignored ✅

chiaberry (approved):
tested and the autocomplete is back!

mddilley (approved):
Womp womp 🎺 Thanks for fixing this as I could have caught it too (livin' on the edge 🏎️ 😅), and glad that Ivonne raised this up. 🙏 Test steps work ✅

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
mddilley: I thought that this returning an array was missing from the docs, but the return is [clearly shown there](https://react-hook-form.com/docs/useform/watch). TIL!

johnclary: 👍

also the fact that using object destructuring on an array doesn't throw an error seems like a quintessential javascript bug 🤯

PR #1083, opened 2023-07-20
Adds DB support for work activities

johnclary:
## Associated issues
- Resolves https://github.com/cityofaustin/atd-data-tech/issues/12523

This PR renames and expands the `moped_proj_contract` table to `moped_proj_work_activity`. It does not alter the editor UI in any way other than renaming the "Contracts" section to "Work activities".

One of the user requirements is that we add tracking for who last updated each record, so I decided to add created/updated tracking fields to the table. I know that we can determine this information quite easily from the activity log, but it seem most reliable to manage this through DB triggers. Let me know what you think: does this new pattern create overhead/redundancy that is already served by the activity log?

## Testing
**URL to test:**  Local

**Steps to test:**
These steps have you create data in the previous schema and then then test the migrations.

1. Checkout/pull this branch and move the head to the commit before these changes:

```shell
$ git fetch
$ git checkout 12523-jc-work-auth-db
$ git checkout e28554e76aa6f5ddabf236af57677217833ddbe3
```

2. Start the hasura cluster and editor

3. Use the Moped editor to edit an existing contract and also create a new one. Do not edit the new one.

4. Without stopping the hasura cluster, checkout the head of this branch and apply migrations and metadata

```shell
# from atd-moped/moped-database
$ git switch - 
$ hasura migrate apply
$ hasura metadata apply
```

5. Use your SQL client to inspect the work activity (formerly contract) records you edited earlier. Make sure the values in the `created_at`, `updated_at`, `created_by`, and `updated_by` column are correct. These have been copied from the activity log.

```sql
select id, created_at, created_by_user_id, updated_at, updated_by_user_id from moped_proj_work_activity order by id desc, updated_at desc;
```

Use this query to view the corresponding data in the activity log:

```sql
select record_id, record_type, operation_type, created_at, updated_by_user_id from moped_activity_log where record_type = 'moped_proj_work_activity' order by record_id desc, created_at desc;
```

6. Hopefully that all looks great. Now use the Moped editor to create and update more work activity records. Again use your SQL client to make sure the created/updated fields are correct.

7. Lastly, you can test the down migration, and then restart the cluster to test the seeds:

```shell
$ hasura migrate apply --down 1
$  ./hasura-cluster restart
```
---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -3399,6 +3253,245 @@
...
+    - name: updated_by_user
+      using:
+        foreign_key_constraint_on: updated_by_user_id
+  insert_permissions:
+    - role: moped-admin
+      permission:
+        check: {}
+        set:
+          created_by_user_id: x-hasura-user-db-id
+          updated_by_user_id: x-hasura-user-db-id
```
johnclary, line 3281: these `set` rules assign the user ID from the session token into these columns. note also that insert/edit permissions are disabled for created by/updated by/created at/updated at.

moped-database/metadata/tables.yaml
```diff
@@ -3399,6 +3253,245 @@
...
+  event_triggers:
+    - name: activity_log_moped_proj_work_activity
+      definition:
+        delete:
+          columns: '*'
+        enable_manual: false
+        insert:
+          columns: '*'
+        update:
+          columns: '*'
```
johnclary, line 3426: i'll open a separate issue to make sure we handle these new columns in the activity log helpers.

frankhereford (approved):
I have run through your _**really excellent**_ testing instructions, and everything performed just as you've described. TIL some `git` shorthand from them too. 🙏 

> Let me know what you think: does this new pattern create overhead/redundancy that is already served by the activity log?

Take me with a grain of salt because I am not sure I recall or understand the user-story that drove this feature, but from a technical standpoint, I think it looks great. 

Very generally, I think the existing "activity log" feature brings a baseline level of service in terms of recording activities as time goes on, and if a particular feature of the application needs a logging solution that is a little bit more involved, then it is totally fine to provide it that. 

Even more, I think it's smart to build out that additional logging in its own table instead of glomming it onto the existing system.

moped-database/metadata/tables.yaml
johnclary: https://github.com/cityofaustin/atd-data-tech/issues/13000

mddilley (approved):
I followed the test instructions, and I see work activity create and edit tracking just like it should along with the new table. These migrations are packed with good stuff! 💪 I also think that this updated by user tracking use case is different enough to justify the new pattern. It is a nice combo of Hasura and postgres features. ✨

PR #1095, opened 2023-08-07
Fix bug where the aerial basemap renders on top of all other layers

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/13097

This PR sorts out the order that basemap, component, and street label layers render. ~These updates introduce some Mapbox errors in the console that I'm continuing to look at, but I wanted to get this code out to review.~ _Update: The errors are fixed now._

## Testing
**URL to test:** 

Local

**Steps to test:**
Testing this is tricky since deploy previews are not on the approved list for using the Nearmap token for the aerial imagery. It is best to look at this one locally.

1. Test locally and go to a project with components and related components (or add components and a subproject with components to an existing project)
2. Go to the project summary page and use the basemap control to show the **Aerial** layer. You should:
   - see the basemap update to aerial imagery
   - see the project's component features in blue (or related components in green) on top of the the aerial basemap
   - see the street labels on top of both the aerial basemap and the component and related features
3. Go to the project Map tab and repeat the same steps. You can also try adding new components or editing existing ones.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/BaseMapSourceAndLayers.js
mddilley, line 1: The key changes in this file are:
- adding a components placeholder layer so that all component feature layers can refer to its id with `beforeId` instead of the street labels layer id
- updating the render order of the street label, components placeholder, and aerial basemap layers so that they exist before the next one refers to the previous one's id with `beforeId`
   1. street label layer renders
   2. component placeholder renders and its `beforeId` prop is set to the street label layer id
   3. aerial basemap renders and its `beforeId` prop is set to the component placeholder layer id

johnclary (commented):
@mddilley i think this is a really clever fix, and the layers are definitely stacking correctly now ✨ 

ugh, those errors are rough 😞 what gives? 

i tried the smallest amount of debugging. first i tried putting an actual geojson feature in the placeholder layer, but that had no effect.

a few meandering thoughts:
- i wonder if there's a problem with multiple layers referencing the same `beforeId`? this doesn't seem to be the case since your approach has obviously fixed the layer ordering problem :/
- i see in [this comment](https://github.com/visgl/react-map-gl/discussions/1541#discussioncomment-2453862) that the maintainer suggests putting some state around whether or not the `beforeId` is ready. this scenario in question looks slightly different, but i know the `Map` instance has methods to check if sources and styles are loaded. 

this is frustrating—thanks for digging into it. i'm happy to rubber duck if it'd be helpful!

mddilley:
@johnclary Yeah!!! the errors were such a pain to see! 😵 I'll dig some into your suggestions, but you also made me think about adding the placeholder layers when the map first loads with the Map instance methods. Maybe this will beat out the rest of the layers that load into the map when the layer components render? I'll give it a shot. Thanks for helping me think through this!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useWatchForPlaceholderLayers.js
```diff
@@ -0,0 +1,24 @@
...
+ * @property {Boolean} hasPlaceHolderLayers - indicating if placeholder layers have loaded
+ * @property {Function} checkIfMapHasPlaceHolderLayers - callback for react-map-gl onLoaded prop
+ * to check if placeholder layers have loaded
+ */
+export const useHavePlaceholderLayersLoaded = (mapRef) => {
+  const [hasPlaceHolderLayers, setHasPlaceHolderLayers] = useState(false);
+
+  const checkIfMapHasPlaceHolderLayers = () => {
+    const hasPlaceholderLayers = Boolean(
+      mapRef?.current?.getLayer("components-placeholder")
```
mddilley, line 17: this layer name could be pulled out into a parameter, but this is pretty specific to the components placeholder layer. Maybe a change for down the road - not sure 🤷

mddilley:
@johnclary I tried watching for layers on map load. It works for the summary map, but I'm still having issues with errors in the Map tab. It relies on using `mapRefState` (which is what we already did in the summary map) vs plain old `mapRef` and is making me feel like justification is needed. The best explanation that I have is we need to load the ref into state so that we can watch its value for change. Going to keep looking at this. 🔍

mddilley:
This is ready for review again, everyone! I think the error messages that we were seeing are gone. I updated the code with one more placeholder layer so the component layers and aerial basemap aren't both targeting the same layer id.

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -363,49 +369,53 @@ export default function TheMap({
...
-        />
-        <CTNSourcesAndLayers
-          isCreatingComponent={isCreatingComponent}
-          isEditingComponent={isEditingComponent}
-          isDrawing={isDrawing}
-          linkMode={linkMode}
-          ctnLinesGeojson={ctnLinesGeojson}
-          ctnPointsGeojson={ctnPointsGeojson}
-        />
+        {hasPlaceHolderLayers && (
```
mddilley, line 372: This is where we check to make sure the component placeholder layer is present before the component layers load beneath it.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useWatchForPlaceholderLayers.js
johnclary, line 1: might as well name this file `useHavePlaceholderLayersLoaded.js`?

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -338,6 +343,7 @@ export default function TheMap({
         onClick={onClick}
         cursor={cursor}
         mapStyle={basemaps[basemapKey].mapStyle}
+        onLoad={checkIfMapHasPlaceHolderLayers}
```
johnclary, line 346: ok, so the map load event triggers the layer check and `hasPlaceHolderLayers` state update—cool.

assuming this `load` event only ever fires once, is there a situation where `checkIfMapHasPlaceHolderLayers` would not set `hasPlaceHolderLayers` to `true`? or to put it another way, would it be more clear to use a generic `hasMapLoaded` state instead?

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
johnclary: awesome that this works!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useMapRef.js
```diff
@@ -0,0 +1,17 @@
...
+export const useMapRef = () => {
+  const [mapRefState, setMapRefState] = useState(null);
+  const mapRef = useCallback((mapInstance) => {
+    if (mapInstance !== null) {
+      // Store instance as the value of current just like a ref would
+      setMapRefState({ current: mapInstance });
+    }
+  }, []);
+  return [mapRef, mapRefState];
+};
```
johnclary, line 17: i hadn't noticed this hook before. it's a big TIL for me—super slick!

johnclary (approved):
no more errors! @mddilley i'm so glad your were able to get this placeholder layer working because it's an awesome fix.

i have a question about the maybe using a generic `hasMapLoaded` state instead, but i'm not sure if i'm thinking about this correctly. curious to hear your thoughts and otherwise this is ready to ship!

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
mddilley: I did some testing, and this is a good catch. I wrote the code to specifically check for that layer, but that check isn't necessary for the errors to not occur. It's misleading and makes this look more complex than it is.

I'll make this more general and add comments about why the layers are waiting on the map to fully load before rendering component layers. ✅

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useWatchForPlaceholderLayers.js
mddilley: Oh yep, updated the name of the hook but not the file name. Thanks for catching that 🙏 I'll update this too when I make the change that I mentioned below.

chiaberry (approved):
this seems like such a tricky bug to fix, but somehow you make it look easy!

johnclary (approved):
LGTM!!!!!! 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1096, opened 2023-08-08
Adds MUG tracking to user profiles

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13274

I decided to add a "note" column as well, because there are a handful of special users in our DB that would benefit from a note. Eg. the "unknown" user ([discussion](https://austininnovation.slack.com/archives/CNUEPKLB1/p1691170753237639)).

## Testing
**URL to test:**  Local

**Steps to test the Moped user API changes:**
I'm not sure how to test the API changes. I've tried in the past to run the test fixtures but have not been able to configure the local setup. I am happy to be on standby to patch this code on staging. Any suggestions?

**Steps to test the Moped editor changes:**
1. Start the Hasura cluster and editor.
2. Navigate to the staff page and observe the new MUG member column on the staff list. Observe that Patricia is listed as "Yes" to being a MUG member.
3. Because Patricia is a non-login user, you can edit her profile without interacting with the user API. So edit Patricia's profile so that she is not a MUG member and add a note. Save.
4. Observe that the staff list reflects Patricia as a non-mug member. Click to edit her profile again to bring up the staff edit form and observe that your changes have persisted. Switch her back to a MUG member and save.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -4443,88 +4443,98 @@
       permission:
         check: {}
         columns:
-          - is_coa_staff
+          - cognito_user_id
+          - date_added
           - email
+          - first_name
+          - is_coa_staff
+          - is_deleted
+          - is_user_group_member
+          - last_name
+          - note
+          - picture
+          - roles
+          - title
           - user_id
           - workgroup_id
-          - roles
-          - first_name
-          - last_name
-          - picture
-          - title
-          - date_added
-          - cognito_user_id
-          - is_deleted
   select_permissions:
     - role: moped-admin
       permission:
         columns:
-          - is_coa_staff
+          - cognito_user_id
+          - date_added
           - email
+          - first_name
+          - is_coa_staff
+          - is_deleted
+          - is_user_group_member
+          - last_name
+          - note
+          - picture
+          - roles
+          - title
           - user_id
           - workgroup_id
-          - roles
-          - first_name
-          - last_name
-          - picture
-          - title
-          - date_added
-          - cognito_user_id
-          - is_deleted
         filter: {}
         allow_aggregations: true
     - role: moped-editor
       permission:
         columns:
-          - is_coa_staff
+          - cognito_user_id
+          - date_added
           - email
+          - first_name
+          - is_coa_staff
+          - is_deleted
+          - is_user_group_member
+          - last_name
+          - note
+          - picture
+          - roles
+          - title
           - user_id
           - workgroup_id
-          - roles
-          - first_name
-          - last_name
-          - picture
-          - title
-          - date_added
-          - cognito_user_id
-          - is_deleted
         filter: {}
         allow_aggregations: true
     - role: moped-viewer
       permission:
         columns:
-          - is_coa_staff
+          - cognito_user_id
+          - date_added
           - email
+          - first_name
+          - is_coa_staff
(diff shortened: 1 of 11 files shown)


moped-database/metadata/tables.yaml
johnclary, line 1: sorry about the diff. only observable change should be adding `is_user_group_member` and `note` permissions. admin user has insert/update/select while all other roles have select only.

moped-editor/src/views/staff/components/NonLoginUserActivationButtons.js
```diff
@@ -59,7 +60,8 @@ const NonLoginUserActivationButtons = ({
   };
 
   const handleUpdateNonLoginUser = async () => {
-    const { password, __typename, ...restOfFormValues } = formValues;
+    const { password, __typename, ...restOfFormValues } =
+      transformFormDataIntoDatabaseTypes(formValues);
```
johnclary, line 64: i uncovered a bug! this was causing non-login-user's role name to be saved as a string instead of an array when editing non-login users. this only recently surfaced because a different bug was blocking non-login-user editing. you can undo this change and test it locally.

chiaberry (approved):

moped-editor/src/views/staff/StaffForm.js
```diff
@@ -324,6 +325,55 @@ const StaffForm = ({
...
+              control={control}
+              render={({ field: { name, value, onChange } }) => (
+                <RadioGroup
+                  aria-label="roles"
+                  name={name}
+                  value={value ? "true" : "false"}
+                  onChange={(e) => {
+                    const newValue = e.target.value === "true";
+                    return onChange(newValue);
+                  }}
```
johnclary, line 344: this the RFH [transform and parse](https://react-hook-form.com/advanced-usage#TransformandParse) method where you mutate the value back and forth between the form input value and the value saved in state.

johnclary:
👋 sorry about that @KaroEngstrom!

mddilley (approved):
This all looks good to me! I also wonder about testing API changes in this project. I'll drop some thoughts in a comment in this PR. Thanks for adding this feature as it recently came up in the VZ project! The note field is a really nice touch. ✨

mddilley:
Testing the API - I've run into this issue before and here are some thoughts/downsides. The trouble always seems to come down to keeping the Moped DB, Cognito user pool, and DynamoDB all in sync.
1. Deploy a test instance
    - the test API makes updates to the test DB, staging Cognito pool, and staging DynamoDB table
    - this can leave the staging DB out of sync with the staging Cognito pool and staging DynamoDB table updates
2. Run the API locally with the local DB
   - same problem but uses the local DB instead
3. Just test it in staging and then patch if needed before a release
4. Using a tool like [LocalStack](https://github.com/localstack/localstack) could be interesting to create local AWS resources but I'm not sure how far that would get us (or if we can cover both Cognito and DynamoDB)
5. Anything else

I'm probably missing some things here, but it would be cool to brainstorm on this. We don't update this API often but I know that I stumble here too.

johnclary:
thanks @mddilley!! i think it would be really cool to get the local API running along with the test fixtures. y'all wrote those test fixtures for good reason! and that would mitigate the potential cognito/dynmo/postgres de-sync issues.

PR #1097, opened 2023-08-10
[Enhancement] Ability to set component completion date without setting a component phase

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/13120

## Testing
**URL to test:** 

https://deploy-preview-1097--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**

Add a new component. Toggle the "Use component phase" 
Only add a completion date, but leave the phase blank. 
Save.
Open the component attributes again, and check that the completion date is there, and that the toggle is selected. 

Open an existing component and repeat the steps above. Confirm the completion date saves, even if there is no phase. 

Screenshot 2023-08-10 at 2 04 45 PM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -97,7 +97,11 @@ const ComponentForm = ({
 
   const componentOptions = useComponentOptions(optionsData);
   const phaseOptions = usePhaseOptions(optionsData);
-  const [component, phase] = watch(["component", "phase"]);
+  const [component, phase, completionDate] = watch([
+    "component",
+    "phase",
+    "completionDate",
+  ]);
   const subphaseOptions = useSubphaseOptions(phase?.data.moped_subphases);
   const internalTable = component?.data?.feature_layer?.internal_table;
   const isSignalComponent = internalTable === "feature_signals";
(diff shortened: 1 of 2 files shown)


roseeichelmann (approved):
Yayyy test steps check out! Kinda unrelated but I noticed this a lot while testing -

There are a lot of projects with map features that currently have no work type, even though work type is now a required field. So when you click to edit an attribute immediately the save button is already greyed out and you cant save it. Idk if there should be some type of backfilling on features without a work type assigned?

image

johnclary (changes requested):
@chiaberry this looks + works great 🙌 

not related to your work, there is this odd behavior where you can toggle the "Use component phase" off even if the phase or completed date fields are populated.

Screenshot 2023-08-11 at 11 36 47 AM

would it be possible to just disable that toggle is there is a value for phase or completion date? that seems like the easiest/most intuitive fix to me 🤷‍♂️

johnclary:
>  Idk if there should be some type of backfilling on features without a work type assigned?

@roseeichelmann thanks for pointing this out! i will ask our users what they prefer. we can definitely backfill everything, otherwise folks will indeed need to select a work type the next time they edit a component.

mddilley (approved):
Tested this out on the preview, and it works fantastically! 🚀

johnclary (approved):
Beautiful 🙌 

🚢 🚢 🚢 🚢 🚢

PR #1098, opened 2023-08-10
Adds council districts to project features and geography views

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13311

This PR creates the mechanisms to associate projects with council districts. I'm looking forward to your feedback on this approach! Here are some implementation considerations:

1. I opted to persist council districts at the feature level instead of querying for them on the fly.

2. Because this approach persists district relationships at the feature level, we still need to do some additional computation/querying to surface council districts at the project and component level since one project has many components and one component has many features. This might warrant additional optimization.

Please also note that all of the changes to the moped editor are only temporary to help with testing. Viewing council district in the app is out of scope for this PR.

## Testing
**URL to test:** Local

**Steps to test:**

### First step: test the backfill

The migration is going to backfill council council districts for existing components, so you can test that by spinning up Moped at the beginning of this branch's history.

```shell
$ git fetch
$ git checkout 13311-jc-council-district-triggers
$ git checkout e28554e76aa6f5ddabf236af57677217833ddbe3
$ ./moped-database/hasura-cluster start
```

1. Navigate to the map tab of `Demo Project #1` and everything should look normal with two components on the map.

2. You need to manually fix the `features` table ID sequence because of a bug in the seed data. This fix is applied in my PR [here](https://github.com/cityofaustin/atd-moped/blob/b9b87bdb58351adb161b1620b12c6b847fba41a4/moped-database/seeds/1602292389297_initial_seed_staging.sql#L356-L360), but you have to do it manually because you already applied the old/buggy seeds when you started the cluster 👆 

```sql
# run this using your SQL client
SELECT
    setval('features_id_seq', (
            SELECT
                MAX(id)
                FROM features));
```

3. Now you will apply this branch's migrations.

```shell
# from atd-moped/moped-database
$ git switch - 
$ hasura migrate apply
$ hasura metadata apply
```

4. Refresh the map tab of `Demo Project #1` and observe that both components are listed as being in District 3. Nice! 

5. You can also use your SQL client or Hasura console to inspect the `project_geography` view, which has a `council_districts` column. 

### Second step: test cases

Now you will test a bunch of editing tasks while making sure that council district IDs are correct. While you do this, note that the district labels on the component map are placed on the district centroid, _so the label for district 9 is on the edge of district 1, and the label for district 1 is kinda hard to see_.

1. Create a new bike lane component. Select street segments within one council district. Save and confirm the district is listed correctly in the component list. Edit the bike lane and select additional segments in other districts.  Save and confirm the district is listed correctly in the component list.

2. Repeat those steps, but use line drawing instead/in addition to street segments.

3. Create a new Bike Box component. Select points in one or more districts, then save and confirm the districts are listed correctly. Repeat those steps using drawn points.

4. Lastly, try drawing a line and point-based component outside of any council district. Make sure that no district is associated with the component.

5. As one additional last step, test the updated seed data by completely restarting your cluster:

```shell
$ ./hasura-cluster restart
````

---
#### Ship list
- [ ] Code reviewed 

### CODE REVIEW

File: moped-database/migrations/1691605248601_council-district-trigger/up.sql [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+    WITH inserts_todo AS (
+        SELECT
+            NEW.id AS feature_id,
+            districts.council_district AS council_district_id
+        FROM
+            layer_council_district AS districts
+        WHERE
+            ST_Intersects(districts.geography, NEW.geography)
+        OR
+            ST_Crosses(districts.geography::geometry, NEW.geography::geometry)
```
johnclary, line 43: together, the combined intersects and crosses methods cover all the scenarios for lines:
- Line is contained by the polygon 
- Line is exactly tracing the edge of a polygon
- Line passes through the polygon (but has no vertices inside it)

moped-database/views/project_geography.sql
```diff
@@ -1,25 +1,46 @@
...
+  moped_components.component_id AS component_archtype_id,
+  moped_proj_components.project_component_id AS component_id,
+  moped_proj_components.is_deleted,
+  moped_project.project_name,
+  feature_layers.internal_table AS "table",
+  feature_layers.reference_layer_primary_key_column AS original_fk,
+  moped_components.component_name,
+  uniform_features.attributes,
+  uniform_features.geography,
+  uniform_features.council_districts
```
johnclary, line 16: this is the only change to this view

moped-database/views/uniform_features.sql
```diff
@@ -1,62 +1,98 @@
...
 SELECT
     feature_signals.id,
     feature_signals.component_id,
     'feature_signals'::text AS "table",
-    json_build_object(
-        'signal_id', feature_signals.signal_id, 'knack_id', feature_signals.knack_id, 'location_name', feature_signals.location_name, 'signal_type', feature_signals.signal_type) AS attributes,
-    feature_signals.geography
+    json_build_object('signal_id', feature_signals.signal_id, 'knack_id', feature_signals.knack_id, 'location_name', feature_signals.location_name, 'signal_type', feature_signals.signal_type) AS attributes,
+    feature_signals.geography,
+    districts.council_districts
```
johnclary, line 11: it seems to make sense to leave `council_districts` outside of the `attributes` object, since that object contains properties that are specific to this feature type.  this comment will hopefully make more sense if you inspect the rows returned from the `uniform_features` or `project_geography` 😅

moped-database/views/uniform_features.sql
```diff
@@ -1,62 +1,98 @@
...
-WHERE (
-    feature_signals.is_deleted = FALSE)
+    LEFT JOIN (
+        SELECT
+            d.feature_id,
+            array_agg(d.council_district_id) AS council_districts
+        FROM
+            features_council_districts d
+        GROUP BY
+            d.feature_id) districts ON districts.feature_id = feature_signals.id
```
johnclary, line 21: we're bringing each feature's council districts into an array of district IDs. i am pretty worried about the performance of this join! i think it might be running a full table scan on `features_council_districts`.

i am open to trying to optimizing it!

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
johnclary, line 1: all of the changes to `moped-editor` are temporary to help with testing. i will drop these commits.

moped-database/migrations/1691605248601_council-district-trigger/up.sql
```diff
@@ -0,0 +1,157 @@
...
+  USING GIST (geography);
+
+-- create table to hold feature-district associations
+CREATE TABLE public.features_council_districts (
+    id serial PRIMARY KEY,
+    feature_id integer NOT NULL,
+    council_district_id integer NOT NULL,
+    FOREIGN KEY ("council_district_id") REFERENCES "public"."layer_council_district" ("council_district")
+        ON UPDATE CASCADE ON DELETE CASCADE
+);
```
johnclary, line 13: note that we cannot establish a foreign key constrain from this table's `feature_id` column to the `features.id` column. This is a [limitation of using table inheritance](https://www.postgresql.org/docs/current/ddl-inherit.html#DDL-INHERIT-CAVEATS).

moped-database/migrations/1691605248601_council-district-trigger/up.sql
```diff
@@ -0,0 +1,157 @@
...
+        OR
+            ST_Crosses(districts.geography::geometry, NEW.geography::geometry)
+    )
+        INSERT INTO features_council_districts (feature_id, council_district_id)
+            SELECT * FROM inserts_todo;
+    RETURN NEW;
+    END;
+$$;
+
+CREATE OR REPLACE FUNCTION public.update_point_council_district()
```
johnclary, line 56: i opted for separate point vs line functions, although it would actually be possible to combine these using an `IF` block that checks the special `TG_TABLE_NAME` variable....

moped-editor/src/queries/components.js
johnclary, line 1: reminder that i am going to drop all changes to `moped-editor/` - this is for demo purposes only :)

moped-database/seeds/1602292389297_initial_seed_staging.sql
johnclary, line 1: sorry for the mega-diff. the seed data inserts the council district data, and it was doing so **after** we inserted the project features, so new triggers were having no affect.

this fixes that. i really regret the decision to put the council districts in the seed data. this should live in a migration. i wouldn't mind moving them 😇

johnclary:
👋 i have updated the PR instructions (see step 2 in the first testing section). thank you @chiaberry!

chiaberry (approved):
Thanks for including the temporary UI elements to help test the council districts

moped-database/migrations/1691605248601_council-district-trigger/up.sql
mddilley: This is tricky stuff! Thanks for the context and comments here. 🙏

moped-database/migrations/1691605248601_council-district-trigger/up.sql
mddilley: We've been thinking about this in VZ a lot lately too. I think separate functions is a good call since we don't have a good feel for if this logic will be recycled yet (or at least I don't 😅)

moped-database/migrations/1691605248601_council-district-trigger/up.sql
mddilley: Also TIL that `TG_OP` is available to handle different operations in triggers. 🔖

moped-database/views/project_geography.sql
mddilley: looking at the diff below - do you use a SQL formatter? I've been looking for one of those in my life and could use a rec 🙏

moped-database/views/uniform_features.sql
mddilley: agreed! doesn't feel like a good fit in attributes 👍

moped-database/views/uniform_features.sql
mddilley: hmmm not sure if I can offer more than questions here. It looks like having feature council districts in this view would impact what we could bring into `project_list_view` and `project_geography`. I think I see a workaround for `project_geography` but would this affect our ability to get council districts into the project list view? I might be on the wrong track.

mddilley (approved):
Followed the test steps and was able to add and edit components in multiple and no council districts! 🙌 The front end work really made this a lot easier to test so thanks for adding it in. 🙏 🎉

moped-database/views/project_geography.sql
johnclary: same! i still haven't found a formatter i like for VSCode, but [TablePlus](https://tableplus.com/) has a prettifier built into it's query interface, so i copypasta from that 😅

moped-database/views/uniform_features.sql
johnclary: you're right, the hasura-based approach does not solve for the project list view, and realistically i think we're just going to need this column on this this view. i was browsing some SO posts and does seem like Postgres is pretty good at optimizing this kind of join on it's own. i think it's safe to ship this and revisit it later.

johnclary:
i'm going to fix conflicts and merge this, because i'm pretty eager to see this run against staging. @frankhereford i know you've been super busy but it would be great to get your thoughts on this when you get around to it.

PR #1105, opened 2023-08-17
Deletes "Construction Inspection" work type

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13426

## Testing
**URL to test:** Local

1. Start your Moped
2. Create a new signal component. On the component attributes form, confirm that "construction inspection" is not available as a work type.
3. You can also peep the data dictionary and there should be no trace of this component work type.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1692285495049_delete_work_type_construction_inspection/up.sql
+++ b/moped-database/migrations/1692285495049_delete_work_type_construction_inspection/up.sql
@@ -1,0 +1,1 @@
+DELETE FROM moped_work_types WHERE KEY = 'construction_inspection';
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1692285495049_delete_work_type_construction_inspection/up.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM moped_work_types WHERE KEY = 'construction_inspection';
```
johnclary, line 1: typically we would not hard delete a work type: there is a `is_deleted` column. but this feature has not reached production yet 😄

johnclary:
thank you @mddilley!!!

PR #1107, opened 2023-08-17
Edit existing component type

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12534

## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -100,11 +100,11 @@ const ComponentForm = ({
   const isLineRepresentation =
     initialFormValues?.component?.data?.line_representation;
 
-  const componentOptions = useComponentOptions(optionsData);
+  const unfilteredComponentOptions = useComponentOptions(optionsData);
   const componentOptionsFilteredByLineRepresentation =
     useComponentOptionsFilteredByLineRepresentation({
       shouldFilterOptions: isEditingExistingComponent,
-      options: componentOptions,
+      options: unfilteredComponentOptions,
       isLineRepresentation,
     });
   const phaseOptions = usePhaseOptions(optionsData);
@@ -168,7 +168,11 @@ const ComponentForm = ({
           <ControlledAutocomplete
             id="component"
             label="Component Type"
-            options={areOptionsLoading ? [] : componentOptions}
+            options={
+              isEditingExistingComponent
+                ? componentOptionsFilteredByLineRepresentation
+                : unfilteredComponentOptions
+            }
             renderOption={(props, option, state) => (
               <ComponentOptionWithIcon
                 key={option.value}
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -106,7 +106,7 @@ const ComponentForm = ({
   const componentOptionsWithoutSignals = useComponentOptionsWithoutSignals(
     componentOptionsFilteredByLineRepresentation
   );
-  console.log(componentOptionsWithoutSignals);
+
   const phaseOptions = usePhaseOptions(optionsData);
   const [component, phase, completionDate] = watch([
     "component",
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
@@ -69,13 +69,6 @@ const ComponentForm = ({
   onSave,
   initialFormValues = null,
 }) => {
-  const doesInitialValueHaveSubcomponents =
-    initialFormValues?.subcomponents.length > 0;
-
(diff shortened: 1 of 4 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
```diff
@@ -55,6 +55,40 @@ export const useComponentOptions = (data) =>
...
+    if (!shouldFilterOptions) return options;
+
+    return options.filter(
+      (component) => component.data.line_representation === isLineRepresentation
+    );
+  }, [shouldFilterOptions, options, isLineRepresentation]);
+
+/**
+ * Take options returned by useComponentOptions and filters out signal components.
+ * @param {Array} options Component Autocomplete options
```
mddilley, line 81: Just realized that I've been leaning on VSCode docstring syntax highlighting here and leaving out dashes after the parameter name. 
`* @param {Array} options Component Autocomplete options`
vs.
`* @param {Array} options - Component Autocomplete options`

If this bother you, just say the word, and I'll update this.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -68,9 +70,6 @@ const ComponentForm = ({
   onSave,
   initialFormValues = null,
 }) => {
-  const doesInitialValueHaveSubcomponents =
```
mddilley, line 71: I changed this up to just watch if the currently selected component type has subcomponents since we are switching between component types now.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -87,15 +86,27 @@ const ComponentForm = ({
   const areFormErrors = Object.keys(errors).length > 0;
 
   // Get and format component and subcomponent options
-  const {
-    data: optionsData,
-    loading: areOptionsLoading,
```
mddilley, line 92: This loading state was only being used to prevent a MUI autocomplete error about it needing options. The hooks that provide the options feed an empty array now so this check isn't needed.

johnclary (changes requested):
@mddilley this is looking and working great for me! 🙌 

this is outside the scope of your work, but i do not like the way we reset subcomponent and work type options when the component type changes, and i'm starting to lean in favor of relying on the users to make the appropriate edits to other fields if they switch component types. i worry that users will not always notice when we reset the work type input to the default value, and we will cause bad data to be saved.

i mention this also because there is a bug in form validation if you follow these steps:
1. Open the edit attribute modal for a component
2. Remove component's the work type so that the input is blank. This should cause the validation to trigger because the input is required.
3. Change the component type, which will cause the work type input to populate again
4. Observe that the work type input is still failing validation despite being populated.

i'm curious what you think about this, and if we should handle it in a separate issue, but i'm leaning toward stripping out the hooks that are reseting the work type and subcomponent state based component type changes. or if it's not trivial, at least disabling those hooks when editing?

here's a screenshot of that validation bug:
Screenshot 2023-08-21 at 11 54 54 AM

mddilley:
@johnclary Thanks for catching that bug! 🙏 I feel the same way about making edits and having the form clear out data in response in a way that isn't obvious. Totally open to removing that hook altogether or disabling. Going to check it out on and see what works best! 👍

Maybe flagging those fields would be another way? Validation for subcomponent choices that are compatible with the component choice + error message that draws the user to a field to "fix" it (which should probably be a spinoff issue). I'll try to see if there are patterns around this more complex part of the form.

mddilley:
Just pushed changes so that the **subcomponents** and **work type(s)** don't reset on component change when editing. It was a small change, and it looks like we could use more complex Yup validation to make sure things "match up" based on selected choices  and their corresponding options, but it would take a little more digging/experimentation.

chiaberry:
I was running through the test steps, and got confused when i changed the component type yet the subcomponent type didnt reset, but looking at the comments and changes requested, it looks like I shouldnt expect for the subcomponent to be reset? 
Screenshot 2023-08-23 at 11 10 31 AM
I switched from driveway to bike lane, added the subcomponent, saved and then edited again to make it driveway.

mddilley:
@chiaberry - that is totally right. I need to update the test steps. Thank you!

johnclary:
@chiaberry sorry, i messaged a bit with Mike on Slack about the form behavior and didn't update any notes here. 

The desired behavior is that we do not reset the work type or subcomponents when switching component type, even if the previously selected work type or subcomponents are not compatible with the new component type. It's up to the user to make sure their data is accurate.

I will open a backlog issue to use Yup/form validation to enforce the appropriate work types and subcomponents based on component type. This should

johnclary (approved):
This is ready to ship—thank you for iterating on the behavior! 🙌 🚢 🚢 🚢 🚢 🚢 🚢

chiaberry (approved):
it works as expected!

roseeichelmann (approved):
test steps check out :ok_hand:

mddilley:
thanks y'all 🚀

PR #1115, opened 2023-08-30
Global search is affected by previous advanced search

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/7659

This PR fixes a bug where the config for the advanced search and the global search were intertwined. My leading theory is that GQLAbstract was using the same object reference for both the global search and the advanced search since the global search recycled the project list view config. So, both components were mutating the same object. Pretty gnarly! 

[Here are the lines](https://github.com/cityofaustin/atd-moped/blob/caffff6ffac648ba034f13b0163dd68f81533599/moped-editor/src/libs/GQLAbstract.js#L11-L12) where the GQLAbstract class assigns the object reference to its instance property called `initConfig` instead of creating a new reference. This remains a bug with GQLAbstract, but I didn't want to make such a large change in this PR. And, I could definitely use some backup on this theory. 🙏

**Before**:
`ProjectsListViewQueryConf` 🔗 gqlabstract config for advanced search 🔗  gqlabstract config for global search

**After**:
`ProjectsListViewQueryConf` 🔗 gqlabstract config for advanced search
`NavigationSearchQueryConf` 🔗  gqlabstract config for global search

## Testing
**URL to test:** 

https://deploy-preview-1115--atd-moped-main.netlify.app/moped

**Steps to test:**
1. Do an advanced search from the projects page like Project ID is 228 if testing locally or Project ID is 1930 if testing in the deploy preview
2. Navigate to a project summary page from a project in the search results
3. Use the global search to find other projects using a search string like demo or another project ID
4. You should see that other projects populate in the results dropdown

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
+++ b/moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
@@ -159,6 +159,12 @@ const NavigationSearchInput = ({ input404Class }) => {
     projectSearchQuery.current.config.options.useQuery
   );
 
+  /* Clear input and reset the gqlabstract 'or' config */
+  const clearSearchInput = () => {
+    setSearchTerm("");
+    projectSearchQuery.current.resetFull();
+  };
+
   // when magnifying glass icon is clicked, show search bar and initiate animation
   const handleMagClick = () => {
     showSearchInput(true);
@@ -169,9 +175,7 @@ const NavigationSearchInput = ({ input404Class }) => {
   const handleDropdownClose = () => {
     setSearchResultsAnchor(null);
     showSearchInput(false);
-    setSearchTerm("");
     setPopperEntered(false);
-    projectSearchQuery.current.resetFull();
   };
 
   // show popper results when search input gets focus
@@ -182,7 +186,7 @@ const NavigationSearchInput = ({ input404Class }) => {
 
   // initiate exiting animation when clicking outside of search bar / popper results
   const startSlideAway = () => {
-    setSearchTerm("");
+    clearSearchInput();
     setPopperEntered(false);
     toggleSlideIn(false);
   };
@@ -202,8 +206,10 @@ const NavigationSearchInput = ({ input404Class }) => {
     switch (key) {
       // On Escape key, clear the search
       case "Escape":
-        setSearchTerm("");
         setSearchResultsAnchor(null);
+        // Slide away the search bar and reset the input
+        toggleSlideIn(false);
+        clearSearchInput();
         break;
       // On Enter key, initialize the search
       case "Enter":
--- a/moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
+++ b/moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
@@ -140,6 +140,15 @@ const NavigationSearchInput = ({ input404Class }) => {
   const classes = useStyles();
   const divRef = React.useRef();
   let projectSearchQuery = new GQLAbstract(ProjectsListViewQueryConf);
+  console.log(projectSearchQuery);
+
+  // TODO: Refactor this so that the we don't reuse the saved GQLAbstract parameters (filters to ID 228)
+  // when viewing a project summary.
+  // TODO: Create a separate GQLAbstract for the project summary page?
+  // TODO: Two separate components based on view? Maybe...
+  // need to figure out how the filters are persisted between views
+  // because we still want to be filtered down when going back to the
+  // project list view
 
   // Toggle Text Input or magnifying glass
   const [searchInput, showSearchInput] = useState(false);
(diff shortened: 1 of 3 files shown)


moped-editor/src/libs/GQLAbstract.js
mddilley, line 1: These changes came from Prettier so they can be ignored.

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
```diff
@@ -139,7 +139,9 @@ const useStyles = makeStyles((theme) => ({
 const NavigationSearchInput = ({ input404Class }) => {
   const classes = useStyles();
   const divRef = React.useRef();
-  let projectSearchQuery = new GQLAbstract(ProjectsListViewQueryConf);
+  const projectSearchQuery = React.useRef(
+    new GQLAbstract(NavigationSearchQueryConf)
+  );
```
mddilley, line 144: I went with a ref here because I kept running into an issue where the component kept holding on to the previous search input value in the `projectSearchQuery` GQLAbstract instance. It seemed unintentional, but I can update to keep that behavior if needed.

Without ref:
https://github.com/cityofaustin/atd-moped/assets/37249039/66fe43b4-5d34-49b7-9d51-41f80d55aa7a

With ref:
https://github.com/cityofaustin/atd-moped/assets/37249039/a341eca8-da09-46c2-9092-efe1be54bff7

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
```diff
@@ -152,10 +154,16 @@ const NavigationSearchInput = ({ input404Class }) => {
...
 
   const [loadSearchResults, { called, loading, data }] = useLazyQuery(
-    projectSearchQuery.gql,
-    projectSearchQuery.config.options.useQuery
+    projectSearchQuery.current.gql,
+    projectSearchQuery.current.config.options.useQuery
   );
 
+  /* Clear input and reset the gqlabstract 'or' config */
+  const clearSearchInput = () => {
```
mddilley, line 162: I combined these two actions since they happen a lot. We could also change up what happens between uses of the search here too.

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
```diff
@@ -198,8 +206,10 @@ const NavigationSearchInput = ({ input404Class }) => {
     switch (key) {
       // On Escape key, clear the search
       case "Escape":
-        setSearchTerm("");
         setSearchResultsAnchor(null);
+        // Slide away the search bar and reset the input
```
mddilley, line 210: I made this change because I noticed that the search box stops responding when:
- enter a search term to populate results
- press ESC while the results are showing
- the input remains open but no longer responds to searches

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchQueryConf.js
```diff
@@ -0,0 +1,66 @@
...
+ * The Query configuration (now also including filters)
+ * @constant
+ * @type {object}
+ * @augments NavigationSearchFiltersConf
+ * @default
+ */
+export const NavigationSearchQueryConf = {
+  options: {
+    useQuery: {
+      fetchPolicy: "cache-first", // Default is "cache-first", or use "no-cache"
```
johnclary, line 11: totally irrelevant, but for whatever reason the cache policy does not seem to work for queries generated  by the abstract. we have https://github.com/cityofaustin/atd-data-tech/issues/10799 to address that.

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
johnclary: thanks for this! i agree that closing the search input should clear the results. this is an improvement over the production behavior which confusingly persists the previous search resuts! 


out of scope, but i think the ideal behavior is that clicking on a search result hides the search results menu, but does not hide or clear the search input. that way you can click back into the search input to navigate to another item in your result. i'll open a separate issue for that :)

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
johnclary: excellent—this feels right! (i watched the vides after my other comment, blelow)

johnclary (approved):
@mddilley the bug is gone! ✨ 

thanks for untangling this! it's cool to see the use of `ref` come in again to solve this—super tricky!

this looks ready to ship to me! 🚢 🚢 🚢 🚢 🚢

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
mddilley: awesome - that sounds great 👍

chiaberry (approved):
awesome! I think your theory makes a lot of sense, it confused me how and why this bug was happening. I'm so glad you untangled it.

mddilley:
thanks for the reviews - glad this will make the release 🙌

PR #1125, opened 2023-09-11
Show component location description in component list item

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/13774

## Testing
**URL to test:** 

you will have to test locally, since there are database changes

**Steps to test:**

Add a new non signal component. 
Add a location description in the form. After saving the component, the description will be visible 

Screenshot 2023-09-13 at 9 42 02 AM

For a signal component, the location description is updated with the signal's location. You should be able to edit it. 

Test changing the location description for existing components as well. There is a follow up issue to backfill the existing signal location descriptions https://github.com/cityofaustin/atd-data-tech/issues/13800

---
#### Ship list

- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -60,7 +70,9 @@ export default function ComponentListItem({
           primary={primary}
           secondary={
             <>
-              <>{secondary}</>
+              <span className={!isExpanded && classes.truncatedText}>
+                {secondary}
+              </span>
```
chiaberry, line 75: when the ListItem is clicked on, (aka expanded), the truncated style is removed so the entire description is visible.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -164,6 +169,13 @@ const ComponentForm = ({
     disable: isEditingExistingComponent,
   });
 
+  useResetDependentFieldOnSignalChange({
+    parentValue: watch("signal"),
+    dependentFieldName: "locationDescription",
+    valueToSet: signal ? getSignalOptionLabel(signal) : "",
```
johnclary, line 175: i worry that we might nullify information that a user intended to keep in this field. E.g., they pick a signal component, edit the location description, then change the component to something else.

to be safe, what if we also `watch` the `location_description` and use that in `valueToSet` instead of `""` (if it's not null)? does that make sense? but that might make the hook go loopy 🤷

johnclary (changes requested):
🙌  @chiaberry this is looks great and works great! i'm curious what you think about the changes i requested.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/form.js
```diff
@@ -387,3 +387,45 @@ export const useResetDependentFieldOnAutocompleteChange = ({
...
+}) => {
+  // Track previous value to compare new value
+  const [previousParentFormValue, setPreviousParentValue] =
+    useState(parentValue);
+    
+  // when the parent value changes, compare to previous value
+  // if it is different, reset the dependent field to its default
+  useEffect(() => {
+    // keep update from firing if the parent value hasn't changed
+    if (parentValue?.properties === previousParentFormValue?.properties) return;
```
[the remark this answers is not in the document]
chiaberry: I could change the original hook to have a comparison variable and have it be "value" for the autocompletes and "properties" for the signal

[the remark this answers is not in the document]
johnclary: that sounds good to me!

mddilley (commented):
I've tested this out and am ready to approve when ready again. I noticed two quirks that may or may not need patching (and I think are because of existing code):
- when I delete a signal component's Location Description, save, and then edit and paste text into the location description, I see the value of the Signal field populate. I can then paste over that value.
- side-scrolling happens in the component list items when there are reallllllyyyyy long values in the Description field or Location Description fields

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -132,38 +144,54 @@ const ComponentForm = ({
...
+  useResetDependentFieldOnParentFieldChange({
+    parentValue: watch("signal"),
+    dependentFieldName: "locationDescription",
+    comparisonVariable: "properties",
+    valueToSet:
+      signal && locationDescription?.length === 0
+      // if the signal exists and the locationDescription is empty, set to option label
+        ? getSignalOptionLabel(signal)
+        : locationDescription,
+    setValue,
```
chiaberry, line 192: @mddilley I changed this to prevent the behavior you were describing

what was happening before was the signal existed so it was using the option label as the description instead of whatever location description was being input into the empty field.

johnclary: thanks @chiaberry for reworking this! completely backtracking on what i asked for yesterday, i think we want the location description to always update if the signal changes, or clear/become empty if the component changes from signal to something else.

i'm not following this other problem of pasting text into the field, so hopefully rolling back to the previous behavior is not in conflict with that!

mddilley: this is partially me too! The paste bug that I noticed is definitely an edge case so, if it gets in the way of making this work the way that we want, I say punt. I'll send John a screencap that I have with some other keyboard events that don't work so great with the form so we can address them later.

johnclary: @chiaberry after looking at some of the wonkiness that Mike found, i think there's an issue with the equality check in the updated `useResetDependentFieldOnParentFieldChange` hook

This returning `false` when trying to compare two signal objects:

```javascript
parentValue?.[comparisonVariable] ===  previousParentFormValue?.[comparisonVariable]`
```

I'm still committed to seeing this one hook handle all the field inputs! I think some lodash might make this easier because you could dot chaining and make compare variable `"properties.id"`?

```javascript
get(parentValue, comparisonVariable) ===  get(previousParentFormValue, comparisonVariable)`
```

johnclary (approved):
That did the trick—this is ready to go 🙌  🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

mddilley:
@chiaberry looking at this right now and will ✅ shortly

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentForm.js
```diff
@@ -132,38 +142,54 @@ const ComponentForm = ({
...
+    comparisonVariable: "value",
     valueToSet: defaultFormValues.work_types,
     setValue,
     disable: isEditingExistingComponent,
   });
 
+  useResetDependentFieldOnParentFieldChange({
+    parentValue: watch("signal"),
+    dependentFieldName: "locationDescription",
+    comparisonVariable: "properties.id",
```
mddilley, line 184: Ah this is great! I often forget how useful lodash get can be.

mddilley (approved):
Works beautifully! 🙌 🚀 🚢

johnclary:
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1126, opened 2023-09-13
Ability to move a component to another project

mddilley:
## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/MoveComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/MoveComponentForm.js
@@ -27,7 +27,7 @@ export const useProjectOptions = (data) =>
   }, [data]);
 
 const validationSchema = yup.object().shape({
-  projectId: yup.string().required(),
+  projectId: yup.object().required(),
 });
 
 const MoveComponentForm = ({ onSave, projectId: currentProjectId }) => {
@@ -36,7 +36,7 @@ const MoveComponentForm = ({ onSave, projectId: currentProjectId }) => {
     control,
     formState: { errors },
   } = useForm({
-    defaultValues: { projectId: currentProjectId },
+    defaultValues: { projectId: null },
     mode: "onChange",
     resolver: yupResolver(validationSchema),
   });
--- a/moped-editor/src/views/projects/projectView/MoveComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/MoveComponentForm.js
@@ -4,30 +4,22 @@ import { useForm } from "react-hook-form";
 import { yupResolver } from "@hookform/resolvers/yup";
 import { Button, Grid } from "@mui/material";
 import { CheckCircle } from "@mui/icons-material";
-import { ComponentOptionWithIcon } from "./utils/form";
-import ControlledAutocomplete from "./ControlledAutocomplete";
+import ControlledAutocomplete from "./ProjectComponents/ControlledAutocomplete";
+import { PROJECT_OPTIONS } from "src/queries/project";
 
 import * as yup from "yup";
 
 const validationSchema = yup.object().shape({
-  projectComponentId: yup.string.required(),
+  projectId: yup.number().required(),
 });
 
-const ComponentForm = ({
-  formButtonText,
-  onSave,
-  initialFormValues = null,
-  projectComponentId,
-}) => {
+const MoveComponentForm = ({ onSave, projectId }) => {
   const {
-    register,
     handleSubmit,
     control,
-    watch,
-    setValue,
     formState: { isDirty, errors },
   } = useForm({
-    defaultValues: { projectComponentId },
+    defaultValues: { projectId },
     mode: "onChange",
     resolver: yupResolver(validationSchema),
   });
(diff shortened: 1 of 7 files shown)


mddilley:
@johnclary 
- updated delete modal close button to not clear clicked component so the list item stays expanded. all the other actions work this way
- redirect on selection? could be jarring

moped-editor/src/views/projects/projectView/ProjectComponents/DeleteComponentModal.js
```diff
@@ -32,15 +32,13 @@ const useStyles = makeStyles((theme) => ({
...
   clickedComponent,
-  setClickedComponent,
   setIsDeletingComponent,
   onDeleteComponent,
 }) => {
   const classes = useStyles();
 
   const onClose = () => {
     setIsDeletingComponent(false);
-    setClickedComponent(null);
```
mddilley, line 43: This prevents the click component item in the list from collapsing when closing the delete modal. All the other modals work this way.

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
```diff
@@ -73,49 +75,52 @@ const ProjectComponentsList = ({
...
-                />
-              )}
-            </ListItem>
+            <Stack
+              spacing={2}
+              direction="row"
+              justifyContent="flex-end"
+              my={1}
+              // estimating alignment with zoom ListItemSecondaryAction button
+              mr={2.5}
```
mddilley, line 84: Not great, but I didn't see a way to emulate the spacing of the zoom button above without refactoring it too.

chiaberry (commented):
this predates your work, but i noticed the other modals have `<DialogContent dividers={true}>` but the delete component modal does not. Its just more obvious now that i was clicking through all the modals

moped-editor/src/views/projects/projectView/ProjectComponents/DeleteComponentModal.js
chiaberry: i like this, thanks for making it consistent

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
```diff
@@ -1,10 +1,11 @@
 import ComponentListItem from "./ComponentListItem";
-import Button from "@mui/material/Button";
+import IconButton from "@mui/material/IconButton";
```
chiaberry, line 2: thanks for also updating these. i know we have an issue to address this, but chipping away at them in other issues is ✨

chiaberry (approved):
I agree that triggering a redirect would be too jarring, I'm trying to brainstorm if there are ways to help users navigate to the components new project (or if users would even want to do that here?) A new tab with the project feels excessive...

this is great! Cant find ways to break it

moped-editor/src/views/projects/projectView/MoveComponentForm.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+  const projectOptions = useProjectOptions(data);
+
+  error && console.error(error);
+
+  return (
+    <form onSubmit={handleSubmit(onSave)}>
+      <Grid container spacing={2}>
+        <Grid item xs={12}>
+          <ControlledAutocomplete
```
johnclary, line 58: yay this component keeps getting put to work!

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
johnclary: looks good to me!

moped-editor/src/views/projects/projectView/MoveComponentForm.js
```diff
@@ -0,0 +1,86 @@
...
+      <Grid container spacing={4} display="flex" justifyContent="flex-end">
+        <Grid item style={{ margin: 5 }}>
+          <Button
+            variant="contained"
+            color="primary"
+            startIcon={<CheckCircle />}
+            type="submit"
+            disabled={areFormErrors}
+          >
+            Save
```
johnclary, line 78: @mddilley i think think i would prefer the button label to say "Move component" just to be extra clear.

johnclary (changes requested):
It's so cool to see this come to life! 🙌 

> I removed showing the existing project in the dropdown since it is filtered from the choices and the project # is visible in the header with the modal open. It feel right to me but open to changing.

You're totally right—having the empty input makes it super clear what the user needs to do ✅ 

> For signal components, I kept the map button but disabled it to keep the UI consistent between list items. Maybe another tooltip that explains editing isn't available for signal components could help?

Great idea—yes please! I'm struggling with the wording: `Signal locations can only be changed by editing the component details`?

> I updated the delete modal close button to not clear the clicked component so the list item stays expanded when closing the delete modal. All the other modals work this way so now they all behave the same way.

✨ 

> I thought about a redirect after moving a component but that seems too jarring.

Agreed. What if we re-centered the map extent to fit the new bounds of the project? Kind of like we do when you save a new component? That might also be too jarring. It might be best for the user to see very clearly that the component has disappeared.

Alternatively or additionally, what if we showed a toast that confirms that "Component moved to Project #123"?


I've also been thinking about the activity log behavior for this kind of edit. Currently, only the destination project picks up the move component activity. So if a user moves a component to the wrong project, they're going to have a really hard time tracking it down.

I wonder if we might be able to create another event that fires only in this specific case, and it can be associated with the the old project ID. i'll scope that separately.

Screenshot 2023-09-13 at 12 10 04 PM

moped-editor/src/views/projects/projectView/MoveComponentForm.js
mddilley: cool! i like that a lot. I'll make that update. 🙏

mddilley:
@johnclary just pushed these changes:
- added the tooltip on the disabled map button for signal components (went with the suggestion above)
- keep modal open after moving component and display a link to the project it moved to

mddilley:
working on the updates described in this thread https://austininnovation.slack.com/archives/CNUEPKLB1/p1694637975089969?thread_ts=1694455435.251499&cid=CNUEPKLB1

moped-editor/src/views/projects/projectView/ProjectComponents/DeleteComponentModal.js
johnclary: ditto!

johnclary (approved):
@mddilley thanks for iterating on this a bit. it looks fantastic! i am going to ask one last design question in Slack but this is ready to ship as-is for tomorrow's release 🙌

PR #1131, opened 2023-09-15
Rework signal feature in project list view query

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13857

## Testing
**URL to test:** Local

It's going to be hard to notice the performance improvements without a large number of projects. But definitely you want to test that the list view loads normally.

1. Start yer Moped
2. Make sure the project list view loads normally.
3. Make all columns visible.
4. Try some advanced search filters

To test the query performance itself you can connect to the read replica!

1. Connect to the prod read replica
2. Run the `up` version of the view sql. (exclude the "CREATE OR REPLACE VIEW..." bit).
3. Run the `down` version of the view sql (exclude the "CREATE OR REPLACE VIEW..." bit).

I am seeing a roughly 50% speed up versus the old list view query. ~500ms vs 1000ms once the query is warm.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1694809881268_moped_project_updated_at_index/up.sql
+++ b/moped-database/migrations/1694809881268_moped_project_updated_at_index/up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX moped_project_updated_at_idx ON moped_project (updated_at);
--- a/moped-database/migrations/1694809881268_moped_project_updated_at_index/up.sql
+++ b/moped-database/migrations/1694809881268_moped_project_updated_at_index/up.sql
@@ -1,1 +1,2 @@
 CREATE INDEX moped_project_updated_at_idx ON moped_project (updated_at);
+CREATE INDEX moped_project_name_idx on moped_project (project_name);
--- a/moped-database/migrations/1694809058492_list_view_optimize/down.sql
+++ b/moped-database/migrations/1694809058492_list_view_optimize/down.sql
@@ -1,4 +1,4 @@
--- latest version 1694809058492_list_view_optimize
+-- rolls back to version 1686243614837_create_work_auth_table
 DROP VIEW project_list_view;
 
 CREATE OR REPLACE VIEW public.project_list_view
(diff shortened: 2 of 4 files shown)


moped-database/views/project_list_view.sql
```diff
@@ -47,14 +47,14 @@ AS WITH project_person_list_lookup AS (
     mp.interim_project_id,
     string_agg(DISTINCT me2.entity_name, ', '::text) AS project_partner,
     string_agg(task_order_filter.value ->> 'display_name'::text, ','::text) AS task_order_name,
-    (SELECT JSON_AGG(feature.attributes) -- this query finds any signal components and those component's features and rolls them up in a JSON blob
+    (SELECT JSON_AGG(json_build_object('signal_id', feature_signals.signal_id, 'knack_id', feature_signals.knack_id, 'location_name', feature_signals.location_name, 'signal_type', feature_signals.signal_type))
         FROM moped_proj_components components   
-        LEFT JOIN uniform_features feature
-          ON (feature.component_id = components.project_component_id)
+        LEFT JOIN feature_signals
+          ON (feature_signals.component_id = components.project_component_id)
```
johnclary, line 53: here's the important change. we used to join up the `uniform_features` view, which itself is a union of all the `feature_signal`, `feature_intersection`, etc... tables. that union is expensive because of limitations around JSON aggregation and i think also because of limitations around table inheritance indexing.

in any case—we're not targeting just the `feature_signals` table which makes our query dramatically faster.

tbd how we will handle adding additional feature tables and feature properties to the list view, if that becomes necessary....

moped-database/migrations/1694809881268_moped_project_updated_at_index/up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX moped_project_updated_at_idx ON moped_project (updated_at);
+CREATE INDEX moped_project_name_idx on moped_project (project_name);
```
johnclary, line 2: `updated_at` is the default column we use for sorting in the editor. i also figured it was wise to index on project name since we're so often searching against it.

chiaberry (approved):
running it locally with prod data, it feels so much faster! opening and closing the advanced filters goes faster too. 

thank you for working the query!

PR #1134, opened 2023-09-18
Add parent project names and subprojects to Project List View

chiaberry:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/13675
- https://github.com/cityofaustin/atd-data-tech/issues/13905

## Testing

Test Locally

**Steps to test:**

I have tested this with prod data, but you don't have to necessarily test with it. 

Screenshot 2023-09-18 at 4 47 45 PM

But you do need to have some projects that are parent projects. In the project list view, the last columns in the table are the Parent project column, that will show a link to the projects parent project, and another column that will show Yes if the project is a parent itself. 

Initially we were going to show the list of child projects, so I wrote the query to return a list of child project ids. I left the query as is, but if we want to only query if a child project exists, I can rework the query. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -534,8 +534,37 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
       field: "interim_project_id",
       hidden: hiddenColumns["interim_project_id"],
-      emptyValue: "-"
-    }
+      emptyValue: "-",
+    },
+    {
+      title: "Parent project",
+      field: "parent_project_id",
+      hidden: hiddenColumns["parent_project_id"],
```
johnclary, line 542: would you make both of these new columns hidden by default? would you also make these columns hidden by default as well?

```
construction_start_date
completion_end_date
project_tags
```

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -534,8 +534,37 @@ const ProjectsListViewTable = ({ query, searchTerm }) => {
...
+      emptyValue: "-",
+      render: (entry) => (
+        <RouterLink
+          to={`/moped/projects/${entry.parent_project_id}`}
+          state={{
+            filters: Object.keys(filters).length
+              ? btoa(JSON.stringify(filters))
```
johnclary, line 549: TIL [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/btoa)!

it looks like every cell in the table is using the same value for this link state. if you assigned this value outside the `columns` declaration you might see a speedup with rendering.

```jsx
# earlier
const linkStateFilters =  Object.keys(filters).length ? btoa(JSON.stringify(filters)) : false;


# inside the columns render definition
<RouterLink
          to={`/moped/projects/${entry.parent_project_id}`}
          state={{ filters: linkStateFilters }}>
```

johnclary (changes requested):
Looks great—this will be a nic lil improvement! I have asked for minor changes.

🙌

johnclary (approved):
🙏 🚢 🚢 🚢 🚢

johnclary:
thanks @chiaberry!

mddilley (approved):
Tested this locally, and it is great to have this info in the list view. The migrations work up and down as well. Good to go! 🚢 🚀

PR #1140, opened 2023-09-27
[Bug] CSV export is broken because commas are not escaped properly

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/13978

## Testing
**URL to test:** 

https://deploy-preview-1140--atd-moped-main.netlify.app/moped/projects

or locally with prod data

**Steps to test:**
Click download button next to search input. Check that the csv export makes sense, a good way to quickly eyeball this is by scrolling horizontally to the end and seeing that there aren't extra columns with empty headers. In prod right now, commas that are in the status update column are being interpreted as delimiters and inserting extra columns. 

It turns out that different applications can parse csvs differently, and I found that just using one set of quotes to parse commas wasn't working when I opened it in Numbers. So I uploaded it to excel online (https://www.office.com/launch/excel) and that did a better job of handling the singular set of quotes. 

**Question**

the file that is downloaded is named `project_list_view2023-09-27blahblahtodaystime.csv` Do you think we should prepend the word moped to the file name?

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -3,6 +3,7 @@ import {
   filterProjectTeamMembers,
   filterProjectSignals,
   filterTaskOrderName,
+  filterStatusUpdate,
 } from "./helpers.js";
 
 /**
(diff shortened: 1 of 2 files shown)


frankhereford (approved):
This works great. I've stuck commas in strange places, and this output renders correctly for me in Excel & with `vd`. I pulled up [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180), and this output looks to spec. 💯 🎯 

Thank you!

roseeichelmann (approved):
yay formatted perfectly in excel!

PR #1143, opened 2023-09-29
Fix unique key error

chiaberry:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/signalProjectTable/RenderSignalLink.js
+++ b/moped-editor/src/views/projects/signalProjectTable/RenderSignalLink.js
@@ -5,7 +5,7 @@ const RenderSignalLink = ({ signals }) => {
   return (
     <span>
       {signals.map((signal, index) => (
-        <React.Fragment key={signal.signal_id}>
+        <React.Fragment key={signal.component_id}>
           {signal?.knack_id ? (
             <Link
               href={`https://atd.knack.com/amd#signals/signal-details/${signal.knack_id}`}
--- a/moped-database/migrations/1695996662086_signal_component_ids/down.sql
+++ b/moped-database/migrations/1695996662086_signal_component_ids/down.sql
@@ -1,0 +1,183 @@
+DROP VIEW project_list_view;
+
+CREATE OR REPLACE VIEW public.project_list_view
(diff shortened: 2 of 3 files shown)


moped-database/views/project_list_view.sql
mddilley, line 1: Think the comment at the top of this file needs to be updated with this new migration version

moped-editor/src/views/projects/signalProjectTable/RenderSignalLink.js
```diff
@@ -5,7 +5,7 @@ const RenderSignalLink = ({ signals }) => {
   return (
     <span>
       {signals.map((signal, index) => (
-        <React.Fragment key={signal.signal_id}>
+        <React.Fragment key={signal.component_id}>
```
mddilley, line 8: Reminds me that I made a hacky workaround in another branch. 🙃 I'm going to remove that, and I'll pull in this work when it gets merged. I don't see the error come up, but I do think that `feature_signals.id` is guaranteed to be unique. I was trying to refresh myself on the columns and found some records that share `component_id` but their soft delete status prevents some from returning from the query. Could a double-check on that, though.

Screenshot 2023-09-29 at 1 19 11 PM

mddilley (commented):
Works great with prod data locally. 🎉 I have one question about the id.

moped-database/views/project_list_view.sql
chiaberry: oops, i never pushed up that commit 🙃

moped-editor/src/views/projects/signalProjectTable/RenderSignalLink.js
chiaberry: oh good call, ill switch to that instead of component id

mddilley (approved):
I see no errors with prod data locally - thanks for fixing this! 🚢 🙌

johnclary (approved):
quite a lot of work to squash a console error—thank you for doing this!!! 🙌

chiaberry:
not just a console error, could lead to rendering issues!!

PR #1144, opened 2023-09-29
Use project list view query hooks for search

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14026

## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1147, opened 2023-10-04
Use project list view query hooks for advanced search

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14027

## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -118,17 +118,26 @@ const generateEmptyField = (uuid) => {
...
+ * @param {Function} handleAdvancedSearchClose - Used to close the advanced search
+ * @param {Object} filtersConfig - The configuration object for the filters
+ * @param {Function} setSearchFieldValue - Used to set the search field value
+ * @param {Function} setSearchTerm - Used to set the search term for simple search
  * @return {JSX.Element}
  * @constructor
  */
-const GridTableFilters = ({
-  query,
-  filterState,
```
mddilley, line 128: One of the bigger changes that I made in this component was breaking `filterState` apart since it was an object that contained both the state set function and the state for filters. It also renamed them. I ended up passing those around separately using their original names.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -196,12 +204,12 @@ const GridTableFilters = ({
       const filtersNewState = { ...filterParameters };
 
       // Find the field we need to gather options from
-      const fieldIndex = query.config.filters.fields.findIndex(
+      const fieldIndex = filtersConfig.fields.findIndex(
```
mddilley, line 207: This is also a bigger change. One initializing a gqlabstract instance, this filter config was copied into the instance from the `PROJECT_LIST_VIEW_FILTERS_CONFIG` object and accessed that way. Now, the fields are retrieved from the original `PROJECT_LIST_VIEW_FILTERS_CONFIG` object.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -408,8 +415,11 @@ const GridTableFilters = ({
   const handleApplyButtonClick = () => {
     filterQuery.set("filter", btoa(JSON.stringify(filterParameters)));
     history.push(`${queryPath}?filter=${filterQuery.get("filter")}`);
-    filterState.setFilterParameters(filterParameters);
+    setFilters(filterParameters);
     handleAdvancedSearchClose();
+    // Clear simple search field in UI and state since we are using advanced search
+    setSearchFieldValue("");
+    setSearchTerm("");
```
mddilley, line 422: This the change that we discussed earlier this week in dev sync. Now, when you click search after setting some advanced filters, the search term is cleared from the UI search box and from the state that is used to build the graphql query.

moped-editor/src/components/GridTable/Search.js
```diff
@@ -72,30 +72,43 @@ const history = createBrowserHistory();
...
   const classes = useStyles();
   const queryPath = useLocation().pathname;
   const divRef = React.useRef();
 
+  /**
+   * The contents of the search box in SearchBar
+   * @type {string} searchFieldValue
+   * @function setSearchFieldValue - Sets the state of the field
+   */
+  const [searchFieldValue, setSearchFieldValue] = useState(searchTerm);
```
mddilley, line 110: I moved this state up the component tree from `SearchBar` to `Search` so that it could be passed from `SearchBar` to both `Search` and `Filters` since `Filters` needs to clear this state when the advanced search is triggered.

moped-editor/src/components/GridTable/Search.js
```diff
@@ -206,10 +219,7 @@ const Search = ({
    * Clears the simple search when switching to filters
    */
   const handleSwitchToAdvancedSearch = () => {
-    searchState.setSearchParameters({
-      value: "",
-      column: "",
-    });
```
mddilley, line 212: This seemed outdated from a time when the simple search used both a column and a search term value. The columns used in the simple search box are defined in `PROJECT_LIST_VIEW_QUERY_CONFIG.columns` so I updated this with the new state setter defined in `useSearch`.

moped-editor/src/components/GridTable/Search.js
```diff
@@ -267,9 +277,10 @@ const Search = ({
               className={classes.gridSearchPadding}
             >
               <SearchBar
-                query={query}
-                searchState={searchState}
```
mddilley, line 271: Same here as with the filter state, `searchState` was an object containing both the state and state setter for the simple search term. I broke it out into two props.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -97,127 +102,60 @@ const handleColumnChange = ({ field }, hidden) => {
...
-    if (Array.from(filterQuery).length > 0) {
-      try {
-        return JSON.parse(atob(filterQuery.get("filter")));
-      } catch {
-        return {};
-      }
-    }
-    return {};
-  }, [filterQuery]);
+const buildStatusBadge = ({ phaseName, phaseKey }) => (
```
mddilley, line 111: Just moving this outside `ProjectListViewTable` since it didn't have any dependencies inside the component.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -97,127 +102,60 @@ const handleColumnChange = ({ field }, hidden) => {
...
-
-  // Resets the value of "where" "and" "or" to empty
-  query.cleanWhere();
-
-  // For each filter added to state, add a where clause in GraphQL
-  // Advanced Search
-  Object.keys(filters).forEach((filter) => {
-    let { envelope, field, gqlOperator, value, type, specialNullValue } =
-      filters[filter];
+  /* Project list query */
```
mddilley, line 132: This is where I decided to move the hooks that build the separate pieces of the query outside of `useProjectListView`. See https://github.com/cityofaustin/atd-moped/pull/1144#discussion_r1343210020.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Stores objects storing a random id, column, operator, and value.
+   * @type {Object} filters
+   * @function setFilter - Sets the state of filters
+   * @default {if filter in url, use those params, otherwise {}}
+   */
+  const [filters, setFilters] = useState(initialFilterState);
+
+  const advancedSearchWhereString = useMemo(() => {
```
mddilley, line 38: The code in this function is copy/paste from the existing code. Only difference is that it returns a string to place in the graphql query instead of mutating data in the gqlabstract instance.

moped-editor/src/components/GridTable/Filters.js
johnclary: super minor, but you can use `Array.prototype.find()` instead of `findIndex()` to grab the field details object in one line.

johnclary: also, btw, this is a great change!

moped-editor/src/components/GridTable/Filters.js
johnclary: 👍 works great

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -605,18 +625,25 @@ const GridTableFilters = ({
               <Hidden mdDown>
                 <Grid item xs={12} md={1} style={{ textAlign: "center" }}>
                   <IconButton
-                    disabled={Object.keys(filterParameters).length === 1 && !filterComplete}
+                    disabled={
+                      Object.keys(filterParameters).length === 1 &&
+                      !filterComplete
```
johnclary, line 630: i don't know if this is helpful or necessary, but fwiw at a glance it looks like the `filterComplete` value makes more sense as something returned from `useMemo` (or not even memoized 😈 ) vs a useEffect + state var.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
johnclary: good stuff!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -97,127 +102,60 @@ const handleColumnChange = ({ field }, hidden) => {
...
-  );
+  const { filterQuery, filters, setFilters, advancedSearchWhereString } =
+    useAdvancedSearch();
 
-  const linkStateFilters = Object.keys(filters).length
-    ? btoa(JSON.stringify(filters))
-    : false;
+  const linkStateFilters = useMemo(() => {
+    return Object.keys(filters).length ? btoa(JSON.stringify(filters)) : false;
+  }, [filters]);
```
johnclary, line 158: 👍

moped-editor/src/views/projects/projectsListView/test/projectListView.test.js
johnclary, line 1: ✨ we should make use of this, huh?

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
johnclary: it's very cool to see this in action!

since it has so much logic, you might move this out into a separate function like you did with other pieces of this hook.

johnclary (approved):
> f you test stacking the advanced and simple searches in the opposite order then the conditions in the query do both apply

interesting! yep, this is definitely pre-existing weirdness. i am comfortable leaving as-is for now.

@mddilley this works just like i expect to! this code is soooo much nicer than the old abstract. i think we're good to go here! 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
mddilley: yes, also think that would be for the best. Making this update right now 🙏

moped-editor/src/components/GridTable/Filters.js
mddilley: thx - and good catch! I'll update that to condense it down. 🚀

chiaberry (approved):
a lot that you had to untangle ✅

PR #1149, opened 2023-10-06
Adds length column to line features

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14155

## Testing
**URL to test:**  Local

1. Create a project component using a mix of selected street segments and drawn lines. 
2. Use google maps (or your favorite mapping platform) to measure the approximate distance of your components.
Screenshot 2023-10-06 at 5 01 07 PM

3. Compare those measurements to the length stored in the DB. This query will sum up `length_feet` from all line features in a component:

```sql
SELECT
    project_id,
    project_component_id,
    description,
    sum(length_feet) AS total_component_length
FROM (
    SELECT
        project_id,
        mpc.project_component_id,
        mpc.description,
        length_feet
    FROM
        feature_street_segments
    LEFT JOIN moped_proj_components mpc ON mpc.project_component_id = feature_street_segments.component_id
WHERE
    feature_street_segments.is_deleted = FALSE
UNION ALL
SELECT
    project_id,
    mpc.project_component_id,
    mpc.description,
    length_feet
FROM
    feature_drawn_lines
    LEFT JOIN moped_proj_components mpc ON mpc.project_component_id = feature_drawn_lines.component_id
WHERE
    feature_drawn_lines.is_deleted = FALSE) uniform_length
GROUP BY
    project_id,
    project_component_id,
    description;
```

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1696622988854_add_feature_length_column/up.sql
+++ b/moped-database/migrations/1696622988854_add_feature_length_column/up.sql
@@ -1,7 +1,7 @@
 ALTER TABLE feature_street_segments
     ADD COLUMN length_feet INTEGER GENERATED always AS (
-Round(ST_Length(geography) * 3.28084)) stored;
+ST_Length(geography) * 3.28084) stored;
 
 ALTER TABLE feature_drawn_lines
     ADD COLUMN length_feet INTEGER GENERATED always AS (
-Round(ST_Length(geography) * 3.28084)) stored;
+ST_Length(geography) * 3.28084) stored;
--- a/moped-database/migrations/1696622988854_add_feature_length_column/up.sql
+++ b/moped-database/migrations/1696622988854_add_feature_length_column/up.sql
@@ -1,0 +1,7 @@
+ALTER TABLE feature_street_segments
+    ADD COLUMN length_feet INTEGER GENERATED always AS (
+Round(ST_Length(geography) * 3.28084)) stored;
+
+ALTER TABLE feature_drawn_lines
+    ADD COLUMN length_feet INTEGER GENERATED always AS (
+Round(ST_Length(geography) * 3.28084)) stored;
--- a/moped-database/migrations/1696624355279_add-length-to-views/up.sql
+++ b/moped-database/migrations/1696624355279_add-length-to-views/up.sql
@@ -79,7 +79,7 @@ SELECT
     feature_drawn_lines.id,
     feature_drawn_lines.component_id,
     'feature_drawn_lines'::text AS "table",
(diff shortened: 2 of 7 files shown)


moped-database/migrations/1696624355279_add-length-to-views/up.sql
```diff
@@ -0,0 +1,95 @@
+CREATE OR REPLACE VIEW "public"."uniform_features" AS
```
johnclary, line 1: we do not have to drop the view bc the schema is not changing, because the new length column is tucked into the existing `attributes` 😅

moped-database/migrations/1696622988854_add_feature_length_column/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE feature_street_segments
+    ADD COLUMN length_feet INTEGER GENERATED always AS (
+Round(ST_Length(geography) * 3.28084)) stored;
+
+ALTER TABLE feature_drawn_lines
+    ADD COLUMN length_feet INTEGER GENERATED always AS (
+Round(ST_Length(geography) * 3.28084)) stored;
```
johnclary, line 7: i tested this migration locally against production data and it is lightening fast. we should have no concerns with a migration timeout when we deploy it to production.

johnclary:
after looking at this anew i decided to add `length_feet` as a column to the geography views instead of tucking it inside `attributes`. i will update that example SQL query momentarily.

chiaberry (approved):
TIL my street is 501 feet long

mddilley (approved):
Works great! My Google Maps estimate came within 20 feet of my 1350 foot component in the local DB. I remember some early talk about having these measurements so it is cool to see it materialize! 🚢 ✨

moped-database/migrations/1696622988854_add_feature_length_column/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE feature_street_segments
+    ADD COLUMN length_feet INTEGER GENERATED always AS (
+ST_Length(geography) * 3.28084) stored;
```
frankhereford, line 3: Meters, of course! 3.28084 feet in a meter. I was confused for a second until I ran down that `ST_Length(geography)` is always in meters. 👍

frankhereford (approved):
Thanks John. This is a neat feature, and really nice use of the generated column and keeping the view as light as possible by avoiding any projecting the geography down into geometry to get the linear distance. 🚢🚢🚢

johnclary:
thanks y'all!

PR #1151, opened 2023-10-07
Show error alert when component is unmapped

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14160

## Testing
**URL to test:** Local

1. Start your local Moped andnrun these SQL statements to delete features from components:

```SQL
delete from feature_street_segments where 1=1;
delete from feature_intersections where 1=1;
```

2. Open project 227's map: `https://localhost:3000/moped/projects/227?tab=map`

3. Observe the error alerts in the component list. Map each component and and observe the alert disappear.

4. Repeat these steps with drawn lines and drawn points.

5. What should the alert text say?  Is "Component is not mapped" ok?

Screenshot 2023-10-07 at 2 07 59 PM
 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
@@ -9,6 +9,7 @@ import ListItemSecondaryAction from "@mui/material/ListItemSecondaryAction";
 import ZoomInIcon from "@mui/icons-material/ZoomIn";
 import IconButton from "@mui/material/IconButton";
 import ListItemText from "@mui/material/ListItemText";
+import ErrorOutlineIcon from "@mui/icons-material/ErrorOutline";
 import { COLORS } from "./mapStyleSettings";
 import {
   useComponentListItemText,
@@ -59,7 +60,7 @@ export default function ComponentListItem({
         className={classes.listItem}
         ref={component._ref}
       >
-        {Icon}
+        {isComponentMapped ? Icon : <ErrorOutlineIcon color="error" />}
         <ListItemText
           className={classes.listItemText}
           primary={primary}
@@ -78,15 +79,17 @@ export default function ComponentListItem({
           </IconButton>
         </ListItemSecondaryAction>
       </ListItem>
-      {!isComponentMapped && (
-        <ListItem dense className={classes.listItem}>
-          <ListItemText
-            primary={<Alert severity="error">Component is not mapped</Alert>}
-          />
-        </ListItem>
-      )}
       <Collapse in={isExpanded}>
         <List component="div" disablePadding dense>
+          {!isComponentMapped && (
+            <ListItem dense className={classes.listItem}>
+              <ListItemText
+                primary={
+                  <Alert severity="error">Component is not mapped</Alert>
+                }
+              />
+            </ListItem>
+          )}
           {component.description && (
             <ListItem className={classes.nested}>
               <ListItemText secondary={component.description} />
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/utils/componentList.js
```diff
@@ -20,3 +20,20 @@ export const useComponentListItemText = (component) =>
...
+export const useIsComponentMapped = (component) =>
+  useMemo(
+    () =>
+      component.feature_drawn_points?.length > 0 ||
+      component.feature_drawn_lines?.length > 0 ||
+      component.feature_intersections?.length > 0 ||
+      component.feature_signals?.length > 0 ||
+      component.feature_street_segments?.length > 0,
+    [component]
+  );
```
johnclary, line 39: this feels like a clear over-use of useMemo: we don't need a stable reference to this value (it's Boolean) and the computation is not expensive. i'm going to remove the memo.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -73,6 +78,13 @@ export default function ComponentListItem({
           </IconButton>
         </ListItemSecondaryAction>
       </ListItem>
+      {!isComponentMapped && (
+        <ListItem dense className={classes.listItem}>
+          <ListItemText
+            primary={<Alert severity="error">Component is not mapped</Alert>}
```
chiaberry, line 84: This works for me -- I was considering proposing a call to action of sorts, "add component to map", but that should only be to the owner of the project and I am hoping that its self evident that that is how to remove the alert.

chiaberry (approved):

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
mddilley: I like that call to action too if we are trying to motivate users to make the alert go away by mapping. Also wanted to see what y'all think about updating the icon next to the component type when not mapped, and then showing the alert when it is expanded? Not sure about the double icons though.

Something like:
Screenshot 2023-10-09 at 10 37 29 AM

Then expanded:
Screenshot 2023-10-09 at 10 40 40 AM

mddilley (approved):
Tested locally, and I see the alert appear on unmapped components and disappear after mapping them. Great to have visual feedback here! 🚢 🚀

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
johnclary: oooh i love it! will make that change 🚀

johnclary: @mddilley do you have this coded out? if so...push?

frankhereford (approved):
Tested this out and it works just as expected. Even as is, I think the message is good. I the need to click on the component to dig down to find the mapping button is a little hidden, but nothing major. Can the red warning UI block be clickable to take you right into editing the map data for that component?

Both @chiaberry's and @mddilley's suggestions are great, and I'm all for incorporating those fully. I think Chia and I are calling out the same or very similar things. Mike's warning icon suggestion is 💯.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
mddilley: git stash pop and pushed ✅

johnclary:
Thanks, y'all for the v helpful feedback. How about adding the red error styling to the map edit icon button as well?
Screenshot 2023-10-09 at 1 17 46 PM

The alert component does take an `action` prop, but the button inside the alert is awkward:

Screenshot 2023-10-09 at 1 21 15 PM

mddilley:
i like that red map edit a lot, John! 🚀 Paints a trail right to where to map the component. 📍

johnclary:
OK! calling this done. thanks again @mddilley for the commit as well 🙌

PR #1152, opened 2023-10-09
Use project list view query hooks for csv export

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14068

This PR ports the existing CSV export generation into a custom hook. Lots of the existing logic was reused but moved out of the `Search` component.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Spin up the stack locally [with production data](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-MIQvl_rKnZ_-wHRdp4J/dev-guides/how-tos/how-to-load-production-data-into-a-local-instance). Seed data can be used but there will be fewer projects to test this feature on. You can also compare prod exports to the export that come out of this branch if you use prod data.
2. Try exporting data with different order by columns
3. Try exporting data with simple search filters applied
4. Try exporting data with advanced search filters applied
5. You should see the number of records in the export match the total number of results returned and filtered down to your searches.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Search.js
+++ b/moped-editor/src/components/GridTable/Search.js
@@ -57,7 +57,6 @@ const history = createBrowserHistory();
 
 /**
  * Renders a table search component with a search bar and search filters
- * * @param {GQLAbstract} query - The GQLAbstract object as provided by the parent component
  * @param {Object} filters - The current filters from useAdvancedSearch hook
  * @param {Function} setFilters - Set the current filters from useAdvancedSearch hook
  * @param {Object} parentData - Response data (if any) from the parent component
@@ -72,7 +71,6 @@ const history = createBrowserHistory();
  * @constructor
  */
 const Search = ({
-  query,
   filters,
   setFilters,
   filterQuery,
@@ -161,7 +159,7 @@ const Search = ({
               {queryConfig.showExport && (
                 <Button
                   disabled={
-                    (parentData?.[query.config.table] ?? []).length === 0
+                    (parentData?.[queryConfig.table] ?? []).length === 0
                   }
                   className={classes.downloadCsvButton}
                   onClick={handleExportButtonClick}
@@ -187,7 +185,6 @@ const Search = ({
       >
         <Paper className={classes.advancedSearchPaper}>
           <Filters
-            query={query}
             filters={filters}
             setFilters={setFilters}
             filterQuery={filterQuery}
--- a/moped-editor/src/components/GridTable/Search.js
+++ b/moped-editor/src/components/GridTable/Search.js
@@ -1,28 +1,13 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { createBrowserHistory } from "history";
 import { useLocation } from "react-router-dom";
 
-import {
-  Box,
-  Button,
-  CircularProgress,
-  Dialog,
-  DialogActions,
-  DialogContent,
-  DialogContentText,
-  DialogTitle,
-  Grid,
-  Paper,
-  Popper,
-} from "@mui/material";
+import { Box, Button, Grid, Paper, Popper } from "@mui/material";
 import SaveAltIcon from "@mui/icons-material/SaveAlt";
 import Filters from "src/components/GridTable/Filters";
 import SearchBar from "./SearchBar";
 import makeStyles from "@mui/styles/makeStyles";
-import { useLazyQuery } from "@apollo/client";
-import { format } from "date-fns";
-import Papa from "papaparse";
 
 const useStyles = makeStyles((theme) => ({
   root: {
--- a/moped-editor/src/components/GridTable/Search.js
+++ b/moped-editor/src/components/GridTable/Search.js
@@ -82,6 +82,7 @@ const history = createBrowserHistory();
  * @param {Function} setSearchTerm - Set the current search term from useSearch hook
  * @param {Object} queryConfig - The query configuration for the current table
  * @param {Object} filtersConfig - The filters configuration for the current table
+ * @param {Function} handleExportButtonClick - The function to handle the export button click
  * @return {JSX.Element}
  * @constructor
  */
@@ -97,6 +98,7 @@ const Search = ({
   setSearchTerm,
   queryConfig,
   filtersConfig,
+  handleExportButtonClick,
 }) => {
   const classes = useStyles();
   const queryPath = useLocation().pathname;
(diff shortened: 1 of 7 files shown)


moped-editor/src/components/GridTable/Search.js
```diff
@@ -109,103 +94,6 @@ const Search = ({
    */
   const [searchFieldValue, setSearchFieldValue] = useState(searchTerm);
 
-  /**
```
mddilley, line 112: All of this moved to `useExportCsv.js`

moped-editor/src/components/GridTable/Search.js
```diff
@@ -236,34 +124,6 @@ const Search = ({
...
-        setTimeout(() => {
-          // Update the state
-          setDialogOpen(false);
-          downloadFile(csvString);
-        }, 1500);
-      }
-    },
-    // eslint-disable-next-line
-    [dialogOpen, downloading, loading, query.limit, getExport]
-  );
```
mddilley, line 265: This useEffect was a workaround for an issue with Apollo Client running the `onCompleted` hook on every render. I demo'd using `onCompleted` today before I realized that it would trigger a CSV download every time the table order or filters changed. Luckily, `getExport` the function that `useLazyQuery` returns to trigger the query fetch now returns a promise that exposes the data object. You can see that further down in `handleExportButtonClick`.

moped-editor/src/components/GridTable/Search.js
```diff
@@ -337,30 +196,6 @@ const Search = ({
           />
         </Paper>
       </Popper>
-      <Dialog
```
mddilley, line 340: Moved this up the tree into `ProjectListViewTable` since it doesn't have any dependencies in this component anymore.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -442,9 +447,10 @@ const ProjectsListViewTable = ({ query }) => {
 
   const columnsToReturn = Object.keys(PROJECT_LIST_VIEW_QUERY_CONFIG.columns);
 
-  const { query: projectListViewQuery } = useGetProjectListView({
+  const { query: projectListViewQuery, exportQuery } = useGetProjectListView({
```
mddilley, line 450: This hook now builds both a query with current filters, order, and pagination and also an export query that doesn't include the pagination parameters. We can adjust how this works in the future, but this is what we do in prod right now. Also open to splitting out the the export and non-export query building if anyone feels like it makes sense.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
mddilley, line 1: This hook now builds both a query with current filters, order, and pagination and also an export query that doesn't include the pagination parameters. We can adjust how this works in the future, but this is what we do in prod right now. Also open to splitting out the the export and non-export query building if anyone feels like it makes sense.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
```diff
@@ -0,0 +1,145 @@
...
+          <DialogContentText>
+            Preparing download, please wait.
+          </DialogContentText>
+        </Grid>
+      </Grid>
+    </DialogContent>
+    <DialogActions>
+      <Button onClick={handleDialogClose} color="primary">
+        Cancel
+      </Button>
```
mddilley, line 38: Maybe we should removed this cancel button? It doesn't cancel the CSV download. It only closes the modal.

moped-editor/src/components/GridTable/Search.js
johnclary: @mddilley i'm seeing some odd behavior that sounds similar to what you're describing here.

1. Type `stassney` in the search bar and press `enter` to search
2. Click the download button to download—download works as expected
3. Use the backspace key to clear the search box
4. You are prompted for download again 🙃

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
johnclary: Hmm—yeah this kinda weird. Is it possible have the 'cancel' button actually cancel the download? And I am curious what would/should happen if the download fails for some reason (most likely because the Hasura request bombed or timed out).

Separately, I noticed that if you cancel a download you will not be able to download a file again until you refresh the page.

Curious what you think, Mike—I don't want the cancel to blow up the scope of this work bc i think it's an edge case.

moped-editor/src/components/GridTable/Search.js
mddilley: i'm going to take a  look at this today - thanks for the heads up! 🙏

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
mddilley: I'd have to look at how to the cancel flow works, but it definitely took some effort to catch the cancel button before the export was done. Makes me wonder how often someone would try to do that? Maybe an alert on failure would be good to improve this? I like your idea of handling the error!

mddilley:
y'all - I'm removing review requests for now. Need to work on a couple of bugs that John has found. 🙏 🐛

moped-editor/src/components/GridTable/Search.js
johnclary: nevermind! i didn't have your most recent commits locally 🤦

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
```diff
@@ -0,0 +1,145 @@
...
+  </Dialog>
+);
+
+/**
+ * Downloads the contents of fileContents into a file
+ * @param {string} fileContents
+ */
+const downloadFile = (fileContents, queryTableName) => {
+  const exportFileName =
+    "moped-" + queryTableName + format(Date.now(), "yyyy-MM-dd'T'HH:mm:ssxxx");
```
johnclary, line 49: @mddilley i'd like to simplify the file name a bit. currently you get something like this:
- `moped-project_list_view2023-10-10T13_22_06-04_00.csv`
and i'd like to instead have something like this:
- `moped_projects_2023-10-10.csv`

My reasoning is:
- short filenames are better
- we don't need a timestamp in the filename. if folks download multiple exports on the same day they rename the files as needed
- `project_list_view` does really mean anything to anyone except us

how does that sound to you?

johnclary (changes requested):
this works really great once you pull the latest commits! ✨  i love this refactor—thank you for untangling it! 

i just have a minor request related to the file name. otherwise this is ready to go!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
chiaberry: i agree with this reasoning!

mddilley: sounds great! i can't come up with a good reason to keep the timestamp around especially since the OS covers that too if it _really_ became important to someone. I'll make the change and see what everyone else thinks. 👍

johnclary (approved):
🙌 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectsListView/ProjectsListView.js
```diff
@@ -18,12 +13,6 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-/**
- * Load Query Configuration as a mutable object
- * @type {GQLAbstract}
- */
-let projectsQuery = new GQLAbstract(PROJECT_LIST_VIEW_QUERY_CONFIG);
-
```
chiaberry, line 26: bye!!!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
chiaberry: my vote is to just remove the cancel button. it is hard to catch anyway

chiaberry: oh its easy to catch if you have the slowest connection, but if it doesnt stop the download from happening, then we should at least rename the button to say "close"

chiaberry (approved):
I am so glad this is moved out of the Search!!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
johnclary: removing it is cool with me 👍

PR #1155, opened 2023-10-11
Add work activity fields to activity log formatter

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/13000

## Testing
**URL to test:** 

https://deploy-preview-1155--atd-moped-main.netlify.app/moped/projects/1759?tab=activity_log

**Steps to test:**

Add and edit work activities, check the activity tag to see that the new columns appear. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/activityLogFormatters/mopedContractsActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedContractsActivity.js
@@ -12,7 +12,7 @@ export const formatContractsActivity = (change) => {
     style: "boldText",
   };
 
-  // add a new contract
+  // add a new work activity
   if (change.description.length === 0) {
     if (contractor) {
       // with contractor populated
@@ -21,26 +21,26 @@ export const formatContractsActivity = (change) => {
         changeText: [
           { text: "Added ", style: null },
           contractorText,
-          { text: " as a new contract", style: null },
+          { text: " as a new work activity", style: null },
         ],
       };
     } else {
       // without contractor populated
       return {
         changeIcon,
-        changeText: [{ text: "Added a new contract", style: null }],
+        changeText: [{ text: "Added a new work activity", style: null }],
       };
     }
   }
 
-  // delete an existing contract
+  // delete an existing work activity
   if (change.description[0].field === "is_deleted") {
     if (contractor) {
       // with contractor populated
       return {
         changeIcon,
         changeText: [
-          { text: "Removed the contract for ", style: null },
+          { text: "Removed the work activity for ", style: null },
           contractorText,
         ],
       };
@@ -48,7 +48,7 @@ export const formatContractsActivity = (change) => {
       // without contractor populated
       return {
         changeIcon,
-        changeText: [{ text: "Removed a contract", style: null }],
+        changeText: [{ text: "Removed a work activity", style: null }],
       };
     }
   }
@@ -82,7 +82,7 @@ export const formatContractsActivity = (change) => {
     return {
       changeIcon,
       changeText: [
-        { text: "Edited the contract for ", style: null },
+        { text: "Edited the work activity for ", style: null },
         contractorText,
         { text: " by updating the ", style: null },
         {
@@ -96,7 +96,7 @@ export const formatContractsActivity = (change) => {
     return {
       changeIcon,
       changeText: [
-        { text: "Edited a contract by updating the ", style: null },
+        { text: "Edited a work activity by updating the ", style: null },
         {
           text: changes.join(", "),
           style: "boldText",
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
mddilley, line 1: I think there might be a column that isn't set in this config? I saw one `undefined` value comes through in the `changes` array when I edited an existing work activites record which creates a blank item getting joined with the others. I can try to reproduce more in the morning!
Screenshot 2023-10-11 at 5 35 40 PM

chiaberry: good catch!! it was "updated_by_user_id"

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -153,7 +154,11 @@ const usePrepareActivityData = (activityData) =>
           const oldData = outputEvent.record_data.event.data.old;
           let changedField = "";
           Object.keys(newData).forEach((key) => {
-            if (newData[key] !== oldData[key]) {
+            if (!!newData[key] && typeof newData[key] === "object") {
+              if (!isEqual(newData[key], oldData[key])) {
+                changedField = key;
+              }
+            } else if (newData[key] !== oldData[key] && key !== "updated_at") {
```
chiaberry, line 161: included comparing to see if there were actual differences in the fields that contain arrays. 

I also had to add this part to skip the "updated_at" column, because its always going to be different. 

I want to refactor how it this all works, but in the interest of getting the release out today, I'm leaving it like this. lemme know if this is okay

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -555,6 +555,24 @@ export const ProjectActivityLogTableMaps = {
       is_deleted: {
         label: "is deleted",
       },
+      interim_work_activity_id: {
+        label: "interim work activity ID",
+      },
+      implementation_workgroup: {
+        label: "implementation workgroup",
+      },
```
johnclary, line 563: this field is not actually implemented in the work activity form. I ultimately decided to use contractor to capture this. it's probably good to leave this here until we remove that field—but would you also rename the "Contractor" field to "Workgroup/Contractor"?

i will open an issue to remove this unused field from the db 😑

moped-editor/src/utils/activityLogFormatters/mopedContractsActivity.js
```diff
@@ -58,10 +59,20 @@ export const formatContractsActivity = (change) => {
...
 
   let changes = [];
+  const fieldsToSkip = ["updated_at", "updated_by_user_id"];
 
   // loop through fields to check for differences, push label onto changes Array
   Object.keys(newRecord).forEach((field) => {
-    if (newRecord[field] !== oldRecord[field]) {
+    // typeof(null) resolves as "object", check that field is not null before checking if object
+    // task orders are in arrays
+    if (!!newRecord[field] && typeof newRecord[field] === "object") {
```
johnclary, line 68: i don't think there's any harm in checking all the fields, but you could compare the task orders' `task_order` property, which is the unique task order ID.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
johnclary: 👍 sounds good to me. it is throwing me off that this code needs to exist twice—so i guess that's what you're referring to.

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -528,7 +528,7 @@ export const ProjectActivityLogTableMaps = {
       },
     },
   },
-  moped_proj_contract: {
+  moped_proj_work_activity: {
```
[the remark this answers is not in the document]
johnclary: I want to request another change, but it would be totally fine to leave it out of scope for now.

We now have the special generated column `reference_id` which would be nice to use to identify the activities in the log. 

So this:

```
Edited the contract for MASTEC ZAPALAC-REED DEVELOPER by updating the task orders
```

becomes:

```
Edited the work activity MPD-1661-156 by updating the task orders
```

In some of the ATSD examples they will have multiple work activities that use the same workgroup/contractor, so it will be best to use the reference ID.

johnclary (changes requested):
This looks great 🙌 

I requested a few minor changes and it would also be fine to leave the bigger changes for a follow-up issue.

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
mddilley: Sounds like a candidate for a spin-off issue to me! I tested the updated code, and I see the deleted rows show in the log as "Removed" 🚀

mddilley (approved):
Tested adding, updating, and deleting work activites, and they are logged as expected! 🚢 🚀

johnclary (approved):
✨ thank you @chiaberry! 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1150, opened 2023-10-07
Ability to map a signal component instead of picking from list

johnclary:
This PR depends on https://github.com/cityofaustin/atd-moped/pull/1094.

## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/12998

## Testing
**URL to test:**  Local

1. Create a signal by picking it from the signal autocomplete input (should work normally)
2. Edit the signal component by clearing the selected signal from the autocomplete input (should result in unmapped component)
3. Map the unmapped signal component (location should be saved as expected)
4. Edit the the signal component you just mapped by picking a signal from the list (the location should be updated with the location you selected)
5. Edit the signal component again by changing it to an **Intersection - Improvement** component (this will result in the component being unmapped)
6. Edit the **Intersection - Improvement** component by switching to a signal component, but leave the signal input blank (should result in the location being unchanged).

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1156, opened 2023-10-11
Show location description in draft/edit component list item

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/14166

## Testing
**URL to test:** 
https://deploy-preview-1156--atd-moped-main.netlify.app/moped/projects/1759?tab=activity_log

**Steps to test:**
1. Go to a projects map view.
2. Create a new component with a location description. When mapping the component in create mode, the component list item should have the Component Type - Subtype on the first line, and the second line should have the location description, similar to the non-editing state of a components list item.
image
4. Save the component and test editing it. The above should also be true.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/DraftComponentListItem.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/DraftComponentListItem.js
@@ -6,15 +6,17 @@ import ListItemText from "@mui/material/ListItemText";
 import Cancel from "@mui/icons-material/Cancel";
 import CheckCircle from "@mui/icons-material/CheckCircle";
 import { COLORS } from "./mapStyleSettings";
+import { useComponentListItemText } from "./utils/componentList";
 
 const DraftComponentListItem = ({
-  primaryText,
-  secondaryText,
+  component,
   onSave,
   onCancel,
   saveButtonText,
   saveButtonDisabled,
 }) => {
+  const { primary, secondary } = useComponentListItemText(component);
+
   return (
     <Box
       borderLeft={7}
@@ -23,7 +25,7 @@ const DraftComponentListItem = ({
       }}
     >
       <ListItem dense>
-        <ListItemText primary={primaryText} secondary={secondaryText} />
+        <ListItemText primary={primary} secondary={secondary} />
       </ListItem>
       <ListItem dense>
         <ListItemText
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/DraftComponentListItem.js
mddilley, line 1: The bug:
1. Create a new component with a location description, map it, and save
2. Edit the same component's location description and save
3. Edit the same component's features using the map and you will see the outdated location description

This is due to the state updates being scattered and hard to grok, but @johnclary  centralized those updates in https://github.com/cityofaustin/atd-moped/pull/1150, and I tested this branch with that branch merged in and the bug is gone. 🪄

mddilley (commented):
Looks great and I love that this makes use of an existing hooks to have less props. 🙌 I found a small bug, but I think it is an existing one that will be fixed by another open PR. Might just mean that we make sure they ship together. Ready to approve when I get some back up on that.

moped-editor/src/views/projects/projectView/ProjectComponents/DraftComponentListItem.js
johnclary: nice—a freebie! 🙌

moped-editor/src/views/projects/projectView/ProjectComponents/DraftComponentListItem.js
```diff
@@ -6,15 +6,17 @@ import ListItemText from "@mui/material/ListItemText";
...
 const DraftComponentListItem = ({
-  primaryText,
-  secondaryText,
+  component,
   onSave,
   onCancel,
   saveButtonText,
   saveButtonDisabled,
 }) => {
+  const { primary, secondary } = useComponentListItemText(component);
```
johnclary, line 18: yay for having a hook ready to handle this.

i just had to code this out in a SQL view for the AGOL integration, and now it occurs to me that we should probably be using a generated column to create the `<name> - <subtype>` string at the DB level. not a big deal and i'm glad to see this hook reused 🚀

johnclary (approved):
looks great and works great! thanks for helping us giving Moped some polish! 🙌 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

mddilley (approved):
Yay! So great to see more component data visible in these flows. ✨

PR #1166, opened 2023-10-16
Ability to create a status update from the Comments tab

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/14256

## Testing
**URL to test:** 
https://deploy-preview-1166--atd-moped-main.netlify.app/moped/ 

**Steps to test:**
1. Go to the "Notes" tab on a project (formerly called "Comments").
2. Toggle between the "Internal Note" and "Status Update" options above the Save button for a new note. Test saving both types of new notes.
3. Check that you see your new note pop up in the notes section under the correct note type toggle.
4. After adding a new status update, check on the project summary tab that it is reflected there in the Status update section as well.
5. Test editing a comment, in editing state you should also see the radio buttons and be able to update the note type.
6. Also test sizing down your screen to see the way the buttons change. I tested on my iPhone and they look fine, lmk if you want them to look differently on a smaller sized screen!

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/queries/notes.js
```diff
@@ -1,7 +1,7 @@
 import { gql } from "@apollo/client";
 
-export const COMMENTS_QUERY = gql`
-  query GetProjectComments($projectNoteConditions: moped_proj_notes_bool_exp!) {
+export const NOTES_QUERY = gql`
+  query GetProjectNotes($projectNoteConditions: moped_proj_notes_bool_exp!) {
```
johnclary, line 4: thanks for updating all these—it's nice lil boost for dev experience.

moped-editor/src/views/projects/projectView/NoteInputQuill.js
```diff
@@ -47,6 +52,23 @@ const CommentInputQuill = ({
     ref?.current.focus();
   }, []);
 
+  /**
+   * Defines the NewNoteTypeButton with a toggle style-change behavior.
+   * @param {Object} props
+   * @return {JSX.Element}
+   * @constructor
+   */
+  const NewNoteTypeButton = (props) => (
```
johnclary, line 61: You can move this component declaration outside of the `NoteInputQuill` component. I think it's fine to leave the component in the same file—IMO it's a judgement call when it makes sense to break components into separate files.

moped-editor/src/views/projects/projectView/NoteInputQuill.js
```diff
@@ -47,6 +52,23 @@ const CommentInputQuill = ({
...
+    <Button
+      color="primary"
+      className={classes.showButtonItem}
+      variant={newNoteType === props.noteTypeId ? "contained" : "outlined"}
+      onClick={() => setNewNoteType(props.noteTypeId)}
+    >
+      {props.children}
+    </Button>
```
johnclary, line 69: I love that you use `props.children` here. I _never_ think to do this and it makes for such cleaner code!

That said...would you mind switching this to the [Radio component](https://mui.com/material-ui/react-radio-button/#direction) with a horizontal layout?

Like this (but plz ignore dark mode):

Screenshot 2023-10-17 at 8 58 19 AM

I'm sorry for not having this spec available for you when you started the work!

moped-editor/src/views/projects/projectView/NoteInputQuill.js
```diff
@@ -61,11 +83,22 @@ const CommentInputQuill = ({
             />
           </Box>
         </Grid>
+        {!editingNote && (
```
johnclary, line 86: oooh it would be nice to be able to edit the note the type. do you think that's reasonable to include in this work? or i'm happy to open a separate issue.

johnclary (changes requested):
This is looking very good! I love the "Notes" rebrand, and this toggle is makes the behavior much more transparent.

moped-editor/src/views/projects/projectView/NoteInputQuill.js
roseeichelmann: i think its reasonable!

moped-editor/src/views/projects/projectView/NoteInputQuill.js
roseeichelmann: oops idk why i assumed radio buttons were just the same kind of buttons as the ones above the comments section. ill fix that! it will be better having this set of buttons be different from the ones for sorting the notes.

moped-toolbox/knack_migration/1_prepare_projects.py
```diff
@@ -37,7 +37,7 @@ def build_note(note, user_name, user_id, note_type=2):
     # 1 = internal note
     return {
         "status_id": 1,  # active / not deleted status
-        # project note types are hardcoded. see here: https://github.com/cityofaustin/atd-moped/blob/9c6cac2598da48f1e06ab20ac0f7e5074b14cc8e/moped-editor/src/views/projects/projectView/ProjectComments.js#L271
+        # project note types are hardcoded. see here: https://github.com/cityofaustin/atd-moped/blob/9c6cac2598da48f1e06ab20ac0f7e5074b14cc8e/moped-editor/src/views/projects/projectView/ProjectNotes.js#L271
```
frankhereford, line 40: This is the silliest little thing, but this deep link points to a time when this file had not been renamed. I think the URL here should be `https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/views/projects/projectView/ProjectNotes.js#L321`. This URL won't work until this is merged, but I think that's what we want, right?

frankhereford (approved):
+1 to John's comments - thanks for cleaning up a lot of naming! I tested this on the deploy preview, and I really like the UX on this - much improved! Code looks good! Thank you Rose!

moped-editor/src/queries/notes.js
mddilley: +1 - really nice to have consistent naming in the UI and code 🔍

moped-editor/src/views/projects/projectView/NoteInputQuill.js
mddilley: yeah, agree here. If this definition stays inside the `NoteInputQuill` component function, it is recreated on every render. Definitely a legacy anti-pattern that we've been slowly chipping away at in the codebase. 🧹

chiaberry (approved):
I'm not in love with the size and placement of the notes / status toggle, but that's not on you! this code is great and thank you for renaming everything.

moped-editor/src/views/projects/projectView/NoteInputQuill.js
mddilley, line 1: Not sure exactly where this is coming from, but I noticed the text labels for the radios is greyed when the editing a note on the Notes tab. The text is darker when creating a new note on the same tab. Maybe there are some styles defined somewhere?

**New note quill input**
Screenshot 2023-10-20 at 11 32 40 AM

**Edit existing note quill input**
Screenshot 2023-10-20 at 11 32 34 AM

mddilley (approved):
Works great - so cool to have more flexibility on what becomes a status update or not! 🚀 🙌

moped-editor/src/views/projects/projectView/NoteInputQuill.js
mddilley: just saw that this is updated now! 🚀

moped-editor/src/views/projects/projectView/NoteInputQuill.js
roseeichelmann: yeahhh ive been trying to figure out where thats coming from and i cant, i can override it by just adding styling to make the text always be black for the radio button labels but i wish i could figure out why its happening in the first place

mddilley: cool, definitely not big enough to block merging this today imo. I think that these lines are a clue. This might have been me in the first place overriding styles to makes the list item icons be the primary text color. https://github.com/cityofaustin/atd-moped/blob/5aff4858422b3f711b60ae71f20c8b406ab3dc9f/moped-editor/src/views/projects/projectView/ProjectNotes.js#L81-L83 Probably inheriting the styles from a MUI component it is wrapped in.

moped-editor/src/views/projects/projectView/NoteInputQuill.js
```diff
@@ -0,0 +1,147 @@
...
+ * @param {function} onChange - Callback function that runs when the note type is changed via button click
+ * @return {JSX.Element}
+ * @constructor
+ */
+const NoteTypeRadioButtons = ({ defaultValue, onChange }) => (
+  <RadioGroup
+    row
+    value={defaultValue}
+    onChange={onChange}
+    sx={{ color: "black" }}
```
roseeichelmann, line 59: couldnt for the life of me figure out why/where the editing mode was inheriting grey text for the radio button labels, so this is the fix i came up withhh. i feel like its a bandaid solution and not great though?

chiaberry: I don't think its a bandaid necessarily, the styles are being inherited from two different places, so its good to specify a consistent color.  I dont know why the top one is using Mui-Paper-root-Mui-Card-root, but the edit one is using a typography style, even tho they are both in cards?

johnclary: agreed :)

johnclary (approved):
this looks great to me! thanks for working through the design changes 🙌 🚢

PR #1167, opened 2023-10-17
Fixes advanced search filter lag and adds query caching

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14257
-  https://github.com/cityofaustin/atd-data-tech/issues/10799

## Testing
**URL to test:** [Netlify](https://deploy-preview-1167--atd-moped-main.netlify.app/)

1. Load the project list view.
2. Navigate to any project except the first project in the project list. 
3. Edit the project description
4. Navigate back to the project list—watch your project jump to the top of the project list
5. Repeat those steps with another project, but this time edit the project description, save, wait 10 seconds, then edit the description again.
6. Navigate back to the project list—your edited project should already be at the top of the project list. This demonstrates the race condition wrt to a project's `updated_at` date. 
7. Increase the rows per page to 1,000, then open the advanced search. It should open without any lag!
8. Apply some advanced search criteria, then navigate to a project and use the `< ALL PROJECTS` link to navigate back to the project list with your advanced filters applied.
9. Test that column sorting and pagination work normally.
10. Add, edit, and map project components. This ensures that the new component object cloning has not broken anything.

I think that's it! 😅 
---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -29,9 +29,9 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
   noResultsMessage: "No projects found.",
   showPagination: true,
   pagination: {
-    rowsPerPageOptions: [250, 1000],
+    rowsPerPageOptions: [100, 250, 1000],
     defaultOffset: 0,
-    defaultLimit: 250,
+    defaultLimit: 100,
```
johnclary, line 34: table rendering is just too dang slow with 250 projects :/

moped-editor/src/App.js
```diff
@@ -22,42 +23,67 @@ const HASURA_ENDPOINT = process.env.REACT_APP_HASURA_ENDPOINT;
...
+          },
+          moped_entity: {
+            keyFields: ["entity_id"],
+          },
+          moped_phases: {
+            keyFields: ["phase_id"],
+          },
+          moped_types: {
+            keyFields: ["type_id"],
+          }
```
johnclary, line 64: besides the `project_list_view`, i added these type policies for demo purposes. it doesn't really matter that we're caching these values until we set the correct `fetchPolicy` for all instances of `useQuery` that consume this data.

moped-editor/src/components/GridTable/SearchBar.js
```diff
@@ -160,6 +160,7 @@ const SearchBar = ({
     <>
       <TextField
         fullWidth
+        autoFocus
```
johnclary, line 163: i'm curious if y'all like this change! i will ask the MUG as well.

in an ideal world, the project list scroll position would persist when you navigate back to it. this `autoFocus` would break that, _I think_.

in order to preserve scroll position we would need to start using the new [data routers](https://reactrouter.com/en/main/routers/picking-a-router). i should probably open an issue for that....

moped-editor/src/App.js
```diff
@@ -22,42 +23,66 @@ const HASURA_ENDPOINT = process.env.REACT_APP_HASURA_ENDPOINT;
...
       <StyledEngineProvider injectFirst>
         <LocalizationProvider dateAdapter={AdapterDateFns}>
           <ThemeProvider theme={theme}>
             <GlobalStyles />
-            {routing}
+            <ProjectListViewQueryContext.Provider
+              value={{ listViewQuery, setListViewQuery }}
+            >
+              {routing}
+            </ProjectListViewQueryContext.Provider>
```
johnclary, line 85: i ended up creating this context after testing lots of different ways to refetch the project list view query from the ProjectView.

the idea is that the context preserves the most recent project list view `gql` so that it can be refetched. despite what the Apollo docs say, this really seems like the only way to do it—because other methods, e.g., refetching a query by name, require that the query itself is still active in a mounted component.

moped-editor/src/queries/project.js
```diff
@@ -96,6 +96,7 @@ export const SUMMARY_QUERY = gql`
       entity_id
       moped_entity {
         entity_name
+        entity_id
```
johnclary, line 99: the cache actually breaks if you define a `typePolicy` don't include the PK property in every instance of the type.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
```diff
@@ -23,9 +23,10 @@ export const makeFeatureFromProjectGeographyRecord = (record) => ({
 export const getAllComponentFeatures = (component) => {
   const allComponentFeatures = [];
 
-  Object.keys(featureTableFieldMap).forEach((key) => {
-    if (component.hasOwnProperty(key)) {
-      allComponentFeatures.push(component[key]);
+  Object.keys(featureTableFieldMap).forEach((featureTableName) => {
+    if (component.hasOwnProperty(featureTableName)) {
+      const feature = cloneDeep(component[featureTableName]);
+      allComponentFeatures.push(feature);
```
johnclary, line 29: apparently this only becomes a problem when you enable caching—you cannot directly modify any object stored in the Apollo cache. they are immutable.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -55,10 +54,10 @@ const ProjectSummaryProjectDescription = ({
         projectId: projectId,
         description: description,
       },
+      refetchQueries: [{ query: listViewQuery }, "ProjectSummary"],
```
johnclary, line 57: So—here's the result of all that fussing with conext! you can pass it to `refetchQueries`. any queries specified here get refetched in the background—thereby updating the cache.

As you can see, we are passing the raw `gql` of the `listViewQuery`, but also the name of summary query, `ProjectSummary`. Wouldn't it be nice if we could pass the name `ProjectListView` instead of fussing with contex? I think so too! :/

[Refetching docs](https://www.apollographql.com/docs/react/data/refetching).

johnclary: In any case—I included this as a working example, because we really do want to make sure we're refetching the list view query in the background, so that the list view table does not re-render a few seconds after we navigate back to it. 

this will make more sense when you test it yourself.
1. Load the list view
2. Navigate to any project **except** the project in the first row of the list view
3. Edit the project's name.
4. Navigate back to the project list view.

Observer how there is a brief lag before the list view re-renders and the project you just ended moves to the top.

What's going on:
1. You make a project edit.
2. Hasura event triggers cause the project's `updated_at` date to update
3. You navigate to the project list
4. The project list data is first loaded from the cache, then it is re-hydrated with network data
5. The project list is sorted by `updated_at`, so the project you just edited is re-ordered to the top.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -204,7 +208,6 @@ const ProjectView = () => {
    */
   const { loading, error, data, refetch } = useQuery(SUMMARY_QUERY, {
     variables: { projectId, userId },
-    fetchPolicy: "no-cache",
   });
```
johnclary, line 211: even though there are parts of this query that can be cached (the lookup values for project_phases, etc) — that cache is not going to stop us from re-querying the DB, because they're bundled together with the project data that we always want to be current.

so we have a backlog task of breaking apart some of our queries so that we can fetch the lookup tables once on app load. we should never need to fetch them more than once per user session. i will open an issue for that 😅

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -381,6 +386,13 @@ const ProjectView = () => {
     data?.moped_project?.[0]?.moped_proj_phases?.[0]?.moped_phase;
   const isProjectDeleted = data?.moped_project[0]?.is_deleted;
 
+  /**
+   * This function exists to enable closing the Map tab
+   */
+  const onCloseTab = useCallback(() => {
+    handleChange(null, 0);
+  }, [handleChange]);
```
johnclary, line 394: implementing caching broke the ability to close the map tab. not sure why exactly, but something to do with the `ProjectView` component only re-rendering because our `client` instance getting rebuilt 🤷

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -15,7 +15,7 @@ import { formatTimeStampTZType } from "src/utils/dateAndTime";
 export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
   options: {
     useQuery: {
-      fetchPolicy: "cache-first", // Default is "cache-first", or use "no-cache"
+      fetchPolicy: "cache-and-network", // Default is "cache-first", or use "no-cache"
```
johnclary, line 18: like i mention in my long comment above—this has the effect of having a project record "jump" to the top of the list view after editing it and then navigating back to the list view. it's not great :/ and we can definitely fix it by being vigilant about refetching in the background. i will open a separate issue for that. i'm curious how urgent y'all think that is...

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -531,59 +270,17 @@ const ProjectsListViewTable = () => {
         {/*Main Table Body*/}
         <Paper className={classes.paper}>
           <Box mt={3}>
-            {loading ? (
+            {!data && !error ? (
```
johnclary, line 273: because we are now using a `cache-and-network` fetch policy, we must stop listening on `loading` - since we want the loading to happen in the background so long as we any data to render in the list.

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -65,3 +76,370 @@ export const resolveHasSubprojects = (array) => {
   }
   return "No";
 };
+
+/**
+ * Various custom components for Material Table
+ */
+export const useTableComponents = ({
```
johnclary, line 83: these new hooks are a bit 😵 but i'm not sure how to make them better 🤷

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
```diff
@@ -12,7 +12,7 @@ export const useGetProjectListView = ({
   advancedSearchWhereString,
 }) => {
   const { query, exportQuery } = useMemo(() => {
-    const query = gql`{
+    const query = gql`query ProjectListView {
```
johnclary, line 15: the `ProjectListView` doesn't do us much good because refetching by it's name doesn't work (see my note above), but it is a best practice to include it anyway.

moped-toolbox/random_project_seeds/index.js
johnclary, line 1: i updated the random seeds tool for benchmarking purposes, but we really shouldn't need it once we the access migration completes and we have thousands of projects to test with.

moped-editor/src/App.js
```diff
@@ -22,42 +23,66 @@ const HASURA_ENDPOINT = process.env.REACT_APP_HASURA_ENDPOINT;
...
+          },
+          moped_entity: {
+            keyFields: ["entity_id"],
+          },
+          moped_phases: {
+            keyFields: ["phase_id"],
+          },
+          moped_types: {
+            keyFields: ["type_id"],
+          },
```
johnclary, line 64: besides the project_list_view, i added these type policies for demo purposes. it doesn't really matter that we're caching these values until we set the correct fetchPolicy for all instances of useQuery that consume this data.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
johnclary: - https://github.com/cityofaustin/atd-data-tech/issues/14310

moped-editor/src/App.js
chiaberry: should this note be as a "todo:" comment in the code?

chiaberry (commented):
I was reviewing this PR last night and had the netlify tab open. Then went back this morning and I was logged out and the page refreshed on its own a few times until I got signed back in automatically. not sure if this always happens, and i was just paying more attention today, or if its related to the changes to useClient and caching? Doesnt feel like its wrong per se, just curious

moped-toolbox/random_project_seeds/index.js
chiaberry: thousands!!! 🤩

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: not waiting on loading seems like a great improvement 😍

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
chiaberry: i actually don't mind the jump at all. maybe because it feels like a confirmation that the edit i just made to the project worked? im probably in the minority tho

chiaberry (approved):
this is such a huge user experience improvement -- not having the whole table show the loading icon when navigating back and forth, soooo much better. 

I kind of take back what i said about the project jumping up to the top of the table, since it doesnt happen all the time (because of the race condition). so yeah, now I do think it feels kinda jarring, but i dont hate it.

moped-editor/src/routes.js
```diff
@@ -121,3 +121,5 @@ export const restrictRoutes = (routes) =>
       return protectedRoute;
     }
   });
+
+export const restrictedRoutes = restrictRoutes(routes);
```
chiaberry, line 125: this does make more sense to export it like this, instead of exporting routes and the function and making the `restrictedRoutes` in App.js, but is there a performance benefit to doing it like this?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
mddilley: Nice! It cool to see `refetchQueries` in action after reading a little about it. 🚀 It is always interesting to face an issue caused by previous logic that was doing just fine on its own! I agree about the DB trigger as a fix.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -164,9 +165,10 @@ const ProjectView = () => {
     : "/moped/projects";
 
   // Get the tab query string value and associated tab index.
-  // If there's no query string, default to first tab in TABS array
-  let activeTabIndex = !!query.get("tab")
-    ? TABS.findIndex((tab) => tab.param === query.get("tab"))
+  // If there's no query string (or the string does not match
```
mddilley, line 168: Glad that you addressed this and that I steered away from messing with this code to fix the project component links. Leaning further into React Router `useSearchParams` would have mega broken this! 😵‍💫

moped-editor/src/views/projects/projectView/ProjectView.js
mddilley: Good to finally see a path to fixing the repeated queries for those lookups! That will be huge. 🙌

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
mddilley: I'm on the fence but lean towards shipping with a focus on fixes coming. I would argue that it is more broken/unoptimized without these changes than with them and having a quirk, but I get that having unexpected things happen in the UI can be annoying and confusing.

moped-editor/src/views/projects/projectsListView/helpers.js
mddilley: I think this is fine and works within the boundaries of what it takes to optimize this table! It also maintains the different pieces that fit into the material table component which I think is still straightforward to see.

mddilley (approved):
This all looks so great to me! The user interactions on the project list view are so fast now. 🏎️💨 Also cool to see a situation where tools like useMemo makes a dramatic difference. A real life example of that situation that is so hard to define in docs! 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeFeatureCollections.js
johnclary: derp—thanks for flagging this. i will update the import :/

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
johnclary: ok—agreed. we'll ship this and continue to make it better 😅

moped-editor/src/views/projects/projectsListView/helpers.js
johnclary: ok—thanks, Mike!

moped-editor/src/routes.js
johnclary: yes—it just seemed like a red flag to have the routing object at the very root of our app be re-constructed on every App render. i doubt it's a noticeable performance improvement, but these little performance leaks do accumulate over time 🤷‍♂️

PR #1170, opened 2023-10-19
[Enhancement] Track if phase start and end dates are tentative or confirmed

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/14300

## Testing
**URL to test:** 

local! 

**Steps to test:**

`select * from moped_proj_phases where is_phase_end_confirmed is true or is_phase_start_confirmed is true`

From the timelines tab:
1. add a new phase. set the start date and/or the end date to a date in the past. Run the above query in your db client to see if your phase appears. 
2. edit an existing phase! see if it appears in the results of your query
3. repeat the steps with a date in the future. phase end and phase start confirmed is false

---
#### Ship list
- [ ] Code reviewed 


johnclary (changes requested):

moped-database/migrations/1697679713991_confirmed_date_trigger/up.sql
```diff
@@ -0,0 +1,25 @@
+-- For each of phase_start and phase_end date, if the date is on or before 
+-- the current date when the record is created or edited then set corresponding
+-- confirmation column to true.
+CREATE OR REPLACE FUNCTION moped_proj_phases_confirmed_dates() RETURNS trigger
+    LANGUAGE plpgsql
+    AS $$ 
+    BEGIN
```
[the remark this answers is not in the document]
johnclary: as Chia and I have been discussing in Slack:

we are saving a timezone-naive date through Moped editor, which stems from [this handler](https://github.com/cityofaustin/atd-moped/blob/19c5556c8d1ab3db9a733ccd2e89d736f75384d1/moped-editor/src/views/projects/projectView/DateFieldEditComponent.js#L15C8-L15C8):

```javascript
format(date, "yyyy-MM-dd")
```

however, the moped database (local, staging, prod) are all on UTC time. (you can verify with the SQL command `show timezone;`. the result is that this function is going to compare the user's local date against the current date in UTC.

we need to stop using `date` columns and use `timestamp with time zone` columns instead. i think we need to live with this bug and address this issue separately.

moped-editor/src/queries/project.js
```diff
@@ -47,6 +47,7 @@ export const SUMMARY_QUERY = gql`
         feature_signals(where: { is_deleted: { _eq: false } }) {
           signal_id
           knack_id
+          id
```
chiaberry, line 50: this is unrelated to this issue, but I forgot to query for the feature id in the summary query when I fixed the unique key error https://github.com/cityofaustin/atd-moped/pull/1143

johnclary: thanks for catching that!

moped-database/migrations/1697679713991_confirmed_date_trigger/up.sql
```diff
@@ -0,0 +1,25 @@
...
+        IF NEW.phase_start <= current_date AND NEW.is_phase_start_confirmed != TRUE THEN
+            new.is_phase_start_confirmed := true;
+        END IF;
+        IF NEW.phase_end <= current_date AND NEW.is_phase_end_confirmed != TRUE  THEN
+            new.is_phase_end_confirmed := true;
+        END IF;
+        IF NEW.phase_start > current_date THEN
```
johnclary, line 14: here, too, you could add the second condition, e.g. `and NEW.is_phase_start_confirmed != FALSE`

johnclary (approved):
🚢 🙏

mddilley:
@chiaberry Still reviewing this but I wanted to check on one thing. I noticed that if I: 
1. create a new phase with a start date in the past and no end date, then save. -> `is_phase_start_confirmed = TRUE` ✅
2. edit to add an end date in the past, then save. `is_phase_end_confirmed = TRUE` ✅
3. then, edit and update both the start date and end date to be in the future, then save. `is_phase_end_confirmed = FALSE` and `is_phase_end_confirmed = TRUE`

Is that what we expect to see?

chiaberry:
@mddilley nope, not at all. Thanks for catching my mistake. pushed the fix

mddilley:
oh nice! That was a needle in a haystack in the code. 🔍 Lots of time comparisons happening!

johnclary:
thanks for catching that, Mike!! 🚢 🚢 🚢 🚢

johnclary (approved):
🚢

PR #1172, opened 2023-10-20
Prep DB for rolling back the Access DB migration

johnclary:
This PR addresses the need to be able to cleanly rollback the Access DB migration.

- Add a  field will allow us totrack any projects created through the Access DB migration so that we can delete them if needed. We can delete this column after we successfully migrate. 
- Disable Hasura event triggers on `delete` operations. We do not allow records to be deleted through Hasura except by our admin/root user. This change removes Hasura `delete` event triggers entirely, because the only conditions under which we are deleting records will be for admin purposes that we presumably to not want to track as events.

In order to completely roll back the migration we'll need to run [these commands](https://github.com/cityofaustin/atd-moped/blob/4838-jc-migrate-access-db/moped-toolbox/access_migration/migrate_data/rollback.sql). I intend to package that up as a single executable.

Steps to test
1. Just start your local DB
2. Use a SQL make sure that deleting project data does not trigger events:

```sql
-- clear all record from the event invocation lgos
truncate hdb_catalog.event_invocation_logs;

-- delete all projects;
delete from moped_project where 1=1;

-- verify no events were invoked from the deletion
select * from hdb_catalog.event_invocation_logs;
```

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -145,8 +145,6 @@
   event_triggers:
     - name: activity_log_feature_drawn_lines
       definition:
-        delete:
-          columns: '*'
         enable_manual: false
         insert:
           columns: '*'
@@ -280,8 +278,6 @@
   event_triggers:
     - name: activity_log_feature_drawn_points
       definition:
-        delete:
-          columns: '*'
         enable_manual: false
         insert:
           columns: '*'
@@ -415,8 +411,6 @@
   event_triggers:
     - name: activity_log_feature_intersections
       definition:
-        delete:
-          columns: '*'
         enable_manual: false
         insert:
           columns: '*'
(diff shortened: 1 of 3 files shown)


mddilley (approved):
Spun up locally and the migrations check out! 🚀

frankhereford (approved):
Starts right up! Code looks great - thank you John.

frankhereford:
@chiaberry pointed out to me that you had removed your request for reviews! Sorry for jumping the gun. I thought you had invited the dev group and it had removed the rest of us when Mike reviewed. My bad - looking forward to this in final form.

Thanks John!

johnclary:
Thanks for the quick reviews Frank and Mike! I just added another change to this PR, updated the test steps, and requested review again. Would y'all mind looking at this again at your convenience?

mddilley (approved):
Tested locally, and I see that record deletes no longer invoke activity log event triggers. 🚀 🚢

chiaberry (approved):
same! test steps check out

johnclary:
Thank you!

PR #1174, opened 2023-10-23
Backfill phase start/end date confirmation status

chiaberry:
## Associated issues

from https://github.com/cityofaustin/atd-data-tech/issues/14350: 

Since we https://github.com/cityofaustin/atd-data-tech/issues/14300 - we need to backfill the existing project phases based on the same logic. For both phase_start and phase_end, if the date is today or earlier it's corresponding confirmation flag can be set to true.

## Testing
**URL to test:** 

**Steps to test:**

I tested by loading prod data, running the migrations and then checking 
`select phase_start, is_phase_start_confirmed, project_phase_id from moped_proj_phases mpp where phase_start > current_date` and `select phase_start, is_phase_start_confirmed, project_phase_id from moped_proj_phases mpp where phase_start < current_date` to see if they made sense. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1698070704301_backfill_confirmed_phase_dates/down.sql
+++ b/moped-database/migrations/1698070704301_backfill_confirmed_phase_dates/down.sql
@@ -1,2 +1,2 @@
-UPDATE moped_proj_phases SET is_phase_start_confirmed = false WHERE phase_start <= current_date;
-UPDATE moped_proj_phases SET is_phase_end_confirmed = false WHERE phase_end <= current_date;
+-- can't undo
+select 0;
--- a/moped-database/migrations/1698070704301_backfill_confirmed_phase_dates/down.sql
+++ b/moped-database/migrations/1698070704301_backfill_confirmed_phase_dates/down.sql
@@ -1,0 +1,2 @@
+UPDATE moped_proj_phases SET is_phase_start_confirmed = false WHERE phase_start <= current_date;
+UPDATE moped_proj_phases SET is_phase_end_confirmed = false WHERE phase_end <= current_date;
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1698070704301_backfill_confirmed_phase_dates/down.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE moped_proj_phases SET is_phase_start_confirmed = false WHERE phase_start <= current_date;
+UPDATE moped_proj_phases SET is_phase_end_confirmed = false WHERE phase_end <= current_date;
```
chiaberry, line 2: actually, I don't think these down migrations will work, because the trigger exists that will automatically change these upon update. 

should i change this file to 

```
-- can't undo
select 0;
```

frankhereford: I think one thing you could do here to work around that trigger is something like prepending this

```
ALTER TABLE moped_proj_phases DISABLE TRIGGER <trigger name>;
```

and then doing the updates, and then enabling it again

```
ALTER TABLE moped_proj_phases ENABLE TRIGGER <trigger name>;
```

johnclary: i'm cool with having an empty down migration on this one. If we're going to undo anything it will involve nuking the triggers entirely.

Thanks for that example, though, Frank. Trigger trickery is going to be increasingly necessary in the Moped DB!

johnclary (approved):
I tested against a local copy of prod and it worked flawlessly. Updating the down migration to select `0` would be good.

frankhereford (approved):
I think this is right on the money! 💯 Thank you Chia!

mddilley (approved):
Tested locally, and all queries for start and end dates and the confirmed booleans match up! 🙏 🚀

PR #1176, opened 2023-10-24
Hide the note type selector when creating a status update

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/14381

## Testing
**URL to test:** 
https://deploy-preview-1176--atd-moped-main.netlify.app/moped/ 

**Steps to test:**
1. Go to the summary tab of a project, click on the status update.
2. You shouldn't see the note type radio buttons here when creating a new status in the modal or when editing an old status from here.
3. Test making a new status from this modal, make sure it saves correctly.
4. Test editing an old status from here and make sure it saves correctly. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectNotes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectNotes.js
@@ -101,7 +101,7 @@ const ProjectNotes = (props) => {
   const classes = useStyles();
   const userSessionData = getSessionDatabaseData();
   const [noteText, setNoteText] = useState("");
-  const [newNoteType, setNewNoteType] = useState(1);
+  const [newNoteType, setNewNoteType] = useState(isStatusEditModal ? 2 : 1);
   const [editingNoteType, setEditingNoteType] = useState(null);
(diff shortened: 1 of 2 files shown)


chiaberry (approved):
test steps check out 🚢

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -306,6 +306,7 @@ const ProjectNotes = (props) => {
                 submitNewNote={submitNewNote}
                 submitEditNote={submitEditNote}
                 cancelNoteEdit={cancelNoteEdit}
+                isStatusEditModal={isStatusEditModal}
```
mddilley, line 309: If we were using this input component in _way_ more places in the app, I could see naming this prop something like `hideTypeToggle` or `showTypeToggle` so that `NoteInputQuill` would only need to know whether or not to show the toggle (and less about the specific context like status modal or not status modal). No change needed here!

johnclary (approved):
Thank you! 🚢

PR #1178, opened 2023-10-25
[Bug] Cannot add a new advanced filter when the operator type is "blank/not blank"

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/14302

## Testing
**URL to test:** 

https://deploy-preview-1178--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

Open the advanced search dropdown
Pick a field that has "is blank" or "is not blank" as an option, such as Team, Partners or Inspector. 
Pick is blank or is not blank as an operator. You should be able to add another filter, in staging/prod right now you cannot. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -578,10 +579,14 @@ const Filters = ({
...
-                            value[filterParameters[filterId].lookup_field]
-                          );
+                          if (value) {
+                            handleSearchValueChange(
+                              filterId,
+                              value[filterParameters[filterId]?.lookup_field]
+                            );
+                          } else {
+                            handleSearchValueChange(filterId, value);
+                          }
```
chiaberry, line 589: this fixes an error i came across that happens if you select a value from an autocomplete and then use the x to remove it. 
Screenshot 2023-10-25 at 3 59 34 PM

mddilley (approved):
Thanks for fixing this! I tested on the deploy and can now stack more filters on top of the blank or not blank operators. 🚢 🙌

moped-editor/src/components/GridTable/Filters.js
johnclary: would you add a comment in the `else` block about what's going on? no value here is just `""`?

PR #1180, opened 2023-10-26
Restores ability to use back button from the project map tab

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14434

## Testing
**URL to test:**  Netlify

1. Visit the project summary for any project. Click around on the map tab, then use your browser's back button to return to the previous tab.
2. Repeat, but use the "Close Map" button to return to the project summary tab.
3. Finally, try navigating to each tab, then use your browser's back button a bunch to make sure you are returned to each previous tab correctly.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectView.js
+++ b/moped-editor/src/views/projects/projectView/ProjectView.js
@@ -228,7 +228,14 @@ const ProjectView = () => {
   const handleChange = useCallback(
     (event, newTab) => {
       if (newTab === 0) refetch();
-      setSearchParams({ tab: TABS[newTab].param });
+      const newTabParam = TABS[newTab].param;
+      setSearchParams(
+        { tab: TABS[newTab].param },
+        // replace history unless navigating to the map tab
+        // so users can use their browser's back button
+        // to return to project summary page
+        { replace: newTabParam !== "map" }
+      );
     },
     [refetch, setSearchParams]
   );
--- a/moped-editor/src/views/projects/projectView/ProjectView.js
+++ b/moped-editor/src/views/projects/projectView/ProjectView.js
@@ -1,7 +1,11 @@
-import React, { useState, useCallback, useContext } from "react";
+import React, { useState, useCallback, useContext, useMemo } from "react";
 import { useMutation, useQuery } from "@apollo/client";
-import { Link as RouterLink, useParams, useLocation } from "react-router-dom";
-import { createBrowserHistory } from "history";
+import {
+  Link as RouterLink,
+  useParams,
+  useSearchParams,
+  useLocation,
+} from "react-router-dom";
 import makeStyles from "@mui/styles/makeStyles";
 
 import {
@@ -129,9 +133,10 @@ function a11yProps(index) {
   };
 }
 
-function useQueryParams() {
-  return new URLSearchParams(useLocation().search);
-}
+const useQueryParams = (search) =>
+  useMemo(() => {
+    return new URLSearchParams(search);
+  }, [search]);
 
 const TABS = [
   { label: "Summary", Component: ProjectSummary, param: "summary" },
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -157,36 +171,22 @@ const history = createBrowserHistory();
  */
 const ProjectView = () => {
   const { projectId } = useParams();
-  let query = useQueryParams();
-  const classes = useStyles();
-  const previousFilters = useLocation()?.state?.filters;
+  let [searchParams, setSearchParams] = useSearchParams();
+  const location = useLocation();
+  const activeTab = useActiveTabIndex(searchParams.get("tab"));
+  const previousFilters = location.state?.filters;
```
johnclary, line 177: it does not seem completely right that we need to all of `useLocation`, `useParams`, and `useSearchParams`—but i didn't want to wade too far into reworking all this 🤷 

in any case, `useSearchParams` seems to be the ticket—it allows us to update search params and listen to changes just like any other state variable.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -219,18 +219,16 @@ const ProjectView = () => {
    */
   const handleChange = useCallback(
     (event, newTab) => {
+      setSearchParams({ tab: TABS[newTab].param });
```
johnclary, line 222: setting search params does push to your browser history, so that back button works as expected.

This function does take an options object that let's you replace the history instead of add to it. E.g.

```javascript
 setSearchParams({ tab: "summary" }, { replace: true });
```

@mddilley i was thinking this might actually be something useful for the project map search params. i feel like i need to spend a full day with the react-router API to get to know all the v6 operations.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -148,7 +148,22 @@ const TABS = [
...
+
+/**
+ * Get the index of the currently active tab
+ * @param {*} tabName - a `tab` name from the url search string
+ * @returns {integer} - the TAB index of the currently active tab, falling back to `0`
+ */
+const useActiveTabIndex = (tabName) =>
+  useMemo(() => {
+    const activeTabIndex = TABS.findIndex((tab) => tab.param === tabName);
+    return activeTabIndex > -1 ? activeTabIndex : 0;
```
johnclary, line 165: if an unknown tab name is used in the search params, we fallback to the summary tab. this behavior was perviously broken. E.g.: https://moped.austinmobility.io/moped/projects/1921?tab=pizza

chiaberry: its not falling back to the summary tab for me.

chiaberry: oh duhh b/c i clicked your link and not the netlify link. ignore me.

chiaberry (approved):
I can go back and forth and not get stuck on the map. 🚢

moped-editor/src/views/projects/projectView/ProjectView.js
mddilley: nice! thanks for making some changes here because I backed away from changing this recently too. Would be great to have a tech debt issue for nav! iirc there are some in-app links that don't use react router too if we haven't squashed them all yet

moped-editor/src/views/projects/projectView/ProjectView.js
mddilley: TIL - didn't catch that option in the docs 🙏

mddilley (approved):
Thanks for fixing this! Good to be leaning more into react router tools here 🚀 🙌

johnclary:
thank you!

PR #1186, opened 2023-10-30
Adds another activity log record type for projects updated via the migration

johnclary:
I've added a separate activity log entry for cases where we have updated an existing project with Access DB data. there are not going to be any other events for these, because the access DB 

Steps to test:
1. Take a look at the code
2. I'm pushing this to the Access migration [test instance](https://4838-jc-migrate-access-db--atd-moped-main.netlify.app/moped/projects). So you can check out [project ID 280](https://4838-jc-migrate-access-db--atd-moped-main.netlify.app/moped/projects/280?tab=activity_log).

Here's the new event I'm adding for projects updated via the migration:
Screenshot 2023-10-30 at 4 38 31 PM

And here's the original event that we have for a project created through the migration:
Screenshot 2023-10-30 at 4 38 42 PM


chiaberry (approved):

mddilley (commented):
Looks great - checked it out in the activity log of the test instance. ✨

PR #1187, opened 2023-10-30
Ability to advanced search using multiple filters as "or"/"any" condition

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11282

This PR:
- adds an advanced search radio group to select either **and** or **or** logic
- updates `useAdvancedSearch` to handle and apply that logic to the advanced filters in the GraphQL query
- updates the code to use React Router [useSearchParams](https://reactrouter.com/en/main/hooks/use-search-params#usesearchparams) to manage the url query strings

## Testing
**URL to test:** 

https://deploy-preview-1187--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to the project list view in the deploy preview
2. Add an advanced filter by click the filter icon in the search box
3. Try searching for:
- **ALL**
- **Lead is COA ATD Project Delivery**
- **Sponsor is COA ATD Active Transportation & Street Design**
- Notice that there are radios for **all** and **any** above the advanced filter rows. Keep it as **all**
- You should see one result that created that is named **#11282 test project** that matches both of the filters
- Check the url and you should see two query parameters: **filter** (with the encoded filter string) and **isOr** which should have the value of **false**
4. Now, try searching for:
- **ANY**
- **Lead is COA ATD Project Delivery**
- **Sponsor is COA ATD Active Transportation & Street Design**
- Change the radios to match **any**
- You should see many results that meet either of the two filters
- Check the url and you should see two query parameters: **filter** (with the encoded filter string) and **isOr** which should have the value of **true**
5. Try combinations of advanced filters like **projects where Arterial Management is either the Lead, Sponsor, or a Partner** from the issue, and try to break it! (You might run into https://github.com/cityofaustin/atd-data-tech/issues/14495 - that fix is coming in the next PR)
6. Try copying your url with **filter** and **isOr** parameters into a different tab and the UI should reflect the advanced filters stored in the encoded string with the correct any/all option selected in the radio buttons
7. Try applying advanced filters to the view and then entering a search string into the search box and pressing enter. You should see the advanced filter UI reset and the query parameters removed.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -21,6 +21,8 @@ import {
 import makeStyles from "@mui/styles/makeStyles";
 
 import { Autocomplete } from "@mui/material";
+import { ToggleButton } from "@mui/material";
+import { ToggleButtonGroup } from "@mui/material";
 import BackspaceOutlinedIcon from "@mui/icons-material/BackspaceOutlined";
 import { LOOKUP_TABLES_QUERY } from "../../queries/project";
 
@@ -138,6 +140,8 @@ const Filters = ({
   filtersConfig,
   setSearchFieldValue,
   setSearchTerm,
+  isOr,
+  setIsOr,
 }) => {
   /**
    * The styling of the search bar
@@ -459,14 +463,41 @@ const Filters = ({
 
   return (
     <Grid>
-      <Grid container justifyContent={"flex-end"}>
-        <IconButton
-          onClick={handleAdvancedSearchClose}
-          className={classes.closeButton}
-          size="large"
+      <Grid container className={classes.gridItemPadding}>
+        <Grid
+          item
+          xs={6}
+          className={classes.filtersContainer}
+          display="flex"
+          justifyContent="flex-start"
         >
-          <Icon fontSize={"small"}>close</Icon>
-        </IconButton>
+          <ToggleButtonGroup
+            color="primary"
+            size="small"
+            value={isOr ? "any" : "all"}
+            exclusive
+            onChange={(e) => setIsOr(e.target.value === "any")}
+            aria-label="And or toggle"
+          >
+            <ToggleButton value="all">all</ToggleButton>
+            <ToggleButton value="any">any</ToggleButton>
+          </ToggleButtonGroup>
+        </Grid>
+        <Grid
+          item
+          xs={6}
+          className={classes.filtersContainer}
+          display="flex"
+          justifyContent="flex-end"
+        >
+          <IconButton
+            onClick={handleAdvancedSearchClose}
+            className={classes.closeButton}
+            size="large"
+          >
+            <Icon fontSize={"small"}>close</Icon>
+          </IconButton>
+        </Grid>
       </Grid>
       {Object.keys(filterParameters).map((filterId) => {
         return (
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -483,11 +483,14 @@ const Filters = ({
   };
 
   console.log(filterParameters);
+  /* First filter is an empty placeholder so we check for more than one filter */
+  const areMoreThanOneFilters = Object.keys(filterParameters).length > 1;
+  const isFirstFilterComplete =
+    Object.keys(filterParameters).length === 1 && filterComplete;
 
   return (
     <Grid>
-      {/* First filter is an empty placeholder so we check for more than one filter */}
-      {Object.keys(filterParameters).length > 1 ? (
+      {areMoreThanOneFilters ? (
         <Grid container className={classes.gridItemPadding}>
           <Grid
             item
@@ -698,10 +701,7 @@ const Filters = ({
               <Hidden mdDown>
                 <Grid item xs={12} md={1} style={{ textAlign: "center" }}>
                   <IconButton
-                    disabled={
-                      Object.keys(filterParameters).length === 1 &&
-                      !filterComplete
-                    }
+                    disabled={!isFirstFilterComplete}
                     className={classes.deleteButton}
                     onClick={() => handleDeleteFilterButtonClick(filterId)}
                     size="large"
@@ -713,10 +713,7 @@ const Filters = ({
               <Hidden mdUp>
                 <Grid item xs={12}>
                   <Button
-                    disabled={
-                      Object.keys(filterParameters).length === 1 &&
-                      !filterComplete
-                    }
+                    disabled={!isFirstFilterComplete}
                     fullWidth
                     className={classes.deleteButton}
                     variant="outlined"
@@ -746,17 +743,15 @@ const Filters = ({
           </Button>
         </Grid>
         <Grid item xs={12} md={1}>
-          {Object.keys(filterParameters).length > 0 && (
-            <Button
-              className={classes.bottomButton}
-              fullWidth
-              variant="outlined"
-              startIcon={<BackspaceOutlinedIcon />}
-              onClick={handleClearFilters}
-            >
-              Reset
-            </Button>
-          )}
+          <Button
+            className={classes.bottomButton}
+            fullWidth
+            variant="outlined"
+            startIcon={<BackspaceOutlinedIcon />}
+            onClick={handleClearFilters}
+          >
+            Reset
+          </Button>
         </Grid>
         <Hidden mdDown>
           <Grid item xs={12} md={7}>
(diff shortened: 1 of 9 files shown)


mddilley:
@johnclary Request you first to check in on the initial design

johnclary:
@mddilley just realized i never circled back on the toggle design. let's go with the radio buttons instead of the button toggle. 
radio (1)

could you hide these until a second filter is added?

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -435,10 +461,11 @@ const Filters = ({
   useEffect(() => {
     if (Object.keys(filterParameters).length === 0) {
       setFilters(filterParameters);
+      handleClearFilters();
       // Add an empty filter so the user doesn't have to click the 'add filter' button
       generateEmptyFilter();
     }
-  }, [filterParameters, setFilters, generateEmptyFilter]);
+  }, [filterParameters, setFilters, generateEmptyFilter, handleClearFilters]);
```
mddilley, line 468: I thought about refactoring this so that the initial filter state has no filter objects in it instead of a blank placeholder with uuid, but I think that is too big and this works for what we are looking to do. It just makes it a little difficult to rely on the count to show/hide or anything else. I ended up creating some booleans vars that describe what is going on more and removing some checks that weren't actually doing anything but idk.

**Initial state**
```
{
    "id": "79f79105-4718-4df4-a614-fb665ec12b52",
    ...emptyFilterDetailsAsPlaceholder
}
```

**State after completing the first filter**
```
{
    "id": "79f79105-4718-4df4-a614-fb665ec12b52",
    ...completeFilterDetails
}
```

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -676,38 +745,34 @@ const Filters = ({
           </Button>
         </Grid>
         <Grid item xs={12} md={1}>
-          {Object.keys(filterParameters).length > 0 && (
```
mddilley, line 679: This check wasn't doing anything because the `useEffect` above makes sure that the `filterParameters` object always has at least one key.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -676,38 +745,34 @@ const Filters = ({
...
+            Reset
+          </Button>
         </Grid>
         <Hidden mdDown>
           <Grid item xs={12} md={7}>
             {""}
           </Grid>
         </Hidden>
         <Grid item xs={12} md={2}>
-          {Object.keys(filterParameters).length > 0 && (
```
mddilley, line 697: Same here, the `Grow` component below has a prop that uses `handleApplyValidation` to control whether it grows in (visible) or not (hidden). The `handleApplyValidation` helper also creates feedback strings about why filters are not valid that we don't display anywhere.

chiaberry:
What do you yall think about changing the color of the labels on the radio buttons to the secondary text color, to match the rest of the labels? Is it too dim?
Screenshot 2023-11-07 at 9 45 20 AM

moped-editor/src/components/GridTable/Filters.js
chiaberry: 🧹  thankyou

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -132,20 +133,20 @@ const generateEmptyField = (uuid) => {
...
+  setIsOr,
 }) => {
   /**
    * The styling of the search bar
    * @constant
    * @default
    */
   const classes = useStyles();
-  const queryPath = useLocation().pathname;
+  let [, setSearchParams] = useSearchParams();
```
johnclary, line 149: TIL—i did not know you can destructure like this!

moped-editor/src/components/GridTable/Filters.js
johnclary: yeah, that is tempting for sure. i can see how the initial filter creates lots of headaches. i will capture this tech debt here: https://github.com/cityofaustin/atd-data-tech/issues/10538

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -62,12 +57,17 @@ const makeAdvancedSearchWhereString = (filters) =>
...
 export const useAdvancedSearch = () => {
-  const filterQuery = useFilterQuery(useLocation().search);
-  const initialFilterState = useMakeFilterState(filterQuery);
+  /* Get advanced filters settings from search params if they exist */
+  let [searchParams] = useSearchParams();
+  const initialFilterState = useMakeFilterState(searchParams);
+
+  /* Determine or/any from search params if it exists */
+  const isOrFromSearchParams = searchParams.get("isOr");
+  const initialIsOrState = isOrFromSearchParams === "true" ? true : false;
```
johnclary, line 69: ✅ i tried to break this feature by setting the url param to `isOr=pizza` and i observed that this was interpreted as the  "match all" setting

moped-editor/src/components/GridTable/Filters.js
johnclary: nice!

johnclary (changes requested):
@mddilley this is fantastic! 

i have encountered one small issue, which is that the "< All projects" links on the project view does not retain the `isOr` param setting.

Screenshot 2023-11-07 at 2 54 01 PM

i think that link state is coming from [here](https://github.com/cityofaustin/atd-moped/blob/45fa650bde12579c8663369eb9a2956e4f4e33b2/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js#L147). hopefully it's not too hard to get this new param in there.


otherwise, this is working perfectly and it's going to make our users v happy!

chiaberry (approved):
I left a note about a style tweak, but besides that it looks great! And I am glad you are refactoring the hooks used from react-router-dom. 

I encountered the existing bug that you are working on but couldnt find any other way to break it!

moped-editor/src/components/GridTable/Filters.js
chiaberry: same!! I was reading MDN and testing stuff.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -175,10 +175,10 @@ const ProjectView = () => {
...
   const location = useLocation();
   const activeTab = useActiveTabIndex(searchParams.get("tab"));
-  const previousFilters = location.state?.filters;
-  const allProjectsLink = !!previousFilters
-    ? `/moped/projects?filter=${previousFilters}`
-    : "/moped/projects";
+  const { queryString: previousProjectListViewQueryString } = location.state;
+  const allProjectsLink = !previousProjectListViewQueryString
+    ? "/moped/projects"
+    : `/moped/projects${previousProjectListViewQueryString}`;
```
mddilley, line 181: Bigger update here, instead of extending this to pass the state of `isOr` along with the state of `filters`, I updated this to get all query strings using React Router so that this will support any future params too.

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -128,8 +128,11 @@ export const useTableComponents = ({
 /**
  * The Material Table column settings
  */
-export const useColumns = ({ hiddenColumns, linkStateFilters, classes }) =>
-  useMemo(
+export const useColumns = ({ hiddenColumns, classes }) => {
+  const location = useLocation();
+  const queryString = location.search;
+
+  const columns = useMemo(
```
mddilley, line 135: This is the source of the `queryString` the comes through the React Router Links that carry state to `ProjectView` so we can link back to the previous filters.

mddilley:
@johnclary I added a couple of comments to explain how I fixed the Link that takes users back to the project list view from the project view with their previously applied filters in the url.

moped-editor/src/views/projects/projectView/ProjectView.js
johnclary: niiiice!

johnclary (approved):
It's fixed! I love seeing how much this code continues to improve. thanks @mddilley!!

mddilley:
thanks, y'all! merging 🚢

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -367,12 +385,16 @@ const Filters = ({
   /**
    * Clears the filters
    */
-  const handleClearFilters = () => {
+  const handleClearFilters = useCallback(() => {
```
roseeichelmann, line 388: i dont think this needs a `useCallback` anymore, in the past this function was called within a `useEffect` but now its not

PR #1191, opened 2023-11-02
Create shared use trail -  dual path component

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14518

## Testing
**URL to test:** Local

**Steps to test:**
1. Create a new map component using: `Trail - Shared use (Paved Dual Path)`
2. Check out the work type and subcomponent options. they should match the component called `Trail - Shared Use Path (Paved)`

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1698950636187_cacasde_component_subcomponent_deletes/up.sql
johnclary, line 1: this allows the component deletion down migration to work

moped-database/migrations/1698950636188_create_shared_use_trail_dual_path_component/up.sql
```diff
@@ -0,0 +1,62 @@
+INSERT INTO moped_components (
+    component_name,component_subtype,line_representation,feature_layer_id
+) VALUES
+    ('Trail', 'Shared Use Path (Paved Dual Trail)', TRUE, 4);
+
+-- insert all "protection type" subcomponents into moped_components_subcomponents
```
johnclary, line 6: this is a kinda fancy way of doing this that we probably want to book mark. it copies the subcomponent and work types of another similar component.

also, it does not rely on us knowing the `component_id` of the component we just created ☝️ instead we get that component ID by query on the component name + subtype.

moped-database/migrations/1698950636188_create_shared_use_trail_dual_path_component/up.sql
```diff
@@ -0,0 +1,43 @@
...
+) INSERT INTO moped_components_subcomponents (component_id, subcomponent_id)
+SELECT
+    mc.component_id, inserts_todo.subcomponent_id
+FROM
+    inserts_todo
+    -- gets the component id of the new component we created
+    LEFT JOIN moped_components mc ON mc.component_name = inserts_todo.component_name
+        AND mc.component_subtype = inserts_todo.component_subtype;
+
+-- insert the same moped_component_work_types work types as the Trail - Shared Use Path (Paved) component
```
johnclary, line 25: this is a kinda fancy way of doing this that we probably want to bookmark. it copies the subcomponent and work types of another similar component.

also, it does not rely on us knowing the `component_id` of the component we just created ☝️ instead we get that component ID by query on the component name + subtype.

chiaberry (approved):
Tested and see the new component type and options -- the sql is fancy and im definitely going to bookmark it

moped-database/migrations/1698950636188_create_shared_use_trail_dual_path_component/up.sql
mddilley: this is really nice, and I agree! 🔖

moped-database/views/project_list_view.sql
```diff
@@ -1,6 +1,4 @@
--- latest version 1697768359465_dedupe_plv_tks
-DROP VIEW project_list_view;
```
mddilley, line 2: iirc there is an edge case that this handled for when replacing a view can fail. I'm going to think about it some more, but it could very well be an artifact that we keep carrying to future versions. 🤔

mddilley: Found this link, and it sounds like what I remember https://dba.stackexchange.com/questions/62813/why-is-it-not-possible-to-alter-a-view-with-a-create-or-replace

mddilley (approved):
Looks great! I see the new component and corresponding subcomponents and work types. It was fun checking out the sql and explore the changes in the DB. ✨ Thanks!

moped-database/views/project_list_view.sql
mddilley: just realized this is coming from a different PR. I'll add these comments there!

chiaberry: yes! this is my fault. When i dropped the project list view, it said that i had to cascade all the other views. but since i was just adding a new column, and not removing or renaming a column, i found that i didnt have to drop the view to replace it

johnclary:
thanks y'all! i merged Chia's branch here but the diff is still wrong :/ the commits all look good so i'm going to proceeed

johnclary:
i merged main and that cleared it up! 🚢 🚢 🚢 🚢

PR #1195, opened 2023-11-07
Add and edit phases using a form instead of inline table editing

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14562
- https://github.com/cityofaustin/atd-data-tech/issues/14342
- https://github.com/cityofaustin/atd-data-tech/issues/15225

This includes the migrations from https://github.com/cityofaustin/atd-moped/pull/1211 and is therefore blocked by it.

## Testing
**URL to test:**  Local

**Steps to test:**
1. `npm install` to install the latest version of the MUI DatePicker
2. Start Moped, ideally with a copy of production. 
3. Create a new project. On the timeline tab, verify that the "potential" phase has a start date of today and no asterisk (`*`) next to it.
4. Delete the potential phase and verify the project status badge changes to **Unknown**
5. Use the **ADD PHASE** menu to open the **Add phase** form. There is a lot to test here:
- Selecting a **Phase** option will reset the **Subphase options**
- If you select a **Start** or **End** date in the past, the **Confirmed** checkbox will default to checked, and vice-versa.
- If you type a date into a date field (rather than selecting with the date picker), it is not possible to save an invalid/partial date.
- It is not possible to save the form without selecting a **Phase**
- If the **Current phase** toggle is checked, it is not possible to save the form without a start date
6. After testing out the form behavior ☝️, save a **current** phase with a confirmed start date and a not-confirmed end date. 
7. The project status should match the phase you selected, and there should be an asterisk next to the phase end date.
8. Hover over the end date to verify the popover appears and says **Estimated end date**
9. Edit the phase by editing all fields and resting ☝️ the expected behaviors.
10. Add a few more phases to the project.
11. On the phase table, test out sorting by start date and end date, and other column headers
12. Lastly, to verify the timestamp migrations, you can inspect those dates in your SQL client, which should all be showing as midnight US/Central time (with a 5 or 6-hour offest). For example:

Before:
before

After:
after

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectPhase/form.js
```diff
@@ -0,0 +1,47 @@
...
+    .required("Phase is required"),
+  subphase_id: yup.number().nullable().optional(),
+  phase_start: yup.string().nullable().optional(),
+  phase_end: yup.string().nullable().optional(),
+  is_current_phase: yup.boolean(),
+  is_phase_start_confirmed: yup.boolean(),
+  is_phase_end_confirmed: yup.boolean(),
+  description: yup
+    .string()
+    .max(500, "Must be less than 500 characters")
```
johnclary, line 16: i think we've talked about this before, but max string length is not something we enforce at the db level, and we probably should 🙃

moped-database/migrations/1704906960001_phase_timestamps/up.sql
```diff
@@ -0,0 +1,496 @@
...
+    SELECT
+        project_phase_id,
+        phase_start AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago' AS phase_start,
+        phase_end AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago' AS phase_end
+    FROM
+        moped_proj_phases) AS dates_to_fix
+WHERE
+    moped_proj_phases.project_phase_id = dates_to_fix.project_phase_id;
+
+-- replace timezone of component completion dates which were imported from Access DB as UTC
```
johnclary, line 30: fixes https://github.com/cityofaustin/atd-data-tech/issues/15225

moped-database/migrations/1704906960001_phase_timestamps/up.sql
```diff
@@ -0,0 +1,496 @@
+-- this migration makes the `moped_proj_phases` timezone aware and corrects the existing date
+-- values which are in central time but stored as UTC
+
+-- we must drop dependent views to change the column types :/
+DROP VIEW IF EXISTS component_arcgis_online_view;
+DROP VIEW IF EXISTS project_list_view;
+
+-- change column types from dates to timestamps with timezones
```
johnclary, line 8: fixes https://github.com/cityofaustin/atd-data-tech/issues/14342

moped-database/migrations/1704906960001_phase_timestamps/up.sql
```diff
@@ -0,0 +1,496 @@
...
+    phase_end = dates_to_fix.phase_end
+FROM (
+    SELECT
+        project_phase_id,
+        phase_start AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago' AS phase_start,
+        phase_end AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago' AS phase_end
+    FROM
+        moped_proj_phases) AS dates_to_fix
+WHERE
+    moped_proj_phases.project_phase_id = dates_to_fix.project_phase_id;
```
johnclary, line 28: this migration is taking `date` values that look like `2023-01-01` and converting them to `timestamptz` values that look like `2023-01-01T00:05:00+00:00` (with the correct timezone offset

moped-database/migrations/1704906960001_phase_timestamps/up.sql
```diff
@@ -0,0 +1,496 @@
...
+    FROM
+        moped_proj_components
+    WHERE
+        -- if the timestamp is 00:00:00 it was imported incorrectly as UTC time
+        completion_date IS NOT NULL
+        AND extract(hour FROM completion_date) = 0) as dates_to_fix
+WHERE
+    moped_proj_components.project_component_id = dates_to_fix.project_component_id;
+
+-- rebuild dropped views to version 1704744986000_substantial_completion_date
```
johnclary, line 48: there are no changes to the views—they are just dropped and re-added

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -0,0 +1,183 @@
...
+};
+
+export const onSubmitPhase = ({
+  data,
+  mutate,
+  currentPhaseIdsToClear,
+  onSubmitCallback,
+}) => {
+  const { project_phase_id } = data;
+  delete data.project_phase_id;
```
mddilley, line 163: I've been recently using destructuring with rest to do this. `const { project_phase_id, ...restOfData } = data;`. It works both ways - just reminded me of code I was writing the other day.

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -0,0 +1,292 @@
...
+  const currentPhaseIdsToClear = useCurrentPhaseIdsToClear(
+    phase.project_phase_id,
+    isCurrentPhase,
+    currentProjectPhaseIds
+  );
+
+  const [phase_start, phase_end] = watch(["phase_start", "phase_end"]);
+
+  /**
+   * Defaults is_phase_start_confirmed to true if date is today or before
```
mddilley, line 73: One question that I thought about when testing the UI, do we need to enforce the checking of the box when the start/end date is today or before? To put it another way, should users be able to uncheck the confirmed box after this hook checks it?

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -0,0 +1,183 @@
...
+      is: true,
+      then: (schema) =>
+        schema.required("Start date is required when phase is current"),
+    })
+    .typeError("Invalid Date"),
+  phase_end: yup.date().nullable().optional().typeError("Invalid Date"),
+  is_current_phase: yup.boolean(),
+  is_phase_start_confirmed: yup.boolean(),
+  is_phase_end_confirmed: yup.boolean(),
+  description: yup
```
mddilley, line 24: **Change request** I just tested this rule in the form, and I think this key needs to be updated to `phase_description`. I was able to submit more than 500 chars and couldn't get the error message to appear.

moped-editor/src/views/projects/projectView/ProjectPhases.js
```diff
@@ -1,411 +1,232 @@
...
+            <Box display="flex">
+              <CheckCircleOutline style={{ color: green[500] }} />
+            </Box>
+          ) : (
+            ""
+          ),
+      },
+      {
+        headerName: "",
+        field: "_edit",
```
mddilley, line 118: **Possible change requested** I noticed that this field name shows up when using the "Manage Columns" menu to show/hide columns. Not sure if this is a big deal or even able to be changed.
Screenshot 2024-01-16 at 4 21 35 PM

Screenshot 2024-01-16 at 4 21 43 PM

moped-editor/src/views/projects/projectView/ProjectPhases.js
```diff
@@ -1,411 +1,232 @@
...
+        <ProjectPhaseDialog
+          phase={editPhase}
+          onClose={() => setEditPhase(null)}
+          onSubmitCallback={onSubmitCallback}
+          phases={data?.moped_phases}
+          currentProjectPhaseIds={currentProjectPhaseIds}
+          projectId={projectId}
+        />
+      )}
       <PhaseTemplateModal
```
mddilley, line 227: **Requested change** When I tested adding from template, I'm seeing an error related to the `current_phase_ids_to_clear`. Probably a small adjustment needed to the state or mutation that fires.
Screenshot 2024-01-16 at 4 28 30 PM

mddilley (changes requested):
This looks so great! Really cool to see another form modal in the mix. I left mostly trivial comments, but I marked a few changes to look at. They seem like small fixes - thanks for knocking out this chunk of work! 🙌

moped-editor/src/views/projects/newProjectView/NewProjectView.js
```diff
@@ -145,7 +145,8 @@ const NewProjectView = () => {
               {
                 phase_id: 1,
                 is_current_phase: true,
-                phase_start: format(Date.now(), "yyyy-MM-dd"),
```
mddilley, line 148: One last thing - there is compile warning about `format` no longer being used in this file.

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -0,0 +1,183 @@
...
+          Object.assign(obj, {
+            [item.subphase_id]: item.subphase_name,
+          }),
+        {}
+      ),
+    [subphases]
+  );
+
+/**
+ * Hoolk which returns an array of project_phase_ids of the project's current phase(s).
```
chiaberry, line 115: type: hook

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -0,0 +1,183 @@
...
+ * Although only one phase should ever be current, we handle the possibilty that there
+ * are multiple
```
chiaberry, line 117: just curious, why do we handle that possibility? is it because its easier to work with an Array?

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
chiaberry: and that will remove it? TIL @mddilley

moped-editor/src/views/projects/projectView/ProjectPhases.js
chiaberry: turns out we can toggle it! 

https://mui.com/x/react-data-grid/column-visibility/#customize-the-list-of-columns-in-panel

moped-editor/src/components/forms/ControlledCheckbox.js
```diff
@@ -0,0 +1,35 @@
...
+import { Controller } from "react-hook-form";
+
+/**
+ * A react-hook-form wrapper of the MUI Checkbox component
+ * @param {object} control - react-hook-form `control` object from useController - required
+ * @param {string} name - unique field name which be used in react-hook-form data object
+ * @param {string} label - the label to render next to the checkbox
+ * @return {JSX.Element}
+ */
+const ControlledCheckbox = ({ name, control, label }) => {
```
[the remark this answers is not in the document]
johnclary: Good call—it would be good to be consistent and make this component ready for future use cases

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
johnclary: nice catch! 🙏

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
johnclary: well, since we don't enforce this constraint at the DB level, my thinking was that it is technically possible for a project to have multiple current phases as result of some busted ETL or data update that happens outside of the Editor.

we could—and maybe should?—create a DB trigger that manages this instead 🤔

moped-editor/src/views/projects/projectView/ProjectPhases.js
johnclary: thanks for catching this. i decided to just disabled this whole menu and the filters. so now it's only possible to sort.

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
johnclary: +1  Chia: that was my thinking, too

mddilley (approved):
Retested and reviewed the latest commits, and this looks super solid! 🚤💨

roseeichelmann (approved):
test steps worked like a charm, this is a huge chunk of work!! very nice :rocket:

chiaberry (approved):
:ship !

johnclary:
this feels like a bit of a yolo merge without waiting for more reviews, but i'm keen to get this out in the wild. any other feedback is still very welcome 🙏

PR #1197, opened 2023-11-07
Adds basic user analytics tracking

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/4839

This PR adds user activity logging by making use of a [custom postgres function](https://hasura.io/docs/latest/schema/postgres/custom-functions/) that is exposed as a mutation through the Hasura API. When a logged-in user mutates this endpoint, the user's ID is logged to a `moped_user_events` table.

This mutation is called automatically when the Moped Editor `<App>` component mounts, thereby providing basic metrics about who is using Moped.

This function also updates a new column on the `moped_users` table called `last_seen_date`. This column can be viewed on the Staff list in the Editor.

## Testing
**URL to test:** Local

1. Login and navigate to the staff page. Notice the **Last seen** column should have a recent timestamp for your user account. Nice!
2. Refresh your page—the **Last seen** column should update accordingly. It may take a second refresh for this to stick. I think there may be a race condition going on there, since it is not certain that the `moped_users` query is going to resolve after  your `last_seen_date` is updated.
3. Without refreshing the page, navigate around the app, clicking on random things, until eventually coming back to the staff list. Your **Last seen** timestamp should not have changed.
4. Use your SQL client to inspect the `moped_user_events` table. It is accumulating one event every time you refresh your page. If you log out of Moped it will also log an event if you log back in.

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1699408061245_create_table_public_moped_user_events/up.sql
johnclary, line 1: when i first opened this PR, it only dealth with the new `last_seen_date` column on the users table. i decided to add this `moped_user_events` table as an afterthought.

so i didn't put too much thought into this events table, but i think it should get the job done for now without adding too much overhead on the app/api/db. and it shouldn't be too hard to extend it if we ever want to up our analytics game. hopefully that's not necessary.

moped-database/migrations/1699408061246_alter_table_public_moped_users_add_column_last_seen_date/up.sql
```diff
@@ -0,0 +1,38 @@
+ALTER TABLE "public"."moped_users"
+    ADD COLUMN "last_seen_date" timestamptz NULL;
+
+COMMENT ON COLUMN moped_users.last_seen_date IS 'Tracks the last time a user loaded the Moped app in their browser. This value is set by the set_last_seen_date function. This value is not 100% reliable because it is updated by an API call that can be blocked by the client.';
+
+-- custom function which can be tracked as a hasura mutation
+-- https://hasura.io/docs/latest/schema/postgres/custom-functions/
+
+CREATE FUNCTION set_last_seen_date (hasura_session json)
+    RETURNS SETOF moped_users
```
johnclary, line 10: i tried a few different ways to condense this into fewer SQL operations, but it's really hard to debug.  Hasura requires that it returns a `moped_users` object, and the graphql endpoint takes the shape of that table.

```graphql
mutation SetLastSeenDate {
  set_last_seen_date {
    user_id
    first_name
    last_name
    is_user_group_member
    last_seen_date
  }
}
```

moped-editor/src/App.js
```diff
@@ -80,11 +81,13 @@ const App = () => {
...
-            >
-              {routing}
-            </ProjectListViewQueryContext.Provider>
+            <ActivityMetrics>
+              <ProjectListViewQueryContext.Provider
+                value={{ listViewQuery, setListViewQuery }}
+              >
+                {routing}
+              </ProjectListViewQueryContext.Provider>
+            </ActivityMetrics>
```
johnclary, line 90: keeping it simple here and just wrapping the entire app underneath this. if we wanted to get fancy, we could push this into specific pages/components and add arguments to the event handler function to log them.

moped-editor/src/components/ActivityMetrics.js
```diff
@@ -0,0 +1,25 @@
...
+  useEffect(() => {
+    if (!user) {
+      // this event can fire when the user is not logged in
+      return;
+    }
+    try {
+      setLastSeen();
+    } catch (error) {
+      console.error("Failed to set the last seen date for the current user.");
+    }
```
johnclary, line 22: an error would happen if Hasura bombed, or if we have user de-sync issue where where the user's ID is not actually in the DB.

moped-editor/src/components/ActivityMetrics.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default function ActivityMetrics({ children }) {
+  const { user } = useUser();
+  const [setLastSeen] = useMutation(LAST_SEEN_MUTATION);
+  useEffect(() => {
+    if (!user) {
+      // this event can fire when the user is not logged in
+      return;
+    }
+    setLastSeen().catch((error) => {
```
johnclary, line 18: an error would happen if Hasura bombed, or if we have user de-sync issue where where the user's ID in the session token is not actually in the DB. if we don't catch the error, the whole app breaks 💥

mddilley: Nice, yes, that is a good call. The one thing that I thought about is if we would want to consider a threshold for how often this mutation could fire. I was able to spam the new table in the DB by reloading the app in the browser repeatedly (but I don't think users would be doing that on purpose or anything). 

Something to mull over because I'm not sure if adding an extra API call to check last seen before the mutation is great or anything!

moped-database/migrations/1699408061246_alter_table_public_moped_users_add_column_last_seen_date/up.sql
mddilley: super cool to see how this is set up when I checked it out in the Hasura console. Thanks for the link too - [accessing session data](https://hasura.io/docs/latest/schema/postgres/custom-functions/#accessing-hasura-session-variables-in-custom-functions) this way is a big TIL for me.

mddilley (approved):
I tested this out locally, and I see the events generating. ✨ This definitely feels like something that we could release tomorrow to test out with live MUG activity. I left one comment but it is the kind of thing that we could easily monitor and iterate on. Very cool stuff!

frankhereford (approved):
John, I've gone over the code here carefully, and tested it to the point of observing it failing correctly because when I login, it's looking for ID 68 in the database, and there's only 1-3. If you want me to resolve that and test further, let me know, but otherwise, I think this is good to ship. Thanks for doing this - I can't wait to see a visualization of moped activity. This is the first step to powerful insights. 

Have we ever considered installing a site analytics package? I'm not pushing for one, but it's interesting to think about.

moped-editor/src/components/ActivityMetrics.js
johnclary: agreed—this is a really good question. let's watch this table once we deploy, and if needed i like your suggestion of throttling this based on the `last_seen_date`!

johnclary:
@frankhereford asked:
> Have we ever considered installing a site analytics package? I'm not pushing for one, but it's interesting to think about.

we have considered it. there's some discussion about it in[#4839]( https://github.com/cityofaustin/atd-data-tech/issues/4839). i particularly though https://www.goatcounter.com/ looks cool.

johnclary:
note that this was heavily modified by https://github.com/cityofaustin/atd-moped/pull/1198

moped-editor/src/components/ActivityMetrics.js
chiaberry: I can attest that the whole app breaks :(

PR #1198, opened 2023-11-09
User activity tracking v2

johnclary:
This is a modified approach to https://github.com/cityofaustin/atd-moped/pull/1197 that uses a column preset to allow users to insert directly into the `moped_user_events` trigger.

## Testing
**URL to test:** Local

Follow the same steps as described in: https://github.com/cityofaustin/atd-moped/pull/1197

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -4517,6 +4517,34 @@
 - table:
     name: moped_user_events
     schema: public
+  insert_permissions:
+    - role: moped-admin
+      permission:
+        check: {}
+        set:
+          user_id: x-hasura-user-db-id
```
johnclary, line 4525: this is a column preset. something a little confusing about this is that using a preset revokes user permissions on the column. per [the docs](https://hasura.io/docs/latest/auth/authorization/permissions/column-presets/):

> the respective fields will also be removed from the generated GraphQL schema for that role.

which is exactly what we want. users cannot modify the `user_id` column on this table.

moped-database/migrations/1699408061245_create_table_public_moped_user_events/up.sql
```diff
@@ -1,10 +1,38 @@
...
+    ADD COLUMN "last_seen_date" timestamptz NULL;
+
+COMMENT ON COLUMN moped_users.last_seen_date IS 'Tracks the last time a user loaded the Moped app in their browser. This value is set by the set_last_seen_date function. This value is not 100% reliable because it is updated by an API call that can be blocked by the client.';
+
+
 CREATE TABLE "public"."moped_user_events" (
     "id" serial NOT NULL,
     "user_id" integer NOT NULL,
-    "description" text NOT NULL,
+    "event_name" text NOT NULL CHECK (event_name in('app_load')),
```
johnclary, line 10: if we want to support additional event types, we would add them to this check constraint

moped-database/migrations/1699408061245_create_table_public_moped_user_events/up.sql
```diff
@@ -1,10 +1,38 @@
...
+    RETURN NEW;
+END;
+$$;
+
+CREATE TRIGGER set_user_last_seen_date_trigger
+    AFTER INSERT ON public.moped_user_events
+    FOR EACH ROW
+    EXECUTE FUNCTION public.set_user_last_seen_date ();
+
+COMMENT ON FUNCTION set_user_last_seen_date IS 'Updates the moped_users.last_seen_date based on inserts into the moped_user_events table';
```
johnclary, line 38: and now we can use a simple trigger to keep `last_seen_date` current.

moped-editor/src/components/ActivityMetrics.js
```diff
@@ -9,7 +9,9 @@ import { useUser } from "src/auth/user";
  */
 export default function ActivityMetrics({ children }) {
   const { user } = useUser();
-  const [setLastSeen] = useMutation(LAST_SEEN_MUTATION);
+  const [setLastSeen] = useMutation(INSERT_USER_EVENT, {
+    variables: { event_name: "app_load" },
+  });
```
johnclary, line 14: per the check constraint, `app_load` is the only allowed value.

mddilley (commented):
All this looks great! I noticed a small issue inside the `useEffect` that is causing two inserts per app load. Looks like the user session is coming through first as a plain object, and, then, a second time as an instance of `CognitoUserSession`? 

That triggers the effect twice, and it makes me wonder if this is an Amplify thing or something our own auth code is doing.

Screenshot 2023-11-09 at 4 35 13 PM

Also wondering if this is reproducible on other people's machines.

johnclary:
@mddilley interesting! i do see that user object changing types as i log each component render, but it only triggers the useEffect once for me. only one table row is inserted. i could convert that `user` object to a bool and make that the dependency, just to be safe 🤔 i will take a closer look next week 🙏

moped-editor/src/components/ActivityMetrics.js
```diff
@@ -21,6 +23,6 @@ export default function ActivityMetrics({ children }) {
         error
       );
     });
-  }, [setLastSeen, user]);
+  }, [setLastSeen, isLoggedIn]);
```
johnclary, line 26: thanks @mddilley for flagging that the `user` object is not a stable reference. i have updated the hook to test for the existence of the user object, which reliably works.

mddilley (approved):
Tested locally, and 🚢 🚢 🚢 🚀!

moped-editor/src/auth/user.js
```diff
@@ -110,11 +110,10 @@ export const initializeUserDBObject = (userObject) => {
       // Then we parse the response
       res.json().then((resData) => {
         if (resData?.errors) {
-          console.error(resData.errors)
+          console.error(resData.errors);
         }
         if (resData?.data?.moped_users) {
           setSessionDatabaseData(resData.data.moped_users[0]);
-          window.location.reload();
```
johnclary, line 117: this lil guy is the reason the Moped app loads then a few seconds later reloads when you first login. and it's also causing an extra activity event to be inserted on login.

@mddilley spotted this, and we have both traced the data flow and think it can be safely removed.

mddilley (approved):
Woohoo! 🙌 I tested logging in, refreshing, and navigating around the app, and I'm seeing app load events insert only when expected. 🚀

frankhereford (approved):
This tested great (using the JD user ID 1 account) and the code looks sound to me. I think this is 🚢. Thank you!

chiaberry (approved):

johnclary:
thanks y'all!

PR #1203, opened 2023-11-09
Add audit fields to `moped_proj_components`

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/14606. 

## Testing

I found this to be easiest to test locally. 

### Steps

* Switch to this branch and spin up your local DB. This is testing the application of the migrations.
* Grab your SQL client of choice and check out this query to get a lay of the land before we observe what this PR does.
```sql
select project_component_id, 
  created_at, 
  created_by_user_id, 
  updated_at, 
  updated_by_user_id,
  description 
from moped_proj_components;
```
* Fire up the editor and add a component via the map to a project. Observe that this works like normal.
* Come back to your DB client and run that above query again.
  * You should observe that all four audit fields are populated. (`created_at`, `created_by_user_id`, `updated_at`,   `updated_by_user_id`)
* Ask yourself this question - do you agree with this PR that `INSERT` is a subset of `UPDATE` and if you agree with populating the update field on create.
* Edit some components and use that above query to track these audit fields and how they are updated.
* Open up the `graphql-engine` console and check the permissions and column preset choices on `moped_proj_components`.
* Run the following two queries and decide if you like the naming choices and general form of the work.
```sql
SELECT event_object_table as table_name,
       trigger_name,
       action_timing as timing,
       event_manipulation as event,
       action_condition as condition,
       action_statement as definition
FROM information_schema.triggers
WHERE event_object_table = 'moped_proj_components';

SELECT pg_get_functiondef(pg_proc.oid)
FROM pg_proc
JOIN pg_namespace ON pg_proc.pronamespace = pg_namespace.oid
WHERE pg_proc.proname = 'update_updated_at_column';
```
---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/1699568640283_squashed-migration-add-audit-field-to-proj-components/up.sql
```diff
@@ -0,0 +1,22 @@
...
+
+CREATE OR REPLACE FUNCTION update_updated_at_column()
+RETURNS trigger
+LANGUAGE plpgsql
+AS $$ 
+BEGIN
+  NEW.updated_at = now();
+  RETURN NEW;
+END;
+$$;
```
frankhereford, line 17: We'll see how it plays out, but my intention is to make this a generic function that works for any table so we can hook it up to whatever we want that has audit fields following our pattern.

johnclary: 💯 this is great!

how would you feel about tearing down the replacing the existing function + trigger we have for the `moped_proj_work_activity` table in this PR?

moped-database/migrations/1699568640283_squashed-migration-add-audit-field-to-proj-components/up.sql
```diff
@@ -0,0 +1,22 @@
+alter table "public"."moped_proj_components" rename column "date_added" to "created_at";
+
+alter table "public"."moped_proj_components" add column "created_by_user_id" integer null;
+
+alter table "public"."moped_proj_components" add column "updated_by_user_id" integer null;
+
+alter table "public"."moped_proj_components" add column "updated_at" timestamptz null;
+
+CREATE OR REPLACE FUNCTION update_updated_at_column()
```
johnclary, line 9: i do not feel strongly about this at all, but i guess i prefer `set_updated_at_column()` since that reads a little more cleanly. i also think it's nice for the trigger name and function to be similar. trigger name is great, by the way

johnclary (approved):
@frankhereford this works beautifully! 

i don't feel strongly about the name changes. let me know what you think about updating the `moped_proj_work_activity` function + trigger in this PR. otherwise i will open a separate issue.

🚢 🚢 🚢 🚢 🚢 🚢 🚢  🙌

moped-database/migrations/1699568640283_squashed-migration-add-audit-field-to-proj-components/up.sql
frankhereford: Yea John, I totally agree with you. i'm not sure where that came from, but I'm furthest from liking it and even thought for a bit what would be better and didn't have anything come to mind. `st_updated_at_column()` is perfect. It says what it does and doesn't re-use words back to back which is a semantic disaster. I'll update this for sure. Thank you!

moped-database/migrations/1699568640283_squashed-migration-add-audit-field-to-proj-components/up.sql
frankhereford: Yep, I'd love to. Will do.

mddilley (approved):
Tested locally, migration works up and down, and I like the naming. Great to see this pattern expanding! 🚀 🙌

About setting both `created_by_user_id` and `updated_by_user_id` on insert, I think that works great unless we foresee needing to query project components that have never been edited. 🤷 Not likely imo.

moped-database/migrations/1699568640283_squashed-migration-add-audit-field-to-proj-components/up.sql
frankhereford: I am going to use the existing [`public.set_updated_at()`](https://github.com/cityofaustin/atd-moped/blame/b96dea54a11544dc07a9d2eee40f0444087f3151/moped-database/migrations/1686243614837_create_work_auth_table/up.sql#L87-L93) function. John wrote that in 4 months ago, and it is a better version of what I initially proposed in this PR because mine turns out had a assignment vs. comparison (`:=` vs. `=`) operator mix-up. 🙈 John's is perfect for this task.

johnclary: oh, cool—that works too! i didn't realize how they were basically the exact same function. glad you caught the assignment bug!

johnclary (approved):
🚢 🚢 🚢 🚢

chiaberry (approved):
🚢 

its gonna be cool to see who is editing components and creating them!

PR #1205, opened 2023-11-10
Fix project view link back to project list view with previous filters

mddilley:
## Associated issues
This fixes the key to access the state passed with React Router. It was the wrong name so the value was `undefined`. 😑 Also, the `filter` and `isOr` query string parameters were not clearing in the url when switching from advanced search to simple search so I patched that up while I was at it.

## Testing
**URL to test:** 

https://deploy-preview-1205--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Create a new project and make sure the project list view loads after the redirect
2. Go to the project list view and add advanced filters
3. Click on the project link or parent project link and make sure that your filters are retained when you click "< ALL PROJECTS" on the project view page
4. When linked back to the project list view with advanced filters active, enter a search value into the simple search box and press Enter. You should see the advanced filters clear and no query string parameters in the url

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Search.js
+++ b/moped-editor/src/components/GridTable/Search.js
@@ -99,7 +99,8 @@ const Search = ({
     setFilters({});
     setIsOr(false);
     setSearchParams((prevSearchParams) => {
-      prevSearchParams.delete("filters");
+      prevSearchParams.delete("filter");
+      prevSearchParams.delete("isOr");
       return prevSearchParams;
     });
   };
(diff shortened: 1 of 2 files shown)


moped-editor/src/components/GridTable/Search.js
```diff
@@ -99,7 +99,8 @@ const Search = ({
     setFilters({});
     setIsOr(false);
     setSearchParams((prevSearchParams) => {
-      prevSearchParams.delete("filters");
+      prevSearchParams.delete("filter");
+      prevSearchParams.delete("isOr");
       return prevSearchParams;
```
johnclary, line 104: cool—i see this mirrors the same behavior as when resetting the filters via the `handleClearFilters` function.

johnclary (approved):
looks great—thanks for making quick work of this!

PR #1207, opened 2023-11-13
Adds additional component tags for ATSD

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/14567

## Testing
**URL to test:** Local

**Steps to test:**
1. Start Moped and confirm these tags are available via the component form
2. Checkout the down migration `hasura migrate apply ---down 1`

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1699899884388_atsd_component_tags/up.sql
+++ b/moped-database/migrations/1699899884388_atsd_component_tags/up.sql
@@ -1,0 +1,7 @@
+ALTER TABLE moped_component_tags DROP CONSTRAINT moped_component_tags_name_key;
+
+INSERT INTO public.moped_component_tags (name, "type", slug)
+        values
+        ( '2024', 'MAP Ped Crossing', 'map_ped_crossing_2024'),
+        ('All', 'Ped Crossing Priority', 'ped_crossing_priority_all'),
+        ('2024', 'Bikeways – Work Plan', 'bikeways_work_plan_2025');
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1699899884388_atsd_component_tags/up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE moped_component_tags DROP CONSTRAINT moped_component_tags_name_key;
+ALTER TABLE public.moped_component_tags add constraint moped_component_tags_name_key UNIQUE ("name", "type");
+
+INSERT INTO public.moped_component_tags (name, "type", slug)
+        values
+        ( '2024', 'MAP Ped Crossing', 'map_ped_crossing_2024'),
+        ('All', 'Ped Crossing Priority', 'ped_crossing_priority_all'),
+        ('2024', 'Bikeways – Work Plan', 'bikeways_work_plan_2025');
```
mddilley, line 8: I think that this slug needs to be updated to say 2024.

Also, tiny detail but I noticed that the existing bikeways work plan tags use a short dash instead of the long dash that is added here. I caught it because it is more noticeable in the UI.
**Bikeways - Work Plan**
**Bikeways – Work Plan**

mddilley (commented):
Worked great! 🚀 I left one comment about a slug update.

moped-database/migrations/1699899884388_atsd_component_tags/up.sql
johnclary: wow—great catch on both counts. thank you!!!!

chiaberry (approved):
I see the new component tags!

mddilley (approved):
Looks ✨! 🛳️

PR #1209, opened 2023-11-15
Only query for columns shown in the project list view

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14298

This PR updates the code to only fetch the data associated with columns shown in the project list view.

## Testing
**URL to test:** 

https://deploy-preview-1209--atd-moped-main.netlify.app/

**Steps to test:**
1. Go to the https://deploy-preview-1209--atd-moped-main.netlify.app/moped/projects
2. Open dev tools because the easiest way to visualize this change is to inspect the GraphQL queries in the network requests
3. Refresh the page and check the columns in the request and check that the columns fetched match up with the columns checked in the column selector UI
   - The config in `PROJECT_LIST_VIEW_QUERY_CONFIG.columns` and the `useColumns` hook can be helpful to match column names to the column labels
   - Also, expect to see `current_phase_key` in each request since it is needed to display the status badge
4. Add/remove columns and inspect requests to see more/less fields fetched
5. Test the `useCurrentData` hook by reloading the page
   - You should see a spinner on first load as the first page is fetching
   - When changing filters, sorting, paginating, or updating columns, you should see no spinner or UI changes other than rows of data updating
6. Test the hidden column preferences by watching the value of `mopedColumnConfig` in your browser's local storage to see columns switch between true and false when toggling them in the table

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -37,67 +30,62 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
...
     defaultDirection: "desc",
   },
-  filters: PROJECT_LIST_VIEW_FILTERS_CONFIG,
-  export: PROJECT_LIST_VIEW_EXPORT_CONFIG,
   search: {
     placeholder:
       "Search by ID, name, description, phase, lead, sponsor, partners, eCAPRIS ID...",
     defaultFieldsOperator: "_or",
   },
   columns: {
```
mddilley, line 38: Lots of changes in this config so I'll try to cover them here
- GridTable config no longer needed: `hidden`, `primary_key`, `icon`, `width`, `filter` (set in the Material Table config so this was sort of a duplicate)

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -176,80 +167,140 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
...
+      showInTable: true,
+    },
   },
-  // This object gets consumed into the GQLAbstract system, and here is the single, un-nested order_by directive. ✅
-  order_by: { updated_at: "desc" },
-  where: null,
-  or: null,
-  and: null,
-  limit: 250,
-  offset: 0,
```
mddilley, line 254: This chunk is old config that is no longer used by this view. The defaults are handled in the hooks added recently.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -46,57 +47,8 @@ const useStyles = makeStyles((theme) => ({
...
- * Keeps localStorage column config in sync with UI interactions
- * @param {Object} column - the MT column config with the `field` prop - aka the column name
- * @param {Bool} hidden - the hidden state of the column
- */
-const handleColumnChange = ({ field }, hidden) => {
-  let storedConfig =
-    JSON.parse(localStorage.getItem("mopedColumnConfig")) ??
-    DEFAULT_HIDDEN_COLS;
-  storedConfig = { ...storedConfig, [field]: hidden };
-  localStorage.setItem("mopedColumnConfig", JSON.stringify(storedConfig));
```
mddilley, line 97: This happens in `useColumns` now so it is responsible for syncing the hidden columns state with user preferences that we save in local storage.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -222,15 +183,6 @@ const ProjectsListViewTable = () => {
     rowsPerPageOptions:
       PROJECT_LIST_VIEW_QUERY_CONFIG.pagination.rowsPerPageOptions,
   });
-  /*
-   * Store column configution before data change
```
mddilley, line 226: Not sure that this is needed without gqlabstract and GridTable. I don't see a regression, but I might not know the whole story here.

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -137,10 +143,35 @@ export const useTableComponents = ({
...
+  }, [hiddenColumns]);
+
+  const columnsToReturnInQuery = useMemo(() => {
+    const columnsShownInUI = Object.keys(hiddenColumns)
+      .filter((key) => hiddenColumns[key] === false)
+      .map((key) => key);
+
+    // Some columns are dependencies of other columns to render, so we need to include them
+    // Ex. Rendering ProjectStatusBadge requires current_phase_key which is not a column shown in the UI
+    const columnsNeededToRender = ["current_phase_key"];
```
mddilley, line 170: I'm just realizing that this column isn't needed if we aren't showing the **Status** column. I'm torn on whether to try to add extra logic to decide when to add this. Maybe down the road when we have more columns with dependencies?

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -153,13 +184,14 @@ export const useColumns = ({ hiddenColumns, classes }) => {
         field: "project_name",
         hidden: hiddenColumns["project_name"],
         render: (entry) => (
-          <RouterLink
+          <Link
+            component={RouterLink}
             to={`/moped/projects/${entry.project_id}`}
             state={{ queryString }}
-            className={classes.colorPrimary}
```
mddilley, line 159: I removed this className so we would have one less param to consume in this hook. One result is that we go back to [the default MUI behavior of having underlined links](https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/) which might not be the worst thing for accessibility. [Github recently did this too.](https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/)

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/usePreviousData.js
```diff
@@ -0,0 +1,21 @@
...
+    if (data) {
+      setPreviousData(data);
+      return data;
+    } else {
+      return previousData;
+    }
+  }, [data, setPreviousData, previousData]);
+
+  return currentData;
+};
```
johnclary, line 21: this is working great!

one side effect of this is that the loading spinner no longer shows when sorting columns, searching, or advance searching. if you throttle your network down to 3G you can see this is probably gonna be a little laggy/jumpy in production.

one idea we discussed is showing the loading spinner elsewhere on the screen, so that we still give visual feedback when the table is updating. i will open a separate issue for that.

johnclary (changes requested):
@mddilley this looks great! it's not as obvious with our smaller number of projects in staging, but this is going to be a serious performance improvement, and it heads off future performance issues as we continue to add columns to this table.

i'm curious too hear your feedback on my suggestions 🚀

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -37,67 +30,62 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
...
     defaultFieldsOperator: "_or",
   },
   columns: {
     project_id: {
-      hidden: true,
-      primary_key: true,
-      searchable: true,
-      sortable: false,
+      searchable: true, // useSearch
+      sortable: true, // useColumns
```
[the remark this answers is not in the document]
mddilley: Sure! This config is used by the **Signal IDs** column config in the memoized `columns` array for Material Table in `useColumns` (line 280 in `helpers.js`). It is the only column that had `sorting` disabled in the existing code so I updated it to use this k/v pair instead. 

Glad you caught this because it felt weird to have only one column use this config. I could see us adding a `sorting` parameter for all the columns fed to Material Table to be more explicit. Idk - what do you think?

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -137,10 +143,35 @@ export const useTableComponents = ({
...
+  );
+
+  const COLUMN_CONFIG = PROJECT_LIST_VIEW_QUERY_CONFIG.columns;
+
+  /*
+   * Sync hidden columns state with local storage
+   */
+  useEffect(() => {
+    localStorage.setItem("mopedColumnConfig", JSON.stringify(hiddenColumns));
+  }, [hiddenColumns]);
```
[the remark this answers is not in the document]
mddilley: I totally agree on this because I left this code thinking - how would I know that the local storage sync was even happening? Adding a separate custom hook at the top level of `ProjectListViewTable` sounds like the answer to me. Thanks!

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -137,10 +143,35 @@ export const useTableComponents = ({
...
-export const useColumns = ({ hiddenColumns, classes }) => {
+export const useColumns = () => {
   const location = useLocation();
   const queryString = location.search;
 
+  const [hiddenColumns, setHiddenColumns] = useState(
+    JSON.parse(localStorage.getItem("mopedColumnConfig")) ?? DEFAULT_HIDDEN_COLS
+  );
+
+  const COLUMN_CONFIG = PROJECT_LIST_VIEW_QUERY_CONFIG.columns;
```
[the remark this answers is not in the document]
mddilley: Just made this update 🙏

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: I cannot remember exactly, but I think this had to do with requests in gqlabstract triggering multiple refreshes when it first loaded. so I agree that it probably isnt needed anymore

chiaberry (approved):
tested, searched for projects based on hidden columns, toggled them on and off, refreshed, all works as expected! 

Thinking about how the signals column isnt sortable, we could write a custom sort function, if we wanted to sort it, but I am not sure what we would sort on. Looking at other columns that contain multiple entries, it looks like we sort on the alphabetical order of the first entry?

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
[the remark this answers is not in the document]
johnclary: ahhh—i'm following now, thanks. i think what you have is a great. the diff was throwing me off but the code is easy to follow 🚢

moped-editor/src/views/projects/projectsListView/helpers.js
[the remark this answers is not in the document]
johnclary: thanks for revisiting this—it looks good!

johnclary:
> Thinking about how the signals column isn't sortable

@chiaberry alphabetical sounds good. shall i log it as a snackoo?

johnclary (approved):
it looks great an works great. let's get this into prod 

🏎️💨

mddilley:
thanks y'all! I made one small update to match a filename with the custom hook name inside of it. Merging!

PR #1213, opened 2023-11-20
Update `moped_project` and `moped_proj_components` on child feature edits

frankhereford:
## Associated issues

This PR hopes to close https://github.com/cityofaustin/atd-data-tech/issues/14617. 

## Testing

Testing can be completed locally.

* While testing, keep in mind that these triggers are set to run on INSERT and UPDATE to the underlying feature tables. An INSERT into one of those tables is an "update"-like operation to the parent project and component records, which is why we won't be editing `created_at` and `created_by_user_id` fields on the project or component records.

* Spin up the local database and make sure migrations are applied cleanly.
* Add a feature to a project, noting the project number.

* You can use your SQL client to check the pertinent fields in the projects and components table by plugging the project number into this query:

```sql
select 
  projects.project_id, components.component_id,
  projects.updated_at as project_updated_at, projects.updated_by_user_id as project_updated_by,
  components.updated_at as component_updated_at, components.updated_by_user_id as component_updated_by
from moped_project projects
left join moped_proj_components components on (components.project_id = projects.project_id)
where projects.project_id = 119
```

* Run this query to make sure the updated fields are updated in the tables. The "project_" fields are in the projects table, so if the project you're editing has multiple components, you'll see the `project_updated_at` and `project_updated_by` value repeated for each component. ✅
* Add new features (components) and update the geometry of existing components, and watch how these updated_by fields are updated by these triggers and function.

## Utility

I made a quick `reload.sh` that was invaluable in constructing this PR. We may want to enhance the `hasura-cluster` script with something along these lines at some point.

```
#!/bin/bash

./hasura-cluster stop
./hasura-cluster start_only
docker stop moped-database-hasura-1
echo 'DROP DATABASE moped; CREATE DATABASE moped' | docker exec -i moped-database-moped-pgsql-1 psql -U moped postgres;
docker exec -i moped-database-moped-pgsql-1 psql -U moped moped < production.sql; 
docker start moped-database-hasura-1
sleep 5
hasura migrate apply
hasura metadata apply
#echo 'UPDATE moped_users set user_id = 1 where user_id = 60;' | docker exec -i moped-database-moped-pgsql-1 psql -U moped moped;
```

---
#### Ship list
- [ ] Code reviewed 


moped-database/docker-compose.arm64.yml
```diff
@@ -15,4 +15,7 @@ services:
             - POSTGRES_USER=moped
             - POSTGRES_PASSWORD=moped
             - POSTGRES_DB=moped
-        command: ["postgres", "-c", "jit=off"]
+        command: ["postgres",
+            "-c", "jit=off",
+            "-c","log_statement=ddl",
+            "-c", "log_min_messages=info"]
```
frankhereford, line 21: This enables `RAISE INFO` to work inside functions to be viewed with `docker logs -f ...`

chiaberry (approved):
I tested using prod data, and it does feel much more intuitive for the project to come to the top of the list when i get back to the projects list -- just like editing anything on the summary table. 

thank you for the comments!

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,69 @@
...
+CREATE OR REPLACE FUNCTION update_parent_records_audit_logs()
+RETURNS TRIGGER AS $$
+DECLARE
+  project_id_variable INTEGER;
+  user_id_variable INTEGER;
+BEGIN
+  -- Select the user ID who last updated the project component
+  SELECT updated_by_user_id INTO user_id_variable
+  FROM moped_proj_components
+  WHERE project_component_id = NEW.component_id;
```
johnclary, line 15: if i'm following, this is always setting the `moped_proj_components. updated_by_user_id` to its previous value, 
whereas we need to be pulling the `updated_by_user_id` value of the current user has made the edits. i didn't write this in scope, but we might need to add audit fields to the `feature_` tables so that we have a current user ID to assign to this column?

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,69 @@
...
+FOR EACH ROW
+EXECUTE FUNCTION update_parent_records_audit_logs();
+COMMENT ON TRIGGER feature_signals_trigger ON feature_signals IS 'Trigger to update parent project and component audit fields';
+
+-- Trigger for feature_street_segments table
+CREATE TRIGGER feature_street_segments_trigger
+AFTER INSERT OR UPDATE ON feature_street_segments
+FOR EACH ROW
+EXECUTE FUNCTION update_parent_records_audit_logs();
+COMMENT ON TRIGGER feature_street_segments_trigger ON feature_street_segments IS 'Trigger to update parent project and component audit fields';
```
johnclary, line 69: nice—it is cool to see the re-usable trigger and have the updates propagate up to the project!

moped-database/docker-compose.arm64.yml
johnclary: nice!

johnclary (changes requested):
@frankhereford i might be missing something, but i think this mechanism needs another look. thanks for working on this!

moped-database/migrations/1700149643858_parent_audit_values/up.sql
frankhereford: John, I may have put a seriously dim-witted comment in there, but I'm going to check / fix it to make sure it's actually representing the current user's ID and not whomever last made an edit on the moped_proj_component's record. I did most of my testing by creating a new component, and I may have massively overlooked this because all of my records were created / edited and whatnot by user_id 1. Thanks for flagging this!! 🙏

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,69 @@
+-- Altering moped_project table to add a new column for tracking the user who last updated a record
+ALTER TABLE moped_project ADD COLUMN updated_by_user_id INTEGER DEFAULT NULL;
+COMMENT ON COLUMN moped_project.updated_by_user_id IS 'User ID of the person who last updated the project';
```
johnclary, line 3: i just want to note that we have https://github.com/cityofaustin/atd-data-tech/issues/14564 to cover the task of setting up a Hasura column preset that should populate the `moped_project.updated_by_user_id` column for any project edit.

frankhereford:
Thanks for referencing this issue @johnclary . I'm working on it as we speak and I am hoping to get it back to you for review this afternoon. Thanks again!

moped-database/migrations/1700149643858_parent_audit_values/up.sql
frankhereford: John -- first off - your comment makes perfect sense. I was pulling the last editor field off the `moped_proj_components` instead of off the feature tables which contained the edited or created data. I've absolutely corrected that with a dynamically constructed query in my function and hooked it up to triggers for the feature tables. This is so absolutely the right way to do this, and I really appreciate you noting this monstrously hair-brained oversight on my part.

I'm excited to ask for your review again -- I think I'm back on track here.

frankhereford:
This no longer applies cleanly into `main` nor to the DB; I'm going to push up a revision here shortly. This is just in case anyone is looking at this now or plans to in the immediate future. Won't take long!

johnclary:
👍 i happen to be testing now but have no incorporated prod data or main. so it's been looking fine to me :)

frankhereford:
I believe this is ready to roll again. I hand removed the `contractor(s)` fields from the metadata, and then merged in `main` to get the `workgroup_contractors` fields in. I probably could have just merged in `main`, but for some reason I felt compelled to get the branch working with the current DB state before I pulled in `main` incase something went sideways.

mddilley (changes requested):
This is super stuff! The `update_parent_records_audit_logs` function is so cool, and the comments are ✨. I requested a couple of small changes unrelated to functionality. I tested creating and updating every type of feature, and I see the create and update timestamp and user ID fields updating just as expected. 👏

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- Trigger for feature_signals table
+CREATE TRIGGER feature_signals_trigger
+AFTER INSERT OR UPDATE ON feature_signals
+FOR EACH ROW
+EXECUTE FUNCTION update_parent_records_audit_logs();
+COMMENT ON TRIGGER feature_signals_trigger ON feature_signals IS 'Trigger to update parent project and component audit fields';
+
+-- Trigger for feature_street_segments table
+CREATE TRIGGER feature_street_segments_trigger
```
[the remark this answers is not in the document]
frankhereford: Absolutely - I'll get those names aligned and will resolve this when that's pushed up. Great catch - thank you.

moped-database/migrations/1700149643859_add_audit_fields_to_unifed_features_view/up.sql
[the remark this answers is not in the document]
frankhereford: Also a huge yes here, thank you for that reminder! Will resolve when pushed up. Thanks!

mddilley:
Forgot to mention that I fully support updating the `hasura-cluster` scripts to be closer to @frankhereford's `reload.sh` script in the PR desc (which is not to far from the VZ local scripts). I have similar commands in my terminal history, and it would be great to get local VZ and Moped spin up more closely aligned.

moped-database/views/project_geography.sql
frankhereford, line 1: I think the reason for these diffs being much bigger than the small changes is that I'm pulling the code out of the DB with something like:

```sql
SELECT pg_get_viewdef('example_view'::regclass, true);
```

Should we define a standard way to do this? I'm not sure how else to get at them.

johnclary: it's a real challenge — and i think the thing we need to do is to all get a linter setup with the same config :|

chiaberry: I think this is another case of a need for consistent SQL linting in our codebase. Ideally whatever is in this file should match what is in the database/migrations. That way if someone were to copy this file and edit it as their migration or pull the view the way you do, it would be the same.

frankhereford: I wonder if the postgres server keeps the formatting of the code that is used to define the view or if it parses it into a syntax tree and then re-codifies it as SQL when you query out the source code. I'm going to do some testing -- maybe we might find it to be a good linter if that pans out.

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,73 @@
...
+  user_id_variable INTEGER;
+  query TEXT;
+BEGIN
+  
+  -- Constructing a dynamic query to select the user ID from the triggered table
+  query := 'SELECT updated_by_user_id FROM ' || quote_ident(TG_TABLE_NAME) || 
+           ' WHERE id' || ' = ' || NEW.id || ';';
+  
+  -- Executing the dynamic query
+  EXECUTE query INTO user_id_variable;
```
johnclary, line 19: this is a TIL for me for sure! in this case, instead of querying this value from the trigger table, i think it can be retrieved from `NEW.updated_by_user_id`?

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,73 @@
...
+
+  -- Update the project component with the current timestamp and user ID
+  UPDATE moped_proj_components
+  SET updated_at = NOW(), updated_by_user_id = user_id_variable
+  WHERE project_component_id = NEW.component_id
+  RETURNING project_id INTO project_id_variable;
```
johnclary, line 25: grabbing the `project_id` at the end of this operation is really elegant! 🔥

moped-database/migrations/1700149643858_parent_audit_values/up.sql
```diff
@@ -0,0 +1,73 @@
...
+  -- Constructing a dynamic query to select the user ID from the triggered table
+  query := 'SELECT updated_by_user_id FROM ' || quote_ident(TG_TABLE_NAME) || 
+           ' WHERE id' || ' = ' || NEW.id || ';';
+  
+  -- Executing the dynamic query
+  EXECUTE query INTO user_id_variable;
+
+  -- Update the project component with the current timestamp and user ID
+  UPDATE moped_proj_components
+  SET updated_at = NOW(), updated_by_user_id = user_id_variable
```
johnclary, line 23: i wonder if it might be better to use the `NEW.updated_at` value, so that we have identical `updated_at` timestamps across all the affected records. i know we're just talking about microseconds here 🙈

moped-database/migrations/1700149643859_add_audit_fields_to_unified_features_view/down.sql
johnclary, line 1: thank you for updating these! i like this best practice of leaning toward exposing these new fields as soon as we implement them. otherwise they get forgotten about, and we forget if/why we did not expose these fields in the views/hasura in the first place. good stuff! 🙌

johnclary (changes requested):
@frankhereford this is working exactly as intended—really nice work! thanks for going the extra distance with code comments, plus the view updates.

i have requested a few really minor changes.

this is out of scope here and does not need to fall on you, but I'd like to start documenting somewhere what triggers we have (including the Hasura event triggers) and how they all relate, especially with regard to these audit fields. i will make note of that task!

moped-database/migrations/1700149643857_add_audit_fields_to_feature_tables/up.sql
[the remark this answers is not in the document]
frankhereford: Great call - thank you for that! I've resequenced them 041bce1f66fd8e512f4cc3c08ede69d5c32bbfdb.

moped-database/metadata/tables.yaml
```diff
@@ -245,80 +245,106 @@
...
   update_permissions:
     - role: moped-admin
       permission:
         columns:
-          - project_extent_id
-          - source_layer
           - component_id
-          - id
           - geography
+          - id
```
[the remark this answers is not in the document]
frankhereford: Thanks for pointing that out. I have removed the `id` column from the insert and update permissions for the various user roles here: 350b370b2f697c84072eeefa91646d91664f1d3e. The intent here is to try to get ahead of that work that @tillyw is doing to try to minimize any conflicts that my code might cause with her work. Tilly, let me know if there's anything else I can do to make things best for you.

moped-database/migrations/1700149643858_parent_audit_values/up.sql
frankhereford: Oh yes, of course! Thank you - that is so the correct way to do this instead of constructing a dynamic query. I'm used to just having an ID to start with and forgot we had the rest of the row as well. 💯  This is addressed in commit baba803d224b27dcb812d75ebf9144c8d589cabc.

moped-database/migrations/1700149643858_parent_audit_values/up.sql
frankhereford: Absolutely. There's no reason in the world not to align them and that change is made in here: 7972d8b9a3a83ba5638e273bb4ce06d299f48890.

chiaberry (approved):
ran through the test steps, and I see the project fields being updated. 

thanks for sharing your utility script, I agree we should add something similar to hasura_cluster. I have a variation on that in my local hasura_cluster that takes the snapshot I have already and uses that to spin up the db.

moped-database/metadata/tables.yaml
```diff
@@ -245,80 +245,102 @@
...
         check: {}
+        set:
+          created_by_user_id: x-hasura-user-db-id
+          updated_by_user_id: x-hasura-user-db-id
         columns:
-          - project_extent_id
-          - source_layer
           - component_id
-          - id
+          - created_by_user_id
```
mddilley, line 253: This one permission snuck in for insert `created_by_user_id` by moped-admin in `feature_drawn_lines`

mddilley (changes requested):
Retested this locally and found one tiny permission change needed. The rest of this is gold 🥇, and I can't wait to ✅!

moped-database/metadata/tables.yaml
frankhereford: Thank you Mike! I appreciate you catching that! I must have miss-clicked in the console UI when I was working on the permissions this morning. Thanks again! ca2ba5d03f62dac1e8fc287ccfb14fb703b3c5e7

mddilley: 🙌🙏🚀  100% - there are tons of boxes to check on this one 😵‍💫

mddilley (approved):
hooray! This looks great, and I can't wait to see us add more auditing like this. 🚀 🙌

johnclary (approved):
@frankhereford looks great to me! thanks for all your work on this one! 🙌 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

frankhereford:
Thanks everyone for all your help on this!

PR #1215, opened 2023-11-20
Exclude deleted components from components column of the project list view 

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/14663

## Testing
**URL to test:** 

locally

**Steps to test:**

spin up your moped. 

Add a component to a project. (Of a component type that doesnt already exist on the project) Delete the component. 
Go back to the project list view, and making sure the components column is visible, confirm that your deleted component doesnt appear. 

If you run the down migration, you should see your deleted component appear. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1700515730257_fix_delete_component_bug/up.sql
+++ b/moped-database/migrations/1700515730257_fix_delete_component_bug/up.sql
@@ -1,4 +1,4 @@
--- latest version 1700515730257_fix_delete_component_bug
+-- latest version 1699978012000_change_contractor_column_name
 DROP VIEW project_list_view CASCADE;
 
 CREATE OR REPLACE VIEW public.project_list_view
@@ -55,7 +55,7 @@ AS WITH project_person_list_lookup AS (
     SELECT
       mpc.project_id,
       string_agg(DISTINCT mc.component_name_full, ', '::text) AS components
-    FROM moped_proj_components mpc
+    FROM moped_proj_components mpc 
     LEFT JOIN moped_components mc ON mpc.component_id = mc.component_id
     WHERE mpc.is_deleted = FALSE
     GROUP BY mpc.project_id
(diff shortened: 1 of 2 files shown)


moped-database/views/project_list_view.sql
```diff
@@ -55,8 +55,9 @@ AS WITH project_person_list_lookup AS (
     SELECT
       mpc.project_id,
       string_agg(DISTINCT mc.component_name_full, ', '::text) AS components
-    FROM moped_proj_components mpc 
-    LEFT JOIN moped_components mc ON mpc.component_id = mc.component_id 
+    FROM moped_proj_components mpc
+    LEFT JOIN moped_components mc ON mpc.component_id = mc.component_id
+    WHERE mpc.is_deleted = FALSE
```
chiaberry, line 60: only real difference is the addition of is_deleted, the other lines I removed empty spaces

moped-database/migrations/1700515730257_fix_delete_component_bug/up.sql
```diff
@@ -0,0 +1,210 @@
+-- latest version 1699978012000_change_contractor_column_name
+DROP VIEW project_list_view CASCADE;
```
mddilley, line 2: This cascade looks to be dropping `component_arcgis_online_view` so it seems like that view needs to be added back in this PR. I feel a little out of the loop on the dependencies here.

mddilley (commented):
One small change needed so that we can test this out locally. The change in the project list view looks great!

moped-database/migrations/1700515730257_fix_delete_component_bug/up.sql
chiaberry: you are right and now im wondering how  i didnt get the same error. fixing now

chiaberry: okay i fixed it now and also have figured out how i missed the error -- i was applying the migration by hand and going up and down and testing that way. not applying the metadata, or spinning it up completely from scratch. good reminder for me to do that before requesting review!

roseeichelmann (approved):
Worked like a charm :sparkles: 

Unrelated to this PR but I feel like the components in the Components column on the project list view could be separated by commas or something. These are three different component types in this sc but its kind of hard to tell

image

johnclary:
@roseeichelmann thanks for flagging that. agreed. i will open an issue

PR #1216, opened 2023-11-22
Update project list view local storage to support new columns and clear outdated ones

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14708

This PR fixes a bug where outdated columns in local storage were breaking the new method of fetching only the data shown in the table. The root of the problem was that we need to clean up outdated columns in user's local storage if we want to rely on those columns to fetch data. This was luckily caught early since `task_order` changed to `task_orders` which surfaced the issue fast. 🏎️ 🙌

Below is a quick summary of the changes and the value of `mopedColumnConfig` local storage for each step.

**Previous method**
```md
// This works because we were fetching all columns every render.
{
   "task_order": false, <- This entry was never consumed by the app and ignored.
   "task_orders": false,
   ...config for any other column set previously (could be incomplete)
}
```

**The bug from [the code update to only fetch shown columns](https://github.com/cityofaustin/atd-moped/pull/1209) which queried all columns in local storage**
```md
{
   "task_order": false, <- This column was fed to the query, and 💥 since it no longer exists.
   "task_orders": false
    ...config for any other column set previously (could be incomplete)
}
```

**The fix**
```md
// Code cleans up local storage and adds settings for any columns not previously set
{
   "task_orders": false,
   ...config for the other 26 columns displayed in the table (mixture of existing user values and defaults)
}
```

## Testing
**URL to test:** 

https://deploy-preview-1216--atd-moped-main.netlify.app/moped/projects

**Steps to test:**
1. Go to the deploy preview and delete your local storage entry called `mopedColumnConfig`
2. Refresh the project list view, and you should see your `mopedColumnConfig` entry back with the 27 columns and boolean values defined in `DEFAULT_HIDDEN_COLS` in the code
3. Edit your `mopedColumnConfig` to contain:
```
{"project_team_members":false,"project_lead":true,"project_sponsor":true,"project_partner":false}
```
4. Refresh and you should see the 27 settings and the values of the columns above retained, and the rest of the selectable columns filled with default booleans.
5. Edit your `mopedColumnConfig` to contain:
```
{"current_phase":false,"interim_project_id":true,"public_process_status":true,"added_by":true,"project_tags":true,"contract_numbers":true,"contractors":true,"project_inspector":true,"project_designer":true,"completion_end_date":true,"construction_start_date":true,"funding_source_name":true,"type_name":true,"task_order":false,"project_feature":true,"parent_project_id":true,"children_project_ids":true,"project_note":false,"project_sponsor":true,"project_partner":true,"updated_at":true,"ecapris_subproject_id":true,"components":false}
```
6. Notice that this value contains a key/value pair for `task_order` which has been renamed to `task_orders`. 
7. Refresh the project list view. You should see `task_order` removed and `task_orders` added. Same as before, you should see the 27 settings and the values of the columns above retained, and the rest of the selectable columns filled with default booleans.
8. Edit your `mopedColumnConfig` to contain:
```
{"thisisnonsense": true}
```
9. Refresh and you should see the nonsense key removed and all of the 27 default settings applied.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
@@ -9,7 +9,7 @@ import {
   useTableComponents,
   useColumns,
   useTableOptions,
-  useHiddenColumns,
+  useHiddenColumnsSettings,
 } from "./helpers.js";
 import { useGetProjectListView } from "./useProjectListViewQuery/useProjectListViewQuery";
 import { PROJECT_LIST_VIEW_QUERY_CONFIG } from "./ProjectsListViewQueryConf";
@@ -91,7 +91,7 @@ const ProjectsListViewTable = () => {
   const { filters, setFilters, advancedSearchWhereString, isOr, setIsOr } =
     useAdvancedSearch();
 
-  const { hiddenColumns, setHiddenColumns } = useHiddenColumns();
+  const { hiddenColumns, setHiddenColumns } = useHiddenColumnsSettings();
 
   const { columns, columnsToReturnInQuery } = useColumns({ hiddenColumns });
 
--- a/moped-editor/src/views/projects/projectsListView/helpers.js
+++ b/moped-editor/src/views/projects/projectsListView/helpers.js
@@ -142,9 +142,16 @@ export const useTableComponents = ({
 
 const COLUMN_CONFIG = PROJECT_LIST_VIEW_QUERY_CONFIG.columns;
 
-export const useHiddenColumns = () => {
+export const useHiddenColumnsSettings = () => {
+  const existingHiddenColumnSettings = JSON.parse(
+    localStorage.getItem("mopedColumnConfig")
+  );
+
+  // TODO: Take the existing hidden columns and merge them with the default hidden columns
+  // TODO: Take the existing hidden columns and remove any that are not in the default hidden columns
+
   const [hiddenColumns, setHiddenColumns] = useState(
-    JSON.parse(localStorage.getItem("mopedColumnConfig")) ?? DEFAULT_HIDDEN_COLS
+    existingHiddenColumnSettings ?? DEFAULT_HIDDEN_COLS
   );
 
   /*
@@ -171,7 +178,7 @@ export const useColumns = ({ hiddenColumns }) => {
 
     // Some columns are dependencies of other columns to render, so we need to include them
     // Ex. Rendering ProjectStatusBadge requires current_phase_key which is not a column shown in the UI
-    const columnsNeededToRender = ["current_phase_key"];
+    const columnsNeededToRender = ["project_id", "current_phase_key"];
 
     return [...columnsShownInUI, ...columnsNeededToRender];
   }, [hiddenColumns]);
--- a/moped-editor/src/views/projects/projectsListView/helpers.js
+++ b/moped-editor/src/views/projects/projectsListView/helpers.js
@@ -133,8 +133,6 @@ const getExistingHiddenColumns = () => {
   return currentHiddenColumnSettings;
 };
 
-const initialHiddenColumnSettings = getExistingHiddenColumns();
-
 const COLUMN_CONFIG = PROJECT_LIST_VIEW_QUERY_CONFIG.columns;
 
 /**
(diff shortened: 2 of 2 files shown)


moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -169,9 +216,12 @@ export const useColumns = ({ hiddenColumns }) => {
...
       .map((key) => key);
 
-    // Some columns are dependencies of other columns to render, so we need to include them
+    // We must include project_id in every query since it is set as a keyField in the Apollo cache.
+    // See https://github.com/cityofaustin/atd-moped/blob/1ecf8745ef13277f784f3d8ba75efa13908acc73/moped-editor/src/App.js#L55
+    // See https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids
+    // Also, some columns are dependencies of other columns to render, so we need to include them.
     // Ex. Rendering ProjectStatusBadge requires current_phase_key which is not a column shown in the UI
-    const columnsNeededToRender = ["current_phase_key"];
+    const columnsNeededToRender = ["project_id", "current_phase_key"];
```
mddilley, line 224: This fixes another bug where Apollo complained about us not fetching the field that it uses to keep track of data in the cache.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -9,7 +9,7 @@ import {
   useTableComponents,
   useColumns,
   useTableOptions,
-  useHiddenColumns,
+  useHiddenColumnsSettings,
```
mddilley, line 12: I changed this to @johnclary's previous suggestion because I like it better. 🚀

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -92,6 +92,49 @@ const filterComponentFullNames = (value) => {
   ));
 };
 
+/**
+ * Get the user hidden column settings from local storage, clear them of
+ * outdated columns, and supplement with remaining default columns
+ * @returns {Object} hidden column settings from local storage
+ */
+const getExistingHiddenColumns = () => {
+  const existingHiddenColumnSettings = JSON.parse(
```
tillyw, line 101: this naming convention makes sense! but I found myself replacing 'existing' with 'previous' as I read through the code because I found the juxtaposition between 'previous' and 'current' to be a little clearer than 'existing' and 'current' which sound very similar in my mind. I definitely understand how 'existing' is more accurate but my suggestion would be to replace it with 'previous' for a more standard comparison.

tillyw (approved):
this is an elegant solution! i ran into the same error yesterday when i was standing up Moped in main. i'm glad y'all caught and fixed this so quickly ✨✨

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -92,6 +92,49 @@ const filterComponentFullNames = (value) => {
...
+    currentHiddenColumnSettings = DEFAULT_HIDDEN_COLS;
+  } else {
+    // Use existing settings to override default hidden columns.
+    // By iterating the defaults, we also remove outdated columns no longer in config.
+    currentHiddenColumnSettings = Object.keys(DEFAULT_HIDDEN_COLS).reduce(
+      (acc, columnName) => {
+        if (existingHiddenColumnSettings.hasOwnProperty(columnName)) {
+          return {
+            ...acc,
+            [columnName]: existingHiddenColumnSettings[columnName],
```
tillyw, line 123: ✨

roseeichelmann (approved):
looks great, thanks for the quick fix mike! :ship:

moped-editor/src/views/projects/projectsListView/helpers.js
mddilley: I agree, and I'm making this change - thank you! 'Previous' sounds much better when I'm reading through this now. 🙏

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -92,6 +92,49 @@ const filterComponentFullNames = (value) => {
...
 
+/**
+ * Get the user hidden column settings from local storage, clear them of
+ * outdated columns, and supplement with remaining default columns
+ * @returns {Object} hidden column settings from local storage
+ */
+const getExistingHiddenColumns = () => {
+  const existingHiddenColumnSettings = JSON.parse(
+    localStorage.getItem("mopedColumnConfig")
+  );
```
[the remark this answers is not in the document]
mddilley: Good catch! I'll add this to keep us away from the dreaded WSOD. 🙏

mddilley:
Thank y'all for the reviews and suggestions! 🙏 Made a couple small updates and ran through the tests again. Merging! 🚀

PR #1217, opened 2023-11-27
Add safe routes to school plan recommendations to component descriptions

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14476

This PR adds a script to add SRTS info from a Google sheet-exported csv file to project component descriptions by matching up SRTS IDs associated with both. Multiple project components can have the same SRTS ID.

## Testing
**URL to test:** 

n/a

**Steps to test:**
1. Follow the instructions in the readme added in this PR to get ready to run this script locally
2. When running the script, comment out line 142 (sleep) - otherwise it will take up to an hour to complete ⏳ 
3. You should see the following output (or similar depending on the prod data dump used)
```
Found 4673 rows in csv file.
Fetching components matching collected SRTS IDs.
Found 6486 components to update. Updating...
Done.

Updated 6486 components.
Found 12 SRTS IDs in csv with no match in database: ['1J - 901', '4C - 008', '3G - 901', '3B - 214', '4G - 002', '5A - 003', '2E - 032', '2K - 626', '3F - 901', 'EarlyOut-05', '3C - 004b', '3G - 016b']
Errors on project component IDs: []
```
4. To confirm that the SRTS info was added to records, look at the following `project_component_id`s in `moped_proj_components` in your local database:
   - id# 104 (example of record with `NULL` for existing description replaced with SRTS by this script)
   - id# 118 (example of record with empty string for existing description replaced with SRTS by this script)
   - id# 180 (example of record with existing description and SRTS info appended by this script)
5. As another way of testing, I modified the script to append an easily searchable string like `abc123` to each component description, and I confirmed that the number of rows that contain that string matches the number of rows modified by the script. You can check it out that way too if you'd like.
6. Last, check the activity log of the projects listed earlier in the steps. You should see updates credited to JD. These will be credited to **Data and Technology Admin** when this script is run on production. See [this discussion](https://austininnovation.slack.com/archives/CNUEPKLB1/p1701283671241449) for more about that.
---
#### Ship list
- [x] Code reviewed 


moped-toolbox/append_srts_desc_to_components_desc/secrets_template.py
```diff
@@ -0,0 +1,12 @@
+HASURA = {
+    "HASURA_GRAPHQL_ENDPOINT": {
```
chiaberry, line 2: the secrets template var names don't match up with the ones in the script, its missing GRAPHQL

mddilley: 👀 Ahhhhh, thanks for catching this! I fixed it in my real secrets file but not the template. 🙏

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
```diff
@@ -0,0 +1,152 @@
...
+        csv_reader = csv.reader(csv_file, delimiter=",")
+
+        line_count = 0
+        for row in csv_reader:
+            if line_count == 0:
+                line_count += 1
+                continue
+            else:
+                rows.append({"srts_id": row[0], "srts_info": row[1]})
+                line_count += 1
```
johnclary, line 59: this works fine—but just noting that [`csv.DictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader) will take care of this for ya.

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
```diff
@@ -0,0 +1,152 @@
...
+
+    return rows
+
+
+# This assumes all rows in the csv have a value for srts_info
```
johnclary, line 64: 👍

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
```diff
@@ -0,0 +1,152 @@
...
+    for row in rows:
+        srts_id = row["srts_id"]
+        srts_info = row["srts_info"]
+
+        existing_components_matched_by_srts_id = make_hasura_request(
+            query=GET_COMPONENTS_BY_SRTS_ID,
+            variables={"srts_id": srts_id},
+            endpoint=HASURA["HASURA_ENDPOINT"][env],
+            admin_secret=HASURA["HASURA_ADMIN_SECRET"][env],
+        )["moped_proj_components"]
```
johnclary, line 91: maybe you could add a condition to this query that excludes any component that contains the string `SRTS Recommendation:` in the description. that way you can run this again if any of the component updates fail for some reason.

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
```diff
@@ -0,0 +1,152 @@
...
+                    {
+                        "project_component_id": project_component_id,
+                        "description": new_description,
+                    }
+                )
+        else:
+            no_match.append(srts_id)
+
+    print(f"Found {len(updates)} components to update. Updating...")
+    for update in updates:
```
johnclary, line 112: one consideration is that each mutation is going to be firing off an activity event to the hasura API. it might be good to add a half second of sleep time between requests, if we're worried about overloading Hasura. iirc we're still running the beefed up Hasura cluster so hopefully it's not a problem.

johnclary (approved):
@mddilley thanks for putting this together! due to my VPN issues, i do not have a current copy of prod to test this on. i did test locally by adding an SRTS ID to one of the demo components, and this worked great.

i will mark this as approved, but i do think updating your Hasura query to check for components with existing SRTS recommendations will be worthwhile.

thanks again!

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
mddilley: Nice, agreed - I'll make that update so that this script is super one-and-done. 🚀 It would be a pain if there were a lot of failures!

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
mddilley: Cool, I'll make this change and [the ones that we discussed](https://austininnovation.slack.com/archives/CNUEPKLB1/p1701283671241449) to make sure we attribute these activities correctly.

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
mddilley: 🐍 🍄

mddilley:
hey all - I pushed some new code based on John's feedback. I added one more step to the test instructions to check the activity log for the updates made by this script. Let me know if y'all have any questions! 🙏

chiaberry (approved):
I was worried the script was hung up, so I added a print statement on each update, and it was working, just very slowly. I changed the sleep to be .1 seconds instead of .5 seconds to test locally. 6486 components at .5 second each is 3243 seconds (54 min)

Checking the components in the db, theres the new recommendations! 
Screenshot 2023-12-05 at 9 31 26 PM

mddilley:
Thanks for the reviews, y'all! 🙏 I'll probably run this script after hours tonight since it takes about an hour. Will bring this up in the release party to see if we are all cool with that.

moped-toolbox/append_srts_desc_to_components_desc/append_srts_desc.py
```diff
@@ -0,0 +1,173 @@
...
+        default="local",
+        help=f"Environment",
+    )
+
+    parser.add_argument(
+        "-t",
+        "--token",
+        type=str,
+        help=f"Bearer token for Hasura endpoint",
+    )
```
johnclary, line 169: nice addition by building this into the CLI! 🔥

PR #1218, opened 2023-11-29
14693 Adds visual feedback while refetching table data

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/14693

this pr adds a loading spinner when refetching data on the projects list table, specifically when the user searches (regular and advanced) or toggles on a new column

## Testing
**URL to test:** 

https://deploy-preview-1218--atd-moped-main.netlify.app/

**Steps to test:**
- go to the projects list view
- open dev tools and set throttling to slow 3g (you might have to make a custom throttling profile to slow network activity down enough to see the loading spinner)
- run a regular search, you should see a spinner pop up briefly in the top left corner of the table above the headers
- follow the same procedure with an advanced filter (try searching 'status' for a single character)
- follow the same procedure by toggling off all but one column, then adding one like 'status' or 'components' (these take a little longer to return than some of the others)

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/SearchBar.js
+++ b/moped-editor/src/components/GridTable/SearchBar.js
@@ -11,6 +11,7 @@ import {
   Icon,
   IconButton,
   Typography,
+  CircularProgress,
 } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import { Search as SearchIcon } from "react-feather";
@@ -100,6 +101,7 @@ const SearchBar = ({
   queryConfig,
   isOr,
   handleSwitchToSearch,
+  loading,
 }) => {
   const classes = useStyles();
 
@@ -206,6 +208,11 @@ const SearchBar = ({
           ),
           endAdornment: (
             <InputAdornment position="end">
+              {!!loading && (
+                <IconButton>
+                  <CircularProgress size="2rem" />
+                </IconButton>
+              )}
               <IconButton
                 onClick={toggleAdvancedSearch}
                 className={clsx({
--- a/moped-editor/src/views/projects/projectsListView/helpers.js
+++ b/moped-editor/src/views/projects/projectsListView/helpers.js
@@ -1,7 +1,7 @@
 import { useEffect, useMemo, useState } from "react";
 import { NavLink as RouterLink, useLocation } from "react-router-dom";
-import { MTableHeader } from "@material-table/core";
-import Link from "@mui/material/Link";
+import { MTableHeader, MTableToolbar } from "@material-table/core";
+import { Box, Link, CircularProgress } from "@mui/material";
 import typography from "../../../theme/typography";
 import parse from "html-react-parser";
 import { formatDateType, formatTimeStampTZType } from "src/utils/dateAndTime";
@@ -147,6 +147,7 @@ const COLUMN_CONFIG = PROJECT_LIST_VIEW_QUERY_CONFIG.columns;
  */
 export const useTableComponents = ({
   data,
+  projectListViewData,
   queryLimit,
   queryOffset,
   setQueryLimit,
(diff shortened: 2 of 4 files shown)


moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -176,9 +177,27 @@ export const useTableComponents = ({
...
+        <Box
+          sx={{
+            display: "flex",
+            justifyContent: "space-between",
+          }}
+        >
+          <Box sx={{ mt: 2, ml: 2 }}>
+            {/* If there is previous data but no current data (i.e., useQuery
+              is refetching the data, then display a progress spinner) */}
+            {!!data && !projectListViewData && <CircularProgress />}
```
mddilley, line 190: iirc we used to have a bug where `loading` that is returned from Apollo `useQuery` used to never switch back to false, but that doesn't seem to be the case anymore. Maybe when we upgraded versions a while back, but, however it happened, it is nice!

How do you feel about using [Apollo's loading state](https://www.apollographql.com/docs/react/api/react/hooks/#loading) here instead?

mddilley (commented):
This looks amazing! So nice to see some feedback that something is going on behind the scenes with the table data. 🚀 I left one comment about updating what state the spinner relies on. I think it is worth chasing.

moped-editor/src/views/projects/projectsListView/helpers.js
tillyw: works perfectly and much smoother, thanks for the suggestion Mike!

roseeichelmann (approved):
Works really well, thanks tilly! :sparkles: 

Unrelated to this PR but when I was testing the last step it was kinda annoying that everytime i toggled a column the side bar would close and I would have to open it again each time to turn off all the columns. Maybe that should stay open so people can toggle a whole lot at once? To be addressed in a diff issue obvi

tillyw:
@roseeichelmann @mddilley i'm able to recreate the issue in the deploy preview environment but it does not occur in my local environment. really odd. going to do some troubleshooting.

chiaberry:
Maybe its causing the toolbar to reload, which closes the column menu?
Is there a different place to put the loading symbol? Or does the styling not work out ?

tillyw:
Chia yeah you're totally right, whenever the loading state changes it triggers the useMemo hook to reload which reloads the entire toolbar component. Going to see if there is a way around that or move the spinner to a different spot.

tillyw:
> Sounds great! Not sure how helpful this is but I started from main branch and noticed that, as soon as I override the toolbar with `MTableToolbar` in `useTableComponents` (without introducing any kind of loading behavior), this regression starts happening. I wonder if that is reproducible, but I'll keep thinking about it.

i was able to reproduce this. the underlying issue seems to be inherent to material table, at least as far as overriding this component. moving the spinner somewhere else  ...

tillyw:
ok! problem solved. i just moved the spinner to john's second preferred spot. i honestly really like it there, it fits well and doesn't get covered up by the advanced search filter modal :)

Screenshot 2023-12-06 at 2 21 09 PM

moped-editor/src/components/GridTable/SearchBar.js
```diff
@@ -206,6 +208,11 @@ const SearchBar = ({
           ),
           endAdornment: (
             <InputAdornment position="end">
+              {!!loading && (
+                <IconButton>
+                  <CircularProgress size="2rem" />
+                </IconButton>
+              )}
```
johnclary, line 215: i love the way this looks! i noticed that the spinner is a clickable button, which i'd like to remove but also isn't critical. 

i think the trick to get the spacing right is to add a separate InputAdornment for the spinner. i will open a follow-up issue!

johnclary (approved):
this is a v fun little enhancement 😎  🚢 🚢 🚢 🚢

moped-editor/src/components/GridTable/SearchBar.js
tillyw: thanks, John! I did not mean to make that a clickable button :0 i'll merge now to get this into the release and make these changes in a follow-up pr

PR #1219, opened 2023-11-30
Fix bug where clicking a drawn line selects the underlying street segment as well

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14569

This PR fixes an issue where clicking on drawn lines or point when creating or editing a component also selected lines or points in the selectable layers they overlapped. When a click event happens, we can access `e.features` to see an array of all features captured in the click in all layers below the mouse. 

The original code only checked the first feature in this array, and the fix was to check if any of the features were drawn before interacting with the selectable CTN layers. This paired with disabling interactive map layers while drawing fixed this up.

Heads up - I found an edge case and captured it in https://github.com/cityofaustin/atd-data-tech/issues/14855

## Testing
**URL to test:** 

https://deploy-preview-1219--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to a new or existing project
2. Create a new line component
   - Click a line from the dotted CTN layer to select it
   - Start drawing a line component and be sure to end it right on top of the line selected in the last step
   - You should see that the draw click no longer deselects the line from the first step
   - Click a spot on the map away from dotted lines and your features
   - Click on your drawn line where it intersects with your selected line. The drawn feature should highlight and the selected line below should remain selected
   - Save your component so you can edit it later
   - You can test this in staging with the same steps to see the bug in action
3. Try similar steps while editing the component that you just saved. You should see the drawn features win every time they are selected and are overlapping selected or unselected CTN layers - except for the edge case that I linked above.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
@@ -117,6 +117,13 @@ export const isDrawnExistingFeature = (feature) =>
   !feature.source.includes("ATD_ADMIN.CTN_Intersections") &&
   !feature.source.includes("ATD_ADMIN.CTN");
 
+export const addDrawnFeatureProperties = (feature) => {
+  feature.properties.sourceLayer = feature.source_layer;
+};
+
+export const isDrawnFeature = (feature) =>
+  feature.properties.sourceLayer?.includes("drawnByUser");
+
 /**
  * Initialize the draw tools state with existing features of the component being edited
  * @param {Object} drawControlsRef - ref for the draw controls
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
@@ -100,12 +100,21 @@ export const getDrawId = (feature) => feature?.properties?.DRAW_ID;
 
 /**
  * Determine if a draft feature is drawn by the user by the draw ID
- * that is added when creating a new drawn feature
+ * that is added when creating a drawn feature that is not yet saved
  * @param {Object} feature - a feature object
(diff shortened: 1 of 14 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
```diff
@@ -100,22 +100,48 @@ export const getDrawId = (feature) => feature?.properties?.DRAW_ID;
...
+ * Determine if a feature was saved previously by checking for the source layer
+ * data that is only present in saved drawn features
  * @param {Object} feature - a feature object
  * @returns {Boolean} - true if feature is drawn by user
  */
-export const isDrawnExistingFeature = (feature) =>
-  !feature.source.includes("ATD_ADMIN.CTN_Intersections") &&
-  !feature.source.includes("ATD_ADMIN.CTN");
+export const isDrawnSavedFeature = (feature) =>
+  feature.properties.sourceLayer?.includes("drawnByUser");
```
johnclary, line 135: not at all in scope of this PR, but i kind of want to make a check constraint that enforces the `drawnByUser` value in the feature tables in the DB. it's weird that we can put any string we want to and it would cause big problems.

chiaberry (approved):
Thanks for the demos in dev sync and keeping us in the loop!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/features.js
mddilley: yeah, this feels like it needs to tightening up to me too! Took me a bit to remember that this string is set in the frontend code (`makeDrawnFeature`), and it is hard to understand how this fits into the big picture.

mddilley:
@johnclary woohoo - thanks for merging! 🚀

PR #1224, opened 2023-12-11
Add fallback error boundary component

chiaberry:
## Associated issues

Discovery Day

https://github.com/bvaughn/react-error-boundary

- https://github.com/cityofaustin/atd-data-tech/issues/14995

## Testing
**URL to test:** 

https://deploy-preview-1224--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

Easier to test locally. 

1. Test the apollo error handler by removing the mopd-admin's permission to one of the fields in the project list view (like project name)

Screenshot 2023-12-18 at 3 35 30 PM

2. Introduce an error to the project list view or the activity log, i.e. try to find the length of null 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/ApolloErrorHandler.js
+++ b/moped-editor/src/components/ApolloErrorHandler.js
@@ -25,7 +25,7 @@ const ApolloErrorHandler = props => {
 
   // Error Variables
   const error = props?.error ?? null;
-  const errorString = JSON.stringify(error);
+  const errorString = JSON.stringify(error, Object.getOwnPropertyNames(error));
   const jwtError = errorString.includes("JWT") || errorString.includes("token");
 
   if (jwtError) {
--- a/moped-editor/src/components/ApolloErrorHandler.js
+++ b/moped-editor/src/components/ApolloErrorHandler.js
@@ -25,7 +25,7 @@ const ApolloErrorHandler = props => {
 
   // Error Variables
   const error = props?.error ?? null;
-  const errorString = JSON.stringify(error, Object.getOwnPropertyNames(error));
+  const errorString = error ? JSON.stringify(error, Object.getOwnPropertyNames(error)) : "";
   const jwtError = errorString.includes("JWT") || errorString.includes("token");
 
   if (jwtError) {
--- a/moped-editor/src/components/ApolloErrorHandler.js
+++ b/moped-editor/src/components/ApolloErrorHandler.js
@@ -40,7 +40,7 @@ const ApolloErrorHandler = props => {
           <CircularProgress color="inherit" />
         </Backdrop>
       ) : error ? (
-        <p>Error: {error.message}</p>
+        <p>Error: {error.message}</p> // here
       ) : (
         props.children
       )}
--- a/moped-editor/src/components/ApolloErrorHandler.js
+++ b/moped-editor/src/components/ApolloErrorHandler.js
@@ -2,6 +2,7 @@ import React from "react";
 import Backdrop from "@mui/material/Backdrop";
 import CircularProgress from "@mui/material/CircularProgress";
 import makeStyles from '@mui/styles/makeStyles';
+import FallbackComponent from "./FallbackComponent";
 
 const useStyles = makeStyles(theme => ({
   backdrop: {
@@ -40,7 +41,7 @@ const ApolloErrorHandler = props => {
           <CircularProgress color="inherit" />
         </Backdrop>
       ) : error ? (
-        <p>Error: {error.message}</p> // here
+        <FallbackComponent error={error}/>
       ) : (
         props.children
       )}
--- a/moped-editor/src/components/FallbackComponent.js
+++ b/moped-editor/src/components/FallbackComponent.js
@@ -1,6 +1,7 @@
 import React from "react";
 import { Box, Container, Typography } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
+import ExternalLink from "./ExternalLink";
 
 const useStyles = makeStyles((theme) => ({
   root: {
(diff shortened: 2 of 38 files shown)


johnclary:
@chiaberry the error boundary is awesome! what do you think about this design?
Screenshot 2023-12-15 at 1 04 55 PM

moped-editor/src/views/projects/projectView/ProjectActivityLog.js
```diff
@@ -215,89 +217,91 @@ const ProjectActivityLog = () => {
...
-                        <Box display="flex" p={0}>
-                          <Box p={0}>
-                            <CDNAvatar
-                              className={classes.avatarSmall}
-                              src={change?.updated_by_user?.picture}
-                              initials={getInitials(change?.updated_by_user)}
-                              // todo: do we want this to not be always gray if its just the initials?
-                              userColor={null}
-                            />
+      <ErrorBoundary FallbackComponent={FallbackComponent}>
```
chiaberry, line 220: the only difference here is wrapping everything in this error boundary.

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -24,7 +25,7 @@ const ApolloErrorHandler = props => {
 
   // Error Variables
   const error = props?.error ?? null;
-  const errorString = JSON.stringify(error);
+  const errorString = JSON.stringify(error, Object.getOwnPropertyNames(error));
```
chiaberry, line 28: https://stackoverflow.com/questions/18391212/is-it-not-possible-to-stringify-an-error-using-json-stringify

this never worked, the errorString was always going to be `{ }`

johnclary (changes requested):
this is looking really good, @chiaberry!!

moped-editor/src/components/ApolloErrorHandler.js
chiaberry: good test of the error boundary!

chiaberry:
I moved the error boundary up to the App component. The drawback to this is you don't get the moped logo/navigation menu. I've seen recommendations to use error boundaries lower in the tree so it doesn't prevent users from using other parts of the app, like if there is a wsod in the activity log, users can still navigate to other projects, etc. 

Screenshot 2023-12-20 at 8 32 38 AM

johnclary:
> I've seen recommendations to use error boundaries lower in the tree so it doesn't prevent users from using other parts of the app

@chiaberry that makes sense. would you suggest we have the boundary at multiple levels? i just want to make sure we handle as many errors as possible.

moped-editor/src/components/FallbackComponent.js
```diff
@@ -0,0 +1,84 @@
...
+  `https://atd.knack.com/dts#new-service-request/?view_249_vars=%7B%22field_398%22%3A%22Bug%20Report%20%E2%80%94%20Something%20is%20not%20working%22%2C%22field_399%22%3A%22Moped%22%2C%22field_400%22%3A%22${error.message}%22%7D`;
+
+const FallbackComponent = ({ error, resetErrorBoundary }) => {
+  const classes = useStyles();
+
+  return (
+    <Box
+      display="flex"
+      flexDirection="column"
+      height="100%"
```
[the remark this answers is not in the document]
mddilley: I read y'all's latest thoughts on this, and I am also split on how high or low in the tree to go with the error boundaries. It is nice to keep the rest of the app going when an error occurs somewhere deeper in the main layout, but that creates decisions to be made about where to place them. 🤔 Maybe the top level of the main views/routes at least? Besides those, the components map jumps out at me since there are a lot of moving pieces, and it is hard to test.

I also noticed that this library has a `resetErrorBoundary` function to reset and try again with a button. I think that would have the same effect as refreshing the page (like the error component says to do) in a lot of our cases, though.

Last thought, we might consider the type of errors that we aim to catch and how/if that could intersect with other ways of catching code errors like TS.

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -40,7 +41,7 @@ const ApolloErrorHandler = props => {
           <CircularProgress color="inherit" />
         </Backdrop>
       ) : error ? (
-        <p>Error: {error.message}</p>
+        <FallbackComponent error={error}/>
```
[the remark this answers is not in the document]
mddilley: Agreed - this makes for a much nicer experience when there is an Apollo error!

mddilley (commented):
So rad! ✨ I see that there is still some conversation on where this is going to land so I dropped some comments in the meantime. Can't wait to see this in action.

johnclary (approved):
I have tested this by adding errors to the TobBar, as well as various project view components. it works great! 

I like @mddilley's suggestions to add in a button to reset the error, and thinking more explicitly about the errors we're handling. I will open a backlog issue for additional enhancements.

chiaberry:
@johnclary there are so many ways to go with enhancements. I'll add some to the backlog issue as i think of them

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -24,7 +25,7 @@ const ApolloErrorHandler = props => {
 
   // Error Variables
   const error = props?.error ?? null;
-  const errorString = JSON.stringify(error);
+  const errorString = error ? JSON.stringify(error, Object.getOwnPropertyNames(error)) : "";
```
mddilley, line 28: TIL about `Object.getOwnPropertyNames`!

mddilley (approved):
Tested locally again, and this is awesome. Love this direction for error handling! Another discovery day exploration makes it way into our code. 🙌 🚢

PR #1225, opened 2023-12-12
Advanced filters with number inputs can break gql query with allowed characters

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/14582

## Testing
**URL to test:** 
https://deploy-preview-1225--atd-moped-main.netlify.app/moped/ 

**Steps to test:**
1. [Current behavior](https://moped.austinmobility.io/moped/projects) for number type filter fields such as ID is that if you type something like "1e1" into the input, the search button will appear after the second `1` and the user can search with this invalid input causing an ugly graphql error to slip into the UI.
2. Now in the deploy preview try to replicate that error. You should find that the search button is disabled until valid input is entered. If invalid input is entered the text box will turn red and give the user an error warning and the search button disables again.
3. Also test adding multiple filters, make one of them have incorrect input. Only the input text box for that field should be red and show the message "Invalid input"

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -173,6 +173,11 @@ const Filters = ({
    */
   const [filterComplete, setFilterComplete] = useState(false);
 
+  /**
+   * Tracks whether the field input value is invalid
+   */
+  const [isInvalidInput, setIsInvalidInput] = useState(false);
+
   /* First filter is an empty placeholder so we check for more than one filter */
   const areMoreThanOneFilters = Object.keys(filterParameters).length > 1;
   const isFirstFilterIncomplete =
@@ -373,7 +378,11 @@ const Filters = ({
    * @param {string} filterId - The FilterID uuid
    * @param {string} value - The value to assign to that filter
    */
-  const handleSearchValueChange = (filterId, value) => {
+  const handleSearchValueChange = (filterId, value, type) => {
+    // Validate input for number type fields and set state accordingly
+    type === "number" && isNaN(value)
+      ? setIsInvalidInput(true)
+      : setIsInvalidInput(false);
     // Clone the state
     const filtersNewState = { ...filterParameters };
     // Patch the new state
@@ -685,16 +694,22 @@ const Filters = ({
                       />
                     ) : (
                       <TextField
+                        error={isInvalidInput(filter)}
+                        helperText={isInvalidInput ? "Invalid input" : ""}
                         key={`filter-search-value-${filterId}`}
                         id={`filter-search-value-${filterId}`}
                         disabled={!filterParameters[filterId].operator}
                         type={
-                          filterParameters[filterId].type
+                          filterParameters[filterId].type === "date"
                             ? filterParameters[filterId].type
-                            : "text"
+                            : null
                         }
                         onChange={(e) =>
-                          handleSearchValueChange(filterId, e.target.value)
+                          handleSearchValueChange(
+                            filterId,
+                            e.target.value,
+                            filterParameters[filterId].type
+                          )
                         }
                         variant="outlined"
                         value={filterParameters[filterId].value ?? ""}
@@ -763,18 +778,17 @@ const Filters = ({
           </Grid>
         </Hidden>
         <Grid item xs={12} md={2}>
-          <Grow in={handleApplyValidation() === null}>
-            <Button
-              fullWidth
-              className={classes.applyButton}
-              variant="contained"
-              color="primary"
-              startIcon={<Icon>search</Icon>}
-              onClick={handleApplyButtonClick}
-            >
-              Search
-            </Button>
-          </Grow>
+          <Button
+            fullWidth
+            className={classes.applyButton}
+            variant="contained"
+            color="primary"
+            startIcon={<Icon>search</Icon>}
+            onClick={handleApplyButtonClick}
+            disabled={isInvalidInput || handleApplyValidation() != null}
+          >
+            Search
+          </Button>
         </Grid>
       </Grid>
     </Grid>
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -173,11 +173,6 @@ const Filters = ({
    */
   const [filterComplete, setFilterComplete] = useState(false);
 
-  /**
-   * Tracks whether the field input value is invalid
-   */
-  const [isInvalidInput, setIsInvalidInput] = useState(false);
-
   /* First filter is an empty placeholder so we check for more than one filter */
   const areMoreThanOneFilters = Object.keys(filterParameters).length > 1;
   const isFirstFilterIncomplete =
@@ -378,11 +373,7 @@ const Filters = ({
    * @param {string} filterId - The FilterID uuid
    * @param {string} value - The value to assign to that filter
    */
-  const handleSearchValueChange = (filterId, value, type) => {
-    // Validate input for number type fields and set state accordingly
-    type === "number" && isNaN(value)
-      ? setIsInvalidInput(true)
-      : setIsInvalidInput(false);
+  const handleSearchValueChange = (filterId, value) => {
     // Clone the state
     const filtersNewState = { ...filterParameters };
     // Patch the new state
@@ -430,6 +421,9 @@ const Filters = ({
             if (gqlOperator && !gqlOperator.includes("is_null")) {
               feedback.push("• One or more missing values.");
             }
+          }
+          if (checkIsInvalidInput(filterParameters[filterKey])) {
+            feedback.push("• One or more invalid inputs.");
           }
         });
       }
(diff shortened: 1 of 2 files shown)


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -491,6 +494,16 @@ const Filters = ({
...
 
+  /**
+   * Returns whether the user input value for the filterParameter is valid
+   * @param {object} filterParameter
+   * @returns {boolean}
+   */
+  const checkIsInvalidInput = (filterParameter) => {
+    // Validate input for number type fields
+    return filterParameter.type === "number" && isNaN(filterParameter.value);
+  };
```
roseeichelmann, line 505: i left the function naming here ambiguous even though it currently only does validation for the number input type. could be extended to other types in the future (thought idk if its necessary bc cant break them as is) or i could update this to `checkIsValidNumber` or something.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -685,13 +699,15 @@ const Filters = ({
                       />
                     ) : (
                       <TextField
+                        error={isInvalidInput}
+                        helperText={isInvalidInput ? "Invalid input" : ""}
```
roseeichelmann, line 703: is this helpful enough or should it be more specific "Invalid input, numbers only" or something

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -685,13 +699,15 @@ const Filters = ({
...
                         key={`filter-search-value-${filterId}`}
                         id={`filter-search-value-${filterId}`}
                         disabled={!filterParameters[filterId].operator}
                         type={
-                          filterParameters[filterId].type
+                          filterParameters[filterId].type === "date"
                             ? filterParameters[filterId].type
-                            : "text"
+                            : null
```
roseeichelmann, line 710: "number" type being used here was causing me issues bc it will automatically set the value to an empty string "" if non ints are in the input and the onChange call wont fire, making it so that i cant check if the input is invalid. [MUI docs](https://github.com/mui/material-ui/blob/master/docs/data/material/components/text-fields/text-fields.md#typenumber) actually recommend avoiding using `type={"number")` for this component and we really only need this distinction here for the "date" type so thats why i modified this

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -763,18 +779,17 @@ const Filters = ({
           </Grid>
         </Hidden>
         <Grid item xs={12} md={2}>
-          <Grow in={handleApplyValidation() === null}>
```
roseeichelmann, line 766: the search button appearing and disappearing kinda threw me off so i replaced this behavior in favor of disabling the search button if there are errors with the filter validation

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -763,18 +779,17 @@ const Filters = ({
...
-            </Button>
-          </Grow>
+          <Button
+            fullWidth
+            className={classes.applyButton}
+            variant="contained"
+            color="primary"
+            startIcon={<Icon>search</Icon>}
+            onClick={handleApplyButtonClick}
+            disabled={handleApplyValidation() != null}
```
roseeichelmann, line 789: disable search button if feedback is returned

moped-editor/src/components/GridTable/Filters.js
johnclary: i like `checkIsValidNumber` or just `isValidNumber`. this is purely a stylistic thing, but fwiw my brain wants this function to check for "isValid" instead of "isInvalid". ie return `true` if the value is a number. 

And this function can be moved outside of the main component since it doesn't have any dependencies in the component scope 😇

moped-editor/src/components/GridTable/Filters.js
johnclary: thanks for catching this. totally agree that the "number" type input only creates headaches. not critical, but it might be cool to test out using the `inputmode` and `pattern` attributes. GOV.UK did a great [write up](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) about this!

moped-editor/src/components/GridTable/Filters.js
johnclary: It'd be good to be more specific. maybe "Must be a number" ?

moped-editor/src/components/GridTable/Filters.js
johnclary: sort of related to the input type discussion further below, i noticed that whitespace characters do not break the search, which was a bit of a surprise. i was going to suggest that we trim whitespace from the user input, but that does not seem worth the trouble.

johnclary (changes requested):
Thanks @roseeichelmann for patching this up! And the disabled search button is a nice  refinement. My feedback is very minor :)

moped-editor/src/components/GridTable/Filters.js
mddilley: Tagging on here to say that the regex used in the existing `removeNonIntegers` helper could be helpful here. I can still enter "1e1" because that string passes the `isNaN` test. 

[TIL that `isNaN` coerces input if needed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description). Isn't JavaScript the best?! 😵‍💫

moped-editor/src/components/GridTable/Filters.js
mddilley: +1 - not seeing that button appear and disappear is ✨

moped-editor/src/components/GridTable/Filters.js
mddilley: Just adding that it makes me so happy to see feedback here 🙌

mddilley (commented):
I dropped some comments, and I'll take another look early next week. The validation being more apparent is 🔥!

moped-editor/src/components/GridTable/Filters.js
roseeichelmann: Oh man :facepalm: at me not testing the input that originally caught this bug hahaha.

Thx for the catch mike, just pushed up some changes using regex instead! The way `isNaN` works is so odd :face_with_spiral_eyes:

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -417,11 +431,14 @@ const Filters = ({
             feedback.push("• One or more operators have not been selected.");
           }
 
-          if (value === null || value === "") {
+          if (value === null || value.trim() === "") {
```
roseeichelmann, line 434: Search button will be disabled if user only entered white space @johnclary

roseeichelmann:
@johnclary @mddilley pushed up some updates addressing y'alls feedback!

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -123,6 +123,20 @@ const generateEmptyField = (uuid) => {
   return { ...defaultNewFieldState, id: uuid };
 };
 
+/**
+ * Returns whether the user input value for the filterParameter is valid
+ * @param {object} filterParameter
+ * @returns {boolean}
+ */
+const checkIsValidInput = (filterParameter) => {
```
mddilley, line 131: I see now that the existing validation was tied to this `filterParameter` object. I'll be taking a close look at the way this component uses its filter state this sprint so this is a nice preview! 🙏

moped-editor/src/components/GridTable/Filters.js
mddilley: @roseeichelmann I re-read John's comment here, and I realized that we did something similar [in this currency component](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/views/projects/projectView/DollarAmountIntegerField.js). `inputmode` would be a good addition here, but it would also add more logic of when/when not to add that attribute based on the filter `type`. 

Feels like future work to handle different `type`s by composing input components instead of toggling prop values like we do now. Curious what others think!

mddilley (approved):
I tested this on the preview, and I am not able to break the UI with invalid inputs any longer. Really nice improvement with the text input too. 🙌 🚀

moped-editor/src/components/GridTable/Filters.js
johnclary: yep. this also makes me appreciate the [Yup-style validation](https://github.com/cityofaustin/atd-moped/blob/f7e7a5f1c0106d1ffecd289eaa5a0c9b6f281b30/moped-editor/src/views/projects/projectView/ProjectWorkActivity/utils/form.js#L22) where you can pass in a message per validation rule. but it definitely makes sense to hold off until we see a bigger need for validation on these filters.

johnclary (approved):
looking good now—thanks @roseeichelmann!! 🚢 🚢 🚢 🚢

chiaberry (approved):
i like the submit button being disabled if the input isnt correct!

PR #1226, opened 2023-12-14
Switch to decoded URL params for tracking project list view state

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14588

This PR updates the current base64 encoded URL params in the project list view to contain less data and be url encoded. React Router handles the encoding since it uses [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and that is the default behavior. 

This PR refactors the format of the data that we set and get from the URL, and it sets the stage to update the format of the `filterParameters` state within the `Filters` component too. For now, I created helpers to go between the previous and new formats as needed and document what the UI expects to be able to show the selected filters in the UI.

**What was previously stored in the URL params**
See the issue linked above

**What is stored with the new changes**
```
[
   { 
      "field":"type_name",
      "operator":"string_equals_case_sensitive",
      "value":"Fiber - New"
   },
   ...other filters
]
```

## Testing
**URL to test:** 

https://deploy-preview-1226--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**
1. Go to the project list view and add advanced filters
2. As you add them, check that the list is filtering and that the url is storing new filters. [This site](https://meyerweb.com/eric/tools/dencoder/) always ends up coming up first when I look for a quick encode/decode tool so it could be helpful to see what is in the url.
3. Test a filter with lookup tables like **Type**
4. Test editing applied filters options (choose new field option, choose new operator option, or update search value)
5. Test deleting filters with the trash can icon
6. Test resetting the filters by clicking the reset button
7. Test making edits to the url string to make sure the app doesn't break when it can't parse the URL params
8. Test filtering, clicking on a project name link in the list view, and then clicking the "< ALL PROJECTS" link to go back to the list view. Your filters should still be present.

---
#### Ship list
- [x] Code reviewed 


mddilley:
@johnclary I still have polish left to do here but feel free to check this out. I'll pick back up on this on Monday. 🚧

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -123,6 +132,57 @@ const generateEmptyField = (uuid) => {
...
+ * @param {Array} filters - filters from the URL {field, operator, value}
+ * @param filtersConfig - The configuration object for the filters
+ * @return {Object}
+ */
+const makeInitialFilterParameters = (filters, filtersConfig) => {
+  if (filters.length === 0) return [];
+
+  return filters.reduce((acc, filter) => {
+    const filterUUID = uuidv4();
+    const { field, operator, value } = filter;
```
mddilley, line 146: These are the three values stored in the URL now. This function takes them and puts together all of the other config needed for the UI.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -123,6 +132,57 @@ const generateEmptyField = (uuid) => {
...
+ * Get configuration about the filters using the query parameters present in the URL
+ * @param {Array} filters - filters from the URL {field, operator, value}
+ * @param filtersConfig - The configuration object for the filters
+ * @return {Object}
+ */
+const makeInitialFilterParameters = (filters, filtersConfig) => {
+  if (filters.length === 0) return [];
+
+  return filters.reduce((acc, filter) => {
+    const filterUUID = uuidv4();
```
mddilley, line 145: After spending a lot of time in this code, I found that the uuid provides a way to reliably point to and edit an existing filter when there are interactions in the UI. A unique ID is used because the filters are stored in an object, and don't have an index or guaranteed order.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -163,7 +223,19 @@ const Filters = ({
    * @function setFilterParameters - Update the state of filterParameters
    * @default {filters}
    */
-  const [filterParameters, setFilterParameters] = useState(filters);
+  const initialFilterParameters = useMemo(() => {
```
mddilley, line 226: Since we aren't storing the full filter configuration in the url anymore, we need to build it for any existing filters in the url.

moped-editor/src/components/GridTable/FiltersCommonOperators.js
```diff
@@ -143,9 +143,9 @@ export const FiltersCommonOperators = {
     type: "date",
   },
   date_less_than_equal_to: {
-    operator: "_lt",
+    operator: "_lte",
     label: "less than or equal to",
-    description: "The date is less than.",
+    description: "The date is less than or equal to.",
```
mddilley, line 148: Small bug fix that I snuck it. 🩹  We were generating the same query for dates less than and dates less than or equal to

moped-editor/src/components/GridTable/SearchBar.js
```diff
@@ -177,13 +179,15 @@ const SearchBar = ({
...
-  if (filterStateActive) {
-    Object.keys(filters).map((parameter) =>
-      filtersApplied.push(filters[parameter]["label"])
-    );
-  }
+  const filterStateActive = filters.length > 0;
+  const filtersApplied =
+    filterStateActive &&
+    filters.map((filter) => {
+      const fieldFilterConfig = filtersConfig.fields.find(
```
mddilley, line 186: We have to grab the filter name to display since we don't have it in state anymore. Not sure how I feel about having to use `find()` here, but changing the shape of `PROJECT_LIST_VIEW_FILTERS_CONFIG.fields` from an array to an object feels like a big change to me. We don't have that many fields to iterate but I'm not writing it off completely.

chiaberry (approved):
I am really excited for this change! you mentioned there will be future refactoring to do, what updates are you thinking?

mddilley:
@chiaberry I touched base with John, and I'll write up an issue to bring to refinement today. The overall idea is to spread the way the state for the query filters was refactored in this PR to the state that manages the filter UI.

moped-editor/src/components/GridTable/Filters.js
johnclary, line 1: This is not related to your work, but while testing I noticed that deleting a filter causes the project search to update immediately, but, if edit an existing filter, the edits are not applied until you click the **Search** button. And if you edit filters but then close the filter menu without click **Search,** the edits are discarded. 

Ideally, deleting a filter would behave the same way, and have no effect until you click the **Search** button. Anyway, I will capture this in the [project search epic](https://github.com/cityofaustin/atd-data-tech/issues/14312) 😅 


Edit: [here](https://github.com/cityofaustin/atd-data-tech/issues/15074) is the issue.

johnclary (approved):
@mddilley this a huge improvement!! i'm so excited to have more manageable filters in place, and for the future enhancements this unlocks. 

i have run through all of your tests cases, and everything is working just as promised! I have tried various means of breaking this feature with crufty URL params, and I have been unsuccessful. 

the code looks very great—i'm sorry that i don't have any substantive feedback for you. thanks also for capturing related work in the [project search epic](https://github.com/cityofaustin/atd-data-tech/issues/14312).

🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/components/GridTable/Filters.js
mddilley: +1 - 100%. The refactor that I've been working on for the internal state should help us get away from building that object, remove the need for the uuids, and mirror the shape of the `filters` state. 🤞

PR #1227, opened 2023-12-18
Add update audit field trigger, tear down event trigger, add preset for updated-by

frankhereford:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


moped-database/metadata/tables.yaml
```diff
@@ -4072,68 +4078,8 @@
...
-      webhook_from_env: HASURA_ENDPOINT
-      headers:
-        - name: x-hasura-admin-secret
-          value_from_env: ACTIVITY_LOG_API_SECRET
-      request_transform:
-        body:
-          action: transform
-          template: |-
-            {
-                            "query": "mutation InsertActivity($object: moped_activity_log_insert_input!, $project_id:Int!, $updated_at:timestamptz ) { insert_moped_activity_log_one(object: $object) { activity_id } update_moped_project_by_pk(pk_columns: {project_id: $project_id}, _set: {updated_at: $updated_at}) { updated_at }}",
```
johnclary, line 4110: My note in the issue description about removing this event was misleading, because we do want to remove the `update_moped_project_by_pk` mutation, but we need to preserve the `insert_moped_activity_log_one` mutation that inserts into the activity log.

johnclary (changes requested):
thanks @frankhereford! this works great. just need to get the activity log event trigger back in place 🙏

moped-database/metadata/tables.yaml
frankhereford: Ah, got it! Thank you for the clarification John. I'll get this fixed up right after dev sync. 🙏

frankhereford: Got it, I think! I went back and forth more times than I want to admit because it was throwing errors if i left the `project_id` and `updated_at` values in the variables payload and also broken if i took them out. Finally,  I realized that I had forgotten to take them out of the graphql arguments themselves. 🤦

GraphQL still gives me a turn-around sometimes. A query language that comes in the form of functions. It's hard for old dogs to learn new tricks!

johnclary (approved):
I've tested this new functionality and it works perfectly—and I see the activity log updating as expected ✅ ✅ ✅ 

Thanks for this, @frankhereford!!!! 🚢

roseeichelmann:
I'm actually still not seeing the activity log updates? Is that just me?

mddilley:
@roseeichelmann I just tested this locally, and I think that the missing link might be logging in to the app using the login in the **Moped Local Editor User Login** 1PW. 

Long story short/my best summary, the activity log expects a user that is in the DB, and that login has a match in the seed data. If you are running locally with prod data, you can run `UPDATE "public"."moped_users" SET "user_id" = 1 WHERE "user_id" = 60;` on the DB in addition to using that login.

mddilley (approved):
Test steps worked great - thanks for adding this! 🙏 🚀 🚢

chiaberry (approved):
👍🏼

PR #1232, opened 2023-12-28
Unify various logins for Moped testing and development

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15039

This PR:
- updates the seed data to match up with the staging admin and editor users stored in 1PW
- adds a migration that fixes a foreign key constraint issue that is blocking the final task in the issue above
- adds a script to help encrypt/decrypt our user claims (also part of the final task in the issue above)

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start Moped locally (using `./hasura-cluster start` and not from production data)
2. Log in using the username/pw stored in the 1PW entry called **Moped Local Editor User Login**
3. Add a status or note to a project or anything else that is tracked in the activity log
4. You should see that there is no foreign key error and that the activity log shows the correct user for the update you made
5. Now, log in using the 1PW entry called **Moped Staging & Production Test Editor** and repeat the same steps
6. Last, start Moped locally with production data
7. Using a DB client, execute the following statement:
```sql
UPDATE moped_users SET user_id = 1 WHERE email = 'transportation.data+mopedadmin@austintexas.gov';
```
8. Do y'all think the last step should happen in a migration? It would have no effect on staging since the record with this email in staging is already `user_id` 1. We just already have to do a manual operation on the production DynamoDB table to make everything match up so I lumped it in with manual changes.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
+++ b/moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
@@ -1,3 +1,3 @@
 ALTER TABLE moped_project_files DROP CONSTRAINT "moped_project_files_created_by_fkey";
 
-ALTER TABLE moped_project_files ADD CONSTRAINT "moped_project_files_created_by_fkey" FOREIGN KEY ("created_by") REFERENCES "public"."moped_users"("user_id") ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE moped_project_files ADD CONSTRAINT "moped_project_files_created_by_fkey" FOREIGN KEY ("created_by") REFERENCES "public"."moped_users"("user_id") ON UPDATE CASCADE;
--- a/moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
+++ b/moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE moped_project_files DROP CONSTRAINT "moped_project_files_created_by_fkey";
+
+ALTER TABLE moped_project_files ADD CONSTRAINT "moped_project_files_created_by_fkey" FOREIGN KEY ("created_by") REFERENCES "public"."moped_users"("user_id") ON DELETE CASCADE ON UPDATE CASCADE;
--- a/moped-database/seeds/1602292389297_initial_seed_staging.sql
+++ b/moped-database/seeds/1602292389297_initial_seed_staging.sql
@@ -20,10 +20,10 @@ SET row_security = off;
 -- Data for Name: moped_users; Type: TABLE DATA; Schema: public; Owner: moped
 --
 
-INSERT INTO public.moped_users (first_name, last_name, title, user_id, workgroup_id, cognito_user_id, date_added, is_coa_staff, email, roles, picture, is_deleted, is_user_group_member, note) VALUES ('Data and Tech', 'Admin', 'Moped Admin', 1, 3, NULL, '2021-03-09 17:08:14+00', true, 'transportation.data+mopedadmin@austintexas.gov', '["moped-admin"]', NULL, false, false, 'This account is used by the DTS for automated tasks and integrations.');
+INSERT INTO public.moped_users (first_name, last_name, title, user_id, workgroup_id, cognito_user_id, date_added, is_coa_staff, email, roles, picture, is_deleted, is_user_group_member, note) VALUES ('Data and Tech', 'Admin', 'Moped Admin', 1, 3, NULL, '2021-03-09 17:08:14+00', true, 'admin@emailhost.xyz', '["moped-admin"]', NULL, false, false, 'This account is used by the DTS for automated tasks and integrations.');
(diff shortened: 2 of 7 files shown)


moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
mddilley, line 1: This change will allow us to update the Data and Tech Admin user (currently `user_id` 29) to `user_id` 1 without violating the existing foreign key constraint.

moped-toolbox/encrypt_or_decrypt_dynamo_db_claims/claims.py
mddilley, line 1: I made this tooling since I was tired of spinning a Python environment up and down, and I also forgot how this worked. Hopefully this helps the next time we find ourselves needing to manually operate on the encrypted claims.

mddilley, line 1: I made this tooling since I grew impatient with spinning a Python environment up and down, and I also forgot how this worked 😅. Hopefully this helps the next time we find ourselves needing to manually operate on the encrypted claims.

moped-editor/src/auth/user.js
```diff
@@ -100,10 +100,7 @@ export const initializeUserDBObject = (userObject) => {
       body: JSON.stringify({
         query: ACCOUNT_USER_PROFILE_GET_PLAIN,
         variables: {
-          userId:
-            config.env.APP_ENVIRONMENT === "local"
-              ? 1
-              : getDatabaseId(userObject),
```
mddilley, line 106: I think that, as long as we test with either the test admin or test editor accounts locally, then we shouldn't need to default to `user_id` 1 anymore. Same with the other change below that defaults to `user_id` when showing the `Profile` component.

moped-toolbox/encrypt_or_decrypt_dynamo_db_claims/claims.py
johnclary: very good to know! 🙌

moped-database/seeds/1602292389297_initial_seed_staging.sql
```diff
@@ -20,9 +20,10 @@ SET row_security = off;
 -- Data for Name: moped_users; Type: TABLE DATA; Schema: public; Owner: moped
 --
 
-INSERT INTO public.moped_users (first_name, last_name, title, user_id, workgroup_id, cognito_user_id, date_added, is_coa_staff, email, roles, picture, is_deleted, is_user_group_member, note) VALUES ('JD', 'Maccombs', 'Software Developer', 1, 18, NULL, '2021-03-09 17:08:14+00', true, 'jd@emailhost.xyz', '["moped-admin"]', NULL, false, false, 'JD was contracted to assist with the initial Moped research and build out.');
```
johnclary, line 23: it would probably be good to use a fake email address instead of the same email we use for a real staging login 🤷‍♂️

moped-editor/src/auth/user.js
johnclary: i'm so glad you addressed this! it only added to the confusion of what the heck is going on with logins 😅

moped-toolbox/encrypt_or_decrypt_dynamo_db_claims/README.md
johnclary, line 1: ✨

moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE moped_project_files DROP CONSTRAINT "moped_project_files_created_by_fkey";
+
+ALTER TABLE moped_project_files ADD CONSTRAINT "moped_project_files_created_by_fkey" FOREIGN KEY ("created_by") REFERENCES "public"."moped_users"("user_id") ON DELETE CASCADE ON UPDATE CASCADE;
```
johnclary, line 3: ah—we've never had to cascade a user ID change!

i don't feel strongly, but i am slightly worried about cascading deletes from the user table. i think it might be unintuitive that a user deletion causes their file data to delete, but no other user-related tables cascade deletes afaik.

johnclary (approved):
this looks great, @mddilley! Lots of bonus content in here ✨ 

I left a comment about cascading deletes—your call though.

🚢 🚢 🚢 🚢

moped-database/migrations/1703786941217_modify_files_user_id_constraint/up.sql
mddilley: cool! yeah, i was a little weary of the delete part too. I'll update it to on work on update. 🙏

moped-database/seeds/1602292389297_initial_seed_staging.sql
mddilley: 👍 Yep, good call. I'll update this too. 👤

chiaberry (approved):
test steps check out, and thank you for removing the defaulting to userid 1 when running locally.

johnclary (approved):
thanks for making those fixes. this looks great!

mddilley:
thanks y'all 🙏 - going to merge this so I can get it into the release and then do the final couple steps after the DB changes make their way to production later in the week.

PR #1236, opened 2024-01-08
Refactor filterParameters state to use search params state shape and filter config

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15049
https://github.com/cityofaustin/atd-data-tech/issues/15074

This PR:
- updates the `filterParameters` state to have the same shape and key-value pairs (field, operator, and value) as the `filters` state that powers the query
- moves a lot of functions defined within the `Filters` component into a helpers file
- removes code that used to access configuration about a field and then store it in state so that it now accessing that configuration as needed

If you want to visualize the biggest change, run locally and log `filterParameters` in the `Filters` component as you add advanced filters. Compare what that looks like with this branch versus `main` branch.

## Testing
**URL to test:** 

https://deploy-preview-1236--atd-moped-main.netlify.app/moped/projects

**Steps to test:**
1. Go to the project list view and add advanced filters
2. Test a filter with lookup tables like **Type**
3. Test editing applied filters options (choose new field option, choose new operator option, or update search value)
4. Test deleting filters with the trash can icon. A new search should not happen until you click the Search button.
5. Test clearing the first **Field** input after selecting a value by using the X icon that appears to the left of the dropdown icon. Both the **Field** and **Operator** values should clear from the UI.
6. Test resetting the filters by clicking the reset button
7. Test making edits to the url string to make sure the app doesn't break when it can't parse the URL params
8. Test filtering, clicking on a project name link in the list view, and then clicking the "< ALL PROJECTS" link to go back to the list view. Your filters should still be present.
9. Add a filter that expects a number value like **ID is** and try entering letters. You should see that the recently-added validation still works.
10. Check that the trash can delete icons and match any/all radios do not appear until more than one filter is added.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -237,11 +237,11 @@ const Filters = ({
       return [generateEmptyField()];
     }
   }, [filters, filtersConfig]);
-  console.log(initialFilterParameters);
 
   const [filterParameters, setFilterParameters] = useState(
     initialFilterParameters
   );
+  console.log(filterParameters);
 
   /* Track toggle value so we update the query value in handleApplyButtonClick */
   const [isOrToggleValue, setIsOrToggleValue] = useState(isOr);
@@ -448,7 +448,7 @@ const Filters = ({
    */
   const handleSearchValueChange = (filterIndex, value) => {
     // Clone the state
-    const filtersNewState = { ...filterParameters };
+    const filtersNewState = [...filterParameters];
     // Patch the new state
     filtersNewState[filterIndex].value = value;
     // Update the state
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -132,6 +132,30 @@ const checkIsValidInput = (filterParameter) => {
 };
 
 /**
+ * It returns true if the GraphQL operator is null type and has no
+ * @param {String} operator - operator to check
+ * @returns {boolean}
+ */
+const isFilterNullType = (operator) => {
+  return operator && OPERATORS_WITHOUT_SEARCH_VALUES.includes(operator);
+};
+
+/**
+ * Check if all filters added have complete values
+ * @param {Array} filterParameters - filter values, fieldd, and operatord
+ * @returns {boolean}
+ */
+const areAllFiltersComplete = (filterParameters) => {
+  return filterParameters.every((filter) => {
+    return (
+      !!filter.field &&
+      !!filter.operator &&
+      (!!filter.value || isFilterNullType(filter.operator))
+    );
+  });
+};
+
+/**
  * Filter Search Component aka Advanced Search
  * @param {Object} filters - The current filters from useAdvancedSearch hook
  * @param {Function} setFilters - Set the current filters from useAdvancedSearch hook
@@ -188,22 +212,10 @@ const Filters = ({
   /* Track toggle value so we update the query value in handleApplyButtonClick */
   const [isOrToggleValue, setIsOrToggleValue] = useState(isOr);
 
-  /**
-   * Tracks whether the user has added a complete filter
-   */
-  const [filterComplete, setFilterComplete] = useState(false);
-
   /* First filter is an empty placeholder so we check for more than one filter */
   const areMoreThanOneFilters = filterParameters.length > 1;
   const isFirstFilterIncomplete =
-    filterParameters.length === 1 && !filterComplete;
-
-  const generateEmptyFilter = useCallback(() => {
-    // Clone state and add empty filter
-    const filtersNewState = [...filterParameters, generateEmptyField()];
-    // Update new state
-    setFilterParameters(filtersNewState);
-  }, [filterParameters, setFilterParameters]);
+    filterParameters.length === 1 && !areAllFiltersComplete(filterParameters);
 
   /**
    * Returns true if Field has a lookup table associated with it and operator is case sensitive
@@ -263,7 +275,10 @@ const Filters = ({
    * Adds an empty filter to the state
    */
   const handleAddFilterButtonClick = () => {
-    generateEmptyFilter();
+    // Clone state and add empty filter
+    const filtersNewState = [...filterParameters, generateEmptyField()];
+    // Update new state
+    setFilterParameters(filtersNewState);
   };
 
   /**
@@ -380,31 +395,6 @@ const Filters = ({
     setSearchFieldValue("");
     setSearchTerm("");
   };
-
-  /**
-   * It returns true if the GraphQL operator is null type and has no
-   * @param {String} operator - operator to check
-   * @returns {boolean}
-   */
-  const isFilterNullType = (operator) => {
-    return operator && OPERATORS_WITHOUT_SEARCH_VALUES.includes(operator);
-  };
-
-  // TODO: We can replace this side effect by checking if each filter in filterParameters
-  // has a value and operator
-  // TODO: Remember to handle filters without values (OPERATORS_WITHOUT_SEARCH_VALUES)
-  /**
-   * This side effect monitors whether the user has added a complete filter
-   */
-  useEffect(() => {
-    filterParameters.forEach((filter) => {
-      if (!!filter.value || isFilterNullType(filter.operator)) {
-        setFilterComplete(true);
-      } else {
-        setFilterComplete(false);
-      }
-    });
-  }, [filterParameters]);
 
   const handleAndOrToggle = (e) => {
     const isOr = e.target.value === "any";
@@ -676,7 +666,7 @@ const Filters = ({
         <Grid item xs={12} md={2}>
           <Button
             // Disable button until the user has added a complete filter
-            disabled={!filterComplete}
+            disabled={!areAllFiltersComplete(filterParameters)}
             className={classes.bottomButton}
             fullWidth
             variant="outlined"
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -625,9 +625,8 @@ const Filters = ({
 
         const { label } = fieldConfig ?? {};
         // const availableOperators =
-        // const operator =
-        // const lookupTable =
-        // const lookupField =
+        const { table_name: lookupTable, field_name: lookupField } =
+          fieldConfig.lookup ?? {};
         // const type =
 
         // support check with isFilterNullType()
@@ -730,23 +729,18 @@ const Filters = ({
                     (renderAutocompleteInput(filterParameters[filterIndex]) ? (
                       <Autocomplete
                         value={value || null}
-                        options={
-                          data[filterParameters[filterIndex].lookup_table]
-                        }
+                        options={data[lookupTable]}
                         disabled={!filterParameters[filterIndex].operator}
                         getOptionLabel={(option) =>
-                          Object.hasOwn(
-                            option,
-                            filterParameters[filterIndex].lookup_field
-                          )
-                            ? option[filterParameters[filterIndex].lookup_field]
+                          Object.hasOwn(option, lookupField)
+                            ? option[lookupField]
                             : option
                         }
                         onChange={(e, value) => {
                           if (value) {
                             handleSearchValueChange(
                               filterIndex,
-                              value[filterParameters[filterIndex]?.lookup_field]
+                              value[lookupField]
                             );
                           } else {
                             // value is null when the Autocomplete selection is cleared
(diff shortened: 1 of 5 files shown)


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -1,6 +1,5 @@
-import React, { useCallback, useEffect, useState, useMemo } from "react";
+import React, { useCallback, useState, useMemo } from "react";
 import { useSearchParams } from "react-router-dom";
-import { v4 as uuidv4 } from "uuid";
```
mddilley, line 3: Since the shape of `filterParameters` is now an array, we can use the index of a filter in the array to point to the filter to edit, delete, etc. Previously, we used an object so we needed to assign a unique ID so that we could point to the right filter in the object.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -96,107 +96,6 @@ const useStyles = makeStyles((theme) => ({
...
-  return { ...defaultNewFieldState, id: uuid };
-};
-
-/**
- * Get configuration about the filters using the query parameters present in the URL
- * @param {Array} filters - filters from the URL {field, operator, value}
- * @param filtersConfig - The configuration object for the filters
- * @return {Object}
- */
-const makeInitialFilterParameters = (filters, filtersConfig) => {
```
mddilley, line 141: We don't need this helper anymore. It took the three values stored in the search parameters (field, operator, and value) and built up the shape of data that the UI expected from our predefined `FiltersCommonOperators` and `PROJECT_LIST_VIEW_FILTERS_CONFIG`. It was a bridge to the UI changes made in this PR. Most of this work happens in the render now.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,108 @@ const Filters = ({
...
-      AUTOCOMPLETE_OPERATORS.includes(field.operator)
-    );
-  };
+  /* Some features like all/any radios require more than one filter to appear */
+  const areMoreThanOneFilters = filterParameters.length > 1;
 
   /**
    * Handles the click event on the field drop-down menu
-   * @param {string} filterId - State FieldID to modify
+   * @param {string} filterIndex - filterParameters index to modify
```
mddilley, line 160: I renamed `filterId` throughout this component to `filterIndex`. Hoping that makes it clearer that we are dealing with array indices instead of the uuids.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,109 @@ const Filters = ({
...
-
-      // Find the field we need to gather options from
-      const fieldDetails = filtersConfig.fields.find(
-        (filter) => filter.name === field
-      );
-
-      if (!fieldDetails) {
-        filtersNewState[filterId] = generateEmptyField(filterId);
-      } else {
-        // Update field & type
```
mddilley, line 310: From here and below in this `handleFilterFieldMenuClick` function, the code was building up new state (similar to what `makeInitialFilterParameters` above was doing). This happens in the return below where `filterParameters.map` happens. The available operators logic here is captured in a new `getAvailableOperators` helper.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,109 @@ const Filters = ({
...
+    const fieldDetails = filtersConfig.fields.find(
+      (filter) => filter.name === field
+    );
+
+    // Update the field and operator values or handle when the clear X icon is clicked
+    if (fieldDetails) {
+      filtersNewState[filterIndex].field = fieldDetails.name;
+      filtersNewState[filterIndex].operator = getDefaultOperator(fieldDetails);
     } else {
-      console.debug(
```
mddilley, line 368: imo this debug message isn't needed any longer since the number of filter rows in the UI and the number of items in the `filterParameters` state is equal. There should always be a corresponding filter to a filter row being interacted with.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -481,103 +273,26 @@ const Filters = ({
...
-   */
-  const isFilterNullType = (field) => {
-    return field.gqlOperator && field.gqlOperator.includes("is_null");
-  };
-
-  /**
-   * This side effect monitors the count of filters
-   * if the count of filters is zero, then resets the state
-   */
-  useEffect(() => {
```
mddilley, line 554: Making sure that an empty filter is in `filterParameters` state now happens on first render when `initialFilterParameters` is determined. Also, since the trash can icon disappears when there is only one filter, we don't have to handle the filter count dropping to 0 that way.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -481,103 +273,26 @@ const Filters = ({
...
-      handleClearFilters();
-      // Add an empty filter so the user doesn't have to click the 'add filter' button
-      generateEmptyFilter();
-    }
-  }, [filterParameters, setFilters, generateEmptyFilter, handleClearFilters]);
-
-  /**
-   * This side effect monitors whether the user has added a complete filter
-   */
-  useEffect(() => {
```
mddilley, line 566: The `filterComplete` state is removed and replaced with the `areAllFiltersComplete` helper. We can use that function to check if all filters are complete during render since it is calculated from `filterParameters`.

johnclary (changes requested):
😍 so much red!

this is such a major improvement, Mike! i ran into one regression, and besides that my comments were pretty minor.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -637,14 +352,33 @@ const Filters = ({
...
-        const isValidInput = checkIsValidInput(filterParameters[filterId]);
+      {filterParameters.map((filter, filterIndex) => {
+        const { field: fieldName, operator, value } = filter;
+
+        /* Get field configuration and values need for inputs */
+        const fieldConfig = filtersConfig.fields.find(
+          (field) => field.name === fieldName
+        );
+        const { label, type } = fieldConfig ?? {};
+        const operators = fieldConfig?.operators;
```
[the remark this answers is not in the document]
mddilley: nah not nits to me - this is great stuff and thank you for taking such a detailed look at this! 🙏 I'll address these comments today.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,109 @@ const Filters = ({
...
+      // if we are switching to an autocomplete input, clear the search value
+      if (shouldRenderAutocompleteInput(lookupTable, operator, loading)) {
+        filtersNewState[filterIndex].value = null;
+      }
     } else {
-      console.debug(
-        `The filter id ${filterId} does not exist, ignoring click event.`
-      );
+      // Reset operator value
+      filtersNewState[filterIndex].operator = null;
```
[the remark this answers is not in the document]
frankhereford: Maybe if the user is using the dev tools to modify the request on the way the server or Postman to create malformed ones for devious intent? I don't think it's something we need to worry about too much, but it did come to mind wrt your question.

frankhereford (approved):
Mike, I have gone over your code, but there is a lot of complexity going on here with a system that I have never really had a chance to work on, so I don't feel really qualified to be able to give you any meaningful feedback, beyond that it looks great. Nothing stuck out to me as even remotely out of place in your work.

I was able to test the deploy preview, and I wasn't able to get it to do anything that it shouldn't. The behavior all felt natural and expected, and nothing broke. I think this is ship shape, and I'm really grateful for this huge amount of work you've done!

moped-editor/src/components/GridTable/Filters.js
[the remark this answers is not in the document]
chiaberry: I'm not sure how that would work, I dont see how the operator would not be in the list of operators, since the select dropdown is populated based on that list of operators. Directly editing the operator in the url to something else doesn't affect this event handler either

moped-editor/src/components/GridTable/Filters.js
mddilley: This is interesting - i think that you might be on to something. I fixed this by adding the two operators that this filters uses ("subprojects_array_is_null" and "subprojects_array_is_not_null") to the list the code looks at when determining whether to show the value input or not (`OPERATORS_WITHOUT_SEARCH_VALUES`). Somehow the prod code handles this case without this change. 🤔

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,109 @@ const Filters = ({
...
-        if (renderAutocompleteInput(filtersNewState[filterId])) {
-          filtersNewState[filterId].value = null;
-        }
-      } else {
-        // Reset operator values
-        filtersNewState[filterId].operator = null;
-        filtersNewState[filterId].gqlOperator = null;
-        filtersNewState[filterId].envelope = null;
-      }
+  const handleFilterOperatorClick = (filterIndex, operator, lookupTable) => {
```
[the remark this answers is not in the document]
mddilley: great call - i'll update them. 👍

moped-editor/src/components/GridTable/Filters.js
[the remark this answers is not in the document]
mddilley: This is a really good catch because this code used to pull double duty and call `handleFilterOperatorChange` from within `handleFilterFieldMenuChange` to reset the operator input's value. So, this block was meant to handle that case. 

TLDR It used to be able to happen, but it cannot anymore. I'll remove it. 🗑️  Thanks, y'all!

moped-editor/src/components/GridTable/Filters.js
mddilley: Added "string_is_blank" and "string_is_not_blank" to that list as well to fix broken **Status is blank** or **Status is not blank** 😑

moped-editor/src/components/GridTable/Filters.js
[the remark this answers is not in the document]
mddilley: This also helped me unearth an issue when switching from existing filter with a search value like **Names contains congress** to one without a search value like **Lead is not blank**. The search value was hanging around in the search params so I made another code update. 🙏 🙏 🙏

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -230,215 +129,109 @@ const Filters = ({
...
-    };
-    // Try to delete the filter by filterId
+  const handleDeleteFilterButtonClick = (filterIndex) => {
+    // Clone the state
+    const filtersNewState = [...filterParameters];
+    // Try to delete the filter by filter index
     try {
       // Delete the key (if it's there)
-      delete filtersNewState[filterId];
+      filtersNewState.splice(filterIndex, 1);
```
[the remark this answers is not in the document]
mddilley: Yup! I checked and this can go and is now gone 👋

moped-editor/src/components/GridTable/Filters.js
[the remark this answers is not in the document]
mddilley: This took me a second, but, yes, I see it now. Yeah, I overlooked how the list of operators are keys in `FiltersCommonOperators`. I'm going to update so that the form input just accesses the label name using the operator key. I can get rid of a helper this way too. 😈

johnclary (changes requested):
@mddilley this batch of changes looks really good—thank you!

i noticed an annoying regression, which does not seem to occur in production currently. The filters break if you switch the filter field from a text-based to a number-based field. For example, create a filter on **Name** **contains** **decker**, then switch the field to **ID**.

Screenshot 2024-01-11 at 12 10 13 PM

you mind taking a look at that?

mddilley:
@johnclary yup, this is a regression related to using * in the `PROJECT_LIST_VIEW_FILTERS_CONFIG` field config to represent [using "all the filters for a specific type"](https://github.com/cityofaustin/atd-moped/pull/1236/commits/205cb33acf49357a44bbca1e5026f7fc278e16a6#diff-e290922fa8581df0703a6896695b704d451d814d5eb327082744c66e960d5afeL139-L148). I'm going to make the two fields that use this shortcut more explicit. `getAvailableOperators` handled this special case but I removed it. 💥 🙃

johnclary (approved):
Yep—this is ready to ship! Thanks @mddilley for all the iterations. 

You love to see it:
Screenshot 2024-01-16 at 12 55 27 PM

🚢 🚢 🚢

mddilley:
Going to merge this one and add a note to [this thread](https://github.com/cityofaustin/atd-moped/pull/1241#discussion_r1453791101) since @tillyw was going to need to merge this branch into that PR's branch.

PR #1239, opened 2024-01-10
Refactor audit tables on `moped_proj_funding`

frankhereford:
## Associated issues

This PR hopes to close https://github.com/cityofaustin/atd-data-tech/issues/15086.

## Testing

This has some DB changes, so this should be tested locally. 

**Steps to test:**

* Spin up the database, optionally replicating from production.
* Observe that migrations and metadata work.
* Spin up your editor.
* Pick a project, and add a funding source.
* Query that record in the DB and see that `created_at` and `created_by_user_id` are populated.
* Edit that record through the editor UI and observe that `updated_at` and `updated_by_user_id` are populated correctly in the `moped_proj_funding` record and that the parent project's record has its `updated_at` and `updated_by_user_id` fields populated. 
* Check out the permissions for `moped_proj_funding` in the hasura console.
* Check out the migration, and check out the down migration. Do they look correct?
* Check the changes to the event trigger for the funding table. Does it correctly no longer manipulate the audit fields in the project table?

---
#### Ship list
- [ ] Code reviewed 


moped-database/.gitignore
```diff
@@ -1,3 +1,4 @@
 atd_moped-full.sql.gz
 *moped_dump.sql
 production.sql
+reload.sh
```
frankhereford, line 4: I'd like to include my reload script in here until we have time to enhance the `hasura-cluster` script with automatic loading from a snapshot functionality.

moped-database/metadata/tables.yaml
```diff
@@ -2523,22 +2535,22 @@
...
-                                    "record_project_id": {{ $body.event.data.new.project_id }},
-                                    "record_data": {"event": {{ $body.event }}},
-                                    "description": [{"newSchema": "true"}],
-                                    "operation_type": {{ $body.event.op }},
-                                    "updated_by_user_id": {{ $session_variables?['x-hasura-user-db-id'] ?? 1}}
-                                },
-                                "updated_at": {{$body.created_at}},
-                                "project_id": {{$body.event.data.new.project_id}}
-                            }
-                        }
```
frankhereford, line 2541: This was outdented way far, so I drew it back inline with the YAML. Everything seems to work fine, but please let me know if that extra whitespace was important. Thanks!

moped-database/metadata/tables.yaml
```diff
@@ -2523,22 +2535,22 @@
...
+              "query": "mutation InsertActivity( $object: moped_activity_log_insert_input!) { insert_moped_activity_log_one(object: $object) { activity_id } }",
```
frankhereford, line 2538: Beyond the formatting, the meaningful change here is the removal of the portion of the mutation which had previously been keeping the update-audit-fields in `moped_project` up to date, which is now done in the database internally.

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
```diff
@@ -0,0 +1,34 @@
...
+
+    -- Update the updated_at and updated_by_user_id fields in the related row of moped_project
+    UPDATE moped_project
+    SET updated_at = NOW(),
+        updated_by_user_id = NEW.updated_by_user_id
+    WHERE project_id = NEW.project_id;
+
+    RETURN NEW;
+END;
+$$;
```
frankhereford, line 25: In the issue for this work, [16086](https://github.com/cityofaustin/atd-data-tech/issues/15086), wisely points out the consideration that we already have a couple of functions in our database related to these audit update fields, and it will be messy and hard to maintain if we end up making a whole bunch of them, especially if the names don't follow a scheme etc. I recognize that consideration and here's what I'm thinking in this PR and going forward. 

* `update_parent_records_audit_logs()` exists, and when it's triggered to run, it wants to update a parent `moped_proj_component()` record and grandparent `moped_project` record.
* `set_updated_at()` exists, and when it runs, it wants to update _self_, meaning whatever table triggered the f unction call. 
* This PR creates `update_self_and_project_updated_audit_fields()` which updates _self_ and a parent `moped_project` record.

I think if you think about these functions in the context of what chain of tables they are going to take to get back up to moped_project, you will have discovered all the cases for which we're going to create a new function. Neither of the existing functions covered the "self + `moped_project`" case, so a new function was needed, but this new function should work for all the tables that are one step away from `moped_project`.

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
```diff
@@ -0,0 +1,34 @@
...
+
+COMMENT ON FUNCTION public.update_self_and_project_updated_audit_fields() IS 'Function to update the updated_at field in the current row of moped_proj_funding and the updated_at and updated_by_user_id fields in the related row of moped_project.';
```
frankhereford, line 27: I think we should establish a naming convention for these triggers, and I regret not having the foresight to see that earlier. Should that be spun off or something to wrap into this issue? I don't want to scope creep unless y'all think i should be. I don't have one thought up yet to propose, but i could. I am leaning towards spin-off, but I'm cool with anything.

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -546,7 +546,6 @@ const ProjectFundingTable = () => {
                 objects: {
                   ...newData,
                   project_id: projectId,
-                  added_by: userId,
```
frankhereford, line 549: The `graphql-engine` instance is configured by this PR to not accept any audit field values from insert mutation, because it figures out who did what based on the authentication token given to the endpoint. That's why this came out and nothing went back in. I appreciate you all considering this and thinking about if i'm looking at this right.

moped-database/.gitignore
johnclary: rather than ignore it, any reason not to commit it? 😇

frankhereford:
```
#!/bin/bash

#docker exec -it -e PGPASSWORD=$MOPED_RR_PASSWORD moped-database-moped-pgsql-1 pg_dump -U $MOPED_RR_USERNAME -h moped-read-replica.austinmobility.io atd_moped > production.sql

./hasura-cluster stop
./hasura-cluster start_only
docker stop moped-database-hasura-1
echo 'DROP DATABASE moped; CREATE DATABASE moped' | docker exec -i moped-database-moped-pgsql-1 psql -U moped postgres;
docker exec -i moped-database-moped-pgsql-1 psql -U moped moped < production.sql; 
docker start moped-database-hasura-1
sleep 5
hasura migrate apply
hasura metadata apply
#echo 'UPDATE moped_users set user_id = 1 where user_id = 60;' | docker exec -i moped-database-moped-pgsql-1 psql -U moped moped;
```

I think we should probably spend time time integrating it properly into `hasura-cluster` since, at least in its current form, it relies on how I have my shell configured and what i export by default into my shell.

But if you think it'd be helpful, I'm totally game to drop it in there!! Just say the word!

johnclary: ok, that's fair! thanks for posting it here :)

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
johnclary: > This PR creates update_self_and_project_updated_audit_fields() which updates self and a parent moped_project record.

Got it—perfect! 

> `set_updated_at()` exists

We will be able to remove that function by re-configuring the `moped_proj_work_activities` table to match what you've done here. I will make sure that gets captured in a follow-up issue 🙏 


> we're going to need a new function for every combination of tables that are required to get back up to `moped_projects`

I have an idea for handling these cases that I shared in Slack. Since that "bubbling up" is not a concern for this PR, I will make sure to capture a potential follow-up in a separate issue.

johnclary (changes requested):
@frankhereford this is really great work. i appreciate your thoughts + guidance on how to manage the function sprawl. my feedback is pretty minor. definitely this is good to open up for wider review! 🙏

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -467,10 +467,10 @@ export const ProjectActivityLogTableMaps = {
       project_id: {
         label: "project ID",
       },
-      date_added: {
+      created_at: {
         label: "date added",
       },
-      added_by: {
+      created_by_user_id: {
```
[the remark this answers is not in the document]
frankhereford: Great call John, I'd never thought of that implication. Thank you!

Can you look at what I've done and see if it passes the test? I dropped a comment in there to leave some info, but it's maybe too long. I'll edit if you want for sure.

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
frankhereford: Thanks for thinking that through with me!

moped-database/metadata/tables.yaml
```diff
@@ -2523,22 +2535,22 @@
...
+                      "record_type":  {{ $body.table.name }},
+                      "activity_id": {{ $body.id }},
+                      "record_project_id": {{ $body.event.data.new.project_id }},
+                      "record_data": {"event": {{ $body.event }}},
+                      "description": [{"newSchema": "true"}],
+                      "operation_type": {{ $body.event.op }},
+                      "updated_by_user_id": {{ $session_variables?['x-hasura-user-db-id'] ?? 1}}
+                  },
+                  "updated_at": {{$body.created_at}},
+                  "project_id": {{$body.event.data.new.project_id}}
```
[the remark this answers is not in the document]
frankhereford: Such a good catch John, thank you! I've cleaned up that template in the tables.yaml file. 

image

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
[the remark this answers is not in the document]
johnclary: looks great 🙏

moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -467,12 +467,31 @@ export const ProjectActivityLogTableMaps = {
...
       },
+      
+
+      // The following two pairs fields represent the same conceptual idea, however, 
+      // the first pair were the original names of the fields and the second pair
+      // are names we later migrated to for these columns. Both need to be defined
+      // as part of this object to support activity log records which contain the 
+      // older field names.
+
+      // original field names
```
chiaberry, line 478: 👍🏼 thank you for including this for future us!

chiaberry (commented):
I tested adding and editing funding sources and signed in as different users, the updated_by_user_id is updating, but for some reason I'm not seeing updated_at timestamp. 
Screenshot 2024-01-18 at 9 42 03 AM

Also I found a pre-existing bug, where I can't update the amount of a funding record, but that also exists in at least staging, so its not from this PR.

frankhereford:
> I tested adding and editing funding sources and signed in as different users, the updated_by_user_id is updating, but for some reason I'm not seeing updated_at timestamp.

Thanks Chia - I will work on replicating that and a fix. I'll let you know what I find asap.

mddilley (changes requested):
This looks, works, and tests perfectly to me. I found one tiny artifact of the old fields in the funding table that I think should be removed, and then I'll ✅ 🚢 ✅ 🚢 asap. Thanks for thinking through these changes so carefully. 🔍 🙌

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
```diff
@@ -0,0 +1,34 @@
...
+CREATE OR REPLACE FUNCTION public.update_self_and_project_updated_audit_fields() RETURNS TRIGGER
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    -- Update the updated_at field in the current row of the triggering table
+    NEW.updated_at := NOW();
+
+    -- Update the updated_at and updated_by_user_id fields in the related row of moped_project
+    UPDATE moped_project
+    SET updated_at = NOW(),
```
johnclary, line 19: looking at this again @frankhereford, I think we probably want to to use the value from `NEW.updated_at` rather than calling `NOW()` again here? just to ensure those two values are 100% identical?

frankhereford: That sounds great John, I'm in there right now addressing some feedback from Mike, and I'll get that in too.

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
```diff
@@ -0,0 +1,34 @@
...
+        updated_by_user_id = NEW.updated_by_user_id
+    WHERE project_id = NEW.project_id;
+
+    RETURN NEW;
+END;
+$$;
+
+COMMENT ON FUNCTION public.update_self_and_project_updated_audit_fields() IS 'Function to update the updated_at field in the current row of moped_proj_funding and the updated_at and updated_by_user_id fields in the related row of moped_project.';
+
+CREATE TRIGGER update_self_and_project_audit_fields_audit_fields
```
johnclary, line 29: @frankhereford 
 taking a second look at this trigger name, can i suggest:

```
update_moped_proj_funding_and_project_audit_fields
```

...or similar? to distinguish this trigger from future triggers which attach the same function to different tables.

johnclary (changes requested):
Frank, thanks for these updates. Sorry to throw a few more nits your way!

frankhereford:
> Frank, thanks for these updates. Sorry to throw a few more nits your way!

Naw, it's great. I've made the changes that you and Mike have suggested and am testing now. Thanks!

moped-editor/src/views/projects/projectView/ProjectFundingTable.js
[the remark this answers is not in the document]
frankhereford: Thank you Mike - I have removed those references to those fields that no longer need to be removed. Awesome catch; I wouldn't have thought about that!!

moped-database/migrations/1704919187107_moped_proj_funding_rename_columns/up.sql
frankhereford: Great idea - using the first timestamp in the second operation will ensure they are always identical. It's in there.

johnclary:
strange, I also cannot replicate Chia's issue :/

wrt Frank's question:
> so I've updated the code to populate the insert audit fields on insert and the updated audit fields to populate on insert and update

Yep—the trigger behavior looks exactly right to me 🙏 

I think we should ship this feature and keep an eye on the potential problem Chia is seeing. I am going to queue up an issue to fix the pre-existing bug she encountered as well. The funding mutation is all messed up.

johnclary (approved):

mddilley (approved):
Reviewed the latest commits, and this looks 🚢 to me! I also cannot replicate the bug, and I agree with John that we can patch if needed.

chiaberry (approved):
I pulled the latest updates and restarted the db from seed data, and I see the updated at timestamps now 🚢

PR #1240, opened 2024-01-11
Persist search term in project list URL parameters

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15206

This PR adds a `search` param that is set by the project list view search box. It maintains the current way that we use either the search box or the advanced filter individually and never combine the two when filtering the project list results.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. In the search box at the top of the project list view, enter a search term like **lamar** and press enter. You should see matching results in the view.
2. Check the url and you should see the search parameter like `https://localhost:3000/moped/projects?search=lamar`
3. Open the advanced filters by clicking the filter button in the search box
4. Add a filter like **Name contains kramer** and click "Search". The search box should clear, the url should no longer have a param called `search`, and it should be replaced with the `filters` and `isOr` params for the advanced search
5. Type another term in the search box like **braker** and press enter. You should see the `filters` and `isOr` params removed from the url and the `search` param added. The list view should show matches for your search term.
6. Click into the search box while it is populated and clear it by backspacing until it is blank. When it is blank, you should see the `search` param removed from the url and have all of the projects listed again.
7. Click the global search magnifying glass icon next to the **New Project** button
8. Enter a search term like **mueller** and then click the **More results** link at the end of the list. You should be redirected to the project list view with the search box populated and the `search` param in the url populated with **mueller**. The project list view should show matching results.
9. Click the **Name** link of a row in the results and navigate to the project details. In the top left corner of the project details, click the **< ALL PROJECTS** link, and you should navigate back to the project list view with the previous search term in the search box and `search` param in the url.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/SearchBar.js
+++ b/moped-editor/src/components/GridTable/SearchBar.js
@@ -17,6 +17,10 @@ import {
 import makeStyles from "@mui/styles/makeStyles";
 import { Search as SearchIcon } from "react-feather";
 import clsx from "clsx";
+import {
+  advancedSearchFilterParamName,
+  advancedSearchIsOrParamName,
+} from "src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch";
 import { simpleSearchParamName } from "src/views/projects/projectsListView/useProjectListViewQuery/useSearch";
 
 /**
@@ -149,6 +153,8 @@ const SearchBar = ({
     // Update state to trigger search and set simple search param
     setSearchTerm(searchFieldValue);
     setSearchParams((prevSearchParams) => {
+      prevSearchParams.delete(advancedSearchFilterParamName);
+      prevSearchParams.delete(advancedSearchIsOrParamName);
       prevSearchParams.set(simpleSearchParamName, searchFieldValue);
       return prevSearchParams;
     });
--- a/moped-editor/src/components/GridTable/SearchBar.js
+++ b/moped-editor/src/components/GridTable/SearchBar.js
@@ -127,7 +127,6 @@ const SearchBar = ({
   };
 
   const handleSearchValueChange = (value) => {
-    console.log(value, value === "" && searchFieldValue !== "");
     if (value === "" && searchFieldValue !== "") {
       handleClearSearchResults();
     } else {
--- a/moped-editor/src/components/GridTable/SearchBar.js
+++ b/moped-editor/src/components/GridTable/SearchBar.js
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { useSearchParams } from "react-router-dom";
 
 import {
   Button,
@@ -16,6 +17,7 @@ import {
 import makeStyles from "@mui/styles/makeStyles";
 import { Search as SearchIcon } from "react-feather";
 import clsx from "clsx";
+import { simpleSearchParamName } from "src/views/projects/projectsListView/useProjectListViewQuery/useSearch";
 
 /**
  * The styling for the search bar components
@@ -106,6 +108,7 @@ const SearchBar = ({
   filtersConfig,
 }) => {
   const classes = useStyles();
+  let [, setSearchParams] = useSearchParams();
 
   /**
    * Attempts to retrieve the default placeholder for the search input field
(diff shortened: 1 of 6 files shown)


moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchResults.js
```diff
@@ -55,13 +55,13 @@ const NavigationSearchResults = ({
...
+        // Link to project list view with simple search populated
         <>
           <Divider />
           <ListItemLink
-            to={`/moped/projects/`}
-            // send searchTerm in location state
-            state={{ searchTerm: searchTerm }}
+            to={`/moped/projects?search=${searchTerm}`}
             onClick={handleDropdownClose}
+            reloadDocument
```
mddilley, line 64: This forces a browser redirect because client-side navigation doesn't re-render the project list view. That render is needed to initialize the search term from the url param. I think this is alright, but I'm open to other suggestions!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useSearch.js
```diff
@@ -28,10 +31,11 @@ export const getSearchValue = (column, value, queryConfig) => {
 };
 
 export const useSearch = ({ queryConfig }) => {
-  // create URLSearchParams from url
-  const navigationSearchTerm = useLocation()?.state?.searchTerm;
+  /* Get simple search from search params if it exists */
+  let [searchParams] = useSearchParams();
+  const simpleSearchValue = searchParams.get(simpleSearchParamName);
 
-  const [searchTerm, setSearchTerm] = useState(navigationSearchTerm ?? "");
```
mddilley, line 34: We used to pass the search term from the global search through [React Router link state](https://reactrouter.com/en/main/components/link#state). This PR adds a search parameter to the url that we can now consume here instead.

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchResults.js
johnclary: cool—i'm glad you thought of this feature! this works as i expect it to, and i think forcing a re-render makes sense.

i'm not sure what an alternative might be. maybe a a global search param context that would cause the consumers to re-render?

anyway, this looks good to me!

johnclary (approved):
Niiice! i love that the project list doesn't reset when you navigate to back to it. this really is such a major improvement to the project search experience. 

🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/components/GridTable/SearchBar.js
```diff
@@ -106,6 +112,7 @@ const SearchBar = ({
   filtersConfig,
 }) => {
   const classes = useStyles();
+  let [, setSearchParams] = useSearchParams();
```
roseeichelmann, line 115: im not familiar with this pattern of `let [, something] =`, what does that do?

roseeichelmann (changes requested):
Most of the test steps check out but when i do the global search i get an error for a brief second that says "data is undefined" i think its just bc its fetching slowly? maybe theres somewhere in the code that needs an `await` or something like that

[simplescreenrecorder-2024-01-16_18.22.26.webm](https://github.com/cityofaustin/atd-moped/assets/67130963/a9ec430d-995b-434f-96c6-fd14031eba6d)

mddilley:
@roseeichelmann Thanks for catching that! I tried replicating and couldn't, but, looking at the code and the error in your capture, I think there is an edge case where our existing checks in [these lines](https://github.com/cityofaustin/atd-moped/pull/1240/commits/cbfe38da30530e123566942cc23efcd2c5ce6162#diff-812ec89db1d4e53a21291824ccb6fa189624e77a86f2655054703b2b1aedc256R342-R344) fail. I added a fallback to try to patch it up. Could you retest when you get a chance? 🙏

moped-editor/src/components/GridTable/SearchBar.js
mddilley: just caught this comment. It is a way to skip an item when destructuring an array. [This post](https://www.samanthaming.com/tidbits/13-skip-values-in-destructuring/) explains it way better than I could! If I assigned the first item (which is a getter for `searchParams`), I would get a compile warning about an unused variable.

chiaberry (approved):
this is such a great improvement! 

I tried replicating Rose's bug and couldnt get it to happen either.

moped-editor/src/components/GridTable/SearchBar.js
frankhereford: TIL! I wouldn't have thought in a hundred years `,,,,` could be a valid bit of javascript, and there it is. Thank you so much for sharing that link, it's so helpful.

roseeichelmann (approved):
Yay im not getting that error anymore i just get a white screen for a brief second as it loads. All the test steps check out, this is great! :ship:

mddilley:
thanks y'all! merging 🚢

PR #1241, opened 2024-01-13
makes all case sensitive operators insensitive

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15189

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


tillyw:
we may need to make a separate issue for this but it seems that date-based queries (modified, etc.) are not working

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
johnclary, line 1: @tillyw would you merge in Mike's PR #1236? He removed the `"*"` operator from the numeric operators, so i think  need to manually add the new `number_does_not_equal` operators to them: https://github.com/cityofaustin/atd-moped/pull/1236/files#diff-d30e0b6ab350635d0f070d49b20e7d8a06190609484ce0423db44c34468011db

johnclary (changes requested):
@tillyw this works perfectly. i really appreciate that you cleaned up the naming conventions and added the new operator for number types. 

There is just a minor potential conflict with Mike's PR.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: @tillyw I just merged the PR that John linked to (https://github.com/cityofaustin/atd-moped/pull/1236) so `main` branch is caught up and can be pulled into this one.

chiaberry (commented):
Thanks for making everything consistent! Once you address John's feedback about merging main and adding the new operator to the number cases, I will approve this PR.

tillyw:
thanks, y'all! i updated the pr to be consistent with Mike's code! there is one caveat: the ``_neq`` operator does not return null values, so if a user searches for something like ``Interim (MPD) access ID is not 234`` they'll get back any records with other Interim access IDs but not records that don't have those values. We need to chain an ``_or`` ``_is_null`` statement to fix that, which I'm looking into now and how that would fit into ``GridTable``. If we're concerned about causing data inconsistencies I can just take out the ``number_does_not_equal`` for now since we didn't have it before. This shouldn't affect ``ID`` filters because all records should have IDs.

johnclary:
thanks @tillyw for flagging that edge case! I can live with that edge wrt to `null` values. Let's ship this!

tillyw:
tl;dr the ``_neq`` operator does not return null values so the ``number_does_not_equal`` operator returns incomplete data and i don't know that ``GridTable`` is able to parse chained operators (i might be mistaken). should we just scrap the new operator and look into this further for the next release?

tillyw:
> thanks @tillyw for flagging that edge case! I can live with that weirdness wrt to `null` values. Let's ship this!

gotcha, thanks @johnclary ! merging now

PR #1242, opened 2024-01-13
Adds more fields to component view

johnclary:
- https://github.com/cityofaustin/atd-data-tech/issues/15349

### Steps for testing

Using an editor and admin login locally:
1. Test that the project list loads. Toggle on the **Status update** and **Partners** fields. 
2. Add a status update to a project, then confirm the status update appears in the **Status update** column of the project list.
3. Add a component to a project that you know is within any council district. Use your SQL client to inspect that the `component_arcgis_online_view.is_within_city_limits` column is `TRUE`. 
4. ☝️ Edit the component by moving its features to be outside of any council district. Verify that `is_within_city_limits` changes to `FALSE`.
5. Create a line-based component (e.g., Bike Lane) that is roughly a mile long. Use Google Maps (or what have you) to measure the distance of the component you created. Use your SQL client to verify the distances matches the value of `length_miles_total` in the DB.


moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
johnclary, line 1: it would be nice to merge/remove redundancy between this config with the `ProjectListViewQueryConf`, the `useColumns` helper, and also the hidden column settings.

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -8,28 +8,36 @@ CREATE OR REPLACE VIEW component_arcgis_online_view AS (
...
         mc.component_name_full, 
-        subcomponents.subcomponents, 
-        work_types.work_types, 
+        'placeholder text' as component_categories,
+        subcomponents.subcomponents as component_subcomponents, 
+        work_types.work_types as component_work_types, 
         component_tags.component_tags, 
-        mpc.description AS component_description, 
-        mpc.is_deleted is_project_component_deleted, 
-        plv.is_deleted is_project_deleted,
```
chiaberry, line 23: these are removed because they are always false in the view, right?

johnclary: exactly. removed so as to (hopefully) avoid any confusion.

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -8,28 +8,36 @@ CREATE OR REPLACE VIEW component_arcgis_online_view AS (
         comp_geography.feature_ids, 
         mpc.component_id, 
         comp_geography.geometry, 
-        comp_geography.line_geometry, 
+        comp_geography.line_geometry,
         comp_geography.signal_ids, 
         council_districts.council_districts, 
+        'placeholder text' as council_districts_searchable,
```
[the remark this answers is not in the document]
johnclary: The main thing is that we're trying to do is cut down on the number of times Andrew has to update the AGOL layers with schema changes. 

So having a column called `council_districts_searchable` with every row value as the string `"placeholder text"` enables us to give Andrew an export of the updated data that has the correct schema, and we can have fake data flowing in the meantime until we actually implement this.

chiaberry (approved):
went through test steps ✅

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -198,8 +198,16 @@ export default function TheMap({
...
+    const source = SOURCES[clickedFeatureSource];
+    if (!source) {
+      // this can happen if the mapped is zoomed out such that AGOL/CTN feature layers are not visibile
+      // but the underyling draft edit component has captured a click
+      // we can't handle this click right now, so just return
+      // todo: manage layer interactivity based on zoom level + if the feature is drawn?
+      return;
+    }
+    const sourceFeatureId = source._featureIdProp;
+    const databaseTableId = source.databaseTableId;
```
johnclary, line 210: fixes a totally unrelated bug that Chia mentioned in [Slack](https://austininnovation.slack.com/archives/CNUEPKLB1/p1705959692637309)

PR #1250, opened 2024-01-23
Moped and Data Tracker projects sync

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8536

This PR adds an ETL for syncing Moped project data to the Data Tracker `projects` table. I updated the existing files to follow the patterns that John started up for the AGOL ETL and also reused some really nice existing helpers 🙏🙏🙏 I decided against a test flow that uses local Moped and local Airflow because I forgot that we have the same overlapping port 8080 issue that we ran into with the VZ local stack. I don't think that it is important enough to update at the moment.

## Testing
**URL to test:** 

**Steps to test:**
Choose your own adventure:
- [Test with local Airflow, the staging Moped editor, and a test Data Tracker app](https://github.com/cityofaustin/atd-airflow/pull/203). This tests updating synced projects. This is the most straightforward path.

 **or**

 - Follow [these steps in the readme](https://github.com/cityofaustin/atd-moped/blob/011aac09c3fce779db304cb55a0bd4f79ad9e8ea/moped-etl/data-tracker-sync/README.md) to replicate how I developed these changes with the local Moped stack and test Data Tracker app. This tests both syncing new projects and updating synced projects. This is a bit more complicated, and I'd appreciate any feedback on how understandable this is.

---
#### Ship list
- [x] Code reviewed 


moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+        update_moped_project_knack_id(moped_project_id, knack_record_id)
+
+    # Find all projects that are synced to Data Tracker to update them
+    synced_moped_projects = find_synced_moped_projects(
+        last_run_date=args.date, is_test=args.test
+    )
+
+    # Update synced Moped projects in Data Tracker and skip those just created
+    updated_knack_records = []
+    updates_to_skip = [record["moped_project_id"] for record in created_knack_records]
```
mddilley, line 156: We are checking for Moped project records that have been updated since the last successful Airflow DAG run. This script updates Moped project records itself since we need to add the Knack project record ID in a Moped project's `knack_project_id` column in order to mark it as synced. 

I still feel like this opens the door to updating Knack records with Moped data that hasn't actually been updated outside of setting the Knack project ID. I might need some rubber ducking here. 🤔

moped-etl/data-tracker-sync/process/queries.py
mddilley, line 1: Also wanted to note that I briefly tried to use [graphql fragments](https://graphql.org/learn/queries/#fragments) here but got stuck on template strings and `.format()` running into issues with the GraphQL curly braces. I think we would need to add a library to handle this better so I abandoned the idea.

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+        data = make_hasura_request(
+            query=GET_TEST_UNSYNCED_PROJECTS,
+            variables={"project_id": TEST_MOPED_PROJECT_ID},
+        )
+        unsynced_projects = data["moped_project"]
+    else:
+        data = make_hasura_request(query=GET_UNSYNCED_PROJECTS)
+        unsynced_projects = data["moped_project"]
+
+    logger.info(f"Found {len(unsynced_projects)} unsynced projects")
```
johnclary, line 42: this logging is super helpful when triaging Airflow errors 🙏

moped-etl/data-tracker-sync/data_tracker_sync.py
johnclary: i consider this an unrelated problem, but i've had mixed results with the activity logging. As you can see in this screenshot, the first activity event used a default behavior, while the second formatted the entry correctly:
Screenshot 2024-01-25 at 1 23 11 PM

i think i'm ok with continuing to have this event in the log, but we'll need ot fix that bugginess. i can open a separate issue for it.

johnclary (changes requested):
@mddilley this looks fantastic! i ran through the testing using the instructions in the README, and they worked great. for thoroughness, i also made a few changes to the script while testing, so that it processed multiple signal records. that all worked as expected 💯 

my feedback is very minor, and i'm just holding back approval until we settle on what (if anything) to do about the duplicate record processing of new projects.

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+)
+
+KNACK_DATA_TRACKER_APP_ID = os.getenv("KNACK_DATA_TRACKER_APP_ID")
+KNACK_DATA_TRACKER_API_KEY = os.getenv("KNACK_DATA_TRACKER_API_KEY")
+TEST_MOPED_PROJECT_ID = os.getenv("TEST_MOPED_PROJECT_ID")
+
+KNACK_DATA_TRACKER_PROJECT_OBJECT = "object_201"
+
+
+def find_unsynced_moped_projects(is_test=False):
```
chiaberry, line 28: I'd really appreciate a brief comment that explains what an unsynced moped project is, like "returns list of moped projects that do not have knack project ids associated with them", i know we can check the queries but....

moped-etl/data-tracker-sync/data_tracker_sync.py
chiaberry: @johnclary this is reminding me of the random activity log entry that showed up as "updated is deleted" but I couldnt reproduce it

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+    parser.add_argument("-t", "--test", action="store_true")
+
+    args = parser.parse_args()
+
+    log_level = logging.DEBUG if args.test else logging.INFO
+    logger = get_logger(name="moped-knack-sync", level=log_level)
+    logger.info(
+        f"Starting sync. Will update records updated in Moped since {args.date}"
```
chiaberry, line 195: I'm not sold on this phrasing, but I cant come up with anything better. 

Updates records in Data Tracker and Moped that have changed since last update? thats not much better....

moped-etl/data-tracker-sync/data_tracker_sync.py
chiaberry: I was stuck on the original question for a while, I didn't understand how there wasn't an infinite loop, but it finally clicked that the updated_at timestamp is only changed on the moped project here: `update_moped_project_knack_id` and not in the update_knack_project flow. 
I agree its not ideal, but seems better than having to update the trigger on every column renaming or addition. 
Will it be confusing in Data Tracker to see the records being updated twice, and a day apart?

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "-d",
+        "--date",
+        type=str,
+        default=datetime.now(timezone.utc).isoformat(),
+        help=f"ISO date string (in UTC) of latest updated_at value to find project records to update.",
```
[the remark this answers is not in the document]
mddilley: True, i'll remove since it could add some confusion! 🙏

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        created_knack_records.append(
+            {
+                "moped_project_id": moped_project_id,
+                "knack_record_id": knack_record_id,
+            }
+        )
+
+        # Update Moped project with Knack record ID of created record
+        update_moped_project_knack_id(moped_project_id, knack_record_id)
```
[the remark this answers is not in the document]
mddilley: Good call! 🧠 I'll log that right in here so we can inspect the DAG logs.

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+    knack_record_id = updated["id"]
+    return knack_record_id
+
+
+def main(args):
+    app = knackpy.App(
+        app_id=KNACK_DATA_TRACKER_APP_ID, api_key=KNACK_DATA_TRACKER_API_KEY
+    )
```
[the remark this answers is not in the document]
mddilley: This is a TIL for me! I didn't realize the difference and am happy to make the update. 🚀

moped-etl/data-tracker-sync/data_tracker_sync.py
mddilley: I like this a lot! I agree that it is hard to link the create/sync/update/unsync usage and that won't get easier with time. I'll add some comments. 🙏

moped-etl/data-tracker-sync/data_tracker_sync.py
mddilley: You and me both @chiaberry! I snuck this in last minute to log the arg coming out of Airflow. I'll update to: 
**Starting sync. Creating Knack records for Moped projects not synced and updating synced Knack records with latest project data from Moped.**
I think that is getting there. What do you think?

moped-etl/data-tracker-sync/data_tracker_sync.py
chiaberry: thanks!!

moped-etl/data-tracker-sync/data_tracker_sync.py
mddilley: awesome - thanks for the feedback! it inspired me to get proper docs in these function too. 🙏

moped-etl/data-tracker-sync/data_tracker_sync.py
johnclary: capturing the record sync date is an interesting idea, but i can't quite see how it all works together. if the sync date is always after the DAG run start date, and the `updated_at` date is always after both of those, i think we're still stuck without a way to know for sure if a record has changed since the last sync. but i'm not super confident about that 🤷

johnclary (approved):
looks good! i really like you solution for making this testable—that will really help us the next time we need to revisit this ETL. let's get this deployed!! 🚢 🚢 🚢

chiaberry (approved):
Approved!

moped-etl/data-tracker-sync/data_tracker_sync.py
mddilley: I agree - it also sounds like it would make it more convoluted. I'm down to ship and see how this works out. Maybe something more solid will float our way! 🤞

PR #1252, opened 2024-01-24
Add an activity event to track when when the Dashboard page loads

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/15392

## Testing
**URL to test:** Local
You will use your SQL client to inspect activity events as they are logged.

1. Start Moped locally. Log out, so that you're looking at the login screen.
2. Open your SQL client, and delete any records in the `moped_user_events` table:

```sql
DELETE FROM moped_user_events WHERE TRUE;
``` 
3. Login to Moped. There should now be one event in the `moped_user_events` table. The event name should be `app_load`.
4. Navigate to the **Dashboard** page. There should now be a second event in the `moped_user_events` table with the name `dashboard_load`.
5. While still on the **Dashboard** page, refresh the page in your browser. This will create two new events: `app_load` and `dashboard_load`.
6. Now navigate to the **Staff** page and verify that no new activity events have been created.
 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1706117943942_add_dashboard_load_event/down.sql
+++ b/moped-database/migrations/1706117943942_add_dashboard_load_event/down.sql
@@ -1,0 +1,4 @@
+ALTER TABLE moped_user_events DROP CONSTRAINT moped_user_events_event_name_check;
+
+ALTER TABLE moped_user_events
+    ADD CONSTRAINT moped_user_events_event_name_check CHECK (event_name in('app_load'));
--- a/moped-database/migrations/1706117943942_add_dashboard_load_event/down.sql
+++ b/moped-database/migrations/1706117943942_add_dashboard_load_event/down.sql
@@ -1,3 +1,5 @@
+DELETE FROM moped_user_events where event_name = 'dashboard_load';
+
 ALTER TABLE moped_user_events DROP CONSTRAINT moped_user_events_event_name_check;
 
 ALTER TABLE moped_user_events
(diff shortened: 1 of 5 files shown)


moped-database/migrations/1706117943942_add_dashboard_load_event/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE moped_user_events DROP CONSTRAINT moped_user_events_event_name_check;
+
+ALTER TABLE moped_user_events
+    ADD CONSTRAINT moped_user_events_event_name_check CHECK (event_name in('app_load', 'dashboard_load'));
```
mddilley, line 4: ✨

mddilley (changes requested):
Love seeing the tracker grow so easily to handle new events!! 🚀 🙌 The down migration doesn't work as is, but, imo, we move forward and never go back! 🏎️💨💨 Probably need a comment that there is no going back? Not sure how we want to handle that.

chiaberry (approved):
I'm looking forward to finding out who uses the dashboard page and how often

moped-database/migrations/1706117943942_add_dashboard_load_event/down.sql
```diff
@@ -0,0 +1,6 @@
+DELETE FROM moped_user_events where event_name = 'dashboard_load';
```
mddilley, line 1: Cool, that makes sense to me! 🙏

mddilley (approved):
Ready to go and to start collecting new events. 👀 🙌

johnclary:
🍿

PR #1256, opened 2024-01-25
Phase editing usability enhancements

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15398

## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectNotes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectNotes.js
@@ -90,6 +90,9 @@ const useStyles = makeStyles((theme) => ({
 // The zeroth item in the list is intentionally blank; the notes are 1-indexed.
 const projectNoteTypes = ["", "Internal Note", "Status Update"];
 
+export const STATUS_UPDATE_TYPE_ID = 2;
+export const INTERNAL_NOTE_TYPE_ID = 1;
+
 const ProjectNotes = (props) => {
   const isStatusEditModal = props.modal;
   // use currentPhaseId if passed down from ProjectSummaryStatusUpdate component,
@@ -101,7 +104,9 @@ const ProjectNotes = (props) => {
   const classes = useStyles();
   const userSessionData = getSessionDatabaseData();
   const [noteText, setNoteText] = useState("");
-  const [newNoteType, setNewNoteType] = useState(isStatusEditModal ? 2 : 1);
+  const [newNoteType, setNewNoteType] = useState(
+    isStatusEditModal ? STATUS_UPDATE_TYPE_ID : INTERNAL_NOTE_TYPE_ID
+  );
   const [editingNoteType, setEditingNoteType] = useState(null);
   const [noteAddLoading, setNoteAddLoading] = useState(false);
   const [noteAddSuccess, setNoteAddSuccess] = useState(false);
@@ -109,7 +114,7 @@ const ProjectNotes = (props) => {
   const [noteId, setNoteId] = useState(null);
   const [displayNotes, setDisplayNotes] = useState([]);
   const [filterNoteType, setFilterNoteType] = useState(
-    isStatusEditModal ? 2 : 0
+    isStatusEditModal ? STATUS_UPDATE_TYPE_ID : 0
   );
   const [isDeleteConfirmationOpen, setIsDeleteConfirmationOpen] =
     useState(false);
@@ -320,8 +325,12 @@ const ProjectNotes = (props) => {
               control={<span />}
             />
             <NoteTypeButton noteTypeId={0}>All</NoteTypeButton>
-            <NoteTypeButton noteTypeId={1}>Internal Notes</NoteTypeButton>
-            <NoteTypeButton noteTypeId={2}>Status Updates</NoteTypeButton>
+            <NoteTypeButton noteTypeId={INTERNAL_NOTE_TYPE_ID}>
+              Internal Notes
+            </NoteTypeButton>
+            <NoteTypeButton noteTypeId={STATUS_UPDATE_TYPE_ID}>
+              Status Updates
+            </NoteTypeButton>
           </Grid>
         )}
         {/*Now the notes*/}
--- a/moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
+++ b/moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
@@ -24,6 +24,10 @@ export const phaseValidationSchema = yup.object().shape({
     .nullable(),
   project_phase_id: yup.number().nullable().optional(),
   project_id: yup.number().required(),
+  status_update: yup
+    .string()
+    .max(500, "Must be less than 500 characters")
+    .nullable(),
 });
 
 const DEFAULT_FORM_VALUES = {
@@ -37,6 +41,7 @@ const DEFAULT_FORM_VALUES = {
   phase_description: null,
   is_current_phase: false,
   project_id: null,
+  status_update: null,
 };
 
 /**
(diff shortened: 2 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -177,3 +184,21 @@ export const onSubmitPhase = ({
...
+
+  mutate({
+    variables: {
+      objects: [
+        {
+          project_note: DOMPurify.sanitize(status_update),
+          project_id,
+          added_by_user_id: user_id,
+          project_note_type: STATUS_UPDATE_TYPE_ID,
+          phase_id,
```
mddilley, line 199: @johnclary iirc the phase added to a status update is the current phase. What are your thoughts on how this should behave? I think that I missed the mark on the business logic by passing the phase selected in form on submission.

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -118,31 +148,45 @@ const ProjectPhaseForm = ({
...
+    console.error("Phase mutation error", mutationState.error);
+    console.error("Status update mutation error", addStatusUpdateState.error);
     return (
       <Grid container spacing={2}>
-        <Grid item xs={12}>
-          <Alert severity="error">
-            Something went wrong. Refresh the page to try again.
-          </Alert>
-        </Grid>
+        {mutationState.error ? (
```
mddilley, line 168: Here is a screenshot of what this would look like if both mutations errored for some reason. I don't think we've handled errors coming from multiple submissions at once. Looking for feedback! 🙏
Screenshot 2024-01-26 at 4 13 23 PM

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
johnclary: ah, right. we do indeed want the status update tagged with the project's current phase, so the behavior here would mirror the behavior from status updates elsewhere in the app.

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
johnclary: Thanks for thinking through these error cases! some of the branching logic might get ugly, but I'm wondering what you think about combining the two mutations into a single request on submit—so as to handle this as an all-or-nothing operation?

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -90,6 +90,9 @@ const useStyles = makeStyles((theme) => ({
 // The zeroth item in the list is intentionally blank; the notes are 1-indexed.
 const projectNoteTypes = ["", "Internal Note", "Status Update"];
 
+export const STATUS_UPDATE_TYPE_ID = 2;
+export const INTERNAL_NOTE_TYPE_ID = 1;
```
johnclary, line 94: thank you for centralizing these values. i will open an issue to get them into a lookup table in the DB

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -258,13 +311,25 @@ const ProjectPhaseForm = ({
...
-            <ControlledSwitch
-              name="is_current_phase"
+        <Grid item xs={12} sx={{ marginY: theme.spacing(2) }}>
+          <Divider />
+        </Grid>
+        <Grid item xs={12}>
+          <FormControl fullWidth error={!!formErrors?.status_update}>
+            <ControlledTextInput
+              fullWidth
+              label="Status Update"
```
johnclary, line 321: Would you make the label sentence case? `Status update`

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -258,13 +311,25 @@ const ProjectPhaseForm = ({
...
+              multiline
+              rows={3}
+              name="status_update"
               control={control}
-              label="Current phase"
+              size="small"
             />
```
johnclary, line 327: you add helper text for this input?

> Optionally add a status update to the project

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -196,6 +240,15 @@ const ProjectPhaseForm = ({
...
           </FormControl>
         </Grid>
+        <Grid item container justifyContent="flex-start">
+          <FormControl>
+            <ControlledSwitch
+              name="is_current_phase"
+              control={control}
+              label="Current phase"
+            />
+          </FormControl>
```
johnclary, line 250: ok—i'm cool with leaving the placement here! would you add some helper text 
> Set this phase as the project's current phase

johnclary (changes requested):
I really like these improvements, Mike! I have a few minor things, and then some bigger questions that I don't have answers to. I'd be happy to discuss them. Thanks for working on this form!

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -118,31 +148,45 @@ const ProjectPhaseForm = ({
...
+   * Defaults phase_start to today if current phase is true and there is no phase_start
+   */
+  useEffect(() => {
+    if (is_current_phase === true && phase_start === null) {
+      const today = new Date();
+      today.setHours(0, 0, 0, 0);
+
+      setValue("phase_start", today);
+    }
+  }, [is_current_phase, phase_start, setValue]);
```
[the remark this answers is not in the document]
mddilley: Oh yep, thanks for catching that. I'll spend some time testing some solutions today. 🙏

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
```diff
@@ -60,14 +66,38 @@ const ProjectPhaseForm = ({
...
+    onSubmitPhase({
+      data: phaseData,
+      currentPhaseIdsToClear,
+      mutate,
+      onSubmitCallback,
+    });
+
+    if (status_update) {
+      const { project_id, phase_id } = phaseData;
+      const { user_id } = userSessionData;
```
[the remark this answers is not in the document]
mddilley: A spinoff would be great! I was getting confused about this not being set by a trigger and then forgot to actually check if this was captured in an issue anywhere.

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
mddilley: I think that combining into one mutation is the right call. I'll make that change and get it pushed up!

moped-editor/src/views/projects/projectView/ProjectPhaseForm.js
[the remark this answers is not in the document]
mddilley: I spent a little time testing an override of the `onChange` handler and then ended up finding that RHF lets us [pass custom `onChange` handlers when using `register`](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.16.0) returned by `useForm`. I don't think that this limits what we are trying to achieve with the controlled switch (but i could use another set of eyes). This lets us push the `useEffect` logic over into some control flow in the `onChange` handler. Most of that is captured in [this commit](https://github.com/cityofaustin/atd-moped/pull/1256/commits/e203e36b68be97f08c132deb49026e8e083b6157).

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
mddilley: Cool - I pushed an update to store the current phase of the project in the new note record unless the user is submitting a new current phase.

chiaberry (approved):
Ran through the test steps, also tried resetting the start date after i selected current phase, and your fixes to address the circular loop worked.

PR #1258, opened 2024-01-26
Ability to download project list with visible columns only

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15304

## Testing
**URL to test:** 
https://deploy-preview-1258--atd-moped-main.netlify.app/moped/ 

**Steps to test:**

1. Go to the projects list view, click on the Download button.
2. A dialog should appear, select that you only want to download **visible** columns. The options dialog should close and the downloading dialog should open.
3. Check the csv you downloaded and make sure it didn't download all the columns but just the ones that you have visible in the list view table.
4. Click the download button again and this time download **all** columns, check the csv that was downloaded and make sure all columns are there not just visible ones.
5. Test removing or adding columns to the list view table and exporting more csvs.
6. Test the cancel button/clicking away.

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -114,8 +121,10 @@ const ProjectsListViewTable = () => {
 
   const { query: projectListViewQuery, exportQuery } = useGetProjectListView({
     columnsToReturn: columnsToReturnInQuery,
-    exportColumnsToReturn: Object.keys(PROJECT_LIST_VIEW_EXPORT_CONFIG),
-    exportConfig: PROJECT_LIST_VIEW_EXPORT_CONFIG,
+    exportColumnsToReturn:
+      columnDownloadOption === "visible"
+        ? columnsToReturnInQuery
+        : Object.keys(PROJECT_LIST_VIEW_EXPORT_CONFIG),
```
roseeichelmann, line 127: If the user selected the "visible" option then only export the columns that are being returned in the table query, otherwise export all the columns in the export config

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -135,13 +144,21 @@ const ProjectsListViewTable = () => {
 
   const data = useCurrentData(projectListViewData);
 
-  const { handleExportButtonClick, dialogOpen } = useCsvExport({
+  const {
+    handleExportButtonClick,
+    handleRadioSelect,
+    handleContinueButtonClick,
+    handleOptionsDialogClose,
+  } = useCsvExport({
```
roseeichelmann, line 152: i found this pattern of returning the functions from `useCsvExport` here and then using them later in the props of the dialog components in this file to be kinda weird but i just followed that pattern from what was here already ?

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -135,13 +144,21 @@ const ProjectsListViewTable = () => {
...
     query: exportQuery,
     exportConfig: PROJECT_LIST_VIEW_EXPORT_CONFIG,
     queryTableName: PROJECT_LIST_VIEW_QUERY_CONFIG.table,
     fetchPolicy: PROJECT_LIST_VIEW_QUERY_CONFIG.options.useQuery.fetchPolicy,
-    limit: queryLimit,
-    setQueryLimit,
```
roseeichelmann, line 144: these aren't actually used anywhere in `useCsvExport`

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
frankhereford: I'm far from 100% on this, but FWIW, maybe because the arguments that need to go into that custom hook are only available in this component, and it's easier to drill down these handler callbacks than make so much of the state of `ProjectListViewTable` a prop to `CsvDownloadOptionsDialog. I'm curious too.

frankhereford (approved):
Rose, this worked great, and thank you very much for working on it. I've gone over your code, and it looks good to me, and the testing worked out.

The only thing that I have noticed is that the CSV, both for all columns and for selected ones, have a column called `Current Phase Key`, and the contents are an "underscored" version of the status. It looks like a token we're using in our JS that isn't intended to be displayed to the user. I think this should probably be skipped over, but it's not hurting anything either. Not enough imho to keep this from shipping.

image



Thanks again!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
```diff
@@ -29,6 +35,42 @@ export const CsvDownloadDialog = ({ dialogOpen }) => (
...
+export const CsvDownloadOptionsDialog = ({
+  dialogOpen,
+  handleDialogClose,
+  handleContinueButtonClick,
+  handleRadioSelect,
+}) => (
+  <Dialog open={dialogOpen} onClose={handleDialogClose}>
+    <DialogContent>
+      <FormControl>
+        <RadioGroup defaultValue={"visible"} onChange={handleRadioSelect}>
```
johnclary, line 47: you've opted to go with an "uncontrolled" component in this case. by that i mean that this component does not use the `columnDownloadOption` as a prop, it simply uses a handler function to set the value of `columnDownloadOption` on change.

that works totally fine for these purposes, but, in general, i almost always go with controlled form components, because i've found that over time i often end up re-visiting uncontrolled components when i want to add more complex / reactive behavior. 

for example, if we wanted add a "reset" button this to dialog that resets the radio button to the "default" value, we would not be able to do that without turning this into a controlled component.

if you want to put this into action, it is as easy as providing the `columnDownloadOption` as a prop to `CsvDownloadOptionsDialog`, and then passing that prop to `value` prop of the `RadioGroup` component.

once you do that, one effect that you'll notice is that radio selection  is persisted when you close the modal and open again.

that's my long-winded way of saying that i would prefer to see you make this a controlled component, but it is not critical. there are some [React docs](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components) on this subject that illustrate this much better than i have.

as one last aside, one of the reasons [React Hook Form](https://react-hook-form.com/) is so popular is because writing controlled components for multiple form inputs quickly becomes absurdly cumbersome 😵

johnclary (changes requested):
@roseeichelmann this is looking really good! +1 what Frank said, I am also noticing the pesky `current_phase_key` column in my exports.

one other issue—when i inspect the network activity in my browser, i am seeing that the data is actually being downloaded twice. here's how to reproduce:

1. Click the **Download** button to open the visibility dialog
2. Click the **Download all columns** options—notice that this triggers a network request for the project list data. 
3. Click **Continue** to initiate the download—notice that this triggers a network request for the project list data.

just taking a quick glance at this, i think the first network request (step 2) is firing as result of the input props in `useGetProjectListView`, which is causing `projectListViewQuery` to change and subsequently causing `useQuery` to re-fetch.

hopefully it's not too hard to untangle those requests 🤷

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
mddilley: Yes, this was 100% an attempt on my part to organize chunks of code that were being split out from a large component, and I agree that it feels weird and smells when looking at it now too. 👃

So, in `useCsvExport.js`, I see a component with no state, and a custom hook that contains all of the state and handlers needed for the UI in that component. I think that moving the contents of `useCsvExport` into `CsvDownloadOptionsDialog` would really cut down on the props passed around. I can expand if needed, and thanks for picking up on this!

mddilley (commented):
The UI flow looks great, and thanks for catching that the custom csv hook feels off. I left some suggestions about that, and I agree with John about updating the radio group to be controlled. I think that could be rolled into the rearrangement that I suggested. 🤞 Thanks for keeping this part of the code on the road to being better! 🙏

mddilley:
@roseeichelmann Forgot to note that I also think the extra network requests are coming from the way dependencies are set inside `useGetProjectListView`. The list query and the export query used to be pretty similar so it made sense to calculate them together at the time. But, now it is time for a shiny new feature. ✨

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
```diff
@@ -39,7 +39,18 @@ export const useGetProjectListView = ({
...
+    queryLimit,
+    queryOffset,
+    columnsToReturn,
+    orderByColumn,
+    orderByDirection,
+    searchWhereString,
+    advancedSearchWhereString,
+  ]);
+  const { exportQuery } = useMemo(() => {
+    const exportQueryString = gql`query ProjectListExport {
```
roseeichelmann, line 53: separating this single useMemo into two useMemos so we dont have the `ProjectListView` query recalculating when the `ProjectListExport` inputs change.

roseeichelmann:
@johnclary okay i think i addressed the bulk of your feedback, i changed the radio group to a controlled component, fixed the `current_phase_key` being in the export, and memoized the queries for the project list view and the project list export separately so the radio button selects arent causing the list view `useQuery` to run again.

there are still some unnecessary network requests firing when the radio buttons are selected but now just for the export query. @mddilley and i paired on this for awhile and couldn't figure out exactly why :sweat:

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
johnclary: nice—that's definitely what this hook needs. since the queries are identical except for the limit and offset, would you be up for taking this a step further and making this hook just return a single query, that includes the limit and offset if they are provided?

then in the ProjectListViewTable, you can call the `useGetProjectListView()` hook twice, once for the list query and once for the export query. 

i also don't mind moving that task into a separate issue—i'd like to get that hook DRY'd out.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
mddilley: i wanted to update that Rose helped me see that what I outlined in my last comment wasn't as simple as I was thinking. As this code is growing, I think it could be good to pull the lazy query out of this hook along with the handlers and try to organize the state/components/handlers differently - but, I don't see that as a blocker to shipping this feature at all since this was an existing issue before this PR. 🚀

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
```diff
@@ -3,51 +3,26 @@ import { gql } from "apollo-boost";
...
-          }
-        }
-      }`;
-    const query = gql`
-      ${queryString}
-    `;
-
-    const exportQueryString = gql`{
+  const { query } = useMemo(() => {
+    const queryString = `query ${queryName} {
```
roseeichelmann, line 15: `queryName` is rly helpful for debugging which query is sending network requests, the export or the list view

johnclary:
oops—we overlooked that there were going to be non-trivial conflicts with https://github.com/cityofaustin/atd-moped/pull/1263, so this is going to fall into the next release or a patch release 🙃

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useProjectListViewQuery.js
```diff
@@ -3,41 +3,20 @@ import { gql } from "apollo-boost";
...
-                advancedSearchWhereString && !searchWhereString
-                  ? advancedSearchWhereString
-                  : ""
-              }
-            }
-        ) {
-            ${columnsToReturn.join("\n")}
-        },
-        project_list_view_aggregate (
+  const { query } = useMemo(() => {
```
mddilley, line 14: This works, but the destructuring could go along with the returned object from `useMemo` now that it is only returning one variable.

mddilley (commented):
This is looking great - love the reuse of the hook! 🙌 I am testing this, and I wanted to raise a flag about the export query firing off on every filter change after exporting once.

To reproduce:
1. Go to the list view
2. Add/remove some advanced filters -> only the `ProjectListView` query is requested
3. Download either the full or partial export
4. Add/remove advanced filters and see that both the `ProjectListView` and `ProjectListExport` are being fetched

This seems related to the export request firing off when the radios are toggled (after the first time). 😞 I have a theory that this is related to how we are using `useLazyQuery` that I can share at dev sync.

roseeichelmann:
I am re-requesting reviews after updating the code to use `useApolloClient.query()` instead of `useLazyQuery` after reading some [threads](https://github.com/apollographql/apollo-client/issues/5912#issuecomment-797060422) where other ppl were having the same issue with `useLazyQuery` firing on every re-render (which is what was causing the problem @mddilley noted above). Now if you inspect the network requests the `ProjectListExport` query is only called after clicking "Continue" after selecting the column options for the export, not after changing filters or searching, and if you download with the same columns again the cached data is used so it also doesn't run again then.

mddilley:
@roseeichelmann i just found one more thing that could use a look. When we create `columnsToReturnInQuery`, there are some columns added to that array [that are required to properly render other columns](https://github.com/cityofaustin/atd-moped/blob/141e12abd8ccaa7266c9f56f882b37d269976452/moped-editor/src/views/projects/projectsListView/helpers.js#L131-L141). It looks like `parent_project_name` was added to that array between when this PR started an now so it is sneaking its way into the csv export.

It looks like your code removes `current_phase_key`, but I'm wondering if there is a way to tie these two things together so we don't have to update two places. What do you think?

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
```diff
@@ -93,11 +90,20 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
...
   components: {
-    label: "Components"
-  }
+    label: "Components",
+  },
+  interim_project_id: {
+    label: "Interim MPD (Access) ID",
+  },
+  project_and_child_project_council_districts: {
+    label: "Council districts",
```
roseeichelmann, line 107: thru my latest update i noticed these were missing bc i actually toggled all the columns in testing... which also lead me to the idea that it would be helpful to have a select all button here?
image

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -134,13 +134,22 @@ export const useColumns = ({ hiddenColumns }) => {
...
       "project_id",
       "current_phase_key",
       "parent_project_name",
     ];
+    // If parent project id is selected we also need to include the parent project name in the export
+    const columnsNeededForExport = columnsShownInUI.includes(
+      "parent_project_id"
+    )
+      ? ["project_id", "parent_project_name"]
+      : ["project_id"];
```
roseeichelmann, line 147: this feels kinda messy to me but needed a way to show the `parent_project_name` in the export if the `parent_project_id` column is turned on since when that column is turned out whats visible in the UI is the name of the parent project + link to the project id

roseeichelmann:
@mddilley thanks for the continued thorough reviews and for catching that! i just pushed up some updates so that now there are two separate arrays `allColumnsToReturn` and `visibleColumnsToReturn` returned by `useColumns`, then the appropriate one is passed to the export query depending on what radio button the user selects. this makes sure those extra columns needed to render aren't added when the user wants to export visible columns only

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -87,18 +94,32 @@ const ProjectsListViewTable = () => {
     storageKey: "mopedProjectListColumnConfig",
   });
 
-  const { columns, columnsToReturnInQuery } = useColumns({ hiddenColumns });
+  const { columns, columnsToReturnInListView, visibleColumnsToReturnInExport } =
```
mddilley, line 97: nice - glad to see this worked out, and i think the var name makes it clear to me what is happening here!

moped-editor/src/views/projects/projectsListView/helpers.js
mddilley: I think this is great and a big improvement to have these special cases, where we include columns to make related columns work correctly, near each other in the code. Maybe we will have to extend this in the future, but your comment captures what is going on here and will make that easier imo. 🙏

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
```diff
@@ -51,13 +94,25 @@ const downloadFile = (fileContents) => {
...
+  columnDownloadOption,
+  visibleColumns
+) => {
   // Allocate an empty object
   const entry = {};
+  const columnsToExport =
+    // Based on user selection set columns to download as all or just currently visible
+    columnDownloadOption === "visible"
+      ? visibleColumns
+      : Object.keys(exportConfig);
```
mddilley, line 113: **For a future spinoff and out of scope**: This caught my eye since it seems to be the second time that `visibleColumns` is used to determine whether to include all or just the visible columns even though the project records coming into this function already have all columns or just the visible columns. 

I think that we could get away with iterating the fields in the record to get the needed `exportConfig` instead of iterating the columns in `columnsToExport`. I might be missing a reason that wouldn't work, though. 🤔

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js
```diff
@@ -87,49 +154,64 @@ export const useCsvExport = ({
...
-    setDialogOpen(false);
+  const handleContinueButtonClick = () => {
+    setDownloadOptionsDialogOpen(false);
+    setDownloadingDialogOpen(true);
+    // Fetch data and format, parse, and download CSV when returned
+    client
+      .query({
+        query: query,
+        fetchPolicy: "network-first",
+      })
```
mddilley, line 197: This works so well! I see the query only happen once (and only once!) click and not when toggling the radios. 🥳

mddilley (approved):
Thanks for iterating on this one and working through some really tricky gotchas! I tested locally with prod data, and I can't find any regressions and the partial and full exports work just as expected. 🙌 🚢 🚀

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
mddilley: forgot to mention that I defer to Chia and Amenity on the column additions. I think that i remember @chiaberry  running into a bug with the adding a "select all" button but I'm not 100% on that.

chiaberry: I did run into a bug with the select all button, it didnt actually return all of the columns but instead returned an empty object. maybe the pro version works better?

chiaberry (approved):
thanks for sticking with it, and rolling with all the feedback

johnclary (approved):
thanks, Rose!

PR #1261, opened 2024-01-30
adds phase start and end confirmation to activity log

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15397

this pr adds phase start and phase end confirmation updates to the activity log. it also fixes a small bug that was responsible for displaying extra commas in the activity log for phase updates.

## Testing
**URL to test:** 

https://deploy-preview-1261--atd-moped-main.netlify.app/

**Steps to test:**
- go to a project's timeline, select a phase and invert the checkbox booleans for the phase start and phase end 'confirmed' values
- you should see these changes cleanly reflected in the activity log ✨

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
@@ -53,16 +53,11 @@ export const formatPhasesActivity = (change, phaseList, subphaseList) => {
   // loop through fields to check for differences, push label onto changes Array
   Object.keys(newRecord).forEach((field) => {
     if (newRecord[field] !== oldRecord[field]) {
-      // phase end and start confirmed are in the moped_proj_phases table but not in the UI
-      // ignore changes to that field until its visually represented to users
-      if (
-        field !== "is_phase_end_confirmed" &&
-        field !== "is_phase_start_confirmed"
-      ) {
-        changes.push(entryMap.fields[field]?.label);
-      }
+      changes.push(entryMap.fields[field]?.label);
     }
   });
+
+  console.log(changes);
 
   return {
     changeIcon,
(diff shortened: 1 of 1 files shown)


roseeichelmann (approved):
looks good :sparkles:

moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
```diff
@@ -53,12 +53,9 @@ export const formatPhasesActivity = (change, phaseList, subphaseList) => {
...
     if (newRecord[field] !== oldRecord[field]) {
-      // phase end and start confirmed are in the moped_proj_phases table but not in the UI
-      // ignore changes to that field until its visually represented to users
-      if (
-        field !== "is_phase_end_confirmed" &&
-        field !== "is_phase_start_confirmed"
-      ) {
+      // filter out fields that are not listed in the activity log table maps to prevent
+      // automated field updates (created at, updated at, etc.) from entering the array
+      if (!!entryMap.fields[field]) {
```
chiaberry, line 58: I like this better than the other way we were doing it, where were declaring which ones to skip
`  const fieldsToSkip = ["updated_at", "updated_by_user_id"];`

butI like your way, because we dont have to declare which ones to skip, and it just skips anything that isnt defined. ✨

chiaberry (approved):
we should expand this pattern to all our activity log formatters!

johnclary (approved):
Thanks, @tillyw!

I really like having a built-in handler for when we don't know the field. I also agree with Chia that we should do this everywhere. 🙏 🚢

tillyw:
yay i'm glad y'all like my approach!! :)) i thought this would be a good/simple way to handle fields we haven't defined and i'm glad it will form a pattern for the rest of the components ^_^

PR #1270, opened 2024-02-05
Fixes empty strings in component attributes

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/15635

## Testing
**URL to test:** Local with prod data

**Steps to test:**
1. Start Moped with prod data, but do not apply migrations.
2. Start by getting a point-in time count of `moped_proj_components` with empty strings:
```sql
SELECT
    location_description,
    description,
    srts_id
FROM
    moped_proj_components
WHERE
    location_description = ''
    OR description = ''
    OR srts_id = '';
```
3. Apply migrations and repeat to confirm that no components have empty strings.
4. In the Moped editor, create and edit a component. First, fill in the SRTS ID, but not the location description or description columns. Save, and then edit the SRTS ID field to be null.
5. Double-check the query from step 2—there should still be no results.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
```diff
@@ -134,14 +134,16 @@ const EditAttributesModal = ({
         subphase: makeSubphaseFormFieldValue(clickedComponent.moped_subphase),
         completionDate: clickedComponent.completion_date,
         srtsId:
-          clickedComponent.srts_id?.length > 0 ? clickedComponent.srts_id : "",
+          clickedComponent.srts_id?.length > 0
```
mddilley, line 137: I was able to save an empty string into the `description` column on update I'm pretty sure that the the initial value for `description` a few lines up needs this same code change. I tested it out and ran into a `!` on the `description` field in the `UPDATE_COMPONENT_ATTRIBUTES` mutation so that would need to be removed too to allow `null` to be passed. 🤦‍♂️ to me since I wrote that code. Thank you for looking at this. 🙏

mddilley (changes requested):
One tiny change needed and this is ready for ship! Huge thanks for catching these data quality issues and patching this. 🙏 ❤️

moped-editor/src/views/projects/projectView/ProjectComponents/EditAttributesModal.js
chiaberry: i cant get an empty string to save in the description column, but I was able to get `"     "` saved in the location_description column. 

Also ran into the issue where the description is required. I got thrown off by the mutation being named UpdateSignalComponent too. 
```
export const UPDATE_COMPONENT_ATTRIBUTES = gql`
  mutation UpdateSignalComponent(
    $projectComponentId: Int!
    $componentId: Int!
    $description: String!
```

johnclary: thank you both! i did these things:
- made `description` correctly fallback to `null` instead of `""`
- made `description` nullable in the update mutation
- renamed the mutation from `UpdateSignalComponent` to `UpdateProjectComponent`

mddilley:
@johnclary Ready to ✅ but i think some other code snuck into your last commit. Thanks for pushing the updates! 🙏 Makes me wish the form data munging here was more like what you did for the project phases form. ✨

johnclary:
ugh—my secret knack work order table side project! 🤦 i have removed it.

mddilley (approved):
Looks and works great! Thanks again for cleaning up the form output and the past data too! 🚢

moped-editor/src/queries/components.js
```diff
@@ -204,10 +204,10 @@ export const GET_PROJECT_COMPONENTS = gql`
 // are switched to is_deleted = false by the mutation). It also handles deleting/creating of signal
 // features and intersection/point features as needed when component type is changed
 export const UPDATE_COMPONENT_ATTRIBUTES = gql`
-  mutation UpdateSignalComponent(
+  mutation UpdateProjectComponent(
```
chiaberry, line 207: gracias

chiaberry (approved):

frankhereford (approved):
This looks and worked great John, thank you!

johnclary:
thank you!

PR #1271, opened 2024-02-05
Data Tracker sync continued

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15550

This adds a helper script to evaluate the sync between Moped projects and Data Tracker `projects` table and delete Knack records that have no connection to a Moped project and no connection to other Knack records.

Also, this is a tool for the toolbox. It is not going to be part of the ETL.

## Testing
**URL to test:** 

n/a

**Steps to test:**
Hard and not really useful to test this on staging so I'm going to say the best way is to check out the code. It is unlikely that we will have to run this ever again.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-etl/data-tracker-sync/data_tracker_sync.py
+++ b/moped-etl/data-tracker-sync/data_tracker_sync.py
@@ -48,7 +48,9 @@ def find_unsynced_moped_projects(is_test=False):
         data = make_hasura_request(query=GET_UNSYNCED_PROJECTS)
         unsynced_projects = data["moped_project"]
 
-    logger.info(f"Found {len(unsynced_projects)} unsynced projects")
+    logger.info(
+        f"Found {len(unsynced_projects)} unsynced projects to create in Data Tracker"
+    )
     logger.debug(f"Found unsynced projects: {unsynced_projects}")
     return unsynced_projects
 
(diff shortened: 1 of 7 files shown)


chiaberry (approved):
reviewed the code, thanks for adding the code to the toolbox!

moped-etl/data-tracker-sync/data_tracker_sync.py
```diff
@@ -194,7 +194,7 @@ def main(args):
 
         # Update Moped project with Knack record ID of created record
         logger.info(
-            f"Updating Knack ID {knack_record_id} from Moped project {moped_project_id}"
+            f"Updating Moped project {moped_project_id} with Knack ID {knack_record_id}"
```
mddilley, line 197: This comment had the logic backwards. We are updating the Moped record with the ID from the newly created Knack record.

mddilley:
@chiaberry thanks for the review! i just added some better logging in the last commit that I think make the script logging easier to follow. I found myself having to look at the script code to see what step the ETL was at when running manually from the Airflow side. Hopefully, this makes that better. 🤞

johnclary (approved):
looks great to me @mddilley! thank you! 🚢

mddilley:
thanks y'all, merging so we can get the improved logging running in the ETL. 🙏 🚀

PR #1272, opened 2024-02-07
Replace `moped_proj_milestones` REST connector events with DB triggers

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15488

## Testing
**URL to test:** 

local because it includes migrations

**Steps to test:**
this shouldn't change any ui functionality!
- go to a project's timeline and add a new milestone
- confirm that the update is registered in the activity log
- edit the milestone and check the activity log
- go to the new milestone record in the db and confirm the ``created_at``, ``updated_at``, ``created_by_user_id`` and ``updated_by_user_id`` columns are correct. there should no longer be a ``date_added`` columns
- delete the milestone and confirm in the activity log
- confirm in the db that the record was set to ``is_deleted`` is ``true
- go to the dashboard, select a followed project and click the milestone progress circle to add a new milestone, edit and delete it

note: if you are using seed data you may come across a bug where the first milestone you add isn't saved. this only seems to occur when using seed data so should only affect us devs, and it is a preexisting bug that we may want to look into!

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectActivityLogTableMaps.js
```diff
@@ -151,9 +151,6 @@ export const ProjectActivityLogTableMaps = {
       milestone_name: {
         label: "name",
       },
```
tillyw, line 153: it should be okay to delete this entry because the formatter will ignore any undefined fields but lmk if we need to keep it!

moped-database/seeds/1602292389297_initial_seed_staging.sql
```diff
@@ -76,24 +76,24 @@ INSERT INTO public.moped_proj_funding (proj_funding_id, project_id, created_at,
 -- Data for Name: moped_proj_milestones; Type: TABLE DATA; Schema: public; Owner: moped
 --
```
tillyw, line 78: updated seed data for the renamed field, i thought about adding seed data for the new used ``added_by_user_id`` and ``updated_by_user_id`` fields but i wasn't sure if that was necessary

frankhereford (approved):
# ✅ 
I have some observations which are outside the scope of this issue below. This is 🚀 🚀 🚀 . Thank you Tilly!

### 🤔 Activity log `refetch()`?
When I created a milestone and clicked over to the activity log, it was there, ✅ . When I went back and edited the milestone and then returned to the activity log, the edit event wasn't represented on the page until I refreshed it. Do you think we should trigger a refetch() on that activity log query on an edit event for milestones, and eventually for all our editing? For sure though, this isn't a regression, I just tested this same behavior when editing the description.

moped-editor/src/utils/activityLogFormatters/mopedMilestonesActivity.js
```diff
@@ -45,7 +45,11 @@ export const formatMilestonesActivity = (change, milestoneList) => {
   // loop through fields to check for differences, push label onto changes Array
   Object.keys(newRecord).forEach((field) => {
     if (newRecord[field] !== oldRecord[field]) {
-      changes.push(entryMap.fields[field]?.label);
+      // filter out fields that are not listed in the activity log table maps to prevent
+      // automated field updates (created at, updated at, etc.) from entering the array
+      if (!!entryMap.fields[field]) {
+        changes.push(entryMap.fields[field]?.label);
+      }
```
johnclary, line 52: thank you!

johnclary (approved):
🎯 nailed it! thank you for getting this set up!!!

wrt to @frankhereford's comment
> Do you think we should trigger a refetch() on that activity log query on an edit event for milestones, and eventually for all our editing?

Yes—thanks. I will open a backlog issue. In my testing just now, the activity events were properly refetching for milestones. But I know I've seen the same issue you're describing. in the past. It would be great to figure out what's going on and get the activity log to consistently show the latest changes.

johnclary:
ohh @frankhereford i think we are actually be experiencing the race condition that occurs when the activity log is refetched vs the amount of time it takes for the Hasura event trigger to fire off the insert mutation that actually creates the new activity log events. i do think that in some ideal future state we generate the activity log inserts via DB mutation instead of hasura event trigger 😞

mddilley (approved):
Walked through all of the test steps locally, and this works so great! 🚢 🙌

johnclary:
Thank you Tilly—i'm merging this 🚢 🚢 🚢

PR #1273, opened 2024-02-07
Adopt `sqlfluff` for linting / formatting SQL

frankhereford:
## Associated issues

_none_, however, this thread stands in its place: https://austininnovation.slack.com/archives/CNUEPKLB1/p1707325797465899.

## A comment

Sorry for this very long PR message. It's both detailing the issue, the scope of my work to address the issue, and testing, all in one. 🙏 

## Problem statement

Before I work on a view, I pull the current state of the view out of the database with a query like:

```sql
SELECT definition
FROM pg_views
WHERE schemaname = 'public' AND viewname = 'example_view';
```

I do this because I want to be assured that I am working from the latest version of the view. I recognize that we track the latest source in the repo, but I have ingrained habits to want to work from the database's understanding of the view, as that is what my changes will be applied to.

However, doing this, I'm essentially put in a place where i have to blast away the contents of one of these files and replace it with what the DB says is the current state of the view, and this destroys any hope of getting a concise diff of the changes. 

## Solution

This PR adopts the `sqlfluff` SQL linter. It's able to act both as a formatter of SQL, tidying it up to be compliant to the standard set forward in the `.sqlfluff` configuration. It's also acts as a linter (through a VSCode extension or a command line invocation), so you get red-squiggly lines when you're doing something that isn't a best practice. 

Note, the view files have been formatted in this PR, but I have made no material change to them beyond what can be fixed automatically, so if you open them up in VSCode with the extension enabled to lint so you get the "problems" (red squiggly lines) in VSCode, you're going to see we have a /lot/ of places we can tighten up our queries. If you are stressed by "red" in the code mini-map in VSCode, take a deep breath before you look in here. There is one exception to this "no meaningful changes" statement, and I will address it in the testing section. 

## Installation

You'll need to get `sqlfluff` installed on your computer and in the path. If you're using `brew`, you can `brew install sqlfluff`. Make sure it's in your path by doing `which sqlfluff` in your terminal, and if it gives you the path to program, then you're good to go.

VSCode users: The extension I'm using to provide linting and formatting capabilities in VSCode is [here](https://marketplace.visualstudio.com/items?itemName=dorzey.vscode-sqlfluff).

image

Users of other editors can adapt the following commands to best suit their needs. These are being run from the `moped-editor/moped-database/views` folder. 

Lint (ie, tell me what i'm doing wrong):

`sqlfluff lint --config ../../.sqlfluff project_list_view.sql`

Format (ie, clean up all my indention and fix things you can fix automatically):

`sqlfluff format --config ../../.sqlfluff project_list_view.sql`

## Discussion

We need to agree on exactly what set of linting and formatting rules we want to act under. I have proposed what I think is best in the `.sqlfluff` configuration file. I'm eager to discuss any changes y'all want to make to that. The `sqlfluff` tool has [superb documentation](https://docs.sqlfluff.com/en/stable/gettingstarted.html), and you can jump right to the rules [here](https://docs.sqlfluff.com/en/stable/rules.html#rule-references.qualification).

Here's what I care about:

* Capitalizing reserved words. `SELECT 'sandwich' AS lunch;` and not `select 'sandwich' as lunch;`
* Consistent indentation. I don't care what the indention rules are, just that it's consistent. 

### ⚠️ Important
I have had to fix a problem in our `component_arcgis_online_view` file which I detailed in slack in the first few sentences [here](https://austininnovation.slack.com/archives/CNUEPKLB1/p1707329741548609). I am unable to offer meaningful test instructions for this PR without fixing that. If this PR is rejected, then I will create a stand-alone PR to fix this file.

@johnclary, because I have altered this file, and despite having been careful in reconstructing it, I am going ask if you'll give it a close look to make sure that the `component_arcgis_online_view` and the `project_list_view` views are aligned to what you expect them to be with your recent changes. These two views were the ones mentioned in the file I fixed up. Let me know if I can help with this.

#### Practical test:

* Spin up your local database and apply a production snapshot to it
* Apply any migrations that are in `main` and not in production.
* Apply metadata because why not
* Drop all the views
```sql
drop view component_arcgis_online_view;
drop view project_geography;
drop view project_list_view;
drop view uniform_features;
drop view current_phase_view;
```
* Add the views back by copying them from the files in this PR. Apply them in the opposite order they were dropped.
* Test the app to make sure our view-driven features are working

#### Esoteric test:

* Look at [this file](https://github.com/cityofaustin/atd-moped/blob/71438056ca71a1e7563d29052d238bb975972fba/moped-database/views/uniform_features.sql), as an example.
  * Do you like how it is indented?
  * What is easiest to read for you?
  * Does this format help you see how the query works and builds up data to return?
  * Would you prefer other indention or formatting rules?
    * Are they available through [configuration](https://docs.sqlfluff.com/en/stable/configuration.html) of sqlfluff?

---
#### Ship list
- [ ] Code reviewed 👉 please DO review `.sqlfluff` and feel free to review the SQL, but it's an uphill battle at best.


moped-database/views/component_arcgis_online_view.sql
frankhereford, line 1: This file got more than formatting. This PR fixes a copy/paste typo as well.

johnclary (approved):
thanks @frankhereford for getting us onboard with a SQL linter!!! this library works beautifully. this is a very nice level-up for us! 🚢 🚢 🚢 🚢 🚢

chiaberry (approved):
yay! thank you for setting us up with this

.sqlfluff
mddilley, line 1: This looks great to me and works with my VSCode setup. Capitalization and indentation 🙌

moped-database/views/current_phase_view.sql
mddilley, line 1: Using this file as an example, I noticed we are generating some linter warnings like `Unqualified reference 'project_id' found in select with more than one referenced table/view.sqlfluff(RF02)`. We can discuss later, but I'm curious if we would want to address these. I definitely don't want to ignore them through settings because they are helping me understand the DB views better and how we could be more explicit. 💯

mddilley (approved):
I followed the test steps and autoformatting is working on my machine. ✨🚀 This really makes the views much more readable, and I'm glad we are gaining consistency here. Thanks for getting this set up! 🍄 🚢

moped-database/views/current_phase_view.sql
frankhereford: > Using this file as an example, I noticed we are generating some linter warnings like `Unqualified reference 'project_id' found in select with more than one referenced table/view.sqlfluff(RF02)`. We can discuss later, but I'm curious if we would want to address these later on. I definitely don't want to ignore them through settings because they are helping me understand the DB views better and how we could be more explicit. 💯

Thanks Mike - I've seen that warning in many places as well. I think to address this specific one, we would need to be explicit about exactly which table a column is being drawn from. For example, where we might use `select project_id` where the database is able to infer which table is providing that column, the linter is suggesting that we say `select moped_project.project_id`. Still in this example, while the database is able to infer the source of the column in the queries current form, the linter wants us to be explicit. The rational, I imagine, for this linter warning is to prevent us from enhancing the query down the road and enhancing it into a state where that inference isn't possible anymore. Another possible reason would be just for simple readability -- the reader doesn't have to run the query planner in their head to figure out where that column is coming from -- rather, the source table of the column is right there in the `SELECT` clause.

To your thoughts about if we would want to address these later on, I would support doing so. I think a big part of adding a linter is to come into compliance with it, and letting it guide us into using best practices. Speaking just for myself, my happy place is a VSCode that doesn't have any linting problems listed in the problems section.

frankhereford: Mike, I was moving on to the next thing, but then it dawned on me that I may have failed to address a question you were raising. I do think we should get compliant with the linter, but I have not done so in this PR to try to keep from scope-creeping it out. I totally do think we should create a spin-off issue to do this refinement of our queries, and if y'all want me to do that as part of this PR, I'd be more than happy to.

mddilley: You are all good - thanks for responding! I agree that it is definitely a spin-off task. All of this is 100% and ready to 🚢

PR #1274, opened 2024-02-08
Add council districts to project list

frankhereford:
## Associated issues

This PR aims to close https://github.com/cityofaustin/atd-data-tech/issues/10892.

## TIL 

`unnest()` & `array_agg(DISTINCT)` are a powerful combination to roll up a distinct set of values.

### Local testing instructions

* Spin up the local database, optionally pulling down a copy of production data for better testing.
* Apply migrations/metadata
* Query the `project_list_view` like so, and spot check council district data until you're confident this is working as intended. 
```sql
# edit, i forgot this when i first opened up this PR, sorry!

select project_id, project_description, self_alone_council_districts, self_and_children_council_districts
from project_list_view 
where project_id = ?
```
  * There are only a handful of projects which have more council district coverage through their children than they do as stand-alone entities. Look for `project_id` `86` as a notable example where the project is just a point and which has children that have far flung points in other districts. There are also projects which have no geometry themselves, but have children with geometry. These will be depicted with an array with a single null element for the `self` list and the appropriate council districts in the `self + children` list. 
* Do you like that the new column is exposed via the graphql endpoint?
* Spin up the editor, and observe that you can add and remove this column from the project list
* Do you like where I put it in the list, at the end?  How about the naming and the render() function?
* You should be able to wipe out your local storage key/value `mopedColumnConfig` and observe that it's off by default.

---
#### Ship list
- [ ] Code reviewed 


moped-database/views/project_list_view.sql
frankhereford, line 1: A concise diff ✨

moped-database/metadata/tables.yaml
frankhereford, line 1: Almost all the changes in this file are `graphql-engine` moving around the items in the list. It looks like it's moving them into alphabetical order though, so that's great. I guess we're just slowing working through the entries in this file as we update them to get to where we get meaningful diffs.

The things to look out for here are the addition of two new fields:

* `self_alone_council_districts`
* `self_and_children_council_districts`

moped-database/migrations/1707416196011_add_council_districts_to_project_view/down.sql
```diff
@@ -0,0 +1,519 @@
+-- no "IF EXISTS" because if it doesn't exist, we have bigger problems and we want this to bail
+DROP VIEW public.project_list_view CASCADE;
```
frankhereford, line 2: Here, and in the up migration, you'll see that I need to drop the view `CASCADE`, and it cascades onto the AGOL view that depends on the project list view. This is why the migrations have to reestablish both views.

moped-database/views/project_list_view.sql
```diff
@@ -82,6 +81,45 @@ moped_proj_components_subtypes AS (
...
+        FROM project_geography
+    ),
+
+    parent_child_project_map AS (
+        SELECT
+            parent_projects.project_id,
+            unnest(
+                ARRAY[parent_projects.project_id]
+                || array_agg(child_projects.project_id)
+            ) AS self_and_children_project_ids
```
frankhereford, line 102: This little trickery is constructing an array of all a given project's children id + its own id, making it a set of all IDs involved in project family.

moped-database/views/project_list_view.sql
```diff
@@ -236,7 +274,9 @@ SELECT
         GROUP BY ptags.project_id
     ) AS project_tags,
     concat(added_by_user.first_name, ' ', added_by_user.last_name) AS added_by,
-    mpcs.components
+    mpcs.components,
+    districts.self_alone_council_districts,
+    districts.self_and_children_council_districts
```
frankhereford, line 279: I went real explicit on these column names because it's an uncommon distinction to be making, and I'm _very_ open to suggestions for names.

frankhereford:
Sorry for the noise re reviews on Friday. I finished up the DB work, thinking that was the entirety of the issue, but it's also to include the value in the project list. I'm amending this PR right now. Thanks!

moped-database/views/project_list_view.sql
johnclary: how about: `project_council_districts` and `project_and_child_project_council_districts` ?

moped-database/views/project_list_view.sql
johnclary: very cool! TIL

moped-database/views/project_list_view.sql
```diff
@@ -82,6 +81,45 @@ moped_proj_components_subtypes AS (
...
+
+project_district_association AS (
+
+    WITH
+    project_council_district_map AS (
+        SELECT DISTINCT
+            project_id,
+            unnest(council_districts) AS council_district
+        FROM project_geography
+    ),
```
johnclary, line 94: one other issue related to `uniform_features`: it appears that there is a bug in that view (courtesy of yours truly 🤦 ) such that deleted council districts are not excluded from the view. i was able to reproduce this by creating a project with two components, each in a different district, and then deleting one of the components.

johnclary (changes requested):
@frankhereford this looks fantastic! we're going to make a lot of Moped user's happy with this enhancement.

i have bene testing with prod data, and beautifully fast on my local machine. I may indeed be wringing my hands too much wrt to the `uniform_features` view and shaving milliseconds off the query time. i'm curious what you think about that. 

in any case, there is a small bug that causes deleted components' council districts to appear in the view.

other than that this looks is ready to ship!

moped-database/views/project_list_view.sql
```diff
@@ -82,6 +81,45 @@ moped_proj_components_subtypes AS (
...
+),
+
+project_district_association AS (
+
+    WITH
+    project_council_district_map AS (
+        SELECT DISTINCT
+            project_id,
+            unnest(council_districts) AS council_district
+        FROM project_geography
```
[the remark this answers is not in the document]
frankhereford:
```sql
select moped_project.project_id, moped_proj_components.project_component_id, features.id as feature_id, project_geography.geography
from moped_project
left join moped_proj_components on (moped_project.project_id = moped_proj_components.project_id)
left join features on (moped_proj_components.project_component_id = features.component_id)
left join project_geography on (moped_project.project_id = project_geography.project_id and features.id = project_geography.feature_id)
where moped_proj_components.is_deleted is false;
```

@johnclary, thank you for this comment. The path you suggested works great, but I am wondering if you may have intended to suggest a different route that doesn't include `project_geography` and `uniform_features` by extension.

I think that there is certainly a performance cost for using the `uniform_features` view, as it does pull in **all** the underlying feature storage tables when you're asking for data from specific ones. However, it's the polymorphism complication coming back up; one doesn't know at query-authoring-time which `feature_` table you're going to ultimately be accessing for each component, and it's through the process of making them all uniform and `union`ing them that allows for the query to be written as a single query and not a pair of them tied together in some sort of loop. 

I think the performance cost can be pushed into one of two places,but not avoided. It's in the application looping over the components and looking up which `feature_` table to find the geography, or it's in the `uniform_features` table which is that same loop, just unrolled and pushed down into the database.

This stuff is complicated to me and I apologize if I'm being dense and missing the point. Thank you again for your help!

moped-database/views/project_geography.sql
```diff
@@ -23,6 +23,7 @@ FROM moped_project
 INNER JOIN moped_proj_components ON moped_project.project_id = moped_proj_components.project_id
 INNER JOIN moped_components ON moped_proj_components.component_id = moped_components.component_id
 INNER JOIN feature_layers ON moped_components.feature_layer_id = feature_layers.id
-INNER JOIN uniform_features ON moped_proj_components.project_component_id = uniform_features.component_id;
+INNER JOIN uniform_features ON moped_proj_components.project_component_id = uniform_features.component_id
+WHERE moped_proj_components.is_deleted IS false;
 
-COMMENT ON VIEW public.project_geography IS 'The project_geography view merges project-specific data with the unified geographical features from the uniform_features view. It links projects with their respective geographical components, including type, attributes, and location.';
+COMMENT ON VIEW public.project_geography IS 'The project_geography view merges project-specific data with the unified geographical features from the uniform_features view. It links projects with their respective geographical components, including type, attributes, and location.'; -- noqa: LT05
```
frankhereford, line 29: This is what it looks like to carve out an exception to a linting rule. This line is /way/ too long, but I think it's the lesser of two evils than doing some convoluted use of `' ||<newline>'` or something like that to break the long string up. 

I do want to keep this linting rule, I think it's always better to fit in an 80 character terminal if you can, and this rule will keep us breaking our long clauses up over lines, which is much clearer to me.

moped-database/metadata/tables.yaml
frankhereford: Great choices, I have adopted them in this commit: [c0c1a5a](https://github.com/cityofaustin/atd-moped/pull/1274/commits/c0c1a5a64f5202fed8fd5d676c24e5323d61e054)

moped-database/views/project_list_view.sql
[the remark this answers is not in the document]
johnclary: > I am wondering if you may have intended to suggest a different route that doesn't include project_geography and uniform_features by extension.
ugh, sorry for that confusing typo. i updated my comment.

my thinking, which may be flawed, was that you could bypass the `feature_` tables in this case, because (thanks to inheritance) it is the `features` table that holds all the feature IDs in one place along with their project component. so that table alone can serve as the bridget to roll up council districts into projects. 

borrowing form your SQL:
```sql
SELECT
    moped_project.project_id,
    moped_proj_components.project_component_id,
    features.id AS feature_id,
    features_council_districts
FROM
    moped_project
    LEFT JOIN moped_proj_components ON (moped_project.project_id = moped_proj_components.project_id)
    LEFT JOIN features ON (moped_proj_components.project_component_id = features.component_id)
    LEFT JOIN features_council_districts on features_council_districts.feature_id = features.id
WHERE
    moped_proj_components.is_deleted IS FALSE;
```

moped-database/views/project_list_view.sql
frankhereford: Thanks again for this feedback. Commit [ead951d](https://github.com/cityofaustin/atd-moped/pull/1274/commits/ead951d74cfc7ec582a0bbac108d5d92e0c696d4) addresses it, I believe.

moped-database/views/project_list_view.sql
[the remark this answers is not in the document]
frankhereford: Thanks again for this feedback. Commit [ead951d](https://github.com/cityofaustin/atd-moped/pull/1274/commits/ead951d74cfc7ec582a0bbac108d5d92e0c696d4) addresses it, I believe.

moped-database/views/project_list_view.sql
johnclary: thanks for that!

moped-database/views/project_geography.sql
johnclary: agreed!

johnclary (approved):
this one is ready for prime time! thanks for going through a few iterations!!!

🚢 🚢 🚢 🚢 🚢

moped-database/views/project_list_view.sql
```diff
@@ -1,6 +1,5 @@
--- latest version 1706897389736_fix_sub_comp_date_dash
```
chiaberry, line 1: i like being able to see when it was last updated, not sure if omitting this from the file was intentional

frankhereford: Thank you for pointing that out Chia, that's helpful. I removed it, and didn't replace it because I am planning on including that comment hint about what's the most recent migration that sets a view as part of my work today on https://github.com/cityofaustin/atd-data-tech/issues/15704, but I should have updated it instead in this PR. I'll stick it back in there - thank you again.

chiaberry (approved):
I think users will be glad to see this!

mddilley:
@frankhereford I started testing this, and I'm looking at project ID 86, and I'm seeing the same districts for both the `project_council_districts` and `project_and_child_project_council_districts`. This project has no components of its own so should we expect to see the same districts in both? I might be missing the difference between the two columns. Also noting that the Council Districts shown in the summary view show project + children so maybe this is what is expected.
Screenshot 2024-02-14 at 12 38 32 PM

frankhereford:
@mddilley wrote

> I'm looking at project ID 86 and seeing the same districts for both the `project_council_districts` and `project_and_child_project_council_districts` columns

Thanks for the report Mike, I'm checking into it right now. 🙏

frankhereford:
Huge 🙏 to @mddilley for catching that I had accidentally duplicated the query that computes districts of self + child into the column that should only report self's districts. This has been addressed, and the table aliases in the CTE have been redone to make it more clear what is going on.

Additionally, in investigating the above stuff, I found that I had failed to exclude features which had been soft-deleted, and had only addressed components which had been soft deleted. I think using the unified_geography may have been doing that  previously, but with John's more performant solution, it needs to have both of those soft-delete states coded into the CTE.

I'm going to pull these changes into the other PR that's downstream of this momentarily, but I think this one is once again ready for review, and with a 👍 from Mike, ready to ship. Thanks again Mike for catching that oversight.

johnclary:
thanks, y'all!!!

PR #1275, opened 2024-02-12
Use Data Grid in the Projects List 

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/15353

This PR migrates the project list table from Material Table to [DataGrid](https://mui.com/x/react-data-grid/)

It does not satisfy all of the requirements from the issue linked above, namely it does not allow users to pin columns. That is a feature available only in the paid version of Data Grid. 

It also does not have consistent row height, but instead I let it have auto row height depending on the row's content. 

## Testing
**URL to test:** 
https://deploy-preview-1275--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

I want to come up with a better test guide than "do everything you can on the table", but that's really what has to be done to test this. 

I am testing viewing every column, toggling column visibility off and on. Sorting by various columns, advanced filters and normal filters. 

Questions I have:
1. Do you like how the columns are ordered? We have been for the most part tacking on new columns at the end, should "Created by" for example be at the end of the table? https://mui.com/x/react-data-grid/column-ordering/ The pro version offers column ordering
2. We can search and advanced search by the project description, but as far as I can remember, we have never showed it in the table -- should we add it? Or have we showed it in the table but decided to remove it? 
3. DataGrid's column visibility model follows a columnName: isVisible pattern, instead of the columnName: isHidden pattern material table does. I changed the logic, but didnt rename the config to be DEFAULT_VISIBLE_COLUMNS, is everyone okay with that? 
4. While testing I noticed that the parent project name sorting isnt working -- its sorting on parent project id instead of the name. I temporarily turned this off, its also broken in prod. I was going to ask if you think we should write a custom sort function for this, but I think I have a way around it. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.js
```diff
@@ -77,7 +77,7 @@ const useColumns = ({ deleteInProgress, onDeleteActivity, setEditActivity }) =>
       {
         headerName: "Description",
         field: "description",
-        minWidth: 150,
+        minWidth: 200,
```
chiaberry, line 80: feels a little scrunched

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -68,10 +66,10 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
       type: "String",
     },
     project_description: {
-      hidden: true,
```
chiaberry, line 71: I searched different places, and I cannot find where this `hidden` is being used or referenced. it was only on four columns, which makes me think its leftover from either GridTable or Material Table

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -68,10 +66,10 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
...
       searchable: true,
       sortable: true,
       label: "Project description",
+      showInTable: false,
```
chiaberry, line 72: the helper function loops through the column definitions looking for showInTable as true, so this isn't any different behavior, just making the definition explicit

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -306,7 +301,7 @@ export const DEFAULT_HIDDEN_COLS = Object.entries(
   PROJECT_LIST_VIEW_QUERY_CONFIG.columns
 ).reduce((acc, [columnName, config]) => {
   if (config.showInTable === true) {
-    acc[columnName] = config.defaultHidden;
+    acc[columnName] = !config.defaultHidden;
```
chiaberry, line 304: https://mui.com/x/react-data-grid/column-visibility/

Data Grids default is all columns visible, setting it false in the column visibility model hides the column.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -92,26 +85,11 @@ const ProjectsListViewTable = () => {
 
   const { hiddenColumns, setHiddenColumns } = useHiddenColumnsSettings({
     defaultHiddenColumnSettings: DEFAULT_HIDDEN_COLS,
-    storageKey: "mopedColumnConfig",
+    storageKey: "mopedProjectListColumnConfig",
```
chiaberry, line 88: renamed instead of trying to use the old one, also as we potentially add more datagrid tables and storable columns, having a more granular name seems like a good idea

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -227,22 +185,56 @@ const ProjectsListViewTable = () => {
...
-                  data={data.project_list_view}
-                  columns={columns}
-                  title=""
-                  options={tableOptions}
-                  components={tableComponents}
-                />
-              </Card>
-            ) : (
-              <span>{error ? error : "Could not fetch data"}</span>
+          <Box sx={{ height: "600px", padding: "14px", marginTop: "24px" }}>
```
chiaberry, line 188: we have the sx prop in a few of our components sprinkled about, I'm torn between leaving this here for legibility vs moving it into a classes.styleName

frankhereford (approved):
## 💯 🚀 🚢 

Wow, this is a world of improvement. I've spent an hour or so with it, using it and going over the code, and this is just so great. I can't wait for this to ship!

## Your questions

1) I never had a thought that I'd reorder them, so I think yes. I don't have the real-user-experience though, so take me with a grain of salt here.

1) I think it's OK. Search, to me, is asking the machine to look into the unknown and find things that are related, which is what we're doing here. However, if we were to have a "filter these results" input, it would be totally wrong to include matches on things not being shown to the user. 👍 

1) Totally fine with me. Maybe stick a comment on line 304 in `moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js` to explain the negation?

1) Cool - great catch

## Final Thoughts

The only thing that I wanted to ask about is the hard-coded `600px` height that encloses the whole thing, and see what you thought about trying to make this table fit into the available height of the viewport. 

## 🙏 

Thanks again Chia - this is fantastic work!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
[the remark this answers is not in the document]
chiaberry: ahhh thank you for the screenshot, i was wondering what it would look like on a different / bigger monitor than the ones ive been using, ill try with the vh!

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -144,59 +122,39 @@ const ProjectsListViewTable = () => {
...
   });
 
-  const tableOptions = useTableOptions({ queryLimit, data });
-
-  const sortByColumnIndex = columns.findIndex(
-    (column) => column.field === orderByColumn
-  );
-
   /**
    * Handles the header click for sorting asc/desc.
```
mddilley, line 126: omg this is so nice to have the sort direction arrows working!!! 🎉 🙌

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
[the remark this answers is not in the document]
mddilley: this is a great catch, Frank! I agree that vh would be a good route. I faced this same problem with the [projects map layout](https://github.com/cityofaustin/atd-moped/pull/1276), and it made me wonder if we should spin off an issue to make this uniform across our main views like dashboard, projects, project page, and staff. idk maybe some of them needed different treatment. 🤷

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -227,22 +185,56 @@ const ProjectsListViewTable = () => {
...
+                  setHiddenColumns(newModel);
+                }}
+                slots={{
+                  toolbar: ProjectListToolbar,
+                }}
+                slotProps={{
+                  columnsPanel: {
+                    // Hiding buttons because when I toggle "show all" its only setting one column to visible
+                    // instead of including every column in the object
+                    disableShowAllButton: true,
```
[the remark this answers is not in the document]
mddilley: Yeah, same, i was initially confused when this button cleared the table completely, but I also think it is a nice way to start fresh!

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -157,346 +134,277 @@ export const useColumns = ({ hiddenColumns }) => {
...
+        width: COLUMN_WIDTHS.xsmall,
       },
       {
-        title: "Name",
+        headerName: "Name",
         field: "project_name",
-        hidden: hiddenColumns["project_name"],
-        render: (entry) => (
+        flex: 2,
+        minWidth: COLUMN_WIDTHS.xlarge,
```
mddilley, line 157: this is a cool setting to be able to set each cell width behavior. i like! 😎

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -157,346 +134,277 @@ export const useColumns = ({ hiddenColumns }) => {
...
+            to={`/moped/projects/${row.parent_project_id}`}
             state={{ queryString }}
             sx={{ color: theme.palette.primary.main }}
           >
-            {entry.parent_project_name}
+            {row.parent_project_name}
           </Link>
         ),
+        flex: 1,
+        sortable: false,
```
mddilley, line 385: Part of me wonders if this should be pushed up into `COLUMN_CONFIG`? Idk I found finding the right mix of the existing `PROJECT_LIST_VIEW_QUERY_CONFIG` config and the config needed for the table library to be challenging when i was looking at this a few months ago.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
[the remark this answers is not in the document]
chiaberry: I changed it to be dynamic, thank you for that recommendation. Let me know if you think it can refined more

mddilley (approved):
What an upgrade! I tested locally and pagination, sort, column toggle, column visibility and the data requests, cells, links, and everything work just like expected. Still super excited about the working sort arrows 🎉 🚢

For your questions:
1. i think keeping the same column order is great for now
2. can't remember the backstory on description but showing it could make a big blob of text
3. i think this is okay. not too much of a lift to understand the hide/show logic at all

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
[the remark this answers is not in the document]
mddilley: oops my comment is outdated now. The new code looks ✨

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -157,346 +134,276 @@ export const useColumns = ({ hiddenColumns }) => {
...
       },
       {
-        title: "Parent project",
-        field: "parent_project_id",
-        hidden: hiddenColumns["parent_project_id"],
-        emptyValue: "-",
-        render: (entry) => (
+        headerName: "Parent project",
+        field: "parent_project_name",
+        renderCell: ({ row }) => (
```
chiaberry, line 374: sorry to make this change as you were reviewing this, but I swapped the field from the parent_project_id to the parent_project_name so the sorting would work. previously it was sorting on the parent project id and displaying the name. now it sorts on the name and displays the name.

mddilley: all good! i can take a closer look tomorrow, but I'm not sure if the sort is acting quite like we expect it too. i think that I'm seeing a difference between the order of the returned results and what is displayed.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -227,22 +185,63 @@ const ProjectsListViewTable = () => {
...
+                columnVisibilityModel={hiddenColumns}
+                onColumnVisibilityModelChange={(newModel) => {
+                  setHiddenColumns(newModel);
+                }}
+                slots={{
+                  toolbar: ProjectListToolbar,
+                }}
+                slotProps={{
+                  columnsPanel: {
+                    // Hiding buttons because when I toggle "show all" its only setting one column to visible
```
johnclary, line 223: is this a bug, or a some limitation in the way we manage column visibility?

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -227,22 +185,63 @@ const ProjectsListViewTable = () => {
...
+                getRowId={(row) => row.project_id}
+                disableRowSelectionOnClick
+                rows={data.project_list_view}
+                onSortModelChange={handleSortClick}
+                disableColumnFilter
+                paginationMode="server"
+                paginationModel={{
+                  page: queryOffset / queryLimit,
+                  pageSize: queryLimit,
+                }}
```
johnclary, line 238: maybe DataGrid handles this better than Material Table, but this prop should probably be memoized so we don't start to regress back toward the way things were before we optimized the project list props.

johnclary: looking at this some more, i see the `sx` and a few other props are not memoized. i guess this is all fine 🤷

johnclary (commented):
this looks fantastic @chiaberry! i tried very hard to break it and i could not. really nice work.

there are a few minor adjustments needed with the layout. the biggest issue for me is the outer scrollbar that allows the search bar to scroll partly under the top bar. there should not be any scrollbars needed on this page except within the data grid. 

i would also like to see the vertical space tightened up, and i think the card border around the bordered data grid looks a bit off. i have annotated a screenshot for you.

Screenshot 2024-02-28 at 11 22 15 AM

moped-editor/src/views/projects/projectsListView/helpers.js
chiaberry: I reverted back to its old ways, and turned off the sorting. This can be a refinement later, I wanna get this in to moped.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: https://github.com/mui/mui-x/issues/11656#issuecomment-1939606662

I think its a bug

chiaberry:
@johnclary I'll work on getting rid of the scroll and shrinking the gray space. 

I kept the padding around the table so it matched the padding around the search, it felt off to me without it, but I tend to like more space in general. I'll get a screenshot without it and see what the other half of Chimenity thinks

amenity (commented):
This looks so good, @chiaberry !! 😍 

### Bug? 
I'm not able to horizontal scroll when my cursor is over the column heading row. 

### Comments
- Is it straightforward to remove "Manage columns" from the individual column menus? It's pretty redundant. Screenshot 2024-02-28 at 3 29 16 PM

### Answers 
> Do you like how the columns are ordered?

I'd also never considered it.  "Created by" definitely belongs at the end, but I'm really not sure about the others. In any case, I think that it will be a nice thing to clean up at some point. We can prioritize the most used fields but also keep the groups of related fields together. Right now, I'm too rusty on terminology to even hazard a guess. 

> can search...by project description...should we add it?

I think that would be logical, but can wait until someone requests it. Also, once PIO gets involved, I imagine they might clean up some of the descriptions, i.e. bond-funded projects. In that case, they'd probably appreciate being able to pull that content in a CSV download.

chiaberry:
@amenity I cant scroll horizontally when im over the headers either! 

I can remove the "manage columns" option, but out of the box it also removes the "hide column" option
Screenshot 2024-02-28 at 4 52 11 PM

Want me to go ahead and do that?

amenity:
Just told @chiaberry in Slack that I'm okay losing "Hide column" along with "Manage columns" since the "Columns" button, where both of those things can be accomplished, is sticky in the upper left.

PR #1277, opened 2024-02-12
Add council districts to advanced search

frankhereford:
## Associated issues

This PR aims to close: https://github.com/cityofaustin/atd-data-tech/issues/10893

## Testing

This is going to inherit the need to do local testing from its upstream branches.

**Steps to test:**

1) Spin up the stack, and it's helpful to grab a copy of production data since you'll want a lot of different projects to search in.
1) Head on over to the project listing page and do some advanced searching. 
* Can you search for a single district's projects?
* Can you break the input for the new advanced search by putting wacky stuff in the input box? 
* Important: Can you still do searches with search operators that use the other types of data? 

### A brief diversion about this point
I found this line in our codebase:

```javascript
value = type in ["number", "boolean"] ? value : `"${value}"`;
```

I don't think it was doing what we thought. The `in` operator in JavaScript checks for property names in objects, not values in an array. I thought this would fly at first too, as written, because arrays are a subset of objects, but apparently not in this way.

image

So.., we're using a variable called type which is very easy to misconstrue with the operator `typeof`. For me, I initially read that as, if the string variable named `type` is in the array of strings containing `'number'` and `'boolean'` then set `value` to itself, as a non-op. If the condition is false, interpolate the string variable `value`'s contents into a string wrapping it in double quotes, but the `in` operator was causing it to always be false.

Needless to say, I got deeply confused here for a minute, so please give this an extra close look.

---

#### Ship list

- [ ] Code reviewed


moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -73,7 +73,17 @@ const makeAdvancedSearchWhereFilters = (filters) =>
...
+            // eslint-disable-next-line no-self-assign
+            value = value;
+          } else if (type === "array") {
+            // a virtue of this stringification of an array, we actually support
+            // comma delimited lists of districts and `graphql-engine` will do the
+            // correct set algebra
+            value = `[${value}]`;
+          } else {
+            value = `"${value}"`;
+          }
```
frankhereford, line 86: Please see main PR comment, but short-story here is that `in` isn't doing what we thought it was. I believe `.includes()` fills the role of checking if an element exists in a list.

moped-editor/src/components/GridTable/FiltersCommonOperators.js
```diff
@@ -168,4 +168,32 @@ export const FiltersCommonOperators = {
...
+    type: "array",
+    envelope: false,
+  },
+  council_districts_array_is: {
+    operator: "_eq",
+    label: "Is",
+    description: "Project has components in, and only in, specific council districts",
+    type: "array",
+    envelope: false,
+  },
```
frankhereford, line 198: I am a little queasy here about making these operators that are specific to the council districts, but I couldn't figure out a way to make the existing array-type, field specific operator's descriptions work with the council district's needs. If you look just above these declarations, you can see the existing ones I'm referring to.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -73,7 +73,17 @@ const makeAdvancedSearchWhereFilters = (filters) =>
 
           // If it is a number or boolean, it does not need quotation marks
           // Otherwise, add quotation marks for the query to identify as string
-          value = type in ["number", "boolean"] ? value : `"${value}"`;
+          if (["number", "boolean"].includes(type)) {
+            // eslint-disable-next-line no-self-assign
+            value = value;
+          } else if (type === "array") {
```
frankhereford, line 79: This construct here is to cause non-ops for these two types of `type.` It's not pretty, made obvious by having to disable that linting line, but .. a switch statement would have to start with switch(true) and then put expressions in the cases -- hardly any more elegant. This at least felt very explicit about what's happening, perhaps even to a fault.

johnclary (changes requested):
this looks great, Frank. it's really interesting to see how these filters get implemented; it's not part of the code i'm familiar with.

my feedback is pretty minor, although i did find a small bug when i put a text character into the search 😈

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -73,7 +73,17 @@ const makeAdvancedSearchWhereFilters = (filters) =>
...
+            // a virtue of this stringification of an array, we actually support
+            // comma delimited lists of districts and `graphql-engine` will do the
+            // correct set algebra
+            value = `[${value}]`;
```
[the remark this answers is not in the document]
frankhereford: I'm totally in agreement with you - I'm going to incorporate the number validation in deference to the original scope of the issue, but I don't think it'd be terrible to spin off a low priority issue to write a custom validator to take advantage of being able to specify multiple districts in a comma delimited list.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
frankhereford: Goodness, that is remarkably better. Thank you!

frankhereford:
Neither here nor there, but I have pulled `main` and each of the parent feature branches through to this one, including feedback changes I've made today.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -414,6 +416,18 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
         "string_is_not_null",
       ],
     },
+    {
+      name: "project_and_child_project_council_districts",
+      label: "Council Districts",
```
chiaberry, line 421: This is title case instead of sentence case, but I also noticed that "Has Subprojects" is also title case and I'm pretty sure I did that, somehow thinking subprojects was a proper noun (?!?!).

chiaberry (approved):
Nice catch fixing that line that's been in moped for a bit -- I couldnt remember if it came from gqlAbstract and GridTable world, so I did a search to see if it was also in VZ. Couldnt find it, but wanted to make sure. 

Left a note about sentence case vs title case, but this is good to go!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
frankhereford: Great catch! I'm about to push up fix to get this right. 🙏

moped-editor/src/components/GridTable/helpers.js
```diff
@@ -31,7 +31,8 @@ export const generateEmptyFilter = () => {
  */
 export const checkIsValidInput = (filterParameter, type) => {
   // If we are testing a number type field with a non null value
-  if (type === "number" && !!filterParameter.value) {
+  // if (type === "number" && !!filterParameter.value) {
```
mddilley, line 34: Could you remove the commented code here? This change looks good to me! Just to confirm - we are validating `array` type inputs for number characters only because the search only supports single districts for now, right? A comment somewhere in here that documents how this works could be helpful to understand how this works in the future.

Edit: Just saw the later discussion about this. A spin off issue is great! Still think a comment would hold us over until that issue is prioritized.

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -73,7 +73,17 @@ const makeAdvancedSearchWhereFilters = (filters) =>
 
           // If it is a number or boolean, it does not need quotation marks
           // Otherwise, add quotation marks for the query to identify as string
-          value = type in ["number", "boolean"] ? value : `"${value}"`;
```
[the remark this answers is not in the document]
mddilley: +1 🙏 - someone was having a case of Python brain here 🐍

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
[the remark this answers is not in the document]
mddilley: This all sounds great to me. I think the number validation works great, and there is even a workaround to searching for multiple districts by stacking rows of filters on the same column for now too.

Screenshot 2024-02-15 at 10 26 51 AM

mddilley (approved):
Tested this out and all of the operators work as expected! Awesome to search this newly-available field! 🚀 🙌

moped-editor/src/components/GridTable/FiltersCommonOperators.js
mddilley: Column specific operators sound good to me too. One of the challenges of refactoring some of this was the use of more general terms like "special case" so I am all for explicit, column-specific naming if we can't find a way to generalize. 🙏

moped-editor/src/components/GridTable/helpers.js
frankhereford: great call, thank you Mike! 

https://github.com/cityofaustin/atd-moped/pull/1277/commits/edc111ff7dc880f71648c3cfe644dff8857f5011

johnclary:
beautiful work! it's headed to main now...! 📦

johnclary (approved):

PR #1278, opened 2024-02-15
Exclude audit field from project phase event trigger listener

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/15856

## Testing
**URL to test:**  Local

**Steps to test:**
1. Create a project phase
2. Edit the phase you created, but only fill in the Status update field
3. Does the activity log ok? 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -3469,7 +3469,21 @@
         insert:
           columns: '*'
         update:
-          columns: '*'
+          columns:
+            - is_phase_start_confirmed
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
good catch

mddilley (approved):
Tested locally and the activity log looks great to me! Really good find and unexpected!!! 🚀

johnclary:
thank you! we should probably do this across the board for all our triggers, just to be safe.

PR #1280, opened 2024-02-16
replace moped_proj_notes rest connectors with db triggers

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15489

## Testing
**URL to test:** 

local

**Steps to test:**
- go to a project's notes section and add a note
- confirm that the update is registered in the activity log
- edit the note and check the activity log
- go to the new note record in the db and confirm the created_at, updated_at, created_by_user_id and updated_by_user_id columns are correct. there should no longer be a date_added or added_by_user_id column
- delete the note and confirm in the activity log
- confirm in the db that the record was set to `is_deleted` is `true`
- try adding a note from the project summary page status update modal
- try adding a note from the phases table

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -2810,7 +2810,7 @@
         foreign_key_constraint_on: project_id
     - name: moped_user
       using:
-        foreign_key_constraint_on: added_by_user_id
+        foreign_key_constraint_on: created_by_user_id
   insert_permissions:
     - role: moped-admin
       permission:
@@ -2819,7 +2819,6 @@
           created_by_user_id: x-hasura-user-db-id
           updated_by_user_id: x-hasura-user-db-id
         columns:
-          - added_by_user_id
           - is_deleted
           - phase_id
           - project_id
@@ -2833,7 +2832,6 @@
           created_by_user_id: x-hasura-user-db-id
           updated_by_user_id: x-hasura-user-db-id
         columns:
-          - added_by_user_id
           - is_deleted
           - phase_id
           - project_id
@@ -2844,7 +2842,6 @@
     - role: moped-admin
       permission:
         columns:
-          - added_by_user_id
           - created_at
           - created_by_user_id
           - is_deleted
@@ -2860,7 +2857,6 @@
     - role: moped-editor
       permission:
         columns:
-          - added_by_user_id
           - created_at
           - created_by_user_id
           - is_deleted
@@ -2876,7 +2872,6 @@
     - role: moped-viewer
       permission:
         columns:
-          - added_by_user_id
           - created_at
           - created_by_user_id
           - is_deleted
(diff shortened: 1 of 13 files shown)


moped-database/metadata/tables.yaml
```diff
@@ -3430,8 +3447,6 @@
   update_permissions:
     - role: moped-admin
       permission:
-        set:
-          updated_by_user_id: x-hasura-user-db-id
```
tillyw, line 3434: this is just moving stuff around you can ignore this

moped-editor/src/views/projects/projectView/ProjectPhase/helpers.js
```diff
@@ -192,7 +192,6 @@ export const onSubmitPhase = ({
         {
           project_note: DOMPurify.sanitize(noteData.status_update),
           project_id,
-          added_by_user_id: noteData.user_id,
```
mddilley, line 195: thanks for catching this update in the phase form! 🙏

mddilley (approved):
I'm seeing the updated behavior for all inserts, edits, and deletes of notes and status updates. I also see the parent project timestamp updating. Hooray! 🙌 🚢

chiaberry (approved):
ran through the test steps, was going to test if edits by other users worked as expected, then i remembered you cant edit other peoples notes
🚢

chiaberry:
> Unrelated to the changes in this PR, but I noticed that, when creating a new note, the `updated_at` timestamp is slightly ahead of the `created_at` timestamp. I expected them to be the same, and I think we could make a small adjustment to `public.update_self_and_project_updated_audit_fields()` to use the value of `created_at` of the new record instead of calling `NOW()`.
> 
> What do you think @chiaberry? Might not be as big of deal as I'm thinking.

Thats a good question @mddilley , I did not notice they were slightly out of sync, and that the updated_at happens BEFORE insert or update, so the updated at is before the created at time. 

On one hand it bothers me because it feels illogical to have something be updated at before its created, but I am trying to come up with scenarios where we would regret these discrepancies, and I cant come up with one. Would we be able to make an adjustment to the function to be able to pass the created_at before insert? I am thinking we cant get the created at time until after its been created and this happens before....

johnclary:
good catch @mddilley i think we should patch that function to use the `created_at` timestamp. you can use the `TG_OP` context variable to see if we're handling an insert vs update 😅

chiaberry (changes requested):
The dashboard query is referencing created_at, I get an apollo error when i navigate to the Dashboard

https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/queries/dashboard.js#L30

tillyw:
> The dashboard query is referencing created_at, I get an apollo error when i navigate to the Dashboard
> 
> https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/queries/dashboard.js#L30

thanks for catching this! i forgot about the dashboard >.< fixed!

chiaberry (approved):
thanks for fixing the dashboard query, the ✅  is back

chiaberry:
going to merge so we can start testing

PR #1287, opened 2024-02-22
fixes deleting phase in activity log

tillyw:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15932

## Testing
**URL to test:** 

https://deploy-preview-1287--atd-moped-main.netlify.app/

**Steps to test:**
- delete a current phase from a project's timeline
- delete a non-current phase from a project's timeline
- check the activity log, the two actions should render the same "x Deleted the phase x"

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/views/project_list_view.sql
+++ b/moped-database/views/project_list_view.sql
@@ -219,10 +219,10 @@ LEFT JOIN project_district_association districts ON mp.project_id = districts.pr
 LEFT JOIN LATERAL (
     SELECT
         mpn.project_note,
-        mpn.date_created
+        mpn.created_at AS date_created
     FROM moped_proj_notes mpn
     WHERE mpn.project_id = mp.project_id AND mpn.project_note_type = 2 AND mpn.is_deleted = false
-    ORDER BY mpn.date_created DESC
+    ORDER BY mpn.created_at DESC
     LIMIT 1
 ) proj_status_update ON true
 WHERE mp.is_deleted = false
--- a/moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
+++ b/moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
@@ -37,19 +37,8 @@ export const formatPhasesActivity = (change, phaseList, subphaseList) => {
 
   let changes = [];
 
-  // loop through fields to check for differences, push label onto changes Array
-  Object.keys(newRecord).forEach((field) => {
-    if (newRecord[field] !== oldRecord[field]) {
-      // filter out fields that are not listed in the activity log table maps to prevent
-      // automated field updates (created at, updated at, etc.) from entering the array
-      if (!!entryMap.fields[field]) {
-        changes.push(entryMap.fields[field]?.label);
-      }
-    }
-  });
-
-  // if the changes array includes a deletion that supersedes any other changes
-  if (changes.includes("is deleted")) {
+  // if the record has been deleted that supersedes any other changes
+  if (newRecord["is_deleted"] === true) {
     return {
       changeIcon,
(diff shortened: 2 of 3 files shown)


tillyw:
there were a few ways to fix this but i thought this made the most sense! this way if any of the field changes include a deletion it will supersede any of the other changes. this might be a good pattern to use for some of the other components, and i think it's worth considering whether we should try to consolidate all these activity log helpers into on component so it will be easier to update patterns (like this or excluding audit fields, etc.) for all activity log events

moped-editor/src/utils/activityLogFormatters/mopedPhasesActivity.js
```diff
@@ -61,6 +48,20 @@ export const formatPhasesActivity = (change, phaseList, subphaseList) => {
     }
   });
 
+  // if the changes array includes a deletion that supersedes any other changes
+  if (changes.includes("is deleted")) {
```
johnclary, line 52: @tillyw i'm not familiar enough with the activity log to know how these entries are built up, but I would be wary of keying on the `is deleted` string if it is a label could theoretically be changed at some point. if possible, it would nice to make this work by keying on the column name `is_deleted` instead.

chiaberry: one idea would be a few lines above this, where we look for the differences, to to check if the field is is_deleted and stop looping. that way we arent tied to the label

tillyw: thanks for pointing that out! i updated the code to use the ``is_deleted`` key and pushed the delete return action prior to looping through any other fields. re-requesting review!

chiaberry (approved):
🚢

johnclary (approved):
nice—thanks @tillyw!

roseeichelmann:
I keep getting this error when i click on any project :weary: happening with all moped test user accounts. strikes me as a permissions problem but idkk

Screenshot from 2024-03-05 16-18-14

chiaberry:
@roseeichelmann ah, this branch needs to be updated to match with what is in staging, I can merge main to it and I will see if that fixes it

moped-database/views/project_list_view.sql
```diff
@@ -219,10 +219,10 @@ LEFT JOIN project_district_association districts ON mp.project_id = districts.pr
 LEFT JOIN LATERAL (
     SELECT
         mpn.project_note,
-        mpn.date_created
+        mpn.created_at AS date_created
     FROM moped_proj_notes mpn
     WHERE mpn.project_id = mp.project_id AND mpn.project_note_type = 2 AND mpn.is_deleted = false
-    ORDER BY mpn.date_created DESC
+    ORDER BY mpn.created_at DESC
```
chiaberry, line 225: these changes appeared after i merged main into the branch, the github workflow changed it

mddilley: this change surprised me because I would expect it to have already been applied to `main` and not be introducing a diff unrelated to this PR. I think this is just the GH workflow catching up to previous changes, though. 🤔

mddilley (approved):
Looks great, Tilly! Thanks for fixing this! 🙌 🚀

roseeichelmann (approved):
thanks tilly! test steps worked perf!

moped-database/views/project_list_view.sql
tillyw: this confuses me too!! was the project list view broken at any point due to this?

tillyw: ok it wasn't, that is a relief. very happy with this workflow!

PR #1288, opened 2024-02-23
Add GitHub workflow to export views to SQL based on the migrations in your branch

frankhereford:
## Associated issues

This PR aims to close https://github.com/cityofaustin/atd-data-tech/issues/15704.

##  Process

I find the easiest way to develop GitHub workflows is to commit directly into `main`, because the state of the repo is a parameter to the functionality you're building. I also knew that while this was under development, I'd need to be creating a new feature branch via the workflow every time I tested it, so that I wouldn't have to re-de-format my SQL as the workflow overwrites the files and commits them back into the repo. I knew I didn't want to accidentally push up 40 useless branches by accident at all. 😅 So, I forked the repo into my personal account so I handle all those concerns. 

⚠️ All that to say, `frankhereford:main` is the feature branch I'm PRing into `cityofaustin:main`. 

### Demonstration

See here: https://github.com/frankhereford/atd-moped/pull/4.

### Manual testing

This one is sort of different in terms of testing. It may be possible to read the code and convince yourself that it can't really harm anything, and let it be merged. The triggering event is pushing on a branch which is not `main` or `production`, and to really simulate that, you need to do the following. It's a mini pain:

* Fork [the upstream repo](https://github.com/frankhereford/atd-moped) in your own account
* Enable actions for the new repo
* Disable all other workflows except the one we're testing here because they need secrets your new repo doesn't/shouldn't have defined.
* Create a feature branch in the new repo and make a migration that changes a view
* When you push up the feature branch with the new migration, you should see your feature branch get commits from the "View Bot" and have the SQL files reflect your migrations.

## Question: Should I format the view SQL files as part of this PR?

I have formatted the existing view files with the workflow in here, so they will not generate unexpected changes for the first person who modifies a moped view after this makes it into `main`. Let me know if y'all'd like this handled differently - thanks!

---
#### Ship list
- [ ] Code reviewed 


.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,77 @@
+name: Export Moped Views to SQL
+
+on:
+  push:
+    branches-ignore:
+      - "main"
+      - "production"
```
frankhereford, line 7: For testing in your personal repo, you can remove these 3 lines to make the workflow trigger on any push. This is handy for development too.

.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+          mkdir -p moped-database/views
+          psql -A -t -c "SELECT table_name FROM information_schema.views WHERE table_schema = 'public';" | \
+          grep -v -E '^geo[a-zA-Z]+y_columns$' | \
+          xargs -I {} sh -c "
+            VIEW_NAME={}
+            MOST_RECENT_MIGRATION=\$(grep -rl --include=up.sql -E 'CREATE (OR REPLACE )?VIEW (public.)?$VIEW_NAME' moped-database/migrations | sort -V | tail -n 1)
+            echo \"-- Most recent migration: \$MOST_RECENT_MIGRATION\" > moped-database/views/\$VIEW_NAME.sql
+            echo "" >> moped-database/views/\$VIEW_NAME.sql
+            psql -A -t -c \"SELECT 'CREATE VIEW ' || '\$VIEW_NAME' || ' AS ' || pg_get_viewdef('\$VIEW_NAME'::regclass, true);\" >> moped-database/views/\$VIEW_NAME.sql
```
frankhereford, line 61: I didn't really want to embed comments in here because I wasn't sure if I'd be commenting out `yaml` or `bash`.  But here's what I would have said for each line, starting with the `mkdir` one.

* `mkdir`: Create the directory if it's missing, ensuring it's there. This should never happen.
* `psql`: Ask the database to give you all the names of views in the public schema
* `grep`: Filter that list to remove the postGIS views listing available geometry and geography columns
* `xargs`: For each one of these views, we're going to iterate, and we're going to go into that iteration with `{}` as the view name.
* `VIEW_`: Save the view name in an environment variable 
* `MOST_`: Recurse search in `up.sql`s for a `CREATE` statement for this view, and pick the most recent one
* `echo`: Write that breadcrumb filename into the SQL output file
* `echo`: Add a linefeed between that and the SQL itself
* `psql`: Query out the view's source code, appending that to the output SQL file

.sqlfluff
```diff
@@ -14,7 +17,7 @@ templater = jinja
 # See https://docs.sqlfluff.com/en/stable/configuration.html#enabling-and-disabling-rules
 # AM04 (ambiguous.column_count) and ST06 (structure.column_order) are
 # two of the more controversial rules included to illustrate usage.
-exclude_rules = ambiguous.column_count, structure.column_order
+exclude_rules = ambiguous.column_count, structure.column_order, layout.long_lines
```
frankhereford, line 20: I'm adding a rule exemption for long lines, which I'm not happy about, but we have something in our `project_list_view` that puts `sqlfluff` into an infinite loop. We may be able to get this rule reinstated, but it felt outside the scope of this project.

.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,77 @@
...
+          "
+
+      - name: Run SQLFluff
+        run: |
+          sqlfluff format --config .sqlfluff moped-database/views/*.sql
+
+      - name: Set Git Config
+        run: |
+          git config --local user.email "moped@view.bot"
+          git config --local user.name "Moped View Bot"
```
frankhereford, line 71: This is the identify of the bot when it creates commits. This data is low key, but I wanted to point it out in case y'all wanted to change the values.

.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Commit and push changes
+        run: |
+          git add moped-database/views/*
+          git commit -m "🤖 Export views"
```
frankhereford, line 76: Here, this commit message actually shows up in a lot of places in the GitHub UI, so here's where I put in a 🤖 emoji. The intent is to make it clear at a glance that this file was created by a program. I'm super open to changing this value too.

.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,77 @@
...
+          echo "PGHOST=localhost" >> $GITHUB_ENV
+          echo "PGPORT=5432" >> $GITHUB_ENV
+          echo "PGDATABASE=moped" >> $GITHUB_ENV
+          echo "PGUSER=moped" >> $GITHUB_ENV
+          echo "PGPASSWORD=moped" >> $GITHUB_ENV
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
```
frankhereford, line 34: This makes the checkout start out in the feature branch push that triggered the workflow.

.github/workflows/export_database_views.sh
frankhereford, line 1: I had this bash script originally embedded in the yaml, but I ended up pulling it out into a helper script for readability and to avoid the escape-quotes insanity of bash inside bash inside yaml.

chiaberry (approved):
Tested! Thanks for the test steps and coming up with this workflow, I always learn a lot reading these bash scripts

.sqlfluff
```diff
@@ -1,5 +1,8 @@
 [sqlfluff]
 
+# Set loop limit for fixing
+runaway_limit = 10
+
```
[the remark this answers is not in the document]
mddilley: cool, sounds good to me! I like the sqlfluff def in their docs heheh
`# Number of passes to run before admitting defeat`

.github/workflows/export_database_views.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Install SQLFluff
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3-pip
+          pip3 install sqlfluff
```
mddilley, line 25: Not saying we should but do you think there is any reason to be explicit about the version here?

moped-database/views/uniform_features.sql
```diff
@@ -132,9 +105,7 @@ LEFT JOIN (
...
         d.feature_id,
         array_agg(d.council_district_id) AS council_districts
-    FROM features_council_districts AS d
+    FROM features_council_districts d
     GROUP BY d.feature_id
-) AS districts ON feature_drawn_lines.id = districts.feature_id
+) districts ON districts.feature_id = feature_drawn_lines.id
 WHERE feature_drawn_lines.is_deleted = FALSE;
-
-COMMENT ON VIEW uniform_features IS 'This view unifies various geographical feature data from multiple tables such as signals, street segments, intersections, drawn points, and lines. It provides a view of these features along with their attributes, geographic details, and council district associations.';
```
mddilley, line 140: Would it make sense to think about extending this to capture these comments too? Not 100% but it feels like the intention here was to make sure the comment survived future drops followed by creating again.

.github/workflows/export_database_views.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+function create_view_file() {
+    local VIEW_NAME=$1
+
+    echo "View: $VIEW_NAME"
+    MOST_RECENT_MIGRATION=$(grep -rl --include=up.sql -E "CREATE (OR REPLACE )?VIEW (public.)?$VIEW_NAME" moped-database/migrations | sort -V | tail -n 1)
```
mddilley, line 7: this is rad. 😎 Do you think there is a chance this would fail if there are quotes around the schema or view name? This is only on my mind because I noticed that double quotes are valid around schema and view or table when I've been working on the data model stuff.

mddilley (approved):
I trust both yours and chia's test of this in repos so I read through the code and also the logging of a completed action. This all looks ready to go to me, and it is super cool to see this idea realized. I had a few questions, and one concern about losing comments on views, but I say 🚢 🚢 🚢 !

.github/workflows/export_database_views.yml
frankhereford: The version of pip? That's a good question. 

Normally, I totally get wanting to put guardrails on versions of libraries that are outside of the core python system like in a requirements.txt, but pip seems like something you always want the most recent version of. Also, since this is coming out of the ubuntu repos, there's really no guarantee that older versions of pip are available -- in fact, I think one should bank on stale packages being removed pretty quick. It's been my experience that linux distribution repos (ubuntu, debian are the ones i know the best) don't have that npm-like, or the pip-like, feature/system of having everything every version available forever.

.sqlfluff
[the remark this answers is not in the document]
frankhereford: Oh, they do have it in there! I didn't even realize that. Here's where I got sideways with this -- when the program does get into a loop and finally quits when it hits the iteration limit, the error that it emits has no mention of this configuration name, and calls it the "loop limit." It's a terrible error text. 

Nothing I could search for based on that error turned me in the right direction, so I ended up just reading the source of the linter to figure out how it worked and that led me to the configuration directive name, `runaway_limit`.

moped-database/views/uniform_features.sql
frankhereford: Yea, it totally should make sense for us to think about it, and thank you for tagging this. I was thinking about it on the way back from the apple store today, and I think the best way to do this is drop the `psql` SQL based extraction of the query driving views, and transition to `pg_dump` which is the more capable, albeit, harder to use tool if you're seeking succinct results.

Doing this is hard enough that I think maybe I'd suggest we spin it off as a future enhancement. 

But - it bears saying - you're 100% right that this will cause us to loose comments on the tables without careful intervention by the dev doing the view update. Not ideal.

.github/workflows/export_database_views.sh
frankhereford: What an insightful question!! I'm going to answer you out of my beliefs from experience but not having just run an experiment to verify my understanding. 

I think we're square here because the process is generating the view-driving-query out of the postgres server's internal representation of the query.  I believe that the PG server outputs SQL in a very specific linted format, which doesn't use quotes. I've found that it refines queries in a whole bunch of ways, almost always adding specificity where humans might have let the query-engine infer things. 

I'll do a quick test tomorrow to firm up my understanding. Thanks again for this very thoughtful question.

.github/workflows/export_database_views.yml
mddilley: sorry, GH's UI isn't making it easy to see but this comment was on the sqlfluff line. I don't really think this is a big concern but wanted to mention it!

.github/workflows/export_database_views.sh
mddilley: Cool, yeah, I was missing the step that it is coming directly from the DB. I only went as far as seeing what formatting would output from the Hasura console. Thanks for checking it out!

moped-database/views/uniform_features.sql
chiaberry: To make sure I'm understanding, in order to not lose the comments, devs will have to go and find the comment from the most recent up migration, and add it to the new migration? 

Only some of our views have comments, its going to be hard to remember which ones we need to go back and find comments for.

frankhereford: Yea - you're right on Chia with your understanding. The way it's configured now, it's just a matter of time until we loose the comments on the views. There's a way to do it right with `pg_dump`, and I'm eager to get that worked out. We can spin that off into an issue to pick up the benefits of this work now, or I would be happy to roll that work into this PR -- either way is great with me.

.github/workflows/export_database_views.sh
frankhereford: This was addressed here, which also got the exports up to date with the latest changes as well: https://github.com/frankhereford/atd-moped/commit/87675e72da3353c1a3040280e6c1296ec6e83e3c

.github/workflows/export_database_views.yml
frankhereford: Oh, totally - thank you Mike for clarifying that. That was me misunderstanding; it makes a lot of sense what you're suggesting. I just don't see any way that could hurt, and even more, it feels like good practice. I bet that `pip3` line can specify the version somehow too, so it'd be easy or we could use redirection to get it on the workflow VM. I'll do my best to PR this in soon.

PR #1293, opened 2024-03-07
Add audit fields to moped_project_files table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/16156

## Testing
**URL to test:** 
local

**Steps to test:**

Edit the description of an existing file. Check that your id is the updated by user id and that the updated timestamp works, and that the project itself has a last updated timestamp that matches. 

Add a new file by using the file url option, you can use a random url or comment like "in g drive", and check that the audit fields make sense. 

Note: I merged Tilly's recent work into this, so its using the new and improved trigger function

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
+++ b/moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
@@ -17,4 +17,4 @@ OR UPDATE ON moped_project_files
 FOR EACH ROW
 EXECUTE FUNCTION public.update_self_and_project_updated_audit_fields();
 
-COMMENT ON TRIGGER update_moped_project_files_and_project_audit_fields ON moped_project_files IS 'Trigger to execute the update_self_and_project_updated_audit_fields function before each update operation on the moped_proj_phases table.';
+COMMENT ON TRIGGER update_moped_project_files_and_project_audit_fields ON moped_project_files IS 'Trigger to execute the update_self_and_project_updated_audit_fields function before each update operation on the moped_proj_files table.';
--- a/moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
+++ b/moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
@@ -1,20 +1,20 @@
-ALTER TABLE moped_proj_files RENAME COLUMN create_date TO created_at;
-ALTER TABLE moped_proj_files RENAME COLUMN created_by TO created_by_user_id;
+ALTER TABLE moped_project_files RENAME COLUMN create_date TO created_at;
+ALTER TABLE moped_project_files RENAME COLUMN created_by TO created_by_user_id;
 
-ALTER TABLE moped_proj_files ADD COLUMN updated_at TIMESTAMP WITH TIME ZONE;
+ALTER TABLE moped_project_files ADD COLUMN updated_at TIMESTAMP WITH TIME ZONE;
 
-ALTER TABLE moped_proj_files
+ALTER TABLE moped_project_files
 ADD COLUMN updated_by_user_id INTEGER
 CONSTRAINT updated_by_user_fkey REFERENCES moped_users (user_id);
 
 
-COMMENT ON COLUMN moped_proj_files.updated_at IS 'Timestamp when the record was last updated';
-COMMENT ON COLUMN moped_proj_files.updated_by_user_id IS 'ID of the user who last updated the record';
+COMMENT ON COLUMN moped_project_files.updated_at IS 'Timestamp when the record was last updated';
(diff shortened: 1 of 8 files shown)


moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -264,7 +263,7 @@ const ProjectFiles = (props) => {
           id="file_description"
           name="file_description"
           value={props?.value ?? ""}
-          onChange={(e) => props.onChange(e.target.value.trim())}
```
chiaberry, line 267: this was preventing users from being able to include spaces in their descriptions

moped-editor/src/components/FileUpload/FileUploadDialogSingle.js
```diff
@@ -204,6 +205,7 @@ const FileUploadDialogSingle = (props) => {
                 <MenuItem value={3}>Estimates</MenuItem>
                 <MenuItem value={4}>Other</MenuItem>
               </Select>
+              <FormHelperText>Required</FormHelperText>
```
chiaberry, line 208: during testing I was confused by why i couldnt save, and its because the file type is required. 

I only added it here, but the whole form could probably use some ux love cc: @amenity

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -186,7 +186,7 @@ const ProjectsListViewTable = () => {
         <Paper className={classes.paper}>
           <Box
             sx={{
-              height: "70vh",
+              height: "75vh",
```
chiaberry, line 189: unrelated to the audit fields, making the project list view table a lil bigger

moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
```diff
@@ -0,0 +1,20 @@
...
+COMMENT ON COLUMN moped_project_files.updated_at IS 'Timestamp when the record was last updated';
+COMMENT ON COLUMN moped_project_files.updated_by_user_id IS 'ID of the user who last updated the record';
+
+CREATE TRIGGER update_moped_project_files_and_project_audit_fields
+BEFORE INSERT
+OR UPDATE ON moped_project_files
+FOR EACH ROW
+EXECUTE FUNCTION public.update_self_and_project_updated_audit_fields();
+
+COMMENT ON TRIGGER update_moped_project_files_and_project_audit_fields ON moped_project_files IS 'Trigger to execute the update_self_and_project_updated_audit_fields function before each update operation on the moped_proj_phases table.';
```
mddilley, line 20: this comment mentions project phases instead of files. copypasta got both of us today! 😄

mddilley (commented):
All of this works great! 🙌 Tested locally, and I'm ready to approve but have one question. I noticed there is still mention of `create_date` and `created_by` in the config in `ProjectActivityLogTableMaps.js`. I remember that we made a decision about this but I don't remember if we are keeping deprecated columns around or not.

chiaberry:
@mddilley we should keep deprecated columns in the activity log table maps so that older activities that reference old column names still can work

mddilley (approved):

frankhereford (approved):
This worked great, and the code looks sharp! Thank you @chiaberry!!

moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE moped_project_files RENAME COLUMN create_date TO created_at;
+ALTER TABLE moped_project_files RENAME COLUMN created_by TO created_by_user_id;
+
+ALTER TABLE moped_project_files ADD COLUMN updated_at TIMESTAMP WITH TIME ZONE;
```
tillyw, line 4: i suggest adding `set default now()` here

tillyw (approved):
this works great! there is one thing i think i should point out! in the instances in my previous pr, the `updated_at` field is set to have a default timestamp of `now()` so when you create a new record the `created_at` and `updated_at` timestamps match. in this case the `updated_at` field shows a value of `null` upon creation of a record. 

Screenshot 2024-03-28 at 4 51 05 PM

this is fine but it is slightly different from the existing pattern.

moped-database/migrations/1709753345334_audit_fields_proj_files/up.sql
tillyw: or i guess if we prefer to have the default `updated_at` value for a new record be `null` then we should drop those defaults from the other tables. either way would make sense to me!

chiaberry: thank you! I want to keep us in the same pattern, so I will add default now here

PR #1299, opened 2024-03-21
Add ability to select related components on the map

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/16334
https://github.com/cityofaustin/atd-data-tech/issues/16332

## Testing
**URL to test:** 

https://deploy-preview-1299--atd-moped-main.netlify.app/moped/projects/1998?tab=map

**Steps to test:**

Navigate to a project that has related components mapped, such as a project that is part of a parent project, or a project that has subprojects. 

From the components list: 
Select one of the projects components, see that the other components on the map now have gray iconography. 
Now select one of the related components, it should behave the same way as the project components. 
The component id of the component you selected should also be in the url.

Repeat the testing by selecting components from the map. Same as before, the other components should go gray, and the component id should be in the url. 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/RelatedComponentsList.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/RelatedComponentsList.js
@@ -8,7 +8,7 @@ const RelatedComponentsList = ({
   editState,
   shouldShowRelatedProjects,
   clickedComponent,
-  setClickedComponent,
+  makeClickedComponentUpdates,
   onClickZoomToComponent,
   allRelatedComponents,
   setIsClickedComponentRelated,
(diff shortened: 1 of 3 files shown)


johnclary (approved):
yay, it works great! 

one minor quirk is when you use a hyperlink to a subproject a component, like [this](https://deploy-preview-1299--atd-moped-main.netlify.app/moped/projects/2000?tab=map&project_component_id=2470).

As you can see in the image, the component is zoomed in correctly, but rendered in blue instead of green:

Screenshot 2024-03-22 at 10 01 04 AM

but the good news is that this is an improvement over production. in prod, you get an error message that the component cannot be found. [Here's an example](https://mobility.austin.gov/moped/projects/3019?tab=map&project_component_id=595).

so this seems like a spinoff issue to me? as another aside, i think that testing these hyperlinks is probably missing from the release QA checklist.

anyway—thanks for making this a lot better, Chia!

roseeichelmann (approved):
yay test steps checked out :heavy_check_mark: :sparkles:

chiaberry:
I pushed up a change to check if the component is related or not, so its highlighted with the right color

johnclary:
> I pushed up a change to check if the component is related or not, so its highlighted with the right color

ooh la la very nice! 🙌

PR #1302, opened 2024-03-23
Add secondary project name to project summary

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/14558.

## Testing

Local testing for this one; there is a new field added to the projects table.

### Testing Process

* Fire up the database, optionally grabbing a production set of data
* Apply migrations and metadata
* Visit any project and edit the name:
  * By clicking directly on the name
  * By choosing "Rename" in the hamburger on the right
* Change a primary project name
  * See that you can not set that primary name to the empty string 
* Define a secondary name
  * See how that is rendered, and note that containing tags are not rendered when it's undefined (the default)
* Blank out a secondary name
* Mess with the viewport to see how things reflow
  * In view mode
  * In edit mode
* Observe the `Snackbar`, do you like the wording? 
---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -200,7 +199,7 @@ const ProjectView = () => {
   const [dialogState, setDialogState] = useState(null);
   const [anchorElement, setAnchorElement] = useState(null);
   const [snackbarState, setSnackbarState] = useState(DEFAULT_SNACKBAR_STATE);
-  const menuOpen = anchorElement ?? false;
+  const menuOpen = Boolean(anchorElement);
```
frankhereford, line 202: I encountered an error when anchorElement was being set to an element throwing an error, because it's later passed in as a prop called `open` that is required to be a boolean. This fixes the error, and TBH, I'm confused why this isn't throwing an error when you go into "edit mode" in production.

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -455,20 +454,16 @@ const ProjectView = () => {
...
-                              updatedCallback={handleNameUpdate} // FLH not sure if this is needed
+                              updatedCallback={handleNameUpdate}
+                              currentPhase={currentPhase}
                             />
-                            <Box>
-                              <ProjectStatusBadge
-                                phaseKey={currentPhase?.phase_key}
-                                phaseName={currentPhase?.phase_name}
-                              />
-                            </Box>
```
frankhereford, line 471: This is pulled into the `ProjectNameComplex` component so that I can continue to have it float to the left up against the project name and number.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford, line 1: It feels a little out-of-style with the application to have multiple input fields in a single component, but in this situation, the multiple input fields interact in a complex way, acting as a single form and sharing submission/cancel buttons. 

I think splitting them out as components to use as children to this one will create unneeded complexity as state is passed up and down while the user goes in and out of edit/view mode. It feels like adding a lot of boilerplate at the expense of having the related logic bundled into one place.

All that said -- that's just how I reasoned with it -- I am **_very_** open to suggestions on how I can better structure this, both as a React component but also as a part of the Moped application.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+    "font-weight": "bold",
+    "padding-top": "0.95rem",
+  },
+}));
+
+const ProjectNameComplex = (props) => {
+  // Incoming props:
+    // projectData={data.moped_project[0]} ✅
+    // projectId={projectId} ✅
+    // editable={true} ❓
```
frankhereford, line 29: I never really found this being used in our previous `ProjectNameEditable` component, but it's passed in as a prop. 🤷

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+    "font-weight": "bold",
+  },
+  projectSecondaryNameEditField: {
+    "font-size": "1.0rem",
+    "font-weight": "bold",
+    "padding-top": "0.95rem",
+  },
+}));
+
+const ProjectNameComplex = (props) => {
```
frankhereford, line 25: I'm very open to suggestions on a better name here. I think of a complex as a set of related things, but maybe that's just for buildings. I'm still thinking about it -- anything y'all would prefer?

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+            />
+          </Box>
+        </Grid>
+
+      </Grid>
+    </>
+  );
+
+  const commonJSX = (
+    <>
```
chiaberry, line 270: does this need the empty tag around it, if it only has one component in it?

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+        });
+        })
+        .finally(() => { // return to the view mode, clear the snackbar after 3 seconds, 
+                         // and alert the parent component of the change
+        props.setIsEditing(false);
+        setTimeout(() => setSnackbarState(DEFAULT_SNACKBAR_STATE), 3000);
+        props.updatedCallback();
+      });
+    }
+  };
```
chiaberry, line 93: can you rerun prettier on this file? the spacing threw me off on when the finally was happening

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
chiaberry: this feels like it must be a holdover from once upon a time when the project name was not editable inline there. i think it can be removed

mddilley (commented):
I like the new error states, and I think you might even be building a case for using React Hook Form here down the road! I left some suggestions on some changes, and I'm curious what you think.

chiaberry:
I believe @amenity mentioned in planning that she didnt like the secondary name being displayed under the primary project name, and instead preferred it also on the same line, but I neglected to note that on the issue, so I apologize for that (and please correct me if I got it backwards)

We also need to be able to assign a secondary project name when a user creates a new project and not only when editing a project name.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+  const [updateProjectNames] = useMutation(UPDATE_PROJECT_NAMES_QUERY);
+
+
+
+  // this function is fired when the user clicks on the check mark to save the names
+  const handleAcceptClick = (e) => {
+    e.preventDefault();
+
+    const projectName = document.getElementById("project_name").value;
+    const secondaryName = document.getElementById("secondary_name").value;
```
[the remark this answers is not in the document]
chiaberry: Thanks for the comment Mike, this stuck out to me, but I was having trouble articulating why I'd prefer using state over the selectors.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: Absolutely - thank you for asking. I just got that installed on my linux machine, which I should have done 1st thing. https://github.com/cityofaustin/atd-moped/pull/1302/commits/e3eab73343446801fade17868ce9083bbd46f509

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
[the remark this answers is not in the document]
frankhereford: I'm so glad y'all said this - thank you for the hint on how to do this right. I felt like I was out there somehow, but I wasn't sure at all where I was going awry.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: No, they can go for sure. I think I have just gotten in the habit of adding them whenever I start defining JSX, but you're 100% right. https://github.com/cityofaustin/atd-moped/pull/1302/commits/85a54b2f7249d2c8f521b6c900c8ba2043e87924

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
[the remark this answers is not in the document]
frankhereford: Totally, now that I'm seeing this, this all makes much more sense. Thanks for this really helpful suggestion. https://github.com/cityofaustin/atd-moped/pull/1302/commits/6829ae511133d647df6f8ad7b5dde8550daea596

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  // this is fired when the snackbar's close button is clicked or by timeout
+  const handleSnackbarClose = (event, reason) => {
+    setSnackbarState(DEFAULT_SNACKBAR_STATE);
+  };
+
+  // Because the layout is non-trivially different between view and edit modes,
+  // we're going to use two different JSX blocks to render the component's content.
+  const viewModeJSX = (
```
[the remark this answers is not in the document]
frankhereford: I hear what you're saying, and some semblance of the approach you're talking about is what I started to do initially but started to run into complications.

I'll try to describe some more detail about my rational: the way our old component worked is that it big blocks of conditions in its JSX flipping between edit mode and view mode, with both modes in one component, so that made the conditional JSX pattern feel in-bounds. Also, I think to break out the input fields here into components, I'm going to need to do the view and edit separately as two components, at least for the secondary name, because the larger layout of the name system visually for view / edit modes is different. I guess you could do it with a prop to control if it shows the edit mode or the view mode and to hide the other, and then use it twice in the page .. but that feels contorted and overly complicated for little benefit, I think ..  like you said, there's a million ways to slice it.

I got to talk to Chia about this briefly earlier, but I don't think I explained the nuance to why I chose to do it this way, or even explained it too clearly here, but not for lack of trying.

If you want to sync on this, I'd love to talk about approaches in handling this in advance of making a revision to this PR, but right now, I still kind of think this slightly monolithic setup is OK given the issue's requirements. But, I do, 100%, recognize what you're saying that this is not ideal. The point you made about how the name I picked indicated that this is too complex and that there are better ways to break this up was verrrry well said. :100: I just don't see the subtleties yet and need some help.

frankhereford:
@mddilley wrote:

> I like the new error states, and I think you might even be building a case for using React Hook Form here down the road! I left some suggestions on some changes, and I'm curious what you think.

Yea, I totally need to get familiar with React Hook Form, it sounds like a really useful library that should be in everyone's tool belt. Thanks for reminding me about it, and I am going to do read some docs on it real soon. :hook: 

Thank you for your always thoughtful suggestions - I believe I have addressed them with commits or comments. I'm always eager to hear what you think - thanks for the fast review on this!

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -1,221 +0,0 @@
...
-  };
-
-  const handleSnackbarClose = () => {
-    setSnackbarState(DEFAULT_SNACKBAR_STATE);
-  };
-
-  return (
-    <Box className={isEditing ? classes.boxField : null}>
-      {isEditing && (
-        <form onSubmit={(e) => handleAcceptClick(e)}>
```
chiaberry, line 141: we can sync about this tomorrow, I am going to try to compare this component to the way you have it set up now. Because as im looking at them now, they are pretty similar except for the extra text field, and removing the form tag

chiaberry: ah, one thing the form tag offered, was being able to submit by hitting enter. I wonder if that functionality should stick around with the secondary name too.

frankhereford: Thanks Chia! I definitely got to a point in this work where I was trying to reuse as much code as I could and started running into complications. At that point, I did change my approach, and began again, this time building from the ground up while referencing the old code. It makes sense to me that I may have introduced subtle changes, and I'll be happy to eliminate any regressions that I've introduced. Thank you - anytime you want to sync is great. :pray:

frankhereford:
@chiaberry wrote:

> I believe @amenity mentioned in planning that she didnt like the secondary name being displayed under the primary project name, and instead preferred it also on the same line, but I neglected to note that on the issue, so I apologize for that (and please correct me if I got it backwards)

Awesome, that's totally cool. If there is a better way we can lay this out, just let me know, and I'll be happy to adjust the PR.


> We also need to be able to assign a secondary project name when a user creates a new project and not only when editing a project name.

:pray: Thank you so much for catching this enormous oversight on my part! 

I have [added the new field](https://github.com/cityofaustin/atd-moped/pull/1302/commits/55c4fc480826e5dc89436ff56f841a559ace75bd) and also [removed the `standard` variant from the components](https://github.com/cityofaustin/atd-moped/pull/1302/commits/8d69896532114f8bc7d7d723538094749149e886). The screenshot in the issue looks different than what the form looks like now for the `<textarea>`, but I looked at the diff, and I don't think I regressed anything, or at least, I can't see how. Was this form revised visually recently?

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
```diff
@@ -25,20 +25,50 @@ const DefineProjectForm = ({
...
+                type="text"
+                fullWidth
+                value={projectDetails.project_description}
+                error={descriptionError}
+                helperText="Required"
+                InputLabelProps={{ required: false }}
+                onChange={(e) =>
+                  handleFieldChange(e.target.value, e.target.name)
+                }
+              />
```
chiaberry, line 70: these need to be pulled out of this conditional, so that if a project is from a signal asset, we can still include a secondary name and a description

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,307 @@
...
+    <>
```
chiaberry, line 137: I dont think you need this react fragment tag

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,307 @@
...
+              </Typography>
+            </Box>
+          </Grid>
+        ) : null}
+      </Grid>
+    </>
+  );
+
+  const editModeJSX = (
+    <>
```
chiaberry, line 199: same for this fragment tag

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,307 @@
...
+        {/* Secondary project name field */}
+        <Grid item xs={12} sm={3} sx={{ paddingRight: "30px" }}>
+          <TextField
+            variant="standard"
+            fullWidth
+            id="secondary_name"
+            label={"Secondary Name"}
+            type="text"
+            value={secondaryName}
+            placeholder={"Project Byline"}
```
chiaberry, line 240: I'm not sold on the "Project Byline" placeholder, I think it would confuse people in thinking it should be the author of the project

chiaberry (changes requested):
I can't add a project using a signal asset, I believe it is because the description is required, but I cannot include one the way its laid out now. 

Also I do like being able to hit enter and submit the project title (you can test this in staging), can you bring that functionality back?

amenity:
@frankhereford @chiaberry - yep! The design in the issue description represents the final specs. FWIW, I always update the issue description itself so that it has the correct specs when I put an issue on deck. But I should also add a comment at the bottom to confirm that other design iterations shown in the comments aren't to be followed. Sorry for the confusion!

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
frankhereford: Absolutely, thank for catching that. I'm going push up a fix for this this afternoon.

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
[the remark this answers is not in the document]
mddilley: Cool, I hear what you are saying too. I'd love to pair on this some today to flesh out what I'm saying more if you have time. Might be easier to get what I'm trying to say across in real time!

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: Sounds great, I've removed both of these empty tag pairs. Thanks! https://github.com/cityofaustin/atd-moped/pull/1302/commits/fee13475f278b7a5f914281d9e99a0ef0bd21165

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: I think this is a great observation, and I changed the `placeholder` value to be "Secondary Name"

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
frankhereford: Thanks again! I put this back together the correct way and also took the opportunity to do some more work to make it match the mockups closer -- this was done by having the `description` field be 5 lines tall at a minimum.

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
frankhereford: I got that 'submit-on-enter' functionality restored. It's very strange - I had to wrap each input field's component in a `<form>` field individually, instead of wrapping the set of fields in a single enclosing form. Maybe this is a react thing?

At least it's working!

chiaberry:
Going through the updates, its looking good so far
> could you also remove the standard variant from the signal input so it matches?


Screenshot 2024-03-27 at 3 39 09 PM

frankhereford:
@chiaberry wrote:

>> could you also remove the standard variant from the signal input so it matches?

Thank you @chiaberry for helping me here - I didn't realize that the Autocomplete component itself is fed a component that it uses as its input.

https://github.com/cityofaustin/atd-moped/pull/1302/commits/1e4484c07a0ca925109619ba225cfc4713cae55a

amenity (changes requested):
This works perfectly, @frankhereford! 😊 👏🏻  

I would like a few style adjustments: 
- Specify the **`standard` variant** of the [MUI text field](https://mui.com/material-ui/react-text-field/#basic-textfield) on both the "New project" page and the Project page.
- Apply **sentence case capitalization** to the field names on the Project page: Update the current field to "Project name" and use "Secondary name" for the new field.  
- Interacting with these fields on the Project page, I think the slightly-smaller font sizing that we had in the specs feels off. 😬 Would it be possible to **make the font size in both fields the same, `1.4rem`**? This is a tad smaller than we have in Production to make space for the additional UI clutter we're adding. 
- Along those lines, I realized that the ✔️ ✖️ are really big right now. Would it be possible to adjust those to `font-size: 1.4rem; margin-top: 32px`?

3 - Proposed implementation

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: Hey Mike,

Thanks again for all your guidance on this Component. I have pulled out the [commonJSX, editModeJSX and viewModeJSX into proper components](https://github.com/cityofaustin/atd-moped/blob/65b04759ea0db073c9942f3f06f584d8c57257e2/moped-editor/src/views/projects/projectView/ProjectNameComplex.js#L168-L320). Do you think I should put them in their own files or leave them in the same file as the main ProjectNameComplex component. I don't think we have an eye to reuse them, so I thought I would ask what you think is best and follow that direction. Thanks!!

mddilley: @frankhereford sorry, i missed this in my notifications! We don't have a super strong pattern here, but I was looking at the project phase components called `ProjectPhases`, `ProjectPhaseForm`, etc and feel like something like `ProjectNameEditable`, `ProjectNameEditableForm`, and `ProjectName` could work if these are split into separate files.

Also, for the common JSX, that snackbar feels like a pretty good fit to stay put and not be its own component to me since it is small amount of code. Total judgement call, though!

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
```diff
@@ -0,0 +1,320 @@
...
+  handleAcceptClick,
+  handleCancelClick,
+  classes,
+  props,
+}) => (
+  <Grid container>
+    {/* Primary project name field */}
+    <Grid item xs={12} sm={6} sx={{ paddingRight: "30px" }}>
+      <form onSubmit={(e) => handleAcceptClick(e)}>
+        <TextField
```
johnclary, line 196: @frankhereford would you add the `autoFocus` prop to to this so that we don't lose that behavior?

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: Mike, these are great suggestions!! Thank you so much for all of your help. I've adopted each of these. 🚀 https://github.com/cityofaustin/atd-moped/pull/1302/commits/180f5502c46bd0205e96b84576de7675995497ca

moped-editor/src/views/projects/projectView/ProjectNameComplex.js
frankhereford: 100%, great catch. Thank you John. https://github.com/cityofaustin/atd-moped/pull/1302/commits/14716e22dcceb44a58e0c781fd2139f283a40e08

mddilley (changes requested):
Frank, I wasn't sure if this was ready for review again, but I made another pass through it. I mentioned you in one requested changes and made some other suggestions.

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -1,209 +1,150 @@
...
 
+  // state to hold values for controlled inputs
+  const [projectName, setProjectName] = useState("");
+  const [secondaryName, setSecondaryName] = useState("");
+
+  // update the state variables when the props change
+  useEffect(() => {
+    setProjectName(props.projectData.project_name || "");
+    setSecondaryName(props.projectData.project_name_secondary || "");
+  }, [props.projectData]);
```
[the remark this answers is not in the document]
mddilley: chatted with Frank and I was wrong about this one. I jumped to thinking about this state already being moved down into `ProjectNameForm` and being able to count on the `isEditing` toggle causing the mount/unmount to re-init these pieces of state from the props.

moped-editor/src/views/projects/projectView/ProjectView.js
chiaberry: I agree with Mike, on both counts. It would be nice to refactor it so that we can keep the badge separate from the name. But could be a spin off issue.

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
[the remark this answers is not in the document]
frankhereford: Mike, I got all the state and logic and whatnot pulled down into the `<ProjectNameForm/>` component, and just like you had originally suggested, the `useEffect` came right out of there. Thanks!

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -1,209 +1,150 @@
...
+    "font-size": "1.4rem",
+    "font-weight": "bold",
   },
 }));
 
 const ProjectNameEditable = (props) => {
   const classes = useStyles();
 
+  // state to hold values for controlled inputs
+  const [projectName, setProjectName] = useState("");
```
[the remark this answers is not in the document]
frankhereford: 100% yes - this came down with the logic while I was chasing that useEffect out of town. Thanks!

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -1,209 +1,150 @@
...
-  editIcon: {
-    cursor: "pointer",
-    marginLeft: "8px",
-  },
-  editIconConfirm: {
+  editIcons: {
     cursor: "pointer",
-    marginTop: "16px",
-    fontSize: "2rem",
+    marginTop: "32px",
```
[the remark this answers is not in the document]
frankhereford: Flexbox worked great - thanks Mike! Also, classes was pulled down into the `ProjectNameForm` component as well.

moped-editor/src/views/projects/projectView/ProjectNameForm.js
```diff
@@ -0,0 +1,98 @@
...
+  secondaryName,
+  handleSecondaryNameChange,
+  handleAcceptClick,
+  handleCancelClick,
+  classes,
+  props,
+}) => (
+  <Grid container>
+    {/* Primary project name field */}
+    <Grid item xs={12} sm={6} sx={{ paddingRight: "30px" }}>
```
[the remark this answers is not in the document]
frankhereford: Thanks - great idea. I have leaned into the spacing prop instead of manual CSS. Thanks Mike.

moped-editor/src/views/projects/projectView/ProjectNameForm.js
```diff
@@ -0,0 +1,211 @@
...
+      {/* Primary project name field */}
+      <Grid item xs={12} sm={6}>
+        <form onSubmit={(e) => handleAcceptClick(e)}>
+          <TextField
+            required
+            autoFocus
+            variant="standard"
+            fullWidth
+            id="project_name"
+            label={"Project Name"}
```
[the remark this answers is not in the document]
frankhereford: Thank you - Amenity had totally specified that in her latest request for changes. 🙏 🙏🙏🙏

chiaberry (approved):
This looks great! I left a note about the input labels, the mockup led you astray :( 

I think users are going to be really happy to have this feature added

moped-editor/src/views/projects/projectView/ProjectNameForm.js
```diff
@@ -0,0 +1,211 @@
...
+          onClick={(e) => handleCancelClick(e)}
+        >
+          close
+        </Icon>
+      </Grid>
+
+      {/* The status badge. Here, we're going to jog it down a bit to make it visually centered
+      along the horizontal midline of the project name input field. */}
+      <Grid item xs={12} md={2}>
+        <Box sx={{ display: "inline", position: "relative", top: "1rem" }}>
```
mddilley, line 200: Since the parent `<Grid item>` here uses flexbox, you could probably get away removing this Box and using `alignContent: end` on the grid item if the intention is to push the status badge down so it aligns with the bottom of the inputs.

amenity (approved):
@frankhereford - looks & works beautifully! Thank you, ty. 🚢 🏆

mddilley (approved):
I tested creating every combination of creating projects with secondary names and also updating in the summary view, and it is working just like described in the test steps. Thanks for all of the work on this one! 🚢 🚀 🙌

frankhereford:
@amenity wrote:

> looks & works beautifully! Thank you, ty. 🚢 🏆

Thank you!! I really appreciate all your patience with me while this got worked on and for all your incredibly helpful feedback!

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -1,209 +1,40 @@
...
     message: null,
     severity: "success",
   };
 
-  const initialProjectName = props?.projectName;
-  const [projectName, setProjectName] = useState(initialProjectName);
-  const [projectNameBeforeEdit, setProjectNameBeforeEdit] =
-    useState(projectName);
-  const [isEditing, setIsEditing] = useState(props?.isEditing ?? false);
+  // used to control the snackbar's contents and visibility
```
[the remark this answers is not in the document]
mddilley: ohhhh "used to" as in "is used to" not "used to" like in the past - sorry my reading is not on point today! 😅

moped-editor/src/views/projects/projectView/ProjectNameForm.js
```diff
@@ -0,0 +1,211 @@
...
+        <form onSubmit={(e) => handleAcceptClick(e)}>
+          <TextField
+            required
+            autoFocus
+            variant="standard"
+            fullWidth
+            id="project_name"
+            label={"Project name"}
+            type="text"
+            value={projectName}
```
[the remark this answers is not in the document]
frankhereford: 100% great call - https://github.com/cityofaustin/atd-moped/pull/1302/commits/cf757c5cf3e041dd793069dcdc080425c42947db

moped-editor/src/views/projects/projectView/ProjectView.js
chiaberry: https://github.com/cityofaustin/atd-data-tech/issues/16776

moped-editor/src/views/projects/projectView/ProjectStatusBadge.js
```diff
@@ -162,7 +161,7 @@ const useChipStyles = makeStyles((theme) => ({
       getStyle(theme, phaseKey ?? "").background,
   },
   leftMargin: {
-    marginLeft: "1rem",
+    // marginLeft: "1rem",
```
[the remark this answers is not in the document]
frankhereford: 🧹🧹 https://github.com/cityofaustin/atd-moped/pull/1302/commits/b7d66bba4a1f26b54c80fac7b9318734e3a1d5cb

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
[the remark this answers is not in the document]
frankhereford: I believe it is; it's passed down into the Snackbar component. If I'm missing something, let's open up an issue and I'll adjust it, but in the interest of the release tomorrow and this passing otherwise, I'm going to leave this here. Thank you!

moped-editor/src/views/projects/projectView/ProjectNameForm.js
frankhereford: I appreciate your feedback on using MUI better; I'm so green at it. 💯 https://github.com/cityofaustin/atd-moped/pull/1302/commits/400416a464148c546dec209e3946881de930efb1

moped-editor/src/views/projects/projectView/ProjectNameEditable.js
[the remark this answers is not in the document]
mddilley: cool - thanks for taking a look 🙏

PR #1303, opened 2024-03-25
Testing the pro version

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/16860

I initially opened this PR to test what the pro version was like before we bought it. But now we have it. 

## Testing

https://deploy-preview-1303--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

you can pin columns to the left or the right, you can resize columns.

You can view 1000 rows per page.

Try to break the table 

If you want to test locally, check out and `npm install`.
create an .env file in the moped-editor folder with the entry `REACT_APP_MUIX_LICENSE_KEY=`, the key is in 1password under MUI. 
 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -20,7 +20,7 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
   noResultsMessage: "No projects found.",
   showPagination: true,
   pagination: {
-    rowsPerPageOptions: [25, 50, 100], // we are limited to max 100 on the data grid community plan
+    rowsPerPageOptions: [100, 250, 1000],
```
chiaberry, line 23: setting this back to what we used to have

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
mddilley, line 1: You had mentioned the grey background in dev sync when pinning the columns to the left or right, and I saw that the header row also has it. I do think that a white background [like in their demo](https://mui.com/x/react-data-grid/demo/) looks nicer and also a lot like what we used to have with the frozen project ID column. 

I wonder if that color is inherited from something in our theme file because it is behaving like the default? Not sure.

Screenshot 2024-04-24 at 10 50 12 AM

mddilley (changes requested):
All this looks great, Chia! So cool to the pinning in action and the pagination limit uncapped. I also tested the license locally, and it works perfectly. MUG is gonna love this upgrade! 🙌

I noticed one small regression - the density toggle isn't having an effect on the table rows anymore.

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: oh good call about the demo -- ill look for how come this is gray when the demo isnt

chiaberry:
I'll remove the density toggle, I dont think anyone wants it anyway

amenity:
Sweet! All works great for me. 

Question, though: I think the grey background is actually kinda noisy, and the reduced contrast reduces legibility, too. What do you think about tweaking these styles so the table colors stay the same?  
- `DataGrid-containerBackground: #fff`
- `DataGrid-pinnedBackground: #fff`

Current Moped - sans Pro:
Screenshot 2024-04-25 at 12 14 02 AM


DataGridPro default:
Data Grid Pro default

Proposed: 
Proposed

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
chiaberry: its from our theme! https://mui.com/x/react-data-grid/style/#theme-header-and-pinned-sections
But when I try to follow these instructions, it wont work. ill keep tinkering

amenity (approved):
Perfection, thanks! ✨ I really appreciate all your research (for a lonnng time now) into how-do-we-progress w/ MUI, @chiaberry 🕵🏻‍♀️ 🙏🏻

mddilley (approved):
Looks awesome with the updates! 🚢 🙌

roseeichelmann (approved):
such a cool feature with the pinning columns!! looks great :)

PR #1308, opened 2024-03-29
15855 handle activity log when parent deleted

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15855

This bug appears to have been related to the way Javascript handles null objects, which is interesting! This fix should hopefully prevent this from popping up again.

## Testing
**URL to test:** 

[deploy preview](https://app.netlify.com/sites/atd-moped-main/deploys/662958dba988a100088cbc35)

**Steps to test:**
- go to a project and add a subproject to it
- remove the subproject
- navigate to the subproject you just removed
- the activity log should say a parent project was removed with the id of the parent project
- try adding/removing other fields from a project and checking the removal statement to make sure it makes sense
- only fields with lookup values should display lookup values, number and id fields should display raw data

here are some examples!
Screenshot 2024-04-30 at 7 59 34 PM

---
#### Ship list
- [x] Code reviewed 


tillyw:
I noticed in testing this PR that the parent project's activity log doesn't handle the user either adding or removing a subproject (e.g., <User> added subproject <id>). So you can see the activity on the subproject but not on the parent project. If we don't have an issue for this would it be worth adding one?

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -108,20 +96,37 @@ export const formatProjectActivity = (change, lookupList) => {
...
-        changeText: [
-          {
-            text: `Changed ${entryMap.fields[changedField]?.label} `,
-            style: null,
-          },
-        ],
-      };
+      // if the new field is null, it is because something was removed
+      if (lookupList[changeData.new[changedField]] == null) {
+        console.log(changeData.old);
```
mddilley, line 104: Could you remove this console log before this gets merged? It was helpful when I tested locally though! 🙏

mddilley (commented):
I checked this out, and the fix is working really great! I've been looking at this for a while, and I think that we may need to keep both checks for `null`. I caught [this bug](https://github.com/cityofaustin/atd-data-tech/issues/16964) which is so closely related to this and introduces a situation where the type of `typeof changeData.new[changedField]` is `number`. 

Let me know if this makes sense, and I'll regroup take another look at this tomorrow. 🙏 😅

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
tillyw: yes, thanks for catching that! i always forget to delete my logs

tillyw:
> I checked this out, and the fix is working really great! I've been looking at this for a while, and I think that we may need to keep both checks for `null`. I caught [this bug](https://github.com/cityofaustin/atd-data-tech/issues/16964) which is so closely related to this and introduces a situation where the type of `typeof changeData.new[changedField]` is `number`.
> 
> Let me know if this makes sense, and I'll regroup take another look at this tomorrow. 🙏 😅

you're totally right! thanks for thinking through that. i'll add the additional checks and re-request review,

mddilley:
@tillyw this morning, i [merged a fix](https://github.com/cityofaustin/atd-moped/pull/1327) to that bug that I mentioned earlier so I hope that makes it easier to test all of the project activities when this branch catches up to `main`. 🙏

tillyw:
I believe I encountered an unrelated bug when testing this. I got this Apollo error when trying to clear our the 'Public Process' field. All other fields worked fine. Has anyone encountered this? 
Screenshot 2024-04-30 at 8 04 55 PM

chiaberry:
@tillyw thanks for pointing out that bug, i made that field nullable in the branch im working on https://github.com/cityofaustin/atd-moped/tree/17027-project-view-snackoos

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -108,20 +119,24 @@ export const formatProjectActivity = (change, lookupList) => {
...
-        changeIcon,
-        changeText: [
-          {
-            text: `Changed ${entryMap.fields[changedField]?.label} `,
-            style: null,
-          },
-        ],
-      };
+      // if the new field is null, it is because something was removed
+      if (lookupList[changeData.new[changedField]] == null) {
```
roseeichelmann, line 126: i think its preferred to use === where possible in javascript

roseeichelmann (approved):
this looks and works great! just have one comment about the equality operator. also, just curious, do we have any plans to add an activity log to the parent project that says when new subprojects are added?

mddilley (approved):
Looks great, Tilly! Thanks for fixing this bug and surfacing another one in the process! 🙏 🙌 🚢

mddilley:
@chiaberry @amenity Do we have an issue for the bug that Tilly described above? If not, it would be the exact same quick fix as https://github.com/cityofaustin/atd-moped/pull/1327.

chiaberry:
@mddilley we dont have an issue for it, but I squeezed it in this PR https://github.com/cityofaustin/atd-moped/pull/1337

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
chiaberry: yup, fix before merging, nice catch Rose.

amenity (approved):
All works well for me, thanks @tillyw. 

Not introduced in this PR and almost too minute to point out: Updating the role for a team member shows as "updated team member _notes_" in the activity log.

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
tillyw: ack embarrassing typo thanks for catching that

tillyw:
> I noticed in testing this PR that the parent project's activity log doesn't handle the user either adding or removing a subproject (e.g., (user) added subproject (id) ). So you can see the activity on the subproject but not on the parent project. If we don't have an issue for this would it be worth adding one?

following up on this question since @roseeichelmann asked it as well! this would definitely be a separate issue but i think worth pursuing - @chiaberry @amenity what do y'all think?

tillyw:
> All works well for me, thanks @tillyw.
> 
> Not introduced in this PR and almost too minute to point out: Updating the role for a team member shows as "updated team member _notes_" in the activity log.

thanks @amenity! @chiaberry made a separate issue for that and i am working on a fix

moped-editor/src/utils/activityLogFormatters/mopedProjectActivity.js
```diff
@@ -20,6 +20,25 @@ export const formatProjectActivity = (change, lookupList) => {
     };
   }
 
+  const getRemovedField = (changedField) => {
```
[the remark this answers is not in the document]
tillyw: thanks for pointing this out! will consider this in future activity log prs

chiaberry:
> > I noticed in testing this PR that the parent project's activity log doesn't handle the user either adding or removing a subproject (e.g., (user) added subproject (id) ). So you can see the activity on the subproject but not on the parent project. If we don't have an issue for this would it be worth adding one?
> 
> following up on this question since @roseeichelmann asked it as well! this would definitely be a separate issue but i think worth pursuing - @chiaberry @amenity what do y'all think?

@tillyw I think it would be worth investigating, iirc we didn't do it from the beginning because of limitations with how the activity log works, but we should revisit it, maybe theres a way to handle it with triggers

PR #1312, opened 2024-04-10
Add "Secondary name" column to Projects page

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16521

## Testing
**URL to test:** 

Local

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectNameEditable.js
```diff
@@ -65,7 +65,7 @@ const ProjectNameEditable = (props) => {
         variables: {
           projectId: props.projectId,
           projectName: projectName,
-          projectNameSecondary: secondaryName,
+          projectNameSecondary: secondaryName.length > 0 ? secondaryName : null,
```
mddilley, line 68: Made this small change so we null out `secondaryName` in the DB if it is removed instead of storing an empty string. The generated column counts on `null` to include or exclude the secondary name part of the full name, and this follows how we treat other text inputs for the most part.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -36,6 +36,20 @@ export const OPERATORS_WITHOUT_SEARCH_VALUES = [
  */
 export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
   fields: [
+    {
+      name: "project_name_full",
+      label: "Full name",
+      placeholder: "Enter full project name",
```
mddilley, line 42: Unrelated but i don't think we use this placeholder value anywhere. Might be nice to create a tech debt snackoo to look into this.

moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -50,13 +50,27 @@ export const PROJECT_LIST_VIEW_QUERY_CONFIG = {
...
+    project_name_full: {
       searchable: true,
       sortable: true,
       link: "project_id",
-      label: "Project name",
+      label: "Full name",
       defaultHidden: false,
       showInTable: true,
+      search: {
+        label: "Search by full name",
```
mddilley, line 61: Same here - not sure this is used anywhere. 🤔

moped-database/migrations/1712697601392_add_full_project_name_column/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE moped_project
+ADD project_name_full text GENERATED ALWAYS AS (COALESCE(project_name || ' - ' || project_name_secondary, project_name)) STORED;
```
mddilley, line 2: I tried using `concat_ws` here which would have been nice, but I found that it is not considered immutable in Postgres so it isn't a fit for generated columns. [Thread here with some details](https://www.postgresql.org/message-id/CA%2Bu7OA6AD9OLziFW%2Bio1HgN8q_XH0o1Y5RyufXYO5%3D0fnhG5zQ%40mail.gmail.com)

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
chiaberry: I'll make a snackoo! this feels like a holdover from old gql abstract configs

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -172,10 +172,22 @@ export const useColumns = ({ hiddenColumns }) => {
...
+        headerName: "Name",
+        field: "project_name",
+        flex: 2,
+        minWidth: COLUMN_WIDTHS.xlarge,
+      },
+      {
+        headerName: "Secondary name",
+        field: "project_name_secondary",
+        flex: 2,
+        minWidth: COLUMN_WIDTHS.xlarge,
```
chiaberry, line 189: this feels too wide to me, but then again I don't know how long project secondary names will be....

chiaberry:
I am testing but can't get the sorting to work on secondary name, and I don't know why yet. I checked what's returned from the query, and the projects are in the right order, but projects with secondary names arent showing up in my table. I am going to try again. 

Also, I tested without deleting my local storage, to get an idea of what users will see, and I see the full name column and the name column. I think that's okay, people can toggle on and off.

mddilley:
@chiaberry I am seeing the same issue with sorting the secondary name. The API returns the correct order but the table isn't displaying them that way. Seems really familiar so I'll take a quick look and see if I can fix it.

mddilley:
I think it is the same thing that we were seeing [here](https://github.com/cityofaustin/atd-moped/pull/1275#discussion_r1505121721). When i scroll to the bottom of the results in the first page of the table, i can see the results with secondary names that should show above the others, but the results returning with `null` for secondary name are bubbling up in the table for some reason. Might be a MUI DataGrid sort config thing? 🤔

chiaberry:
@mddilley you are right! I think I found it, I neglected to put "sortingMode: server" in the config
https://mui.com/x/react-data-grid/sorting/#server-side-sorting

mddilley:
@chiaberry whoa!!! nice find! i was over here looking at custom sorting 😵‍💫, but that fixes it! I'll push up that change so we can test it alongside this PR.

amenity (approved):
👌🏻🚢  Looks sharp! Thanks, @mddilley

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
chiaberry: https://github.com/cityofaustin/atd-data-tech/issues/16783

chiaberry (approved):
Thanks! 🚢

PR #1318, opened 2024-04-15
Refine Projects page advanced filters for date fields 

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16639

## Testing
https://deploy-preview-1318--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

1. Test with the following date fields : Construction start, Completion date, Modified
2. Do you see the new labels, do they make sense?
3. See and test the new is blank option.
4. Test using the "on" operator, previously that wasnt working correctly bc we were searching timestampz with a date string. Now it should return expected dates

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/components/GridTable/FiltersCommonOperators.js
```diff
@@ -107,39 +107,46 @@ export const FiltersCommonOperators = {
...
     description: "The date is equal to.",
     envelope: null,
     type: "date",
   },
   date_greater_than: {
     operator: "_gt",
-    label: "greater than",
-    description: "The date is greater than.",
+    label: "after",
+    description: "The date is after.",
```
roseeichelmann, line 118: i couldnt figure out what these descriptions are used for, maybe just for us? in that case maybe i didnt need to update these descriptions idk

roseeichelmann:
Also I just noticed a bug with our date filtering that pre-dates this PR -- if you are filtering using "On" (or "Is" for current staging/prod), which should mean is equal to, its not actually capturing dates equal to. Example in the video below:

[simplescreenrecorder-2024-04-15_15.06.39.webm](https://github.com/cityofaustin/atd-moped/assets/67130963/5db0fe44-fcb2-4d36-90af-0a9b0a363470)

Should be capturing all the projects modified on 4/12/2024 but its returning 0 projects. It does return those projects if we select "Less than 4/13/2024" for example though.

mddilley (approved):
Looks great! The new wording reads better when talking about dates to me. 🚢  🚀 🙌

chiaberry:
Ah, it doesnt work because its querying for `where: {_and: [{updated_at: {_eq: \"2024-04-12\"}}]} )`, but all of our updated_at's are timestamptz `"2024-04-12T14:30:04.195292+00:00"` .

chiaberry (commented):
the wording makes so much more sense to me now. Thanks for surfacing that bug where dates do not equal timestamps, I'm checking to see if theres a way hasura suggests to handle that.

chiaberry:
@roseeichelmann go ahead and remove the option for "on" aka date equals, since its not working right now. then i will approve!!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -84,6 +85,15 @@ const makeAdvancedSearchWhereFilters = (filters) =>
...
         }
       }
+      // If we are filtering on a date equality we need to create a custom search string to handle the date/timestampz conversion
+      if (type === "date" && gqlOperator === "_eq") {
+        const nextDay = format(
+          addDays(parseISO(value.replaceAll('"', "")), 1),
+          "yyyy-MM-dd"
+        );
+        // Greater/equal to the selected day and less than the next day will return all timestampz for the given date
+        return `_and: [ { ${field}: { ${`_gte`}: ${value} } }, { ${field}: { ${`_lt`}: "${nextDay}" } } ]`;
```
roseeichelmann, line 95: in order to be able to search for all timestampz on a given date we need to create this greater than/less than query to use instead

johnclary: nice fix! this behavior was really surprising to me, since pg SQL will generously handle the casting for you when you do a query like `where modified_date = '2024-04-01'`.  cool that Chia was able to find an open Hasura issue for it 👍

johnclary:
Possibly related issue: https://github.com/cityofaustin/atd-data-tech/issues/9418#issuecomment-1979532412

chiaberry (approved):
im so glad this is going to be fixed in the next release!
🚢

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
mddilley: +1 - this is great! 👏 I also suspect that we will find ourselves in this part for https://github.com/cityofaustin/atd-data-tech/issues/9418. Really nice to be reminded where this happens. 🙏

mddilley (approved):
I retested everything, and the `on` operator is now working as expected. Thanks for taking that on in this PR too! 🙌 🙏 🚢

tillyw (changes requested):
the new wording is much more intuitive, the 'is blank' option is a great addition and thank you for fixing the 'on' option! 'after or on' works as expected for me but for some reason 'before or on' doesn't seem to include results that match the 'on' criteria.

query filter
Screenshot 2024-04-24 at 6 21 08 PM

results
Screenshot 2024-04-24 at 6 20 57 PM

missing records
Screenshot 2024-04-24 at 6 27 28 PM

amenity (approved):
Thanks! Beautiful & logical. ✨ 🧠 📆 🚢

frankhereford (approved):
This looks great, thank you Rose. 

@chiaberry & @amenity -- do you think we should add the negative version of the `is blank` operator, where you can search for projects that have a certain date defined? Something like a `is defined` condition?

chiaberry:
@frankhereford good question Frank, I think we should ask the MUG -- I pictured people searching for "is blank" in order to clean up their data and fill in missing dates. I can't off the top of my head think of why to search "is not blank", maybe to get a list of scheduled work?? but its a good thing to bring up

frankhereford:
@chiaberry wrote:

> I can't off the top of my head think of why to search "is not blank"

Yea - that is a smart way to think about it. 

I had only thought of adding it in terms of boolean-algebra completeness -- like, if we have an operator, shouldn't we have the negation of it. That's a very programmer-brain / naive tact, and I think your approach of thinking what the people using it actually will find useful vs. clutter is super valuable and the way to go. Thanks!!

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -84,7 +85,24 @@ const makeAdvancedSearchWhereFilters = (filters) =>
...
+        );
+        if (gqlOperator === "_eq") {
+          // Greater or equal to the selected day and less than the next day will return all timestampz for the given date
+          whereString = `_and: [ { ${field}: { ${`_gte`}: ${value} } }, { ${field}: { ${`_lt`}: ${nextDay} } } ]`;
+        } else if (gqlOperator === "_lte") {
+          // Less than the next day will give us all the timestampz that are less than or equal to the selected date
+          whereString = `${field}: { ${"_lt"}: ${nextDay} }`;
+        } else if (gqlOperator === "_gt") {
+          // Greater or equal to the next day will give us all the timestampz that are greater than the selected date
+          whereString = `${field}: { ${"_gte"}: ${nextDay} }`;
```
roseeichelmann, line 102: had to handle these other cases as well

roseeichelmann:
@tillyw okay i fixed the bug you found, i needed to handle the before or on case in a custom where string similar to the equal to case. there is still an issue with the "after" operator which is described in [this issue](https://github.com/cityofaustin/atd-data-tech/issues/9418) and will be fixed in a later PR, but this one is ready for re-review!

tillyw (approved):
works great!! thank you for reworking this feature it's so much better now!

chiaberry (approved):
thanks Tilly for the thorough testing and thank you Rose for fixing the filters that I just wanted to delete 😬

PR #1321, opened 2024-04-18
Update and add audit fields to moped proj tags

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16180

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Add some project tags to a project. 
2. Check the `moped_proj_tags` table and see that the new columns have been updated - `updated_at`, `created_at`, `updated_by_user_id`, `created_by_user_id`. 
3. Now check the `moped_project` table, you should see the same values in the audit fields for the project you just added tags to.
4. I dont think there is a way to edit tags right? You can just add or delete them?
5. Test the down migration

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -3612,6 +3612,9 @@
     - role: moped-admin
       permission:
         check: {}
+        set:
+          created_by_user_id: x-hasura-user-db-id
+          updated_by_user_id: x-hasura-user-db-id
         columns:
           - project_id
           - tag_id
@@ -3619,6 +3622,9 @@
     - role: moped-editor
       permission:
         check: {}
+        set:
+          created_by_user_id: x-hasura-user-db-id
+          updated_by_user_id: x-hasura-user-db-id
         columns:
           - project_id
           - tag_id
@@ -3627,26 +3633,38 @@
     - role: moped-admin
       permission:
         columns:
+          - is_deleted
           - id
           - project_id
           - tag_id
-          - is_deleted
+          - created_by_user_id
+          - updated_by_user_id
+          - created_at
+          - updated_at
         filter: {}
     - role: moped-editor
       permission:
         columns:
+          - is_deleted
           - id
           - project_id
           - tag_id
-          - is_deleted
+          - created_by_user_id
+          - updated_by_user_id
+          - created_at
+          - updated_at
         filter: {}
     - role: moped-viewer
       permission:
         columns:
+          - created_at
+          - created_by_user_id
           - id
+          - is_deleted
           - project_id
           - tag_id
-          - is_deleted
+          - updated_at
+          - updated_by_user_id
         filter: {}
   update_permissions:
     - role: moped-admin
@@ -3657,6 +3675,8 @@
           - is_deleted
         filter: {}
         check: {}
+        set:
+          updated_by_user_id: x-hasura-user-db-id
     - role: moped-editor
       permission:
         columns:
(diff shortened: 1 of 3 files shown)


moped-database/metadata/tables.yaml
johnclary, line 1: it looks like you'll also need to tweak the mutation in this table's [activity log trigger](https://github.com/cityofaustin/atd-moped/blob/ba3590822c29e9f040f167d15b48cadf4ce98aa0/moped-database/metadata/tables.yaml#L3692-L3707), since it is configured to update `moped_project`. your audit triggers are replacing that functionality 🧹

moped-database/migrations/1713466319581_moped_proj_tags_audit_fields/up.sql
```diff
@@ -0,0 +1,24 @@
+-- Adds audit columns and trigger to moped_proj_tags
+ALTER TABLE moped_proj_tags
+ADD COLUMN created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+ADD COLUMN created_by_user_id INT4 NULL,
+ADD COLUMN updated_by_user_id INT4 NULL,
```
johnclary, line 5: super minor, but i think it is conventional to declare `integer` instead of `int4`. even though they are effectively identical i think `integer` is SQL standard?

moped-database/metadata/tables.yaml
johnclary: also, i just want to observe that you have opted to not to track the `created_by_user_id` and `updated_by_use_id` relationships in Hasura—so those are not exposed through our graphql API. i don't think we've been in the habit of tracking those relationships at all but i did want mention it.

johnclary (changes requested):
looks great but for the issue with the event trigger mutation 🙏

moped-database/migrations/1713466319581_moped_proj_tags_audit_fields/up.sql
roseeichelmann: gotcha! tbh was copying one of mikes audit PRs for some of this code hehe. also wasnt sure about which columns should be nullable or not, does this look right?

moped-database/metadata/tables.yaml
roseeichelmann: i just tracked them!

mddilley: y'all gave me the idea to check if we forgot to remove the update to `moped_project` in other places. [Found one](https://github.com/cityofaustin/atd-data-tech/issues/16913) and i'll fix it today. 🙏

moped-database/migrations/1713466319581_moped_proj_tags_audit_fields/up.sql
johnclary: in an ideal world, we would backfill these columns so that we could make them `not null` going forward, but we've never discussed it. what you have looks good to me!

moped-database/metadata/tables.yaml
johnclary: nice catch!

moped-database/migrations/1713466319581_moped_proj_tags_audit_fields/up.sql
mddilley: ahhhh i created a copypasta cascade 🍝  fyi - i wondered if it originally stemmed from Hasura, but they also declare with `integer` when you add from the console so that further supports John's thoughts here. thanks for flagging this!

Also, nullable sounds good to me since we won't be able backfill those values for previous project tags.

johnclary (approved):
🚢

mddilley (approved):
I tested this locally with multiple users, and I'm seeing the audit fields captured just as expected. Migrations works great too. Thanks for updating! 🚀 🚢 🙌

PR #1323, opened 2024-04-22
Update tag styles to display inline

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16778

This PR updates the project tags in the summary view to show inline and wrap instead of as a single column list. I used the [MUI multiline chip example](https://mui.com/material-ui/react-chip/#multiline-chip) for chips with long text on smaller screens.

## Testing
**URL to test:** 

https://deploy-preview-1323--atd-moped-main.netlify.app/moped/projects/1909

**Steps to test:**
1. Go to the link above
2. Scroll down to the **Tags** section and:
   - See that the many tags are now displayed inline and not in a single column
   - Click the X icon to remove a tag in the middle of the list and see they adjust to fill the empty space
   - Add a tag and see that it adds to the list at the end
   - **Question**: Should we update the order that the tag chips are displayed? Alphabetical? We order the dropdown ascending by name.
   - Make the screen mobile size (somewhere around 400px width) and see that the text within the chips wraps to multiline if needed

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/TagsSection.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/TagsSection.js
@@ -18,8 +18,8 @@ import AddCircle from "@mui/icons-material/AddCircle";
 import Autocomplete from "@mui/material/Autocomplete";
 import ApolloErrorHandler from "../../../../components/ApolloErrorHandler";
 import DeleteConfirmationModal from "../DeleteConfirmationModal";
-
 import Grid from "@mui/material/Grid";
+import theme from "src/theme";
 
 import {
   TAGS_QUERY,
@@ -164,11 +164,19 @@ const TagsSection = ({ projectId }) => {
           >
             <Grid container spacing={1}>
               {data.moped_proj_tags.map((tag) => (
-                <Grid item spacing={0.5}>
+                <Grid item>
                   <Chip
                     key={tag.id}
                     label={tag.moped_tag.name}
                     onDelete={() => handleDeleteOpen(tag)}
+                    sx={{
+                      height: "auto",
+                      minHeight: theme.spacing(4),
+                      "& .MuiChip-label": {
+                        display: "block",
+                        whiteSpace: "normal",
+                      },
+                    }}
                   />
                 </Grid>
               ))}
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/TagsSection.js
```diff
@@ -29,20 +31,17 @@ const useStyles = makeStyles((theme) => ({
...
-    margin: theme.spacing(0.5),
-  },
   chipContainer: {
     display: "flex",
     justifyContent: "left",
     flexWrap: "wrap",
     listStyle: "none",
-    padding: "1rem 0",
-    paddingLeft: "16px",
+    padding: `${theme.spacing(1)} ${theme.spacing(2)}`,
```
roseeichelmann, line 39: nice touch to use the theme spacing here

roseeichelmann (approved):
looks and works great, tested on a small screen view as well. thanks for thinking of the multiline example. :sparkles:

amenity (approved):
So tidy, thanks! 🧹 🏷️  🚢

chiaberry (approved):
better use of screen real estate for sure
I did notice that the title of the PR is to move the tags box below the subprojects, but its still above the subprojects

mddilley:
@chiaberry oops thanks for catching that. I'll update the title if we decide not to change up the order. 🙏

chiaberry:
@mddilley I am for the idea of moving the tags below subprojects, what do you think @amenity ?

frankhereford (approved):
Mike, this is 🚢 🚢 🚢 🚢 . 

Your question about ordering the tags is an interesting one. I think the order of the possible tags in the selection dropdown is good where it is as alphabetical, because that's the order that is most helpful in finding one out of the group.

As for the order that they are displayed in the list of active tags .. I don't have a super strong opinion, but I would vote to leave it the way it is. I think the order of creation makes the most sense. By ordering by created date, the user has agency in how they want the tags ordered on the page. The first thing that they tag a project with is likely the most important tag they will give it and therefor deserves to be at the top of the list. 

I could be convinced easily though that there is a better way - this is my gut reaction, nothing more.

mddilley:
@frankhereford i like that perspective a lot, and you convinced me on leaving it alone. I could also imagine someone having a hard time finding the tag that they just added if it didn't snap on to the end of the list. 🙏

mddilley:
I created https://github.com/cityofaustin/atd-data-tech/issues/16978 to capture the tags and subprojects switcheroo so we can merge the bigger part of this issue. 🚀

PR #1324, opened 2024-04-22
Improve project map lag when there are a lot of components

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16487

## Testing
**URL to test:** 

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -78,25 +78,27 @@ export default function ComponentListItem({
...
-            </ListItem>
-          )}
-          {component.description && (
-            <ListItem className={classes.nested}>
-              <ListItemText secondary={component.description} />
-            </ListItem>
-          )}
-          {additionalCollapseListItems}
-        </List>
+        {isExpanded ? (
```
mddilley, line 83: This prevents us from rendering the expanded content of every single component in the list, and it was the most noticeable improvement.

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
```diff
@@ -1,197 +1,67 @@
...
-                      aria-label="link"
-                      onClick={copyLinkToClipboard}
-                    >
-                      <LinkIcon />
-                    </IconButton>
-                  </Tooltip>
-                </span>
-              </Tooltip>
+    <List sx={{ paddingTop: 0 }}>
+      {allComponents.map((component) => {
```
mddilley, line 28: This change helps pave a path to virtualizing this list in the future. Previously, we handle project components and related components as two different lists which increased the odds of us seeing two separate scrollable areas. They are now combined and we define which list item component to render for each one depending on if they belong to the current project or a subproject.

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsListItem.js
mddilley, line 1: Most of this code was moved from the previous version of the `ProjectComponentsList` component.

moped-editor/src/views/projects/projectView/ProjectComponents/RelatedComponentsListItem.js
mddilley, line 1: Most of this code was moved from the now removed components called `RelatedComponentsList`. This one is concerned with rendering a single list item instead of a list.

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -105,7 +105,7 @@ export default function TheMap({
         type: "show_error",
         payload: { message: "Zoom in to select features", severity: "error" },
       });
-    } else if (isCreatingComponent || isEditingComponent) {
+    } else {
```
mddilley, line 108: This fixes a bug where the **Zoom in to select features** warning that appear when mapping component features would stick around if you started creating or editing and then pressed the **Cancel** button. It nows disappears when cancelling out.

moped-editor/src/views/projects/projectView/ProjectComponents/index.js
```diff
@@ -275,30 +286,24 @@ export default function MapView({
                 onSaveDraftComponent={onSaveDraftComponent}
                 onSaveEditedComponent={onSaveEditedComponent}
               />
+            </List>
+            {isNotCreatingOrEditing ? (
```
mddilley, line 290: Updating this to toggle rendering here instead of further down in the tree decreased the number of props here.

amenity (approved):
🐎 🗺️ ⚡ 🚢  -- Thanks!

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
```diff
@@ -1,197 +1,67 @@
...
-                </IconButton>
-              </Tooltip>
-            </Stack>
-          }
-        />
-      );
-    })
+        if (isProjectComponent) {
+          return (
+            <ProjectComponentListItem
```
roseeichelmann, line 34: suchh a tiny thing but could we make the name of the file the same as the name of the component? The file is `ProjectComponentsListItem`

roseeichelmann (approved):
tested locally and with the netlify link, the maps with a lot of components are loading super speedy and i tested adding a whole bunch of components to a project as well. thanks for optimizing the performance and for fixing a whole bunch of other bugs, this is awesome work mike!! :rocket: :sparkles:

moped-editor/src/views/projects/projectView/ProjectComponents/ProjectComponentsList.js
mddilley: Thanks for having a sharp eye and catching this. 🙏 🙏 Just pushed this change

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
chiaberry: great catch!

chiaberry (approved):
A lot of really good refactoring here, I am learning a lot more about leveraging good patterns to cut down on unnecessary renders

Its a lot faster comparing the big pedestrian crossings map locally vs in prod

✨

PR #1325, opened 2024-04-23
Use project full name instead of primary name

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/16919

## Testing
**URL to test:** 
you will have to run locally to do the full test, since there are changes to the metadata and the project list view. 

I've switched the project name to use the full project name in the following places:

project list page, parent project name column
project summary page, parent project name
project summary page, subprojects list, what is shown and the options that are in the autocomplete
project name on the map tab
move component selector on the map tab
dashboard project names

**Steps to test:**

Besides visually seeing the full project name shown, there are few places where I've updated the autocomplete choices. 
Test finding projects to move the component to in the map tab. 
Screenshot 2024-04-23 at 4 04 53 PM

Test the subprojects table on the summary page, searching for projects by secondary name too. 
Screenshot 2024-04-23 at 4 06 59 PM

---
#### Ship list
- [x] Code reviewed 
- [ ] ~~Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable~~

### DIFF
--- a/moped-database/migrations/1713896796482_add_full_name_to_list_view/up.sql
+++ b/moped-database/migrations/1713896796482_add_full_name_to_list_view/up.sql
@@ -1,0 +1,423 @@
+DROP VIEW project_list_view CASCADE;
+
+CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
+    SELECT
+        mpp.project_id,
+        string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members
+    FROM moped_proj_personnel mpp
+    JOIN moped_users mu ON mpp.user_id = mu.user_id
+    JOIN moped_proj_personnel_roles mppr ON mpp.project_personnel_id = mppr.project_personnel_id
+    JOIN moped_project_roles mpr ON mppr.project_role_id = mpr.project_role_id
+    WHERE mpp.is_deleted = false AND mppr.is_deleted = false
+    GROUP BY mpp.project_id
+),
+
+funding_sources_lookup AS (
+    SELECT
+        mpf_1.project_id,
+        string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
+    FROM moped_proj_funding mpf_1
+    LEFT JOIN moped_fund_sources mfs ON mpf_1.funding_source_id = mfs.funding_source_id
+    WHERE mpf_1.is_deleted = false
+    GROUP BY mpf_1.project_id
+),
+
+project_type_lookup AS (
+    SELECT
+        mpt.project_id,
+        string_agg(mt.type_name, ', '::text) AS type_name
+    FROM moped_project_types mpt
+    LEFT JOIN moped_types mt ON mpt.project_type_id = mt.type_id AND mpt.is_deleted = false
+    GROUP BY mpt.project_id
+),
+
+child_project_lookup AS (
+    SELECT
+        jsonb_agg(children.project_id) AS children_project_ids,
+        children.parent_project_id AS parent_id
+    FROM moped_project children
+    JOIN moped_project parent ON children.parent_project_id = parent.project_id
+    WHERE children.is_deleted = false
+    GROUP BY children.parent_project_id
+),
+
+work_activities AS (
+    SELECT
+        mpwa.project_id,
+        string_agg(task_order_objects.task_order_object ->> 'display_name'::text, ', '::text) AS task_order_names,
+        string_agg(task_order_objects.task_order_object ->> 'task_order'::text, ', '::text) AS task_order_names_short,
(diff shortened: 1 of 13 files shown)


moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -554,7 +554,7 @@ const ProjectView = () => {
                             data={data}
                             error={error}
                             refetch={refetch}
-                            projectName={data.moped_project[0].project_name}
+                            projectName={data.moped_project[0].project_name_full}
```
chiaberry, line 557: this is what is shown on the map tab

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -182,14 +182,14 @@ const DashboardView = () => {
       width: "10%",
     },
     {
-      title: "Name",
+      title: "Full name",
       field: "project.project_name",
```
mddilley, line 186: Small thing but I think this could be updated to `project_full_name` too in case we ever fall back to not using the custom `render`.

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
```diff
@@ -52,14 +52,14 @@ const SubprojectsTable = ({ projectId = null, refetchSummaryData }) => {
       width: "15%",
     },
     {
-      title: "Name",
+      title: "Full name",
       field: "project_name",
```
mddilley, line 56: Same here - i think this could be updated to `project_name_full`

moped-editor/src/views/projects/projectView/ProjectView.js
mddilley: This made me think of the title that shows in the browser tab. Can't remember where that is set, and I don't know if it is important enough to change tbh. 🤷

mddilley (approved):
All looks great, and I'm happy to see the secondary names spreading through the app. Seems like good motivation for people to add them to me! 🚢 🙌

tillyw (approved):
tested locally and the full name shows up in all the places listed!

moped-database/migrations/1713896796482_add_full_name_to_list_view/up.sql
roseeichelmann, line 1: these views are so gnarly, wish there was an easier way to review diff changes when it comes to something like recreating a view

roseeichelmann (approved):
yay the more names the merrier :sparkles: looks good!

amenity (approved):
Full name everywhere! 🪄 😊 Thanks!

moped-database/migrations/1713896796482_add_full_name_to_list_view/up.sql
chiaberry: thats what the views files are hopefully helping with, and since Frank automated them, I feel better about them matching exactly what changes were made.

moped-editor/src/views/projects/projectView/ProjectView.js
chiaberry: good catch!! and ive updated it

moped-editor/src/views/dashboard/DashboardView.js
chiaberry: you are absolutely right, I've updated them in the both places you pointed out.

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
```diff
@@ -52,14 +52,14 @@ const SubprojectsTable = ({ projectId = null, refetchSummaryData }) => {
...
     },
     {
-      title: "Name",
-      field: "project_name",
-      width: "35%",
-      validate: (entry) => !!entry.project_name,
+      title: "Full name",
+      field: "project_name_full",
+      width: "40%",
+      validate: (entry) => (!!entry.project_name),
```
frankhereford, line 58: I'm not 100% sure this line is the culprit, but I do observe a regression where I can't get the checkbox to turn black when I'm trying to add a new subproject. 
image

frankhereford (changes requested):
This looks great Chia - thank you so much. I can't wait to see how folks use this to adapt moped to their group's needs.

I noticed one tiny regression -- I think the below issue resolves by aligning the field name to include the new `_full` designation, but that's just a guess without trying it.

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
chiaberry: yup, you nailed it. i updated the field but didnt update the validate. thanks for catching this!

frankhereford (approved):
Thanks for fixing that!! 🚀🚀🚀

PR #1326, opened 2024-04-25
Enhance moped database orchestration

frankhereford:
## Intent

This PR intends to close both:

* https://github.com/cityofaustin/atd-data-tech/issues/16443
* https://github.com/cityofaustin/atd-data-tech/issues/16386

## Usage

```
./hasura-cluster replicate
```

## Testing

Check out this branch and fire up your cluster. Everything should work like normal. 

Please copy the `env_template` into `env` and add your moped read replica credentials.

Try out the new command, `./hasura-cluster replicate`. You should end up with a copy of live data, up to date with migrations and metadata. 

Try it again, it should refresh the data but skip the downloading step. It will cache the export in `./snapshots/` for each day, and you can remove the exported `.sql` file to cause it to download the data again.

---
#### Ship list
- [ ] Code reviewed 


moped-database/env_template
frankhereford, line 1: Taking out these deprecated version designations

moped-database/docker-compose.arm64.yml
frankhereford, line 1: These `version` changes are tag-along changes. These changes remove a deprecated designation in the compose files.

moped-database/env_template
frankhereford, line 1: I went back and forth on if we should have a service account to the read replica that is stored in 1PW or to ask devs to put their own read replica credentials in an env file. I went this way because I thought it might be onerous to ask folks to install the 1PW CLI `op` tool. If anyone wants to reconsider that, I'd be eager to revise. 🤔

moped-database/.gitignore
```diff
@@ -1,4 +1,4 @@
 atd_moped-full.sql.gz
 *moped_dump.sql
 production.sql
-reload.sh
```
frankhereford, line 4: This is removing my own reload script that I had put in the ignore file.

moped-database/hasura-cluster
frankhereford, line 1: There's a ton of red in this file. The orchestration script had a replication system in it that relied on opening a `ssh` tunnel through the bastion and then creating a snapshot from that, or alternatively, pulling down a snapshot from S3. I think we may not have adopted this system because of complexity around maintaining the pre-shared keys on the bastion and other credentials, but I am not 100% on that.

moped-database/hasura-cluster
```diff
@@ -533,9 +149,9 @@ function start_only() {
   __get_platform;
   START_MODE=$1;
   if [[ "${START_MODE}" = "" ]]; then
-    docker-compose -f docker-compose.yml -f docker-compose.${PLATFORM}.yml up -d;
+    docker compose -f docker-compose.yml -f docker-compose.${PLATFORM}.yml up -d;
   else
-    docker-compose -f "${START_MODE}" up -d;
+    docker compose -f "${START_MODE}" up -d;
```
frankhereford, line 154: One of the issues in this PR is to transition to using `docker compose` instead of the old way of using the perl script `docker-compose`. The core docker project adopted the compose functionality and brought it into the core package.

moped-database/hasura-cluster
```diff
@@ -707,6 +291,57 @@ function show_banner() {
   echo "- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -";
 }
 
+function __check_export_complete() {
```
frankhereford, line 294: This replication system will check the end of the export files it keeps around to make sure that they didn't get interrupted while being downloaded.

moped-database/hasura-cluster
```diff
@@ -707,6 +291,57 @@ function show_banner() {
...
+    # echo "PostgreSQL database export is complete."
+    return 0
+  else
+    # echo "PostgreSQL database export is not complete."
+    return 1
+  fi
+}
+
+function replicate() {
+  source ./env;
```
frankhereford, line 308: Consume the configurations in `env` by sourcing it.

moped-database/hasura-cluster
```diff
@@ -707,6 +291,57 @@ function show_banner() {
...
+    -U $MOPED_RR_USERNAME \
+    -h $MOPED_RR_HOST \
+    $MOPED_RR_DB > $snapshot_file;
+    
+    # Check if the export is complete
+    if __check_export_complete "$snapshot_file"; then
+      echo "Replication successful.";
+    else
+      echo "Replication failed. The export is not complete.";
+      return 1;
```
frankhereford, line 330: This should halt the program and by returning a non-zero return code, indicate an error.

moped-database/hasura-cluster
```diff
@@ -707,6 +291,57 @@ function show_banner() {
...
+    return 1
+  fi
+}
+
+function replicate() {
+  source ./env;
+  start_only;
+  # Get the current date in YYYY-MM-DD format
+  current_date=$(date +"%Y-%m-%d");
+  snapshot_file="./snapshots/production-$current_date.sql";
```
frankhereford, line 312: This system should get a new file every day, but only once a day. Otherwise it will reuse the same snapshot all day unless you delete it to force it to get a new one.

chiaberry (commented):
edited -- see my next comment, it was user error

~~I am running into an error while testing this, I am troubleshooting~~

```
Replicating production database to local database... (~ 2-4 minutes)
Replication failed. The export is not complete.
```


Also, one step that needs to be added to the test instructions is to create the snapshots folder: 

```
Replicating production database to local database... (~ 2-4 minutes)
./hasura-cluster: line 317: ./snapshots/production-2024-04-29.sql: No such file or directory
```

What do you think about adding an option to use production data, without having to download a new snapshot?

chiaberry (commented):
user error -- i must have checked this out last week, and i had not pulled in the updates, so it was looking for the incorrect last line in the sql dump file

frankhereford:
> one step that needs to be added to the test instructions is to create the snapshots folder

Great call, thank you @chiaberry 🙏 I approached it by making the program ensure that it's created if its not. https://github.com/cityofaustin/atd-moped/pull/1326/commits/15894c7b3575c34ac5fb268304625b3c13911832

chiaberry (approved):
this is a total level up for us

I have a branch started where I am updating the Readmes in moped, I went ahead and updated the readme with steps to use the new `replicate` option https://github.com/cityofaustin/atd-moped/pull/1332

moped-database/hasura-cluster
mddilley: nice touch to close the loop on this issue we see with the VZ stack every once in a while. 🙏

mddilley (approved):
Tested locally using with and without the `-j` flag, and I can't wait to have this merged! Code looks great as well, and thanks for removing so much cruft that doesn't reflect our local dev process anymore. 🙏 This puts us in a much better place. 🙌 🧪 🚀

PR #1276, opened 2024-02-12
Create projects map page layout and route

mddilley:
## This is the collector branch for the projects map MVP

This will not be merged until the MVP is complete and ready for release.

### Merge TODOs
- [x] Add to QA test script
- [x] Merge main and test migrations (update if needed)

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9913
https://github.com/cityofaustin/atd-data-tech/issues/9914
https://github.com/cityofaustin/atd-data-tech/issues/9919

This adds: 
- a **Map** tab to the main view that links to a new route
- a map to start building on
- a collapsible side bar that has a placeholder title and content

## Testing
**URL to test:** 

https://deploy-preview-1276--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Go to the deploy preview and click the new navigation tab called **Map**. You should see that the sidebar is expanded by default.
2. Click the arrow in side menu bar to collapse and expand it. Scroll the placeholder area of the side bar and see that the title and nav arrow stay at the top.
3. Adjust the width to 600px or less or use a mobile device mode in dev tools. You should see that the sidebar is collapsed by default.
4. Click the nav arrow to expand it, scroll the placeholder area, and see that the title and arrow button stay at the top.

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager approved
- [x] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1328, opened 2024-04-26
Move map from Map tab to Project list view

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16966

This one moves the projects map from a tab in the main app view to the project list view where it is toggled on/off with a switch. I'd appreciate any feedback on design, etc. 🙏 No rush since the [next issue](https://github.com/cityofaustin/atd-data-tech/issues/9927) isn't blocked by this.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Go to the Projects table
2. See that there is a switch next to the **Download** export button that toggles the map view
3. See that the map that was in a tab in the [last PR](https://github.com/cityofaustin/atd-moped/pull/1276) now replaces the list when the switch is on
4. I also made changes to the responsive behavior of the row with the search bar, download button, and switch so feedback on that would be great, thanks!

---
#### Ship list
- [x] Code reviewed 


amenity (changes requested):
The map looks beautiful! 😊 🗺️ ✨ 

Responsive-wise, I like the switch and download button bumping down to a new line. It looks like there's still an the buttons are no longer full-width at the smallest breakpoint.

Screenshot 2024-05-07 at 12 07 41 PM

mddilley:
@amenity Thanks for catching that! I fixed that so that all breakpoints are covered now. 🙏

amenity (approved):
👌🏻 Looks great, thanks!

johnclary (commented):
this looks awesome! so exciting this come to life 🚀 

on the small display size, i wonder if we could get the search button, download icon, and map toggle in one row. at that breakpoint you could hide the download button label and only show the icon.

one other observation is that i think the map toggle might be better placed to the left of the search button. i think folks are going to be clicking that much more often than the download button. 

just food for thought! 🙌 🚢 🚢 🚢 🚢 🚢

chiaberry (approved):
Looks good!

chiaberry:
also now that i am looking at the next PR, I like the toggle saying "Map" versus show map.

PR #1334, opened 2024-05-01
adds full name to navigation search

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/16832

## Testing
**URL to test:** 

https://deploy-preview-1334--atd-moped-main.netlify.app/

**Steps to test:**
- find or create a project that has a secondary name with a string that is not contained in the primary name
- search for the secondary name string in the nav search bar
- your result should appear, displaying the full name!

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchQueryConf.js
```diff
@@ -37,6 +37,24 @@ export const NavigationSearchQueryConf = {
...
+    project_name_secondary: {
+      searchable: true,
+      search: {
+        operator: "_ilike",
+        quoted: true,
+        envelope: "%{VALUE}%",
+      },
+      type: "String",
+    },
+    project_name_full: {
```
mddilley, line 49: Since we are checking if the search string is anywhere in all three project name fields, I wonder if we should remove the other two and keep `project_name_full` since it is the combo of the other two. It works the same either way. Really just thinking out loud.

mddilley (approved):
Works great! I found my project using the global search! 🙌 🍕 🚀

Screenshot 2024-05-02 at 5 27 21 PM

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchQueryConf.js
roseeichelmann: makes sense to remove the other two imo!

roseeichelmann (approved):
yay looks good!

chiaberry (approved):
thanks!

amenity (approved):
Looks good! Thanks, @tillyw

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchQueryConf.js
tillyw: makes sense, thanks for thinking of that!

tillyw:
thanks y'all!

PR #1335, opened 2024-05-01
 [Bug] Error happens when not completely removing an advanced search filter

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16889

I addressed the bug and also added a few other improvements. Previously when you deleted a filter it would update the URL search params before you hit apply. Now the URL search params are only updated once you click apply. I also made it so that when you edit filters and dont apply your changes, close out of advanced search, and open again, ur unsaved changes are not persisted which is what users said they would expect. 

## Testing
**URL to test:** 

https://deploy-preview-1335--atd-moped-main.netlify.app/moped/projects

**Steps to test:**
1. Open the advanced search. See that the Search button now says Apply.
2. Create a filter, add an operator and value and hit apply. Open up advanced search again. Clear out the Field part of the filter and exit out of the advanced search without clicking Apply. You should not get an error anymore and when you open it again your unapplied change should not persist. Also the URL search params should not change until you click Apply.
3. Test changing the value of the filter and exiting out without applying, open up advanced search, the unapplied changes should not persist.
4. Test adding multiple filters, deleting some, exiting out without applying, other times with applying.
5. Now when you click the Reset button it will clear the URL search params as well. Before it would delete all the filters/search and update the search results, but the URL would still have the old search params.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -115,6 +115,8 @@ const Filters = ({
   resetSimpleSearch,
   isOr,
   setIsOr,
+  searchParams,
+  setSearchParams,
 }) => {
   /**
    * The styling of the search bar
@@ -122,7 +124,6 @@ const Filters = ({
    * @default
    */
   const classes = useStyles();
-  let [, setSearchParams] = useSearchParams();
 
   const { loading, error, data } = useQuery(LOOKUP_TABLES_QUERY);
 
@@ -283,10 +284,8 @@ const Filters = ({
     setSearchParams((prevSearchParams) => {
       prevSearchParams.delete(advancedSearchFilterParamName);
       prevSearchParams.delete(advancedSearchIsOrParamName);
-
       return prevSearchParams;
     });
-
     resetSimpleSearch();
   };
 
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useState, useMemo } from "react";
+import React, { useState, useMemo } from "react";
 import { useSearchParams } from "react-router-dom";
 import PropTypes from "prop-types";
 import { useQuery } from "@apollo/client";
@@ -108,14 +108,12 @@ const useStyles = makeStyles((theme) => ({
  * @constructor
  */
 const Filters = ({
-  filters,
   setFilters,
   handleAdvancedSearchClose,
   filtersConfig,
   resetSimpleSearch,
   isOr,
   setIsOr,
-  searchParams,
   setSearchParams,
 }) => {
   /**
@@ -129,14 +127,24 @@ const Filters = ({
 
   if (error) console.error(error);
 
+  let [searchParams] = useSearchParams();
+
   /* Consume existing filters or start with an empty filter if none exist */
   const initialFilterParameters = useMemo(() => {
-    if (filters.length > 0) {
-      return filters;
-    } else {
-      return [generateEmptyFilter()];
+    if (Array.from(searchParams).length > 0) {
+      const filterSearchParams = searchParams.get(
+        advancedSearchFilterParamName
+      );
+      if (filterSearchParams === null) return [];
+
+      try {
+        return JSON.parse(filterSearchParams);
+      } catch {
+        return [];
+      }
     }
-  }, [filters]);
+    return [];
+  }, [searchParams]);
 
   /**
    * The current local filter parameters so that we can store updated filters and
@@ -148,7 +156,6 @@ const Filters = ({
   const [filterParameters, setFilterParameters] = useState(
     initialFilterParameters
   );
-  const [searchParameters, setSearchParameters] = useState();
 
   /* Track toggle value so we update the query value in handleApplyButtonClick */
   const [isOrToggleValue, setIsOrToggleValue] = useState(isOr);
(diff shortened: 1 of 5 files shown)


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -122,24 +121,21 @@ const Filters = ({
...
-      return filters;
-    } else {
-      return [generateEmptyFilter()];
-    }
-  }, [filters]);
+  /* Consume existing URL search params or start with an empty filter if none exist */
+  const initialFilterParameters = useMakeFilterState(
+    searchParams,
+    advancedSearchFilterParamName
+  );
```
roseeichelmann, line 134: this keeps advanced search filter changes from persisting when the user exits out without applying. now when this component renders it is pulling from the URL search params initially

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -235,25 +231,9 @@ const Filters = ({
...
 
-    if (remainingFiltersCount === 0) {
-      /* Clear search params since we have no advanced filters */
-      setSearchParams((prevSearchParams) => {
-        prevSearchParams.delete(advancedSearchFilterParamName);
-        prevSearchParams.delete(advancedSearchIsOrParamName);
-
-        return prevSearchParams;
-      });
-    } else if (remainingFiltersCount === 1) {
```
roseeichelmann, line 246: we shouldnt be making any changes to the URL search params when a filter is deleted. the URL should only be updated once the Apply button is clicked.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -274,19 +255,17 @@ const Filters = ({
   /**
    * Reset search box and advanced filters
    */
-  const handleResetFilters = useCallback(() => {
```
roseeichelmann, line 277: i dont think this needs a `useCallback` anymore, previous this function was called within a `useEffect` but now its not

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -108,38 +108,35 @@ const useStyles = makeStyles((theme) => ({
...
 const Filters = ({
-  filters,
   setFilters,
   handleAdvancedSearchClose,
   filtersConfig,
   resetSimpleSearch,
   isOr,
   setIsOr,
+  searchParams,
+  setSearchParams,
```
roseeichelmann, line 118: im not sure exactly why but sharing these with the Search parent component and passing them down from there fixed some issues with the URL search params not clearing on Reset button click

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -305,6 +285,13 @@ const Filters = ({
       setIsOr(isOrToggleValue);
       setFilters(filterParameters);
     } else {
+      /* Clear search params since we have no advanced filters */
+      setSearchParams((prevSearchParams) => {
+        prevSearchParams.delete(advancedSearchFilterParamName);
+        prevSearchParams.delete(advancedSearchIsOrParamName);
+
+        return prevSearchParams;
+      });
```
roseeichelmann, line 294: here, after the user clicks apply, is where we should clear the URL search params if the filters were deleted

moped-editor/src/components/GridTable/helpers.js
```diff
@@ -136,3 +137,29 @@ export const getDefaultOperator = (filterConfigForField) => {
...
+/**
+ * If filter exists in url, get the values and try to parse them
+ * Used to initialize filter state
+ * @param {Object} searchParams - The URL search parameters
+ * @param {String} advancedSearchFilterParamName
+ * @return {Object}
+ */
+export const useMakeFilterState = (
+  searchParams,
+  advancedSearchFilterParamName
```
mddilley, line 150: Small thing that jumped out at me with this `advancedSearchFilterParamName` param is that I kept mixing this up with the specific const that stores the `filters` string used in the url. Maybe naming it something more general could work or maybe it is just my brain 🙃

I also realize that it is only used for that specific search param so maybe it doesn't really matter at all!

mddilley (changes requested):
This is amazing, Rose! This feels so much better and made me realize how confusing some of the existing behavior can be. I stepped through a bunch of filter combos (simple, advanced, and/or) locally, and all of this looks ✨. I added a comment about a small regression that I think should be fixed.

chiaberry:
I agree with @mddilley that we should keep the functionality where a blank filter is automatically included when someone opens the advanced filters. 

I tested with different filters and followed your test steps and I couldnt break it! After that regression is fixed, ill 👍🏼

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -99,47 +99,45 @@ const useStyles = makeStyles((theme) => ({
 
 /**
  * Filter Search Component aka Advanced Search
- * @param {Object} filters - The current filters from useAdvancedSearch hook
  * @param {Function} setFilters - Set the current filters from useAdvancedSearch hook
  * @param {Function} handleAdvancedSearchClose - Used to close the advanced search
  * @param {Object} filtersConfig - The configuration object for the filters
  * @param {Function} resetSimpleSearch - Function to reset the simple search
+ * @param {Object} searchParams - The URL search params
+ * @param {Function} setSearchParams - Function to set the URL search params
```
frankhereford, line 107: Thanks for being diligent and keeping these up to date. I find them easy to miss.

frankhereford (approved):
Thank you Rose. Nothing simple about this code at all, at least not to me. I think this might be a bit bigger than the snackoo label it had!

I've read through your code, and nothing is jumping out at me. Mike's catch on suppressing the 'No filters applied' is very sharp, thank you Mike! Testing out the functionality based on your instructions and poking around with it -- 💯 

Thanks again 🚢 🚢🚢

tillyw (approved):
tested locally and tried a whole variety of states and everything worked as expected! this is a huge improvement and makes a lot more sense to me as a user. i agree it would be nice to have a blank filter appear in the initial state (which would mirror the state you get when you click the 'reset' button). thanks for working through all these different possible states!

roseeichelmann:
@mddilley thanks for catching that regression, you do have a really good eye! just pushed up the fix

amenity (approved):
Really nice! Thanks, @roseeichelmann ✨ 🚢

mddilley (approved):
Amazing! All of this works so well - thanks! 🚢 🙌 🚢

chiaberry (approved):
🚢

PR #1347, opened 2024-05-17
2.10.1 patch

chiaberry:
patch fix

https://github.com/cityofaustin/atd-data-tech/issues/17287
and
https://github.com/cityofaustin/atd-data-tech/issues/17277


mddilley (approved):
🚢 🩹 🙌

PR #1350, opened 2024-05-22
Prevent staff list from showing stale data

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/17286

## Testing
**URL to test:** 

Test by updating login users here. nonlogin users will not work on the netlify app

https://deploy-preview-1350--atd-moped-main.netlify.app/moped/staff 

1. Change a name or title. 
2. Go back to the page list view
3. see that your updates is there
4. navigate back to the profile you just updated. see that the user details matches your updates

To test updating nonlogin users you will have to test locally
same steps as above, but only for a nonlogin user 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/hasura-cluster
+++ b/moped-database/hasura-cluster
@@ -316,8 +316,8 @@ function replicate() {
 
   # If use_any_snapshot is true, use the most recent snapshot
   if $use_any_snapshot; then
-    echo "Using most recent snapshot $snapshot_file"
     snapshot_file="./snapshots/$(ls -t ./snapshots | head -n 1)"
+    echo "Using most recent snapshot $snapshot_file"
   fi
 
   # Check if the file already exists and the export is complete
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -5019,6 +5019,7 @@
           - picture
           - roles
           - title
+          - user_id
           - workgroup_id
   select_permissions:
     - role: moped-admin
@@ -5097,6 +5098,7 @@
           - picture
           - roles
           - title
+          - user_id
           - workgroup_id
         filter: {}
         check: {}
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -5019,7 +5019,6 @@
           - picture
           - roles
           - title
-          - user_id
           - workgroup_id
   select_permissions:
     - role: moped-admin
@@ -5098,7 +5097,6 @@
           - picture
           - roles
           - title
-          - user_id
           - workgroup_id
         filter: {}
         check: {}
--- a/moped-editor/src/views/staff/EditStaffView.js
+++ b/moped-editor/src/views/staff/EditStaffView.js
@@ -44,6 +44,7 @@ const EditStaffView = () => {
     error: userQueryError,
   } = useQuery(GET_USER, {
     variables: { userId },
+    fetchPolicy: "cache-and-network",
   });
 
   if (userQueryError) {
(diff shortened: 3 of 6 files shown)


moped-database/metadata/tables.yaml
```diff
@@ -5019,6 +5019,7 @@
           - picture
           - roles
           - title
+          - user_id
```
chiaberry, line 5022: this fixes a bug that was driving me nuts, and I came across it while I was trying to test this PR. 

if you go into staging/prod, you cannot update a user and it fails silently (try updating the title for this user for example https://moped.austinmobility.io/moped/staff/edit/224). the error says user_id isnt in the mutation, and its because of permissions issues 🙃

johnclary: interesting—why is it necessary to insert the `user_id` if it is generated by DB sequence?

chiaberry: good point -- I initially removed it from the mutation but still kept getting the error and then went down the permissions rabbit hole. It is coming from the form submission I think, let me check why its in here at all.

moped-editor/src/views/staff/EditStaffView.js
```diff
@@ -44,6 +44,7 @@ const EditStaffView = () => {
     error: userQueryError,
   } = useQuery(GET_USER, {
     variables: { userId },
+    fetchPolicy: "cache-and-network",
```
frankhereford, line 47: Thank you for unraveling [the correct fetch policy](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies). It's far from obvious to me which is right. 🙏🙏

moped-database/hasura-cluster
```diff
@@ -316,8 +316,8 @@ function replicate() {
 
   # If use_any_snapshot is true, use the most recent snapshot
   if $use_any_snapshot; then
-    echo "Using most recent snapshot $snapshot_file"
     snapshot_file="./snapshots/$(ls -t ./snapshots | head -n 1)"
+    echo "Using most recent snapshot $snapshot_file"
```
chiaberry, line 320: i was printing the wrong file name....

moped-editor/src/views/staff/components/NonLoginUserActivationButtons.js
```diff
@@ -60,7 +60,7 @@ const NonLoginUserActivationButtons = ({
   };
 
   const handleUpdateNonLoginUser = async () => {
-    const { password, __typename, ...restOfFormValues } =
+    const { password, __typename, user_id, ...restOfFormValues } =
```
chiaberry, line 63: removing it from the payload!

moped-database/metadata/tables.yaml
mddilley: @chiaberry I'll wrap this review in the morning, but don't these permissions need to be removed now that the non-login user form payload is fixed?

moped-editor/src/views/staff/components/NonLoginUserActivationButtons.js
mddilley: This is making a lot more sense out of one of the test failures that Rose saw last release. I thought it was stale data only but it was also the mutation not working. So glad this is patched!

mddilley (approved):
I tested this locally and on the deploy preview, and I'm seeing my edits work and update in all staff views. Thanks for finding a fix to another bug on top of this one in the issue! 🙌  🚀 🚢

tillyw (approved):
tested both in the deploy preview and locally for both login and non-login users and both worked perfectly! interesting to learn about the different fetch policies! thank you

moped-editor/src/views/staff/components/NonLoginUserActivationButtons.js
johnclary: nice!

PR #1355, opened 2024-06-03
[Bug] Alert that a moved component cannot be found in original project is missing

chiaberry:
## Associated issues

Closes https://github.com/cityofaustin/atd-data-tech/issues/17518

## Testing
**URL to test:** 
https://deploy-preview-1355--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**
1. Visit a project (for example
https://deploy-preview-1355--atd-moped-main.netlify.app/moped/projects/2022?tab=map&project_component_id=2627)
2. Move a component from your project to another project, and then try to visit the previous link. Alternately, make up a component id that doesnt exist: (https://deploy-preview-1355--atd-moped-main.netlify.app/moped/projects/2022?tab=map&project_component_id=262300) Make sure you can see the alert. 
3. Test adding a component, but be zoomed out -- verify the "zoom in to select features" alert appears. And disappears when you zoom in. 
4. This project has related project components, but no mapped components of its own. Verify that if you try this link for a non existent project component id, you still see the alert (https://deploy-preview-1355--atd-moped-main.netlify.app/moped/projects/1940?tab=map&project_component_id=236200)

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
@@ -103,10 +103,14 @@ export default function TheMap({
     if (shouldShowZoomAlert) {
       errorMessageDispatch({
         type: "show_error",
-        payload: { message: "Zoom in to select features", severity: "error" },
+        payload: {
+          message: "Zoom in to select features",
+          severity: "error",
+          messageType: "componentZoomError",
+        },
       });
     } else {
-      errorMessageDispatch({ type: "hide_error" });
+      errorMessageDispatch({ type: "hide_zoom_error" });
     }
   }, [
     currentZoom,
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/utils/useComponentLinkParams.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/utils/useComponentLinkParams.js
@@ -63,7 +63,7 @@ export const useComponentLinkParams = ({
     }
 
     // Set clicked component from search parameter once projectComponents data loads
-    if (projectComponents.length > 0) {
+    if (projectComponents.length > 0 || allRelatedComponents.length > 0) {
       const componentFromParams =
         projectComponents.find(
           (component) => component.project_component_id === componentParamId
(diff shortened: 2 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
```diff
@@ -103,10 +103,14 @@ export default function TheMap({
...
         type: "show_error",
-        payload: { message: "Zoom in to select features", severity: "error" },
+        payload: {
+          message: "Zoom in to select features",
+          severity: "error",
+          messageType: "componentZoomError",
+        },
       });
     } else {
-      errorMessageDispatch({ type: "hide_error" });
```
chiaberry, line 109: this is what was hiding the moved component alert -- the useEffect would identify that the zoom alert was not needed, it would dispatch the hide_error action thus hiding the moved component alert.

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useComponentLinkParams.js
```diff
@@ -63,7 +63,7 @@ export const useComponentLinkParams = ({
     }
 
     // Set clicked component from search parameter once projectComponents data loads
-    if (projectComponents.length > 0) {
+    if (projectComponents.length > 0 || allRelatedComponents.length > 0) {
```
chiaberry, line 66: if a project only had related components mapped, it would skip this part of the code and not set the clicked component nor show an error.

roseeichelmann (approved):
thanks for fixing this chia, test steps worked!

moped-editor/src/views/projects/projectView/ProjectComponents/utils/useComponentLinkParams.js
mddilley: When I was testing, i noticed that, if the last component in one project is moved to another project, the alert doesn't show because of the `projectComponents > 0`  check that we are using to make sure the data has loaded here before checking if the param component id is in the returned data or not.

This is an edge case so, while it would be cool to find out, I don't know if it is worth figuring out. what do you think?

moped-editor/src/views/projects/projectView/ProjectComponents/TheMap.js
mddilley: that makes so much sense! 🙌 I like this fix a lot so we are hiding the right alert. It made me start thinking about what a separate alert for the map itself would look like so I wrote up an issue in case it intersects with needs for the projects map some day. https://github.com/cityofaustin/atd-data-tech/issues/17687

mddilley (approved):
I tested this all which ways, and I am no longer seeing the different alerts compete. 🎉 I mentioned one small edge case that doesn't seem super important to me. Thank you for fixing this! 🚢

PR #1356, opened 2024-06-04
[Bug] Adding and then removing a funding amount causes work to be lost

chiaberry:
## Associated issues

closes https://github.com/cityofaustin/atd-data-tech/issues/17596

## Testing
**URL to test:** 

https://deploy-preview-1356--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

Go to the funding tab of a project and select Add a new funding source
In the row, choose a Source and then enter 100 in the Amount field but then delete the amount so it is blank. 
see that it saves correctly.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFundingTable.js
@@ -253,9 +253,7 @@ const ProjectFundingTable = () => {
       PopperComponent={CustomPopper}
       id={props.name}
       options={props.data}
-      renderInput={(params) => (
-        <TextField variant="standard" {...params} multiline />
-      )}
+      renderInput={(params) => <TextField variant="standard" {...params} />}
       getOptionLabel={(option) =>
         // if our value is a string, just return the string instead of accessing the name
         typeof option === "string" ? option : option[`${props.name}_name`]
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectFundingTable.js
```diff
@@ -253,9 +253,7 @@ const ProjectFundingTable = () => {
       PopperComponent={CustomPopper}
       id={props.name}
       options={props.data}
-      renderInput={(params) => (
-        <TextField variant="standard" {...params} multiline />
```
chiaberry, line 257: I removed multiline because there was a warning that shows up
```
ProjectFundingTable.js:649 A textarea element was provided to Autocomplete where input was expected.
This is not a supported scenario but it may work under certain conditions.
A textarea keyboard navigation may conflict with Autocomplete controls (for example enter and arrow keys).
Make sure to test keyboard navigation and add custom event handlers if necessary. 
```

And I dont think the autocompletes need to be multiline for this table

mddilley: Thanks for addressing this! 🙏 I agree that multiline is not needed since the chosen source or program is highlighted in the dropdown if someone needs to double-check their selection.

mddilley (approved):
Works great! I did not lose my data when bailing on a funding amount. 🎉 Thanks for bringing up moving more tables to DataGrid too. I think it would be great to explore the effort it would take to push the remaining uses of Material Table in that direction! 🚢

johnclary (approved):
👍 works as promised—no more error! 🚢

mddilley:
thanks again @chiaberry - merging this to wrap on QA testing! 🚀

PR #1357, opened 2024-06-04
Populate descriptor columns in ArcGIS database view and sort component tags in Data Dictionary

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17639
Closes https://github.com/cityofaustin/atd-data-tech/issues/17432

This PR populates some placeholder columns in the `component_arcgis_online_view` DB view for ATSD and also updates the ordering of the component tags in the Data Dictionary which came from some recent user feedback. 

## Testing
**URL to test:** 

Local and https://deploy-preview-1357--atd-moped-main.netlify.app/moped/dev/lookups#moped-component_tags

**Steps to test:**
1. Go to https://deploy-preview-1357--atd-moped-main.netlify.app/moped/dev/lookups#moped-component_tags and see that the component tags are ordered by **Type** ascending and then **Name** ascending.
2. Start the local stack with production data and then use the query below to see that:
   - `project_name` + `project_name_descriptor` (secondary name) = `project_name_with_descriptor` (full name) in the AGOL view
   - `parent_project_name` in the AGOL view is the same as the parent project's full name
3. You can also test removing the WHERE conditions to see that the `project_name_descriptor` column will show `NULL` if the component's project has no secondary name and the `parent_project_name_with_descriptor` column will show `NULL`  if it doesn't have a parent project
4. Test the down migration, and you should see the descriptor columns go back to having the value `placeholder text`

```sql
SELECT 
	agol.project_component_id,
	agol.project_id,
	agol.project_name,
	agol.project_name_secondary,
	agol.project_name_full,
	agol.parent_project_id,
	agol.parent_project_name_full,
	pp.project_name_full
FROM
	component_arcgis_online_view agol
	LEFT JOIN moped_project pp ON agol.parent_project_id = pp.project_id
WHERE
	TRUE
	AND agol.project_name_secondary IS NOT NULL
	AND agol.parent_project_name IS NOT NULL
	AND pp.project_name_secondary IS NOT NULL;
```

---
#### Ship list
- [x] Code reviewed 


johnclary (approved):
@mddilley this looks good to me—glad to see those placeholders replaced with real data 😅 

for what it's worth, i think it might be good to take this opportunity to do away with the names `project_name_with_descriptor` and `project_name_descriptor` and instead match the Moped column names of `project_name_secondary` and `project_name_full`. 

i doubt NW is attached to those old column names, we just didn't have a convention established when we created those placholders. 

i _think_ that can be done fairly painlessly by having the geo team add those new columns to the layers, and then the old columns can be deprecated later once the AGOL viewer is updated.

chiaberry (approved):
Works great, I got thrown off by the different results after the down migration, but after digging into it, its proof this PR works! 

🚢

mddilley:
@johnclary that sounds like a good change especially since i see that we use the same convention for component names in the component attributes (`component_name` and `component_name_full`). I'm going to feel out the coordination needed for that change (and for future changes). thanks!

johnclary:
> especially since i see that we use the same convention for component names in the component attributes future changes

ah yes—great point!

frankhereford (approved):
Tests out great; code looks 💯. Thanks Mike!

PR #1358, opened 2024-06-07
[Bug] User password requirement error is not translating correctly

chiaberry:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17285

## Testing
**URL to test:** 

https://deploy-preview-1358--atd-moped-main.netlify.app/moped/staff/new

**Steps to test:**

Fill out the new user form, every required field. Try a one letter password, you will see both error messages. 
Now try one that is longer than 8, but still not correct. 

Screenshot 2024-06-10 at 11 32 37 AM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/staff/helpers.js
+++ b/moped-editor/src/views/staff/helpers.js
@@ -68,8 +68,6 @@ export function useUserApi() {
 const errorsToTranslate = {
   "value does not match regex '^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$'":
     "not a valid email format",
-  "value does not match regex '^[a-zA-Z0-9_-!@%^*~?.:&*()[]$]*$'":
-    "password must be at least 12 characters long, it must include at least one lowercase letter, one uppercase letter, one number, and one special character: _-!@%^~?.:&()[]$",
   "value does not match regex '^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$'":
     "password must be at least 12 characters long, include at least one lowercase letter, one uppercase letter, one number, and one special character: _-!@%^~?.:&()[]$",
 };
--- a/moped-editor/src/views/staff/helpers.js
+++ b/moped-editor/src/views/staff/helpers.js
@@ -84,7 +84,7 @@ export const formatApiErrors = (errorsArray) =>
  */
 export const passwordLooksGood = (password) =>
   new RegExp(
-    "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$'"
+    "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$"
   ).test(password);
 
 /**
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/staff/helpers.js
```diff
@@ -68,8 +68,9 @@ export function useUserApi() {
 const errorsToTranslate = {
   "value does not match regex '^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$'":
     "not a valid email format",
-  "value does not match regex '^[a-zA-Z0-9_-!@%^*~?.:&*()[]$]*$'":
-    "password must be at least 12 characters long, it must include at least one lowercase letter, one uppercase letter, one number, and one special character: _-!@%^~?.:&()[]$",
+  "value does not match regex '^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$'":
+    "password must be at least 12 characters long, include at least one lowercase letter, one uppercase letter, one number, and one special character: _-!@%^~?.:&()[]$",
+  "min length is 8": "password is too short",
```
chiaberry, line 73: this error "min length is 8" comes from cognito. I tried using a blank string skip showing this error, but the formatApiErrors function `errorsArray.map((error) => errorsToTranslate[error] || error).join(", ")` evaluates the empty string as false, which shows the "min length is 8" error anyways. 

I welcome edits on what to show instead of this error

mddilley: I think this text is great and leads us in the right direction w/o the confusion of seeing two lengths requirements. 🙏 Could you add a comment that this min length 8 error is coming from a Cognito requirement for future us?

Maybe someday when we have all the time in the world, we could make the API handle returning nice, human-readable error messages. For now, this is the exact amount of effort that I think we should put into this. Thanks for making QA testing better! ✨

moped-editor/src/views/staff/helpers.js
```diff
@@ -83,7 +84,7 @@ export const formatApiErrors = (errorsArray) =>
  */
 export const passwordLooksGood = (password) =>
   new RegExp(
-    "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$"
+    "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$'"
```
mddilley, line 87: I did a double-take on this but I'm pretty sure this added single quote needs to be removed. Above, the whole regex pattern in the returned error is wrapped in single quotes and the pattern used in the api is `^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{12,}$` (no single quote at the end)

mddilley (changes requested):
Looks great to me! Thanks for tracking down that there are actually more error messages to handle here! There is one tiny change needed before 🚢.

moped-editor/src/views/staff/helpers.js
chiaberry: good catch, my faulty copy and paste job from above

mddilley (approved):
🚢 🙌 🙏

frankhereford (approved):
Looks great, and I think this meets the request of the issue perfect. 

But -- on seeing the long chain of concatenated errors during testing -- what would you think about expanding the scope slightly to put those errors in a `<UL>` and capitalizing the first letter of each, since they are sentences. It's the tiniest thing, but it did jump to mind. It could be a spin-off snackoo too.

mddilley:
@frankhereford that sounds like a great idea! Love putting some more effort into making this nicer. I'll open a snackoo and put it in the backlog. We only have a few admins that see this part of the UI so I'd like to cap the time spent on this for now.

frankhereford:
@mddilley wrote:

> We only have a few admins that see this part of the UI

Such a good call - 👍👍👍

PR #1360, opened 2024-06-13
Local PostGIS version bump to 3.4 & eliminate platform specific compose file

frankhereford:
## Associated issues

This PR aims to close https://github.com/cityofaustin/atd-data-tech/issues/17456. 

This work comes in the form of the code changes in the PR as well as a new multi-architecture image that I've built for us [here](https://hub.docker.com/layers/frankinaustin/postgis/12-3.4/images/sha256-ef147e4721f4ca650d0f632e23c74ac76db43598a1eb6268b799b7ff408d661a?context=repo).

## Testing

Local testing on this one, as it pertains only to the local stack. 

* Fire up your database. Did it work?
* Maybe fire up the editor to make sure that `graphql-engine` can connect to it and serve data.
* Give your DB client a whirl -- can you query the local database?
* Spin it down. 👋 

## Note
I'm not sure why this bears the "High Priority" flag, and I can't seem to turn it off. This doesn't feel very high priority to me.

---
#### Ship list


moped-database/docker-compose.yml
```diff
@@ -24,6 +24,24 @@ services:
...
+        environment:
+            - POSTGRES_USER=moped
+            - POSTGRES_PASSWORD=moped
+            - POSTGRES_DB=moped
+        command: ["postgres",
+            "-c", "jit=off",
+            "-c","log_statement=ddl",
+            "-c", "log_min_messages=info"]
+
+
```
frankhereford, line 44: I'm able to pull in the service that was previously defined in the platform specific docker compose files into the main file. This is possible now because I've built a single, multi-arch image that exactly matches the versions we have in production, where as previously, we had a pair of single architecture images, and had to pick the right one based on the chip in the computer running it.

moped-database/hasura-cluster
frankhereford, line 1: This is the undoing of the complexity of picking which compose file based on architecture since we can just use the main, single one now.

moped-database/hasura-cluster
```diff
@@ -27,12 +27,7 @@ function __get_platform() {
 
 function __set_postgresql_image() {
```
mddilley, line 28: Really glad to see the multiarch image introduced here now!! Out of curiosity and definitely for another day, but do you think we need to keep around any of these helpers to get the platform and set the image in the long term? I can make a tech debt issue to do this later.

mddilley (approved):
I spun up the local stack, and all of this works just like before and with less code! 🚢 🙌

moped-database/hasura-cluster
frankhereford: Good call Mike - I thought about that, and waffled back and forth...

I think it's right to yoink them out of there and clean that up, but I had that urge to keep them "in case we needed them" -- but that is a silly instinct in the land of `git`. I appreciate you making that issue - it'll be a quick one.

frankhereford:
@mddilley - I'm not sure if you want to merge this in advance of the release or not -- it certainly doesn't show up in the release to the users so 🤷 

if you do though, please feel free to merge this anytime that you'd like!

moped-database/hasura-cluster
mddilley: cool - that makes sense to me too. i also had that thought about keeping it just in case!

PR #1369, opened 2024-06-25
Incremental updates in Moped > AGOL ETL

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14479

This PR updates the Moped to AGOL ETL to be able to replace features of components whose parent project was updated since a provided timestamp with time zone. This will enable us to pass the last successful DAG run date when running through Airflow so we run it more frequently and without fully replacing the dataset each run.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start local stack with data replicated from production that is up to date. **This is a test on the production AGOL dataset so it is important to have fresh data to do a full replace at the end of the test steps.**
2. In your local editor, add a new component to a project or make an update to a title (any change [that propagates](https://github.com/cityofaustin/atd-data-tech/issues/14242) to the project's `updated_at` timestamp). Note the project ID for later.
3. Export env variables that are in `env_template` in your shell using `EXPORT key=value`. You can find the ATD Scripts Publisher details in [this 1PW entry](https://start.1password.com/open/i?a=J2OIJEXKD5ADBAZSD6M4R4VZ2A&v=ocp5yzivvhbyr226xvbo33eoqi&i=2v4vn2aovktpw2j6gvvk6uw2hm&h=cityofaustin.1password.com).
4. Create a virtual environment in `atd-moped/moped-etl/arcgis`, run: 
```shell
# Install requirements
$ pip install -r requirements.txt

# Run the script with a timestamptz that is 5 minutes ago or long enough to capture your edit
$ python components_to_agol.py -d 2024-06-27T22:43:16.360805+00:00
```
6. You should see you project ID in the script logging (`Deleting features for project ids <project ids>...`) when it detects that your project has changed. You should also see the `Uploading <total number of component features in your project including the new ones> features in chunks of 100...` logged.
7. Now, stop and start the local stack so your edit goes away. Run `python components_to_agol.py -f` to restore the production data in the dataset.

---
#### Ship list
- [x] Code reviewed 


moped-etl/arcgis/components_to_agol.py
```diff
@@ -54,74 +63,192 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
-        # extract geometry and line geometry from component data
-        # for line features, the line geometry is redundant/identical to geometry
-        # for point features, it is the buffered ring around the point as defined
-        # in the Moped component view
-        geometry = component.pop("geometry")
-        line_geometry = component.pop("line_geometry")
-
-        if not geometry:
-            continue
+    if args.full:
```
mddilley, line 70: this first block is the main content of the previous version of the script. It deletes and replaces all component features. I think it would be handy to have around as an option.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -54,74 +63,192 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
+            logger.info("Deleting all existing features...")
+            delete_all_features(feature_type)
+
+            logger.info(
+                f"Uploading {len(features)} features in chunks of {UPLOAD_CHUNK_SIZE}..."
+            )
+            for feature_chunk in chunks(features, UPLOAD_CHUNK_SIZE):
+                logger.info("Uploading chunk....")
+                add_features(feature_type, feature_chunk)
+    else:
```
mddilley, line 138: This block looks only deletes and replaces features from projects that were updated since the last DAG run. That date will be passed when run by Airflow.

moped-etl/arcgis/process/logging.py
mddilley, line 1: this was copied from the other Moped ETL because I thought about adding different logging levels. I'm down to remove if anyone feels strongly against this.

mddilley, line 1: i brought this over from the other Moped Knack ETL because I thought about adding a debug level of logging. I'll remove it if anyone thinks it is too much.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -54,16 +63,15 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
     return feature
 
 
-def main():
-    logger.info("Getting token...")
-    get_token()
-
-    logger.info("Downloading component data...")
-    data = make_hasura_request(query=COMPONENTS_QUERY)["component_arcgis_online_view"]
+def make_all_features(data):
```
mddilley, line 66: this is some of the original code in this ETL that was repeated between full and partial replace routes of this script so i made a function out of it.

moped-etl/arcgis/settings.py
```diff
@@ -2,9 +2,9 @@
 
 LAYER_IDS = {"points": 0, "lines": 1, "combined": 2}
 
-COMPONENTS_QUERY = """
-{
-  component_arcgis_online_view {
+COMPONENTS_QUERY_BY_LAST_UPDATE_DATE = """
+query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!) {
```
mddilley, line 6: Using `component_arcgis_online_view_bool_exp` here lets us pass a date filter for finding records from recently updated projects or no date filter for full replace. You can see that in the `make_hasura_request` calls. Also lets us reuse the same query w/o using a library to create gql fragments or something to that effect.

moped-etl/arcgis/utils.py
```diff
@@ -116,6 +116,32 @@ def delete_features(feature_type):
...
 
+def delete_features_by_project_ids(feature_type, project_ids):
+    """Deletes features from an arcgis online feature service associated with projects in a list ids.
+
+    Args:
+        feature_type (Str): the feature type we're adding: "points" or "lines"
+        project_ids (List): the project ids to delete
+
+    Raises:
+        Exception: if the deletion failes
```
chiaberry, line 127: small typo

moped-etl/arcgis/components_to_agol.py
```diff
@@ -100,28 +110,112 @@ def main():
...
+    parser.add_argument(
+        "-f",
+        "--full",
+        action="store_true",
+        help=f"Delete and replace all project components.",
+    )
+
+    args = parser.parse_args()
+    logger = get_logger(name="components-to-agol", level=logging.INFO)
+
```
johnclary, line 213: it might be good to throw an error here if someone decides to use both `--full` and `--date` in combination: it looks like that would result in the script running both tasks, which wouldn't be a big deal but still 🤷

moped-etl/arcgis/components_to_agol.py
```diff
@@ -100,28 +110,112 @@ def main():
...
+    get_token()
+
+    if args.full:
+        logger.info("Getting component features from all projects...")
+        data = make_hasura_request(
+            query=COMPONENTS_QUERY_BY_LAST_UPDATE_DATE,
+            variables={"where": {}},
+        )["component_arcgis_online_view"]
+
+        all_features = make_all_features(data)
```
johnclary, line 136: minor thing, but it looks like you could cut back on some dupe code by making the calls to `make_hasura_request` and `make_all_features` before you do the branching logic by assigning the `where` statement to a var

moped-etl/arcgis/utils.py
```diff
@@ -116,6 +116,32 @@ def delete_features(feature_type):
...
+    """
+    endpoint = get_endpoint("deleteFeatures", feature_type)
+    joined_project_ids = ",".join(project_ids)
```
johnclary, line 130: i am running into a TypeError in this line, because `project_ids` is an array of ints. back in the main script you do a similar join to create `joined_project_ids`, and it kinda looks like you may have meant to pass that directly into this function. should be a quick fix either way 👍

moped-etl/arcgis/components_to_agol.py
```diff
@@ -100,28 +110,112 @@ def main():
...
+                joined_project_ids = ", ".join(str(x) for x in delete_chunk)
+                logger.info(
+                    f"Deleting {len(project_ids_for_feature_delete)} features with project ids {joined_project_ids} in chunks of {UPLOAD_CHUNK_SIZE}..."
+                )
+                delete_features_by_project_ids(feature_type, delete_chunk)
+
+            features = all_features[feature_type]
+
+            logger.info(
+                f"Uploading {len(features)} features in chunks of {UPLOAD_CHUNK_SIZE}..."
```
johnclary, line 186: the console logging was throwing me off a bit here, because i encountered this:

```shell
 components-to-agol.INFO: Deleting 3 features with project ids 3392, 786, 3523 in chunks of 100...
 components-to-agol.INFO: Uploading 2 features in chunks of 100...
```

and i see that in certain cases there is a difference between the length of `project_ids_for_feature_delete` vs `features`—because `project_ids_for_feature_delete` includes IDs from all feature types (point/line/combined) while the `features` array is specific to the current `feature_type`. 

it's definitely not a bug, but i think you could make the AGOL deletes slightly more optimal by parsing the project IDs to delete out of the `features` list. i'm not sure though—i've confused myself by writing this comment 😛

moped-etl/arcgis/settings.py
johnclary: this is nice!

johnclary (changes requested):
@mddilley this is really great—nice work! i had to patch a little TypeError in order to get the script working, and after that it was smooth sailing.

this is such a dramatic improvement over the full replace. i can't wait to see this deployed 🙌

moped-etl/arcgis/components_to_agol.py
mddilley: good catch - i'll update 🙏

moped-etl/arcgis/components_to_agol.py
mddilley: nice - yes, even better! 🧹

moped-etl/arcgis/utils.py
mddilley: thanks for catching this detail 🙏

moped-etl/arcgis/components_to_agol.py
mddilley: it is confusing! but this makes sense to me and a result of my rushed effort to improve logging. 😅 I'll iron this out.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -1,11 +1,20 @@
 #!/usr/bin/env python
 """Copies all Moped component records to ArcGIS Online (AGOL)"""
 # docker run -it --rm  --network host --env-file env_file -v ${PWD}:/app  moped-agol /bin/bash
```
chiaberry, line 3: I went the docker way and not the virtual env way and had to update this to
`docker run -it --rm  --network host --env-file env_file  -v ${PWD}:/app  atddocker/atd-moped-etl-arcgis:production /bin/bash`

chiaberry (commented):
I ran into the same TypeError as John, but otherwise it was all smooth sailing! This is so exciting and a game changer for MUG and moped.
Screenshot 2024-06-28 at 1 52 20 PM

I tried updating a team members role and saw that the script updated the project in agol, and also updated a components description and saw it in agol too

moped-etl/arcgis/components_to_agol.py
chiaberry: I found it confusing too, trying to figure out what was happening, but then just trusted the script and moved on 😅

mddilley:
Just pushed a few commits and the logging now looks like:
```shell
components-to-agol.INFO: Processing lines features...
components-to-agol.INFO: Deleting all 24 existing features in lines layer for updated projects in chunks of 100...
components-to-agol.INFO: Deleting features with project ids 3392, 288, 290, 291, 292, 289, 3523, 2312, 214, 287
components-to-agol.INFO: Uploading 24 features in chunks of 100...
components-to-agol.INFO: Uploading chunk....
```

chiaberry (approved):
🍾

johnclary (approved):
yep—looks great!! deploy deploy deploy 🚢 🚢 🚢 🚢

roseeichelmann (approved):
followed all the test steps and they worked perfectly, this is super cool! :rocket: :champagne:

mddilley:
just made a small addition to the graphql request for the features in here to start populating `funding_sources` that Frank added in https://github.com/cityofaustin/atd-moped/pull/1370 when this all releases next Monday

PR #1370, opened 2024-06-28
Update funding source data for atsd

frankhereford:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1719582711923_refine_funding_source/up.sql
+++ b/moped-database/migrations/1719582711923_refine_funding_source/up.sql
@@ -16,9 +16,9 @@ CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
 funding_sources_lookup AS (
     SELECT
         mpf_1.project_id,
-        string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
-    FROM moped_proj_funding mpf_1
-    LEFT JOIN moped_fund_sources mfs ON mpf_1.funding_source_id = mfs.funding_source_id
+        string_agg(DISTINCT mfs.funding_source_name, ', ' ORDER BY mfs.funding_source_name) AS funding_source_name
+    FROM moped_proj_funding AS mpf_1
+    LEFT JOIN moped_fund_sources AS mfs ON mpf_1.funding_source_id = mfs.funding_source_id
     WHERE mpf_1.is_deleted = false
     GROUP BY mpf_1.project_id
 ),
--- a/moped-database/migrations/1719582711923_refine_funding_source/up.sql
+++ b/moped-database/migrations/1719582711923_refine_funding_source/up.sql
@@ -1,0 +1,507 @@
+DROP VIEW IF EXISTS component_arcgis_online_view;
+DROP VIEW IF EXISTS project_list_view;
+
+CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
+    SELECT
+        mpp.project_id,
+        string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members
+    FROM moped_proj_personnel AS mpp
+    INNER JOIN moped_users AS mu ON mpp.user_id = mu.user_id
+    INNER JOIN moped_proj_personnel_roles AS mppr ON mpp.project_personnel_id = mppr.project_personnel_id
+    INNER JOIN moped_project_roles AS mpr ON mppr.project_role_id = mpr.project_role_id
+    WHERE mpp.is_deleted = false AND mppr.is_deleted = false
+    GROUP BY mpp.project_id
+),
+
+funding_sources_lookup AS (
+    SELECT
+        mpf_1.project_id,
+        string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
+    FROM moped_proj_funding mpf_1
+    LEFT JOIN moped_fund_sources mfs ON mpf_1.funding_source_id = mfs.funding_source_id
+    WHERE mpf_1.is_deleted = false
+    GROUP BY mpf_1.project_id
+),
+
+project_type_lookup AS (
+    SELECT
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1719582711923_refine_funding_source/up.sql
```diff
@@ -0,0 +1,507 @@
...
+    INNER JOIN moped_proj_personnel_roles AS mppr ON mpp.project_personnel_id = mppr.project_personnel_id
+    INNER JOIN moped_project_roles AS mpr ON mppr.project_role_id = mpr.project_role_id
+    WHERE mpp.is_deleted = false AND mppr.is_deleted = false
+    GROUP BY mpp.project_id
+),
+
+funding_sources_lookup AS (
+    SELECT
+        mpf_1.project_id,
+        string_agg(DISTINCT mfs.funding_source_name, ', ' ORDER BY mfs.funding_source_name) AS funding_source_name
```
frankhereford, line 19: @mddilley -- Sorry if it's strange to tag you in a PR that I'm actively working on, but I wanted to ask you --

I'm not 100% sure you asked for this, but we did circle around the idea, at least, during our conversation yesterday. Do you like this change? For the existing funding column where it creates a comma-delimited list of funding sources names, do you like this change where i now am deduplicating them and ordering alphabetically?

moped-database/views/project_list_view.sql
```diff
@@ -15,7 +15,7 @@ CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
 funding_sources_lookup AS (
     SELECT
         mpf_1.project_id,
-        string_agg(mfs.funding_source_name, ', '::text) AS funding_source_name
```
frankhereford, line 18: @mddilley -- Sorry if it's strange to tag you in a PR that I'm actively working on, but I wanted to ask you --

I'm not 100% sure you asked for this, but we did circle around the idea, at least, during our conversation yesterday. Do you like this change? For the existing funding column where it creates a comma-delimited list of funding sources names, do you like this change where i now am deduplicating them and ordering alphabetically? 

FWIW, i know this issue is now written to add a new column and leave this one be, but since I'm in here .. 😉

moped-database/views/project_list_view.sql
```diff
@@ -15,7 +15,7 @@ CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
...
+        string_agg(DISTINCT mfs.funding_source_name, ', '::text ORDER BY mfs.funding_source_name) AS funding_source_name
```
frankhereford, line 18: @mddilley -- Sorry if it's strange to tag you in a PR that I'm actively working on, but I wanted to ask you --

I'm not 100% sure you asked for this, but we did circle around the idea, at least, during our conversation yesterday. Do you like this change? For the existing funding column where it creates a comma-delimited list of funding sources names, do you like this change where i now am deduplicating them and ordering alphabetically? 

FWIW, i know this issue is now written to add a new column and leave this one be, but since I'm in here .. 😉

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -182,6 +182,7 @@ SELECT
     plv.project_partners,
     plv.task_order_names,
     plv.funding_source_name,
+    plv.funding_source_and_program_names,
```
mddilley, line 185: @frankhereford I like that this is named `funding_source_and_program_names` in the project list view, but could you alias this as `funding_sources` here? That is the name that was added to the AGOL dataset and also aligns with the request from ATSD. 🙏

moped-database/views/project_list_view.sql
mddilley: this looks great! i like that it will declutter the view while also taking away duplicate info that doesn't have the extra context of why it isn't a true duplicate which confused me when we were first looking at this together. We'll also have the source + program column ready to go if we hear that info would be useful in this list too. 🙏

i tested advanced searching and several combinations of funding rows with the same program, etc. all works just as before!

roseeichelmann (approved):
these new columns look great to me! :ship:  :sparkles:

mddilley:
When i was testing, I noticed that both funding source and funding program columns in the `moped_proj_funding` table are both nullable. There are also real projects like 3019 that have funding rows with a program but no source.

I talked with Frank about it, and we agreed to show: only funding source, only funding program, or both depending on what is available. I'll take another look after that update. Thanks for handling this curveball, @frankhereford!

moped-database/migrations/1719582711923_refine_funding_source/up.sql
```diff
@@ -0,0 +1,531 @@
...
+            ', '
+            ORDER BY
+                CASE
+                    WHEN mfs.funding_source_name IS NOT null AND mfp.funding_program_name IS NOT null
+                        THEN concat(mfs.funding_source_name, ' - ', mfp.funding_program_name)
+                    WHEN mfs.funding_source_name IS NOT null
+                        THEN mfs.funding_source_name
+                    WHEN mfp.funding_program_name IS NOT null
+                        THEN mfp.funding_program_name
+                END
```
frankhereford, line 39: @mddilley I left three branches in here, where two of them -- the ones where it only has one of the two fields to display -- where two of them are broken out separately when they could be combined. My logic is that it leaves open an opportunity to add some sort of label to them, because as-is, it's not easy to tell if you're looking at a single program name or a single source name. LMK if you'd like these combined down into one.

mddilley: sounds great! I'll check back in on any refinements needed in my next check-in with Nathan in a couple weeks. Either way, I think the branching makes it easier to see what makes up this field. I've been trying to keep that in mind as we accumulate fields with more complex logic.

mddilley (approved):
I tested this with funding records with source only, with program only, both, or none, and I'm seeing the new column display all data available. Looks so great, and thanks for adding permissions that I missed in a previous PR. 🙏 🚢 🙌

PR #1373, opened 2024-07-08
Update the Team, Tags, and Status filters in the advanced search to use lookup tables

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/17598
https://github.com/cityofaustin/atd-data-tech/issues/18145

This PR:
- updates the **Team** advanced search to default to **contains** operator and support both arbitrary text input or selection from dropdown options
- updates the **Tags** advanced search to default to **contains** operator and support both arbitrary text input or selection from dropdown options
- removes **is** operator from **Team** and **Tags** advanced searches
- adds options to the **Status is** advanced search coming from the distinct values of `phase_name_simple` from the `moped_phases` table

## Testing
**URL to test:** 

https://deploy-preview-1373--atd-moped-main.netlify.app/moped/ or local

**Steps to test:**
1. Test advanced searching with the following fields:
   - Team
      - ~test **Team contains** and then type **m** in the autocomplete and press enter (don't choose from the dropdown). Press **Apply**. You should see results for **Teams contains m**. I'm really open to feedback on how useful this actually is. Do users need to search in this way? I'm on the fence about if this is intuitive or not.~ This functionality was removed
      - test **Team contains** and then start typing your name until you see an option appear that matches. Select the option and then click **Apply**. You should see results for **Team contains < your name >**
   - Tags
      - Repeat the same steps you did for **Team** but with tag values
   - Status
      - test **Status is Potential**, and you should see that this search now has the status options in the dropdown
   - Lead
      - test searching for **Lead is COA ATD** to make sure that the autocomplete still works like it does now in production
      - **is** feels like it should be the default operator for this field. What do y'all think?
   - Full name
      - test searching for **Full name contains test** to make sure the text input search still works like it does in production

---
#### Ship list
- [x] Code reviewed 
- [ ] ~Product manager added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable~


moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -1,16 +1,5 @@
 import { FiltersCommonOperators } from "../../../components/GridTable/FiltersCommonOperators";
 
-/**
- * Array of operators that should be autocompleted
- * @constant
- * @type {array}
- * @default
- */
-export const AUTOCOMPLETE_OPERATORS = [
```
mddilley, line 9: Removing this config shifts setting what operators use the autocomplete input from being global to being specifically set in each field configuration. This way some fields like **Tags** and **Teams** can use the autocomplete with `string_contains_case_insensitive` too

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -116,7 +105,11 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
       defaultOperator: "string_equals_case_insensitive",
       lookup: {
         table_name: "moped_types",
-        field_name: "type_name",
+        getOptionLabel: (option) => option.type_name,
```
mddilley, line 108: I added this function to the column-specific configuration so we can customize what keys the label uses out of the option objects for the dropdown choices. We used to set a `field_name` here in the config which was used but it only supported a single key. We need more than one for columns like **Team** which is first name + last name.

moped-editor/src/components/GridTable/helpers.js
```diff
@@ -51,18 +50,28 @@ export const isFilterNullType = (operator) => {
...
+  loading,
+  lookupOperators
 ) => {
-  return lookupTable && !loading && AUTOCOMPLETE_OPERATORS.includes(operator);
+  return lookupTable && !loading && lookupOperators.includes(operator);
 };
 
+/**
+ * Returns true if free solo should be used for the autocomplete input with the given operator
+ * Contains operators use free solo to support both arbitrary values and values from the dropdown suggestions
```
mddilley, line 69: I think this comment is confusing so I'm thinking about how to word this better. 😅

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -481,33 +499,18 @@ const Filters = ({
...
                     ) ? (
                       <Autocomplete
+                        freeSolo={shouldUseFreeSolo(operator)}
                         value={value || null}
-                        options={data[lookupTable]}
+                        options={data[lookupTable].map((option) =>
+                          getOptionLabel(option)
+                        )}
                         disabled={!filterParameters[filterIndex].operator}
-                        getOptionLabel={(option) =>
```
mddilley, line 490: A lot of these blocks of code had to do with determining what to do if a `field_name` was set in `PROJECT_LIST_VIEW_FILTERS_CONFIG` or not. The `getOptionLabel` function that is specific to each field now takes care of this.

moped-editor/src/components/GridTable/helpers.js
chiaberry: Operators that search on "contains" use free solo to support searching both arbitrary values and values from the dropdown suggestions 

is this any better? I was confused when I first read it b/c I thought it was the verb contains

chiaberry (approved):
I have tried to break it and I can't. I dont know if users want to be able to add arbitrary strings when looking for Team members, maybe if someone didnt know which John they were looking for (there are four users with the first name John), but this might be a stretch. 

for some of the autocomplete fields I don't see how "begins with" or "ends with" would be useful since users can't control the order the tags for example appear, can they?

roseeichelmann:
I wasnt able to put a screenshot cuz the window goes away but i find it odd that when im advanced searching on the Team filter and typing into the options text box, if i start typing "Amenity" i get like 5 identical "Amenity Applewhite"s to choose from. Also when typing my name, i got an option "rose eichelmann" that i clicked on but it didnt actually pull up any projects and i didnt have any other filters set.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -147,6 +140,11 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
       label: "Status",
```
roseeichelmann, line 140: i feel like we only need is and is not here as operators? and maybe is blank if its possible to have a blank status?

chiaberry:
@roseeichelmann hm that is weird that Amenity shows up so many times, when shes only in the staff list twice

the list comes from staff members, so thats why your name shows up even if you arent on any projects

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -38,6 +38,7 @@ import {
   handleApplyValidation,
   isFilterNullType,
   shouldRenderAutocompleteInput,
+  shouldUseFreeSolo,
```
roseeichelmann, line 41: my mind immediately went to https://en.wikipedia.org/wiki/Free_Solo

roseeichelmann (approved):
Couldnt break anything, test steps worked great! I agree with chia about the begins with and ends with operators for some of these fields, and just generally feel like some of these feels could be shortened to just having like 3 operators. :shrug:

mddilley:
@chiaberry Thx for the feedback on the free solo option. I'm going to remove it since I am also not seeing a lot of utility in it.

@roseeichelmann Thx for flagging the multiple users quirk! I'm going to do some testing because it almost seems to be collecting more values as I keep typing the same string in over and over. Might be related to the free solo 🧗‍♂️ option which I'm removing.

moped-editor/src/components/GridTable/Filters.js
mddilley: +1 such an incredible doc!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: I'm going to see if we have an issue related to this or create one if not. I agree that we could sweep through the searchable fields and take another look at whether they make sense! 🧹

mddilley:
Alright y'all - there is a bug in here which comes down to me tearing out the part of the Autocomplete that compares what options are available to what value is selected. The problem is that, as soon as there is a duplicate name (like Amenity's in staging), the component starts getting confused since many options satisfy the test `Test User === Test User`.

I think it is important to fix this since we are bound to have two people with the same first name and last name combo someday! I'll rework this in the morning, but I wanted to keep everyone in the loop.

moped-editor/src/components/GridTable/helpers.js
mddilley: forgot to respond here - that is way better and i wish that i didn't remove the free solo so i could have kept it 😅

mddilley:
I made a small adjustment that assumes that we aren't concerned with showing duplicate options here. I don't think our search needs to be sophisticated enough to tell the difference between two people records with the same details that translate to the same option label like `[{ id: 1, first: "Mike", last: "Dilley"}, { id: 2, name: "Mike", last: "Dilley"}]` which would appear in the UI as `Mike Dilley` and `Mike Dilley`. 

I also can't think of a case where duplicate tags or any other type of record would be relevant either.

chiaberry:
>never-ending Amenitys is an existing bug in MUI 5  

she will rule them all!!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
johnclary: nice!

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -377,12 +388,23 @@ const Filters = ({
...
         /* Check filter row validity */
         const isValidInput = checkIsValidInput(filter, type);
 
+        /* Use option formatter and dedupe items to handle cases like same team member name options
+         * See https://github.com/mui/material-ui/issues/26492
+         **/
+        const options = data?.[lookupTable]
+          ? data[lookupTable].map((option) => getOptionLabel(option))
+          : [];
+        const dedupedOptions = [...new Set(options)];
```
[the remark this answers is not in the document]
johnclary: this looks good, but it's catching my eye that we're doing a bunch of config re-structuring here, during render, that would be better suited for a memoized hook. i don't think it's critical by any means but it might be nice to address that now while code is fresh 🤷

johnclary (approved):
@mddilley this looks great—and is a major improvement to the search experience: so much better than typing "construction" into the status filter!

i haven't spent too much time with the code, so please take my memoization suggestion with a grain of salt 🙏

johnclary:
> I don't think our search needs to be sophisticated enough...

+1 - agreed @mddilley

mddilley:
@johnclary thanks for flagging the memoization. i think that is totally valid so i added it to the [next issue](https://github.com/cityofaustin/atd-data-tech/issues/18230) for other fields that we could improve with this contains + options behavior. Thank you!

PR #1377, opened 2024-07-19
17450 explore replacing react quill

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/17450

## Testing
**URL to test:** 

https://deploy-preview-1377--atd-moped-main.netlify.app/

**Steps to test:**
- go to a project, ideally one you commented on previously when we were using quill
- copy and paste a comment with rich text and links
- save the comment, it should appear in the comments section and the editor should clear out and refocus
- click edit, the editor should appear below the comment populated with the selected comment
- edit the comment and save
- try editing an old comment made using quill to confirm it loads in the new editor

this is a proof of concept at the moment, there are a few kinks to work out and features to add (resizing the editor to fit longer comments, adding a toolbar, a few other rich text and link functionalities). those will be addressed in a followup PR!

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/NoteInputQuill.js
```diff
@@ -78,25 +99,95 @@ const NoteInputQuill = ({
...
+    namespace: 'MyEditor',
+    theme,
+    nodes: [
+      LinkNode,
+    ],
+    onError,
+  };
+
+  // On change, return editor content as HTML
+  const OnChangePlugin = ({ onChange }) => {
```
mddilley, line 113: @tillyw I was taking a preview of this code to see how this works, and I noticed that these plugins are components in themselves. Could you move their definitions outside the `NoteInputQuill` component. You'll probably have to pass some props.

This is cool to see! The MUI DataGrid Pro work that Chia did recently used a similar pattern with a table context that was used in child components.

tillyw: thanks, @mddilley! i just pushed an update that makes these fixes and cuts out some unnecessary code. will request review shortly!

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,236 @@
...
+
+// On edit, load the selected comment and refocus the editor
+const OnEditPlugin = ({ htmlContent, editingNote }) => {
+  const [editor] = useLexicalComposerContext();
+  useEffect(() => {
+    const editorState = editor.getEditorState()
+    const nodeMap = editorState._nodeMap;
+    const textArray = [...nodeMap.entries()][2];
+    const isTextInEditor = !!textArray;
+    const isTextinDatabase = !!htmlContent.replace(/<[^>]*>/g, "");
```
johnclary, line 120: i'm trying to interpret what's going on here. i guess this regex is stripping out HTML characters in case someone saved a comment with hidden html?

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,236 @@
...
+    // If user clicks edit and there is a stored note and a clear editor,
+    // populate editor with stored note
+    editingNote && !isTextInEditor && isTextinDatabase &&
+      editor.update(() => {
+        $getRoot()
+          .getChildren()
+          .forEach((n) => n.remove());
```
johnclary, line 127: might be good to add comments to explain what's going on here 👀

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,236 @@
...
+  const classes = useStyles();
+
+  const initialConfig = {
+    namespace: 'MyEditor',
+    theme,
+    nodes: [
+      LinkNode,
+    ],
+    onError,
+  };
```
johnclary, line 164: it looks like this object can be moved outside of the component

johnclary (commented):
i tested this using the Netlify preview—it works as promised! I was able to copy rich text from elsewhere and paste it into the editor. i didn't notice any issues with rendering or editing any notes.

i left some feedback. i like where this is headed!

moped-editor/patches/react-scripts+5.0.1.patch
mddilley, line 1: I think it is great that you found a way around the build error that happened because we are still in React v17. I am going to start a list of things that we can clean up when we move to React v18 or maybe even Next.js 👿

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,236 @@
...
+  root: {
+    width: "100%",
+    backgroundColor: theme.palette.background.paper,
+  },
+  cancelButton: {
+    margin: theme.spacing(1),
+    position: "relative",
+    color: theme.palette.secondary.dark,
+  },
+  contentEditable: {
```
mddilley, line 37: This looks like a good place to add some styles to make this look more like MUI. Could you add the following? Not overly concerned about going full-blown MUI since the old Quill component had its own theme.
- border radius 4px
- border color the same as the primary text color
- font family to `theme.typography.fontFamily`

Also, could you remove the 300px height set here? The current production editor starts at one line of height and grows as needed. I think that we should keep that behavior to preserve vertical space unless we hear feedback that says otherwise. 🙏

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,236 @@
...
+
+  return (
+    <Container>
+      <Grid container direction="column" spacing={1}>
+        <Grid item xs={12} sm={12}>
+          <Box className={classes.editorWrapper} pt={2}>
+            <LexicalComposer initialConfig={initialConfig} >
+              <RichTextPlugin
+                contentEditable={<ContentEditable className={classes.contentEditable} />}
+                placeholder={<div className={classes.placeholder}>Enter some text...</div>}
```
mddilley, line 178: I like the new placeholder text! Could you add a style to it so that it uses the same font family as the rest of the app? You can access it through the `theme` object (i think it is `theme.typography.fontFamily`?)

moped-editor/src/views/projects/projectView/NoteInput.js
mddilley: I also noticed a regression that we fixed before where the font color of an existing comment in edit mode is the lighter secondary color instead of the primary text color. Looks like we need to explicitly set the color of the font here to be the theme primary text color. I think we determined that it is inheriting the secondary color from its wrapper? Could you add a comment to help us next time around? Thanks!

Screenshot 2024-08-09 at 10 26 31 AM

moped-editor/src/views/projects/projectView/NoteInput.js
mddilley: I think that is the case here. We use the same regex and `replace` call to disable the save button so we don't save empty HTML tags into the database. This was a Quill quirk so I wonder if we have the same issue with Lexical. Tilly, could you check if there is a graceful way to handle this in Lexical? If not, it would be awesome to pull out the two uses of this check into a reusable helper function and add a docstring to it.

Here is that comment:
`// disable save button if no text after removing html tags`

mddilley (changes requested):
This is looking great, Tilly! 🤩 Can't wait to see this with the toolbar! I requested some small styling changes to make this look a little more MUI-ish that you may have already had in mind since this PR is PoC. This works great for all the test steps!

mddilley:
also wanted to note that Lexical add the [`dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) to the parent `p` tags to show that we are using left to right text. The `spans` are set with [`white-space: pre-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#pre-wrap) too so that white space doesn't collapse when captured in the editor. I think both of these are improvements, but I wanted to see if anyone sees anything that I'm missing with the change.

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: after a bit of digging, I found a Lexical method called `$isRootTextContentEmpty` that checks whether a root node is empty, so I added that check _before_ serializing the node into HTML, so it will not be possible to save an empty tag into the db and I was able to drop the regex. thanks for pointing this out!

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: updated all of these, thank you for the guidance!! it looks much nicer ✨

moped-editor/patches/react-scripts+5.0.1.patch
mddilley, line 1: @tillyw I see that this is a diff of what patch-package is doing to webpack.config.js in our `node_modules`, but could you see if a comment at the beginning of this file would break the build?

Something like: `// Generated with patch-package to fix our build when using Lexical with React v17`

mddilley (changes requested):
I reviewed the changes that I requested and see that all were updated except two from John's review. Could you update the two that I tagged you in from John? And the one that I tagged you in?

Thank you for all your work for getting us to a better place with this editor. 🙏

moped-editor/src/views/projects/projectView/EditorTheme.css
```diff
@@ -0,0 +1,432 @@
+/**
+* Code sourced from Lexical Playground:
+* https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+*/
```
chiaberry, line 4: thank you!

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: yes - added a couple step by step comments to clarify this!

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: moved!

moped-editor/patches/react-scripts+5.0.1.patch
tillyw: added the comment and seems to have built without issue!

tillyw:
> I reviewed the changes that I requested and see that all were updated except two from John's review. Could you update the two that I tagged you in from John? And the one that I tagged you in?
> 
> Thank you for all your work for getting us to a better place with this editor. 🙏

done, thank you!!

chiaberry (approved):
Thank you for addressing my requests for more comments and documentation -- It will be good to get this in users hands to see what further refinements this could use!

PR #1378, opened 2024-07-22
Patch v2.15.1

mddilley:
## Associated issues
n/a

When double-clicking a cell and then clicking away to cancel without actually making a value change, the update mutation is still called and ends up creating an activity log entry with the same before and after values. This adds a check for updated form data before calling the mutation.

## Testing
**URL to test:** 

**Steps to test:**
1. Go to a project funding sources table with a funding row or create a row if there isn't one
2. Open your dev tools to watch for network requests then double-click a cell to edit. Without making a change, click outside the table to cancel. You should not see a network request happen.
3. Now, do the same but make an update. You should see a network request for the mutation.

---
#### Ship list
- [ ] Code reviewed 


chiaberry (approved):
thanks for catching it and the speedy fix!!

PR #1382, opened 2024-07-26
Add project development status and date fields to AGOL database view

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/17605

This adds two fields to `component_arcgis_online_view` called `project_development_status` and `project_development_status_date`. They are used to capture the next upcoming stage of project development as well as an estimated status date for ATSD reporting purposes.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Create a new project and add one component
2. Add a **Complete** phase to your project and mark it as the current phase and make sure it has a confirmed start date
3. Run the query below and you should see:
   - `project_development_status` is "Complete"
   - `project_development_status_date` is the start date of your **Complete** phase
4. Now uncheck the **Confirmed** box next to the start date of your **Complete** phase to make it estimated. Make a new **Construction** phase and mark it as the current phase.
5. Run the query below and you should see:
   - `project_development_status` is "Estimated End Date (In Construction)"
   - `project_development_status_date` is the estimated date in your **Complete** phase
6. Update your **Construction** phase to **Planned** (or any phase that does not have a `phase_name_simple` of Complete or Construction in the `moped_phases` table)
7. Run the query below and you should see:
   - `project_development_status` is "Estimated End Date"
   - `project_development_status_date` is still the estimated date in your **Complete** phase

8. Now remove all phases, and then add a milestone that is **Public Meeting** with an estimated completion date.
9. Run the query below and you should see:
   - `project_development_status` is "Estimated Public Meeting Date"
   - `project_development_status_date` is your estimated date
10. Add a **Date completed** date to this same milestone and re-run the query, and you should see the `project_development_status_date` update with the completed date.
11. Delete your milestone and add a phase that is **Construction** or **Scoping** (or any phase that has `phase_name_simple` of Active or Construction) and make it the current phase
12. Run the query below and you should see:
   - `project_development_status` is "Estimated Start of Project Development"
   - `project_development_status_date` is the date that you added to your phase (estimated or confirmed)
13. Test the down migration

```sql
SELECT
	project_id,
	project_component_id,
	project_development_status,
	project_development_status_date
FROM
	component_arcgis_online_view
WHERE
	project_id = <your project id>;
```

#### Ship list
- [ ] Code reviewed 


moped-database/views/component_arcgis_online_view.sql
```diff
@@ -114,22 +114,23 @@ related_projects AS (
     GROUP BY pmp.project_id
 ),
 
-min_phase_dates AS (
```
mddilley, line 117: This CTE was left behind in old work and is not used in this view anymore. 🗑️

moped-database/views/project_list_view.sql
```diff
@@ -196,14 +196,6 @@ SELECT
     work_activities.task_order_names,
     work_activities.task_order_names_short,
     work_activities.task_orders,
-    'placeholder text'::text AS project_development_status,
```
mddilley, line 199: These were mocked out in the project list view, but I have confirmed that they are only going to be used on the component level so I moved them into `componentn_arcgis_online_view`.

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -202,14 +203,25 @@ SELECT
...
+            THEN
+                CASE
+                    WHEN plv.current_phase_simple = 'Construction'::text THEN 'Estimated End Date (In Construction)'::text
+                    ELSE 'Estimated End Date'::text
+                END
+        WHEN lpmd.latest IS NOT null THEN 'Estimated Public Meeting Date'::text
+        WHEN eaocpd.earliest IS NOT null THEN 'Estimated Start of Project Development'::text
+        ELSE null::text
+    END AS project_development_status,
+    get_project_development_status_date(lpmd.latest::timestamp with time zone, eaocpd.earliest, mpc.project_id, coalesce(mpc.completion_date, plv.substantial_completion_date), plv.substantial_completion_date_estimated, current_phase.phase_name_simple)::text AS project_development_status_date,
```
mddilley, line 218: I went with a function here to calculate `project_development_status_date` since the next several fields in this view are different formats of the same value. If anyone can think of a better way to approach this than calling this function for each field, I'd love to hear about it. I wish that we could calculate based on other fields in the same SELECT.

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -114,22 +114,23 @@ related_projects AS (
...
-        SELECT
-            mpp.project_id,
-            min(mpp.phase_start) AS min_phase_start,
-            min(mpp.phase_end) AS min_phase_end
-        FROM moped_proj_phases mpp
-        LEFT JOIN moped_phases mp ON mpp.phase_id = mp.phase_id
-        WHERE mpp.is_phase_end_confirmed = false AND mpp.is_phase_start_confirmed = false AND mpp.is_deleted = false AND mp.phase_name_simple = 'Complete'::text
-        GROUP BY mpp.project_id
-    )
+latest_public_meeting_date AS (
```
mddilley, line 117: There is a comment in the migration but this CTE finds the latest completed or estimated "Public meeting" milestone date.

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -114,22 +114,23 @@ related_projects AS (
...
+    SELECT
+        mpm.project_id,
+        coalesce(max(mpm.date_actual), max(mpm.date_estimate)) AS latest
+    FROM moped_proj_milestones mpm
+    WHERE mpm.milestone_id = 65 AND mpm.is_deleted = false
+    GROUP BY mpm.project_id
+),
 
+earliest_active_or_construction_phase_date AS (
```
mddilley, line 126: This one finds the earliest estimated or confirmed date of any phase with a simple name that is “Active” or “Construction”.

johnclary (changes requested):
I ran into a potential issue during the second testing step. i have a **Complete** phase with an estimated start date, and I have a **Construction** phase with a confirmed start date. When I query the view i see the development status as `Estimated Start of Project Development`.

However, if I mark the start date of the **Complete** phase to confirmed, I see the status as `Estimated End Date (In Construction)`

Screenshot 2024-07-30 at 2 05 27 PM
Screenshot 2024-07-30 at 2 05 31 PM


I stopped my testing there, and I left some possibly helpful feedback about view SQL. thanks for working through this very complicated business logic, Mike!

mddilley:
y'all - i'm removing review requests to work out the bug that John found. i'll request again when i'm done - thanks!

moped-database/migrations/1719872865109_add_work_plan_category/up.sql
```diff
@@ -0,0 +1,551 @@
...
+
+-- Create function to determine project development status date and reuse for other date formats in view
+CREATE OR REPLACE FUNCTION public.get_project_development_status_date(latest_public_meeting_date timestamptz, earliest_active_or_construction_phase_date timestamptz, component_project_id int, substantial_completion_date timestamptz, substantial_completion_date_estimated timestamptz, current_phase_simple text)
```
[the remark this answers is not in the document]
mddilley: thanks for catching this! 🙏 i'll fix.

moped-database/migrations/1719872865109_add_work_plan_category/up.sql
```diff
@@ -0,0 +1,551 @@
...
+    plv.parent_project_id,
+    plv.parent_project_name,
+    plv.parent_project_url,
+    plv.parent_project_name AS parent_project_name_full,
+    rp.related_project_ids_with_self AS related_project_ids,
+    rp.related_project_ids_searchable_with_self AS related_project_ids_searchable,
+    plv.knack_project_id AS knack_data_tracker_project_record_id,
+    plv.project_url,
+    (plv.project_url || '?tab=map&project_component_id='::text) || mpc.project_component_id::text AS component_url,
+    CASE WHEN current_phase.phase_name_simple = 'Complete' THEN 'Complete'
```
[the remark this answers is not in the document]
mddilley: cool, i agree and that does sound better than jamming the case logic in here! another benefit that i always overlook is that the logic in the function can be updated later without dropping and replacing the whole view.

moped-database/views/component_arcgis_online_view.sql
mddilley: thanks for throwing some thoughts about this! very handy since you've been in SQL mode for a while with VZ 🙏 💪  Out of curiosity, I tested out calling the status date function for all the other status date fields and saw about a 100-150ms increase in query time on the full ~12000 rows. I'm going to keep thinking about this for the next issue because I agree that it isn't ideal - especially since there are a bunch of args to pass repeatedly. Just would look kinda yuck.

I'm also taking into account that we will rarely query the full dataset now that we have incremental updates, but I don't know if that is a really good reason to settle and not find a better way. 🤔

mddilley:
@johnclary thanks again for the testing. i updated:
- using `lower()` for the equality check
- created the function `get_project_development_status`
- removed the unused component id parameter 
- removed the check for `substantial_completion_date` that was causing the bug you ran into. not sure why I added that since the original VBA code this was translated from isn't concerned with it. I tested again and am seeing what I expect now.

chiaberry (approved):
I ran through the test steps, they all check out. Down migration works too. 

🚢

frankhereford (approved):
Woa, this was complex logic that you captured through SQL 🔥 

Thank you so much for the world class testing instructions. I followed them very carefully, and all the queries came back exactly like you predicted. I fussed with the dates a little along the way, and still - it worked just as you had described.  

I think the code looks awesome, ~but I have not reviewed it as closely as the testing instructions~ edit: I just finished reading it and +1, I think the functions makes a ton of sense. Esp looking at the [issue's main comment](https://github.com/cityofaustin/atd-data-tech/issues/17605) with the if-else decision tree. Also, thank you for picking off that CTE that wasn't being used anymore. 🧹 🙏 

I think this is fantastic.

johnclary (approved):
test steps all check out now—looks great! 🚢

mddilley:
thanks for your time and reviews on this one! merging now!

moped-database/views/component_arcgis_online_view.sql
johnclary: sorry, i missed this comment from you. yes, 150ms is for sure not a big deal, especially because no human is sitting around waiting for this view to load. maybe a different story if this was powering a UI component.

anyway, i came back to this thread because this exact situation _might_  be a case for a lateral join ([docs](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL)), since the lateral join allows you to "reference columns provided by preceding FROM items."

probably not worth the time to try and tune this any further but i did want to drop that here 👍

mddilley: i was staring at the docs trying to understand that perk of lateral joins the other day when looking at VZ code. this made it click, and i get it now! thanks, John!!!

PR #1391, opened 2024-08-12
Create generated columns to show fund name and FDU format in project funding table

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18450

This PR adds two generated columns to pull the fund name out of the `fund` jsonb column and pull the fund, department, and unit values out of the `fund` and `dept_unit` jsonb columns. We may change the way this data is stored in the future so these fields will give us a way to expose this data consistently over time.

## Testing
**URL to test:** 

Local only

**Steps to test:**
1. Create a new project and add a new funding row in the Funding sources table within the Funding tab. Don't add values to the **Funding** or **Dept-unit** fields in the form yet. Run the SQL shown below on the local database, and you should see that both `fund_dept_unit` or `fund_name` are `null`.
2. Now, add a value for **Funding**. Run the SQL shown below on the local database, and you should see that `fund_name` now shows the value held in the json in the `fund` column. `fund_dept_unit` should still be `null` since there isn't complete data to populate it.
3. Now, choose a value for **Dept-unit**. When you run the sql one last time, you should see both `fund_name` and `fund_dept_unit` have values now, and they should match the values in the `fund` and `unit` columns of the row.
5. Test the down migration.

```sql
SELECT
	project_id,
	fund,
	dept_unit,
	fund_dept_unit,
	fund_name
FROM
	moped_proj_funding
WHERE
	project_id = <your project id>;
```

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
+++ b/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
@@ -1,0 +1,8 @@
+-- Add generated columns for fund_dept_unit and fund_name
+ALTER TABLE moped_proj_funding
+ADD COLUMN fund_dept_unit text GENERATED ALWAYS AS (
+    coalesce(fund ->> 'fund_id', ' ') || ' ' || coalesce(dept_unit ->> 'dept', ' ') || ' ' || coalesce(dept_unit ->> 'unit', ' ')
+) STORED,
+ADD COLUMN fund_name text GENERATED ALWAYS AS (
+    coalesce(fund ->> 'fund_name', ' ')
+) STORED;
--- a/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
+++ b/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
@@ -1,8 +1,8 @@
 -- Add generated columns for fund_dept_unit and fund_name
 ALTER TABLE moped_proj_funding
-ADD COLUMN fund_dept_unit text GENERATED ALWAYS AS (
-    coalesce(fund ->> 'fund_id', ' ') || ' ' || coalesce(dept_unit ->> 'dept', ' ') || ' ' || coalesce(dept_unit ->> 'unit', ' ')
-) STORED,
+ADD COLUMN fund_dept_unit text GENERATED ALWAYS AS (CASE WHEN (fund IS null OR dept_unit IS null) THEN null ELSE
+        coalesce(fund ->> 'fund_id', ' ') || ' ' || coalesce(dept_unit ->> 'dept', ' ') || ' ' || coalesce(dept_unit ->> 'unit', ' ')
+END) STORED,
 ADD COLUMN fund_name text GENERATED ALWAYS AS (
-    coalesce(fund ->> 'fund_name', ' ')
+    CASE WHEN fund IS null THEN null ELSE coalesce(fund ->> 'fund_name', ' ') END
 ) STORED;
--- a/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
+++ b/moped-database/migrations/1723496266685_add_fdu_columns/up.sql
@@ -6,3 +6,6 @@ END) STORED,
 ADD COLUMN fund_name text GENERATED ALWAYS AS (
     CASE WHEN fund IS null THEN null ELSE coalesce(fund ->> 'fund_name', ' ') END
 ) STORED;
+
+COMMENT ON COLUMN moped_proj_funding.fund_dept_unit IS 'Fund, department, and unit numbers concatenated; null if fund or unit is not populated';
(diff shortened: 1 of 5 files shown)


moped-database/migrations/1723496266686_add_bike_share_station_comp/down.sql
frankhereford, line 1: This down migration failed --

```
ERRO skipping applying migrations on database '', encountered:                                                                                                                                                                                                                                 
{
  "error": "query execution failed",
  "path": "$",
  "code": "postgres-error",
  "internal": {
    "arguments": [],
    "error": {
      "description": "Failing row contains (12663, 3626, null, null, f, null, null, null, null, 2024-08-14 16:15:35.214612+00, null, null, 68, 68, 2024-08-14 16:18:17.370734+00).",
      "exec_status": "FatalError",
      "hint": null,
      "message": "null value in column \"component_id\" violates not-null constraint",
      "status_code": "23502"
    },
    "prepared": false,
    "statement": "DELETE FROM moped_components\nWHERE moped_components.component_name = 'Bike Share Station';\n"
  }
} 
FATA[0000] applying migrations failed on database(s):   
```

I think you may have to include the hard-deletion of components which are of the type added in the up migration, but that is just theory.

frankhereford (changes requested):
All the good stuff is great here -- works great, looks great.

It's only on the down migration that I bumped into a problem. If you follow the instructions, and create a new component of the new 'Bike Share Station' type, you enter into a place where you can no longer remove the Bike Share Station component class. And because we use soft-deletes on components, even if you remove it in the app, it's still there from the DB's point of view.

I know this is not the best timing on a "request changes" review. If I can help you resolve this on your behalf to protect your time, I'd be more than happy to.

moped-database/migrations/1723496266685_add_fdu_columns/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Add generated columns for fund_dept_unit and fund_name
+ALTER TABLE moped_proj_funding
+ADD COLUMN fund_dept_unit text GENERATED ALWAYS AS (CASE WHEN (fund IS null OR dept_unit IS null) THEN null ELSE
+        coalesce(fund ->> 'fund_id', ' ') || ' ' || coalesce(dept_unit ->> 'dept', ' ') || ' ' || coalesce(dept_unit ->> 'unit', ' ')
+END) STORED,
```
johnclary, line 5: this looks good. i did a quick scan of all the dept-units we have in the public dataset, and i didn't see a case where the `dept` or `unit` values are null/empty. if there were any empty values, we could end up with some generated values with extra whitespace, and that might start to be an issue if the downstream/PowerBI users are trying to join these records with other data. i think what you have here is a great solution 🚢

mddilley:
@frankhereford It is always a good time for requested changes! Thanks for catching that, and I'm looking at it right now. 🙏

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO moped_components (
+    component_name, line_representation, feature_layer_id
+) VALUES
+('Bike Share Station', FALSE, 5);
```
johnclary, line 4: @mddilley i think you'll need to add some more inserts here to associate this component with the work types and subcomponent (if any) that should be available int he UI. e.g., it has no work types available currently:

Screenshot 2024-08-14 at 12 58 12 PM

this migration probably has some boilerplate you can copy:
https://github.com/cityofaustin/atd-moped/blob/main/moped-database/migrations/1698950636188_create_shared_use_trail_dual_path_component/up.sql

johnclary (changes requested):
looks good—just a few more things needed to bring that component to life 🚲

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
mddilley: yep, i'll update this so that the **required** field can be filled out. 🤦‍♂️  Thanks for catching this because I only thought about the subcomponents, and I forgot that the work types worked this way. 🙏 🙌

mddilley:
@frankhereford @johnclary Thanks for your reviews on this today! I pulled out the code for the Bike Share Station because it is feeling like a lot to me, and I'd rather focus on getting this FDU code shipped tomorrow. I'll target the other issue for a patch or the next release.

Just re-requested y'all's review 🙏

johnclary (approved):
🚢 🚢

frankhereford (approved):
# 🌬️ ⛵ 

Thanks Mike!! Looking forward to the rest of the code making it back in!

PR #1393, opened 2024-08-19
Add Bike Share Station component

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/18589

This PR adds a new component type.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start up the local stack
2. Add the new **Bike Share Station** component to a project and notice that the point icon shows next to it in the **Component Type** dropdown
3. Make sure that the work types available in the component form are: **New, Replacement, and Modification**
4. Make sure the **Subcomponents** dropdown is not shown in the component form since this component type has none
5. Make sure that you can select and draw points on the map when creating and editing
6. Test the down migration

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
+++ b/moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
@@ -1,0 +1,6 @@
+INSERT INTO moped_components (
+    component_name, line_representation, feature_layer_id
+) VALUES
+('Bike Share Station', FALSE, 5);
+
+-- TODO: Add work types for bike share station
--- a/moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
+++ b/moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
@@ -1,6 +1,37 @@
-INSERT INTO moped_components (
-    component_name, line_representation, feature_layer_id
-) VALUES
-('Bike Share Station', FALSE, 5);
+-- add bike share station component type (point) to moped_components table if it doesn't already exist
+INSERT INTO moped_components (component_name, line_representation, feature_layer_id)
+SELECT
+    'Bike Share Station',
+    FALSE,
+    5
+WHERE NOT EXISTS (
+        SELECT *
+        FROM
+            moped_components
+        WHERE (component_name, line_representation, feature_layer_id) = ('Bike Share Station', FALSE, 5)
+    );
 
--- TODO: Add work types for bike share station
+-- insert new, mod, and replacement work types for bike share station in moped_component_work_types table
+WITH inserts_todo AS (
+    SELECT
+        mwt.id AS work_type_id,
+        mc.component_id AS component_id
+    FROM
+        moped_work_types AS mwt,
+        moped_components AS mc
+    WHERE
+        mwt.name IN (
+            'New',
+            'Modification',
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
```diff
@@ -0,0 +1,37 @@
+-- add bike share station component type (point) to moped_components table if it doesn't already exist
+INSERT INTO moped_components (component_name, line_representation, feature_layer_id)
+SELECT
+    'Bike Share Station',
+    FALSE,
+    5
+WHERE NOT EXISTS (
```
mddilley, line 7: this was a TIL for me to check for an existing value before inserting. I have only used `ON CONFLICT` to do something like this before.

chiaberry (approved):
able to add bike share station and the test steps check out!

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
johnclary: +1 — good to know of this approch!

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
```diff
@@ -0,0 +1,37 @@
...
+    );
+
+-- insert new, mod, and replacement work types for bike share station in moped_component_work_types table
+WITH inserts_todo AS (
+    SELECT
+        mwt.id AS work_type_id,
+        mc.component_id AS component_id
+    FROM
+        moped_work_types AS mwt,
+        moped_components AS mc
```
johnclary, line 21: this is also a TIL for me—the implicit `CROSS JOIN` ⚔️

johnclary (approved):
yep—looks great + works great 🚢 🚢

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
roseeichelmann: why do we need to check if the component already exists in the table? just so that if we run the down migration and then up migration again we wouldnt double insert?

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
```diff
@@ -0,0 +1,37 @@
...
+        SELECT *
+        FROM
+            moped_components
+        WHERE (component_name, line_representation, feature_layer_id) = ('Bike Share Station', FALSE, 5)
+    );
+
+-- insert new, mod, and replacement work types for bike share station in moped_component_work_types table
+WITH inserts_todo AS (
+    SELECT
+        mwt.id AS work_type_id,
```
roseeichelmann, line 17: kinda confused about how this sql is working, how does the code know what table `mwt` or `mc` is? what am i missing :sweat_smile:

roseeichelmann (approved):
works great :sparkles:

moped-database/migrations/1723496266686_add_bike_share_station_comp/up.sql
mddilley: yep - that is exactly why! thinking about how to avoid inserting twice became a fun detour.

PR #1394, opened 2024-08-20
Update project list view advanced search funding sources

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/18361

## Update

Going to use this PR to make some small changes to the funding sources filter and come back to the funding program after some more changes. 

To test: 
https://deploy-preview-1394--atd-moped-main.netlify.app/moped/projects?filters=%5B%7B%22field%22%3A%22funding_source_name%22%2C%22operator%22%3A%22string_is_null%22%2C%22value%22%3Anull%7D%5D&isOr=false

Confirm the filter options for the funding sources have been pared down. 

---
## old description

The issue description says both funding source and funding program should search on `funding_source_and_program_names`, but I ran into issues with two different entries in the config searching on the same field. Rather than refactor the entire filter config I instead left the funding source searching on the funding_source_name field. 

There is a quirk (or maybe its a bug?) with having the funding program search on the entire source and program name string: if someone searches "is blank", the results are projects that have neither a source nor a program. This project has only a source: https://deploy-preview-1394--atd-moped-main.netlify.app/moped/projects/1802?tab=funding, but will not come up when searching is program blank. 

Also the AND search behavior may be more restrictive than if both options were searching on the same field, but this feels more correct to me. 

## Testing
**URL to test:** 

https://deploy-preview-1394--atd-moped-main.netlify.app/moped/projects?filters=%5B%7B%22field%22%3A%22funding_source_and_program_names%22%2C%22operator%22%3A%22string_contains_case_insensitive%22%2C%22value%22%3A%22Bikeways%22%7D%2C%7B%22field%22%3A%22funding_source_name%22%2C%22operator%22%3A%22string_contains_case_insensitive%22%2C%22value%22%3A%222012+Bond%22%7D%5D&isOr=false

**Steps to test:**

The link above has a search with Funding source and Funding program. 
Test searching with only one or the other. 
Try searching is blank, is not blank. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -258,7 +258,7 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
       ],
     },
     {
-      name: "funding_source_name",
+      name: "funding_source_and_program_names",
       label: "Funding source",
       placeholder: "Funding source",
       type: "string",
@@ -274,10 +274,29 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
         "string_contains_case_insensitive",
         "string_begins_with_case_insensitive",
         "string_ends_with_case_insensitive",
-        "string_equals_case_insensitive",
-        "string_does_not_equal_case_insensitive",
         "string_is_null",
-        "string_is_not_null",
+        "string_is_not_null"
+      ],
+    },
+    {
+      name: "funding_source_and_program_names",
+      label: "Funding program",
+      placeholder: "Funding program",
+      type: "string",
+      lookup: {
+        table_name: "moped_fund_sources",
+        getOptionLabel: (option) => option.funding_source_name,
+        operators: [
+          "string_equals_case_insensitive",
+          "string_does_not_equal_case_insensitive",
+        ],
+      },
+      operators: [
+        "string_contains_case_insensitive",
+        "string_begins_with_case_insensitive",
+        "string_ends_with_case_insensitive",
+        "string_is_null",
+        "string_is_not_null"
       ],
     },
     {
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -274,22 +274,6 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
       ],
     },
     {
-      name: "funding_source_and_program_names",
-      label: "Funding program",
-      placeholder: "Funding program",
-      type: "string",
-      lookup: {
-        table_name: "moped_fund_programs",
-        getOptionLabel: (option) => option.funding_program_name,
-        operators: ["string_contains_case_insensitive"],
-      },
-      operators: [
-        "string_contains_case_insensitive",
-        "string_is_null",
-        "string_is_not_null",
-      ],
-    },
-    {
       name: "project_status_update",
       label: "Status update",
       placeholder: "Status update",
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -265,17 +265,26 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
...
+        operators: ["string_contains_case_insensitive"],
+      },
+      operators: [
+        "string_contains_case_insensitive",
+        "string_is_null",
+        "string_is_not_null",
+      ],
+    },
+    {
+      name: "funding_source_and_program_names",
```
mddilley, line 277: thanks for hitting the brakes when you realized that we can't have two identical names in the config! I reproduced the quirk that you mentioned locally, and I see how the code does not like having two identical names and starts populating source options for programs. 🫠

How do you feel about keeping the changes to the funding sources in this PR and then we regroup on the funding program advanced search? I'm thinking that we should add another column to the view that is a concatenation of funding program names like we have for funding sources for that work.

That way, we could advanced search for:
- sources
- programs
- sources and programs
- sources or programs
- projects with no sources (Funding sources is null)
- projects with no programs (Funding programs is null)

I could use another set of eyes to see if this sounds right!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -265,17 +265,26 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
...
+        table_name: "moped_fund_programs",
+        getOptionLabel: (option) => option.funding_program_name,
+        operators: ["string_contains_case_insensitive"],
       },
       operators: [
         "string_contains_case_insensitive",
-        "string_begins_with_case_insensitive",
-        "string_ends_with_case_insensitive",
-        "string_equals_case_insensitive",
-        "string_does_not_equal_case_insensitive",
```
mddilley, line 278: Thanks for removing these! I definitely think we should keep this change in this PR since it is a search on a concatenated list.

mddilley (commented):
Thanks for catching the incompatibility of the issue requirements with the way we have this implemented! 🙏 I dropped some thoughts about spinning off the program part of the issue and shipping the rest of the changes that I could use some feedback on.

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: Here is the spinoff issue https://github.com/cityofaustin/atd-data-tech/issues/18763

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -259,23 +259,16 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
...
       placeholder: "Funding source",
       type: "string",
       lookup: {
         table_name: "moped_fund_sources",
         getOptionLabel: (option) => option.funding_source_name,
-        operators: [
-          "string_equals_case_insensitive",
-          "string_does_not_equal_case_insensitive",
-        ],
+        operators: ["string_contains_case_insensitive"],
```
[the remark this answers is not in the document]
chiaberry: i like does not contain better

contains not sounds a little shakespearean 😬

roseeichelmann (approved):
I like the pared down operators. Also, i noticed that this is one of the only fields where the advanced search label "Funding sources" does not match the column header "Funding", which is maybe the tiniest bit confusing?

mddilley:
@roseeichelmann sorry to leave your comment hanging! I agree that it could be confusing. We will be updating the advanced search with a new field for **Funding Programs** too so I added a note in https://github.com/cityofaustin/atd-data-tech/issues/18763. Thanks!!!

PR #1395, opened 2024-08-23
Refactor Funding table columns and autocompletes

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/18669

I reorganized some of the files, made a folder to keep the Funding Table components, and moved the autocomplete components out of the funding table file. I also memoized the columns and fixed the autofocus when adding a new line to the table. 

## Testing
**URL to test:** 

https://deploy-preview-1395--atd-moped-main.netlify.app/moped/projects/2035?tab=funding

**Steps to test:**

Add a new funding source, note that the source field autofocuses. Test out the autocomplete inputs, edit an existing row, etc. Kick the moped tires. 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -382,7 +403,7 @@ const ProjectFundingTable = () => {
     ]);
     setRowModesModel((oldModel) => ({
       ...oldModel,
-      [id]: { mode: GridRowModes.Edit, fieldToFocus: "source" },
+      [id]: { mode: GridRowModes.Edit, fieldToFocus: "funding_source_id" },
```
chiaberry, line 406: I had the wrong field name when I initially set the focus

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -130,9 +106,164 @@ const useFdusArray = (projectFunding) =>
...
+        },
+      },
+    ];
+  }, [
+    data,
+    rowModesModel,
+    handleDeleteOpen,
+    handleSaveClick,
+    handleCancelClick,
+    handleEditClick,
```
johnclary, line 262: i don't think the memoization is working here, because these handler functions are re-created on every render. wrapping them in useCallback() should do the trick 🤷

moped-editor/src/views/projects/projectView/ProjectFunding/LookupAutocompleteComponent.js
```diff
@@ -0,0 +1,70 @@
...
+    padding: theme.spacing(1),
+  },
+}));
+
+/**
+ * Component for dropdown select using a lookup table as options
+ * @param {*} props
+ * @returns {React component}
+ */
+const LookupAutocompleteComponent = (props) => {
```
johnclary, line 21: slick component! it'd be nice to have the props destructured just so it's easy to see what need to get passed to this function

johnclary (approved):
tires kicked ✅  looks great. i left some feedback. 🚢 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
chiaberry: I wrapped them in useCallback and added their dependencies. 

How do you check to see if they are being re-created on every render?

moped-editor/src/views/projects/projectView/ProjectFunding/LookupAutocompleteComponent.js
chiaberry: destructured!

moped-editor/src/views/projects/projectView/ProjectFunding/FundAutocompleteComponent.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+  };
+
+  return (
+    <Autocomplete
+      className={classes.fundSelectStyle}
+      value={value ?? null}
+      // use customized popper component so menu expands to fullwidth
+      PopperComponent={CustomPopper}
+      id={"moped_funds"}
```
mddilley, line 40: tiny tiny thing but these curly braces can go.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -89,19 +83,6 @@ const useStyles = makeStyles((theme) => ({
     top: "1rem",
     right: "1rem",
   },
-  deptAutocomplete: {
```
mddilley, line 92: oh nice! i see that these styles weren't needed anymore. 👋

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
mddilley: good catch, John! yes, i see how these functions would have caused the memo to recalculate every render since their reference would update every render too.

as far as testing goes, i think you could log inside the memo and compare how often you see the logging or something similar. my short and lazy answer is because the docs say so. 🙃

mddilley (approved):
I tested this every which way possible, and I see the funding table functionality intact. Thanks for taking another look at this component and making it all ✨✨✨!!! 🚢

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
johnclary: ditto 🙃  the way i tested this is by adding console logging within the hook and compared that to when i completely deleted the dependency array (square brackets and all)—when there is no dep array any hook will run on every render

PR #1397, opened 2024-08-27
Add 4th layer to the Moped AGOL resource to house exploded multipoint geometry as points

frankhereford:
# Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/17999

# Testing

* Get on this branch and replicate the production DB to your machine
* Make yourself a `env_file` or ask me to `wormhole` you one.
   * Thanks to @mddilley for observing that those with existing env files may have to update this line as follows:
```
HASURA_ENDPOINT=http://host.docker.internal:8080/v1/graphql
```
* `cd moped-etl/arcgis/`
* `docker compose run arcgis;`
  * This will run with the default argument, a sole `-d`, which means update the existing data with changes from the last seven days.
* `docker compose run arcgis -d 2024-08-24T18:33:51.026046+00:00` 
  * This will do a similar data update, handling changes since the supplied datetime.
* `docker compose run arcgis -f`
  * This is a full replace. This is painful, as it's more likely than not that you'll encounter an AGOL error, on their end. When this happens, you just have to try again. And you should try again until it works so the data is restored to the correct state.  It took me maybe 10 tries over the course of half an hour to finally get through a full replace, but it did do it eventually.
    * Update on Sept 10: I just ran this doing a full replace and got it first try, so YMMV for the better. 🤞 
* Review the data on AGOL: https://austin.maps.arcgis.com/home/item.html?id=1c084c8756a84e6db7e2796c98c850a2

# Thanks

Thanks to @mddilley for his help on this one -- brainstorming and rubber ducking. 🙏 🦆 🙏 

Also thanks to @johnclary for his input and request to stick with our tried and true patterns. 💯 

---
#### Ship list
- [x] Code reviewed 


moped-etl/arcgis/utils.py
```diff
@@ -11,6 +11,7 @@
 AGOL_USERNAME = os.getenv("AGOL_USERNAME")
 AGOL_PASSWORD = os.getenv("AGOL_PASSWORD")
 HASURA_ENDPOINT = os.getenv("HASURA_ENDPOINT")
+HASURA_SQL_ENDPOINT = os.getenv("HASURA_SQL_ENDPOINT")
```
frankhereford, line 14: This will require a PR to airflow to include this new environment variable. I guess it could be inferred from the `HASURA_ENDPOINT`, but this felt cleaner to define like everything else.

moped-etl/arcgis/utils.py
```diff
@@ -64,6 +65,33 @@ def make_hasura_request(*, query, variables=None):
...
+    }
+    payload = {"type": "run_sql", "args": {"sql": query}}
+
+    res = requests.post(HASURA_SQL_ENDPOINT, json=payload, headers=headers)
+    res.raise_for_status()
+    data = res.json()
+    try:
+        return data["result"]
+    except KeyError:
+        raise ValueError(data)
```
frankhereford, line 92: New sibling function to 'make_hasura_request()` that takes SQL and executes it and returns the result. Makes me wonder what secures this from people writing arbitrary queries to it with a token extracted from the moped web app. I hope it's restricted to having the `HASURA_ADMIN_SECRET`. But anyway, that's entirely out of the scope of this issue.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -63,23 +65,39 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
-        # extract geometry and line geometry from component data
-        # for line features, the line geometry is redundant/identical to geometry
+
+        # extract geometry and line geometry from component data.
+
+        # for line features, the line geometry is redundant/identical to geometry.
+
         # for point features, it is the buffered ring around the point as defined
-        # in the Moped component view
+        # in the Moped component view.
```
frankhereford, line 99: I added some whitespace here to make these sentences clearly distinct.

moped-etl/arcgis/settings.py
```diff
@@ -1,6 +1,96 @@
 UPLOAD_CHUNK_SIZE = 100
 
-LAYER_IDS = {"points": 0, "lines": 1, "combined": 2}
+LAYER_IDS = {"points": 0, "lines": 1, "combined": 2, "exploded": 3}
+
+EXPLODED_GEOMETRY_SQL_QUERY = """
+SELECT 
+    --ST_GeometryType(dump.geom) AS geometry_type,
+    --dump.path[1] AS point_index, -- ordinal value of the point in the MultiPoint geometry
```
frankhereford, line 8: Leave these in? They feel useful and show how to produce data that doesn't feel obviously available to me -- especially the second line.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -1,17 +1,20 @@
 #!/usr/bin/env python
 """Copies all Moped component records to ArcGIS Online (AGOL)"""
-# docker run -it --rm  --network host --env-file env_file -v ${PWD}:/app  moped-agol /bin/bash
+# docker compose run arcgis;
```
frankhereford, line 3: I gave this a docker compose file for ease of development. I also have removed the `network=host` configuration. I can't see a reason why this needs to be available on the host's network and not tucked away in a docker environment.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -195,15 +255,17 @@ def main(args):
         "-d",
         "--date",
         type=str,
+        nargs="?",
+        const=(datetime.now(timezone.utc) - timedelta(days=7)).isoformat(),
         default=None,
-        help=f"ISO date string with TZ offset (ex. 2024-06-28T00:06:16.360805+00:00) of latest updated_at value to find project records to update.",
+        help="ISO date string with TZ offset (ex. 2024-06-28T00:06:16.360805+00:00) of latest updated_at value to find project records to update. Defaults to 7 days ago if -d is used without a value.",
```
frankhereford, line 261: I gave the `-d` option a default argument of "the last week". This was handy during development and feels like a genuine enhancement, so leave it in?

If you specify the date like usual, the behavior is unchanged of course.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -115,17 +129,81 @@ def make_all_features(data):
...
+                except ValueError:
+                    try:
+                        obj[field] = float(value)
+                    except ValueError:
+                        obj[field] = value
+        result.append(obj)
+
+    return result
+
+
```
frankhereford, line 198: this function both changes the shape of the data coming back to an array of dictionaries, but also casts the data back into ints, floats and strings due to the "creative" use of the SQL endpoint. cc/ @mddilley

johnclary (dismissed):
It's awesome to see this new layer come to life—and of course nice to get some TILs from your postgis skills, Frank!

I'm requesting changes because I'd really strongly prefer that we move this query into our graphql API. I also spotted what i think is an opportunity to simplify this new query by using the attribute data from the pre-existing graphql query. hopefully that helps!

moped-etl/arcgis/components_to_agol.py
frankhereford: My rational was that it would have been a 3 views deep, meaning any changes to the project list view mean dropping it, the 1st AGOL view, and this. Then make the change and have to build them all back up. That coupled with my mistaken belief that this was intended usage when I started got me here. The docs have an example of querying data out with using SQL via this part of the graphql-engine, but in retrospect, it's not meant for anything other than the most simple of queries. 

Anyway, it's all good - i'm removing the request for reviews - Mike gave me the same direction to go back and put a view in, so, I'm working on that today. 

Further down the road, I think a better solution might be to rework the AGOL view so that records don't represent components, but rather they represent single geometries, and then you can aggregate them into components much more easily than exploding pre-aggregated stuff back into single geometries. But -- another view will get us over the finish line and on to the next thing.

Thanks for you input -- you are undoubtedly correct, this is a bit to much to ask of that endpoint usage.

johnclary: Sounds good—thanks, Frank.

frankhereford: This brought its own set of TILs, but this choice was 100% for the best. Thanks again for this request John. 🙏

moped-etl/arcgis/components_to_agol.py
```diff
@@ -59,27 +62,38 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
     return feature
 
 
-def make_all_features(data):
+def make_all_features(data, exploded_geometry):
     """Take a list of component feature records and create Esri feature objects for lines, points, and combined layers in AGOL.
 
     Args:
         data (dict): a list of component feature records
+        exploded_geometry (dict): a dictionary of exploded geometry data from the component_arcgis_online_view
```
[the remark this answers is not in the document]
frankhereford: Great call - thank you. 591b7d9df57ef4d27c2f64e800305e653f9ec4b8

moped-etl/arcgis/components_to_agol.py
```diff
@@ -115,17 +129,81 @@ def make_all_features(data):
...
             all_features["combined"].append(line_feature)
+
+            project_component_id = feature["attributes"]["project_component_id"]
+
+            # Find the elements in exploded_geometry which match the project_component_id
+            matching_elements = [
+                element
+                for element in exploded_geometry
+                if element["project_component_id"] == project_component_id
+            ]
```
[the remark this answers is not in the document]
frankhereford: Yea, that's a good thought for sure, because, at the very least, it'd save bandwidth not pulling down duplicates of the data, just grabbing the geometry and a primary key. 

Since this is an ETL though, I think it's less of a concern as it should be if this were in a reactive application, but I could easily be convinced that it's worth it. @mddilley, do you have direction on if you'd like to see that optimization that John suggested?

moped-database/metadata/tables.yaml
```diff
@@ -1,6 +1,16 @@
...
+  array_relationships:
+    - name: componentToExplodedGeometryComponent
+      using:
+        manual_configuration:
+          column_mapping:
+            project_component_id: project_component_id
+          insertion_order: null
+          remote_table:
+            name: exploded_component_arcgis_online_view
+            schema: public
```
frankhereford, line 13: This is a real relationship, and it's usable when you're looking at a handful of AGOL records, but it's not performant enough to use when asking for the full table. 

Because of this, I am not using the relationship in the program, but rather, just executing two queries. 

But since it's a genuine and real relationship, I'm thinking we should keep it. But I would be happy to zap it out of there too.

moped-etl/arcgis/components_to_agol.py
mddilley: thanks for leveling up the dev environment!

moped-etl/arcgis/components_to_agol.py
```diff
@@ -214,6 +260,11 @@ def main(args):
             "Please provide either the -d flag with ISO date string with TZ offset or the -f flag and not both."
         )
 
+    if not args.date and not args.full:
```
mddilley, line 263: 🙏

mddilley (commented):
About to start testing the ETL but wanted to get these comments out while I do that. This is looking great!

moped-etl/arcgis/components_to_agol.py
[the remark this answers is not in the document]
mddilley: One advantage that I could see here is having to only update one of the queries with a field when it comes time to add a new one. Does this feel like a big lift?

moped-etl/arcgis/components_to_agol.py
```diff
@@ -53,33 +55,52 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
         },
     }
-    feature["geometry"][esri_geometry_key] = geometry["coordinates"]
+    if (esri_geometry_key == "points") or (esri_geometry_key == "paths"):
+        feature["geometry"][esri_geometry_key] = geometry["coordinates"]
+    elif esri_geometry_key == "point":
+        geometry = json.loads(geometry)
+        feature["geometry"]["y"] = geometry["coordinates"][1]
+        feature["geometry"]["x"] = geometry["coordinates"][0]
+    else:
```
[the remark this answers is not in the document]
frankhereford: Mike and I sync'd on this real quick, and we have opted to leave this case logic as is, as it's an adaptation of existing, functional code. 🙏

moped-etl/arcgis/components_to_agol.py
```diff
@@ -115,6 +136,23 @@ def make_all_features(data):
...
+                geometry = record.pop("geometry")
+                esri_geometry_key = 'point'
+
+                feature = make_esri_feature(
+                    esri_geometry_key='point',
+                    geometry=geometry,
+                    attributes=record,
+                )
+
+                feature["attributes"]["source_geometry_type"] = 'point' # it's unclear what this needs to be. 'point' seems the most correct.
```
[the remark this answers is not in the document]
frankhereford: Great idea, thank you! e56855943aefb3ae05919cb2e7799269df3b9fd5

moped-etl/arcgis/components_to_agol.py
```diff
@@ -140,10 +178,16 @@ def main(args):
         variables=variables,
     )["component_arcgis_online_view"]
 
-    all_features = make_all_features(data)
+    exploded_data = make_hasura_request(
```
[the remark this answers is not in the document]
frankhereford: I think the script is massively more likely to bomb out due to a AGOL error than our graphql-engine endpoints, and I would rather focus on making that more resilient than this. Just my two cents, and isn't meant to be a dodge on doing this change.

Like we discussed, I had originally given it a whirl to use one query, but the line quoted below wasn't obvious to me how to extend to a two-unrelated-queries in one request, but I'm sure it's possible. I didn't see anything jumping out at me from Hasura's documentation, so I think the next step is to find some graphql standards or look for other examples in our code that handles this correctly. 

```
query GetExplodedProjectsComponents($where: exploded_component_arcgis_online_view_bool_exp!) {
```

Totally up to you if you would like me to put in the time to get them into one query. Just say the word, and I'm on it.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -53,33 +55,52 @@ def make_esri_feature(*, esri_geometry_key, geometry, attributes):
...
+            by taking the multi-point geometry from the component_arcgis_online_view and "exploding" it into individual points.
 
     Returns:
         dict: An object with lists of Esri feature objects for lines, points, and combined layers
     """
-    all_features = {"lines": [], "points": [], "combined": []}
+
+    # inspect_geometry = exploded_geometry[0]
+    # print(f"exploded_geometry: {exploded_geometry}")
+    # print(f"inspect_geometry: {inspect_geometry}")
```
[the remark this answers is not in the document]
frankhereford: YES! thank you -- this was messy on my part. 2be8ac5a4243e71e238c61fc2560ec721386374d

moped-etl/arcgis/components_to_agol.py
frankhereford: Thanks for this observation. This change was born out of me forgetting to pick one or the other.

moped-etl/arcgis/components_to_agol.py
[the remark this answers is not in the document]
mddilley: Frank and I chatted about John's feedback in combination with [this comment](https://github.com/cityofaustin/atd-moped/pull/1397#discussion_r1755491619). Just to make sure we are all talking about the same thing - i was thinking something like:
```
query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!) {
  component_arcgis_online_view(where: $where) {
    <all of the fields>
  }
  exploded_component_arcgis_online_view(where: $where) {
    <fields needed for the exploded part>
  }
}
```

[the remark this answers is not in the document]
mddilley: @frankhereford @johnclary mentioning y'all because i realized this thread is pretty buried!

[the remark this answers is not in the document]
frankhereford: 💯 That's what I had originally written out and was trying to go with, but then I got hung up on the line:

```
query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!) {
```

I'm unsure what is going on here, and how to expand it to cover both views that we're selecting out of. The `$` in front of the where and this unknown value -- I'm not even sure what to call it. If anyone has some docs they can point me to, I would be massively grateful. I am a graphql newb, for sure.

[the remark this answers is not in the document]
mddilley: ahhh gotcha! i see now. The Hasura docs are thin on this, but it is a way to dynamically set the filter conditions. I'll try to find an example or doc somewhere, but the secret sauce is this part of the script (line 168) where we set the filter conditions based on the value of `args.full`.

```python
variables = (
        {"where": {}} # no filter w/ full replace
        if args.full
        else {"where": {"project_updated_at": {"_gt": args.date}}} # date filter w/ incremental replace
    )
```
I see the issue because we are dealing with two different queries, component and exploded, which require different variable types, `component_arcgis_online_view_bool_exp!` and `exploded_component_arcgis_online_view!`. So it would have to be more like:
```
query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!, $whereExploded: exploded_component_arcgis_online_view_bool_exp!) {
  component_arcgis_online_view(where: $where) {
    <all of the fields>
  }
  exploded_component_arcgis_online_view(where: $whereExploded) {
    <fields needed for the exploded part>
  }
}
```
and then an update to the vars passed in `make_hasura_request`. Going to think about this and come back to it in the morning.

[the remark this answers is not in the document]
johnclary: fwiw, you also have the option of using python string formatting to compose the query / make it reusable.

 but my comment was referring not to the Hasura query, but the actual construction of this view. We have the option of slimming down the new exploded geometry view so that it only contains geometry data, and not any component attribute columns.

in other words, you could change lines 142 to 150 to look like this:

```python
            for record in matching_exploded_geometry_records:
                geometry = record.pop("geometry")
                esri_geometry_key = "point"

                feature = make_esri_feature(
                    esri_geometry_key="point",
                    geometry=geometry,
                    attributes=component,   # use the component properties from the matching component
                )
```

that would have the benefit of having fewer places where we need to maintain the code for this ETL.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -115,6 +131,29 @@ def make_all_features(data):
...
+                esri_geometry_key = "point"
+
+                feature = make_esri_feature(
+                    esri_geometry_key="point",
+                    geometry=geometry,
+                    attributes=record,
+                )
+
+                # it's unclear what this needs to be. 'point' seems the most correct.
+                # please see lines 117-118 for where the 'point' choice was drawn from.
```
johnclary, line 153: can you clarify this comment or remove it? i think the line references are broken.

johnclary (commented):
thanks @frankhereford for reworking this! i tested this by commenting out the AGOL operations, and everything looks good.

i am curious what you and Mike think about trimming down this view so that it's purely geometry-focused. i think it would be worth it to do that to keep future maintenance simple.

i'm requesting changes because I think it would be good to add a tiny bit documentation about this new layer. That would included adding a bullet in the readme ([here](https://github.com/cityofaustin/atd-moped/blob/95a3e313a2cdb9a797cb1e337df803818292f57e/moped-etl/arcgis/README.md?plain=1#L14)) that links to this layer on AGOL along with just a quick explanation that this new layer contains one row/component per feature. A visual/screenshot might be nice as well 🤷 

thanks again for refactoring this code—it's looking sharp!!

johnclary:
ah—i see the [separate PR](https://github.com/cityofaustin/atd-moped/pull/1413#pullrequestreview-2300478053) now wrt the README. thanks!

moped-etl/arcgis/components_to_agol.py
mddilley: i tripped up on this too due to habit - @frankhereford could you add this to the template? i think that is a great idea because, knowing myself, i'll do it again. 🙃

moped-etl/arcgis/components_to_agol.py
mddilley: @frankhereford and I chatted about this in slack, and I think this can go altogether since the lines shifted and probably will again.

mddilley (dismissed):
Looking so great! I requested a few changes based on the discussion. Thanks for continuing on this to get it over the line. 🙏 💪

moped-etl/arcgis/.gitignore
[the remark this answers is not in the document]
frankhereford: I think it certainly can't hurt and is a good idea. 671679c644888cdfd33e4be787405ad9d3f779f8. I almost am of the mind that for new docker entities like this one, I almost want to just symlink the `.gitignore` as `.dockerignore`.. 🤔

frankhereford:
@johnclary wrote:

> it's my own (bad?) habit of expecting my docker container to have access to it's siblings. if we're going to remove this, would you also update the env_template with the correct local Hasura endpoint?

I updated environment template file -- thank you for that. 5b74d4a5311b54bef9852db45f49220cb7c923ac. 

To your observation about using `network=host` or not; I guess my tendency is to try to lean into docker's ability to encapsulate and containerize, which makes me prefer a container not having access to the host's other services and being able to open ports on the host without explicit permission. Maybe this is a reflection of why I'm prone to write `docker-compose.yaml` files; i'm all about cementing that sibling relationship in configuration files, where you can have docker make one container genuinely dependent on another being available and all that. Your habit isn't any more wrong than mine; I think they are just different.

frankhereford:
@johnclary wrote:

> We have the option of slimming down the new exploded geometry view

Absolutely - great call. The view, query, and ETL have been refactored to do this.

moped-etl/arcgis/components_to_agol.py
frankhereford: It's outta there! a61579f63338c2db59d631f8f6f37124c25b968e

moped-etl/arcgis/components_to_agol.py
frankhereford: Great call y'all -- i hadn't expected a default value in there, and it makes total sense that it should be there since it's not secret. This is done.

mddilley:
@frankhereford i merged the base branch (`md-18877-update-proj-dev-status`) into `main` but did not delete so i didn't pull the branch out from under you and if you prefer to manually re-target `main`

moped-database/metadata/tables.yaml
chiaberry: I'm not sure that I fully understand, but if this isn't performant enough to be used in the front end (its not, right?) then why include it?

moped-etl/arcgis/components_to_agol.py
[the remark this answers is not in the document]
mddilley: thanks John for the insight here and Frank for implementing. The latest iteration of the exploded view makes this very maintainable for next time around. 🙏

mddilley (approved):
Thanks for all the work on this, Frank! There is one small clean up item that I mentioned you in, but this is good to ship! I compared the attributes coming out of the AGOL endpoint on a single component (11887), and I see the same project and component level data coming out of the exploded and non-exploded features. 🚢 🙌

moped-database/metadata/tables.yaml
frankhereford: You bet, thanks for this! de42ca4273b442f7c5e96817f77fd0bfea33b7f0

johnclary (approved):
yep, this looks great!! i can't wait to see this live. thank you, Frank!!! 🚢 🚢 🚢 🚢 🚢

chiaberry (commented):
Merged before I could finish my review, but just commenting to say I approve this PR! Thank you for all your work on this, and it is going to be such a great enhancement for ATSD's map

PR #1400, opened 2024-08-29
Add ability to toggle filters on any or all in the search filters chips

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/17956

## Testing
**URL to test:** 

[Deploy Preview with filters applied](https://deploy-preview-1400--atd-moped-main.netlify.app/moped/projects?filters=%5B%7B%22field%22%3A%22project_name_full%22%2C%22operator%22%3A%22string_contains_case_insensitive%22%2C%22value%22%3A%22test%22%7D%2C%7B%22field%22%3A%22project_name_full%22%2C%22operator%22%3A%22string_contains_case_insensitive%22%2C%22value%22%3A%22Mike%22%7D%5D&isOr=false)

**Steps to test:**

Add more than one filter to the advanced search filters. Apply filters
Click on the Matching *all* filters chip. 
Confirm that it switches to any filters, the url params update (the end of the params says `isOr=true`) and the search results update. 
Click again! Search goes back to filtering on all, url param updates and search results update. 

Screenshot 2024-08-29 at 9 34 51 AM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/FiltersChips.js
+++ b/moped-editor/src/components/GridTable/FiltersChips.js
@@ -95,17 +95,12 @@ const FiltersChips = ({
 
   /**
    * Triggered by Filter Chip any/all click
-   * Gets previous IsOr state and sets the IsOr state and the url params state to be the opposite
+   * Sets the IsOr state and the url params state to be the opposite of current isOr state
    */
   const toggleIsOrOnClick = () => {
-    let newIsOrState = null;
-    setIsOr((prevState) => {
-      newIsOrState = !prevState;
-      return newIsOrState;
-    });
-
+    setIsOr(!isOr);
     setSearchParams((prevSearchParams) => {
-      prevSearchParams.set(advancedSearchIsOrParamName, newIsOrState);
+      prevSearchParams.set(advancedSearchIsOrParamName, !isOr);
       return prevSearchParams;
     });
   };
--- a/moped-editor/src/components/GridTable/FiltersChips.js
+++ b/moped-editor/src/components/GridTable/FiltersChips.js
@@ -118,6 +118,7 @@ const FiltersChips = ({
             <Grid item>
               <Chip
                 variant="outlined"
+                color="primary"
                 onClick={toggleIsOrOnClick}
                 label={
                   isOr ? (
--- a/moped-editor/src/components/GridTable/FiltersChips.js
+++ b/moped-editor/src/components/GridTable/FiltersChips.js
@@ -93,6 +93,23 @@ const FiltersChips = ({
     }
   };
 
+  /**
+   * Triggered by Filter Chip any/all click
+   * Gets previous IsOr state and sets the IsOr state and the url params state to be the opposite
+   */
+  const toggleIsOrOnClick = () => {
+    let newIsOrState = null;
+    setIsOr((prevState) => {
+      newIsOrState = !prevState;
+      return newIsOrState;
+    });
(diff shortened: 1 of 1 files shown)


moped-editor/src/components/GridTable/FiltersChips.js
```diff
@@ -101,6 +118,7 @@ const FiltersChips = ({
             <Grid item>
               <Chip
```
mddilley, line 119: @chiaberry could you also make the color="primary" on this isOr chip? John had brought it up in [the discussion](https://austininnovation.slack.com/archives/CNUEPKLB1/p1718826221462969) a while back, and I think it would look great now that I'm seeing this live.

moped-editor/src/components/GridTable/FiltersChips.js
```diff
@@ -93,6 +93,23 @@ const FiltersChips = ({
     }
   };
 
+  /**
+   * Triggered by Filter Chip any/all click
+   * Gets previous IsOr state and sets the IsOr state and the url params state to be the opposite
+   */
+  const toggleIsOrOnClick = () => {
+    let newIsOrState = null;
```
mddilley, line 101: i'm trying to think through this because i'm concerned that assigning this variable in the `setIsOr` call could behave unpredictably since setting state is not synchronous.

would `const newIsOrState = !isOr;` work? I'm thinking that the two subsequent setter functions would be guaranteed to use the same value that way.

mddilley (commented):
Looks and works beautifully!!! Fully interactive chips on deck - awesome. ✨ I have one question about the toggle function, but I might be overthinking.

moped-editor/src/components/GridTable/FiltersChips.js
chiaberry: thank you for thinking through this, I was initially going to have both setter functions just toggle the opposite of their previous states, but then i was concerned that they could get out of sync. 

I'll change it to the const !isOr, the state shouldn't set before that definition, so it should work?

chiaberry: hm does this mean 
```
    const newIsOrState = !isOr;
    setIsOr(newIsOrState);
```
without referencing the prevState?

mddilley: yes, after looking at [the docs](https://react.dev/reference/react/Component#setstate), i don't think that `prevState` is needed since state is only being updated once. i'm also starting to wonder if just using `!isOr` is the way to go since its value won't update until the next render.

johnclary (approved):
finally i can click all the chips!!! looks great and works so great 🚢 🚢 🚢 

possible future enhancement, as i was testing i kept wanting to click on filter chips to open the advanced search menu. having to click all the way over on the right side of the search menu is the slightest bit tedious.

chiaberry:
@johnclary clicking them how you delete them, hmm thinking about how to be able to open or edit them but also retain the delete capability

mddilley (approved):
🚢 ✨✨✨

chiaberry:
Thanks for the set state discussion! I am going to merge this so testing can continue

johnclary:
thanks, Mike! to Chia's point, if having two different click interactions on a small little chip is too much, maybe moving the advanced search toggle to left (like swapping it with the inert search icon). I dunno! thanks for entertaining this idea.

PR #1404, opened 2024-08-29
18390 add toolbar to lexical

tillyw:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/18390

## Testing
**URL to test:** 

local

**Steps to test:**
- add a project note, try enabling and disabling all of the different styling options on different text selections and confirm the styling is being applied and removed as expected and the buttons are reflecting the state of the selected text. make sure to try different list options, toggling those back and forth and applying styling within lists, etc.
- try using the back and forward buttons before adding any text or styling, they should be disabled. add some styling and click back until you reach the original state and the button is disabled, then do the opposite with the redo button.
- try using the clear formatting button and confirm it works as expected
- copy a url to your clipboard, select a piece of text and click the link button - the link should be applied. does this seem like an adequate way of implementing links or do we need something a bit more user friendly?
- try copying an email or some other previously formatted text (make sure it also has hyperlinks in it) and pasting it into the window
- save a note and make sure it appears correctly below and that the editor is cleared after a successful save
- edit an existing note, the editor should appear in the note space and the note should appear as it did in the ui. the toolbar buttons will be a lighter color than when you are adding a new note. does that seem okay or do we need to override that behavior?
- try editing a note that was made using quill, does it appear as expected? can you edit and save it?
- inspect the notes in the backend. do the tags, styles, etc. look okay?

---
#### Ship list
- [x] Code reviewed 


mddilley (commented):
I'll continue taking a look at this more on Monday, but I left some notes for now. Thanks for exploring these uncharted waters! The functionality is all here! 🔭

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,260 @@
...
+  },
+  toolbarButtons: {
+    backgroundColor: theme.palette.primary.main,
+    '&:hover, &.Mui-focusVisible': {
+      backgroundColor: theme.palette.primary.main,
+    },
+  },
+  startIcon: {
+    margin: 0,
+    color: theme.palette.text.primary,
```
mddilley, line 71: I noticed that this color style overrides the active/inactive colors built into the Toolbar theme. You can check that out by removing and seeing that the "redo" button icon is a disabled color until text is entered. I think it would be fine to remove this line, and let the toolbar theme take over.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,360 @@
...
+  center: "center",
+  right: "right",
+  justify: "justify",
+  divider: "divider",
+  undo: "undo",
+  redo: "redo",
+  clear: "clear"
+}
+
+const RICH_TEXT_OPTIONS = [
```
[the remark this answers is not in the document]
mddilley: You had mentioned that you were having issues with the dividers being interactive, and I think it would be totally fine to compose these in the return of the component instead of iterating this options list. Might makes it easier to control of what ends up wrapped in a button group or not.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,360 @@
...
+    const formatTypes = Object.keys(selectionMap);
+    formatTypes.forEach((formatType) => {
+      if (selectionMap[formatType]) {
+        onAction(formatType);
+      }
+    });
+  };
+
+  useEffect(() => {
+    // to do: move this function outside of useEffect without causing warnings
```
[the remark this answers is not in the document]
mddilley: Just wanted to update that this makes more sense to me now that I read https://lexical.dev/docs/concepts/listeners and https://lexical.dev/docs/concepts/commands. I also found this code example which looks like provided some good inspiration. https://github.com/ozanyurtsever/verbum/blob/master/src/plugins/ToolbarPlugin/ToolbarPlugin.tsx. I think it would be awesome to point to some links in this code since it is new to all of us. 📚

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,260 @@
...
+  return (
+    <Container>
+      <Grid container direction="column" spacing={1}>
+        <Grid item xs={12} sm={12}>
+          <LexicalComposer initialConfig={initialConfig} >
+            <ToolbarPlugin noteAddSuccess={noteAddSuccess} classes={classes} />
+            <Box className={classes.editorWrapper} pt={2}>
+              <RichTextPlugin
+                contentEditable={<ContentEditable className={classes.contentEditable} />}
+                placeholder={<div className={classes.placeholder}>Enter some text...</div>}
```
mddilley, line 201: Could you pass `null` for this placeholder? We don't have any placeholder text in the the current editor, and we haven't had reports of it being confusing (yet 🙃).

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -0,0 +1,260 @@
...
+  cancelButton: {
+    margin: theme.spacing(1),
+    position: "relative",
+    color: theme.palette.secondary.dark,
+  },
+  contentEditable: {
+    width: "100%",
+    padding: "16px 16px",
+    border: "4px solid",
+    borderRadius: "8px",
```
mddilley, line 46: In other places where we want the nice MUI border styles, we have used [Paper](https://v5.mui.com/material-ui/react-paper/) component an elevation of 2. I think if you remove these border styles and wrap `<ContentEditable />` with Paper then it would match up nicely with the rest of the Notes UI.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,360 @@
...
+        break;
+      case "clear":
+        clearFormatting();
+        break;
+      default:
+        break;
+    }
+  }
+
+  // to do: wrap this in a useEffect hook
```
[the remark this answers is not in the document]
mddilley: The only thing that I could think of when poking around the Lexical API is if there is a way to clear the formatting or EditorState with the `editor` instance kind of the same way commands are dispatched. I couldn't get this to work with useEffect, and I kept getting stuck where `clearFormatting()` is calling `onAction()` from within `onAction()`. We probably need more eyes on this part.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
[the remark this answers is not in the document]
tillyw: actually found a solution to the dividers issue! i like iterating through the options list so i'll stick with that :)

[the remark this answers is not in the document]
tillyw: removed those extra options!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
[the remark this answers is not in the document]
tillyw: I was able to resolve it by wrapping `updateToolbar()` in its own `useCallback()` hook

tillyw:
> I'll continue taking a look at this more on Monday, but I left some notes for now. Thanks for exploring these uncharted waters! The functionality is all here! 🔭

thanks @mddilley! I made the changes you requested along with a couple of additional optimizations. re-requesting review!

moped-editor/src/views/projects/projectView/PlaygroundEditorTheme.css
```diff
@@ -0,0 +1,427 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+ .PlaygroundEditorTheme__ltr {
```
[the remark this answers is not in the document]
tillyw: sorry i missed this comment! will clean this up

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,290 @@
...
+        clearFormatting();
+        break;
+      default:
+        break;
+    }
+  }
+
+  // Clear editor formatting when note is saved successfully. Prop is passed down from
+  // parent component so the toolbar will rerender on save
+  if (noteAddSuccess) {
```
tillyw, line 212: perhaps it's a bit awkward to have this logic free floating here before the `useEffect`, but it functions just as it would in the `useEffect` since the prop is passed down from the parent component. placing this inside the `useEffect` causes a continuous loop which can be fixed but may require a refactor of the parent component. not sure if this is worth tackling in this pr but wanted to point that out and welcome any thoughts!

moped-editor/src/views/projects/projectView/EditorTheme.css
mddilley, line 1: This initially felt like a lot of CSS to take on, but I'm realizing that we did the same thing with Quill and its [Snow theme that we imported](https://github.com/cityofaustin/atd-moped/blob/c5934212ec4b0eda5b52bfd3092a95a70bb66519/moped-editor/src/views/projects/projectView/NoteInputQuill.js#L13). We'll also have the added benefit of updating any of these styles if it ever comes to that like we have done with the Mapbox draw library style configuration.

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -35,22 +41,32 @@ const useStyles = makeStyles((theme) => ({
     color: theme.palette.secondary.dark,
   },
   contentEditable: {
-    height: "300px",
     width: "100%",
-    padding: "8px 8px",
-    border: "1px solid #fff"
+    padding: "16px 16px",
```
mddilley, line 45: these jumped out at me this time around. Could you update these to use `theme.spacing(2)` so this uses our MUI default scaling factors?

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -35,22 +41,32 @@ const useStyles = makeStyles((theme) => ({
...
+    fontSize: "1rem",
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.text.primary,
   },
   placeholder: {
```
mddilley, line 50: I don't think this `placeholder` style is used anymore. Could be worth a check!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,290 @@
...
+    })
+    return isLink;
+  };
+
+  const handleLinkUpdate = () => {
+    const isLinkApplied = selectionMap["link"];
+    isLinkApplied ?
+      editor.dispatchCommand(TOGGLE_LINK_COMMAND, null)
+      :
+      navigator.clipboard.readText().then(val => {
```
mddilley, line 136: I stumbled a little when trying to create a link and unintentionally captured my clipboard to a link. I like this ternary to toggle between link and no link, and I'm wondering the toggle could be between:
1. If highlighted text isn't already a link, the text turns into a link (whether it is valid or invalid - just like production code which doesn't worry about validating what is entered as a link)
2. If highlighted text is already a link, the link is removed from the text

I think that the editor dispatches could happen in the context of `editor.update` so that you can access `selection.getTextContent()` and use that to set the `url` value. Overall, I'm worried about introducing a more complex interaction that involves the clipboard. 100% open to other opinions here in case I'm missing something!

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -88,7 +102,8 @@ const OnChangePlugin = ({ onChange }) => {
   useEffect(() => {
     return editor.registerUpdateListener(({ editorState }) => {
       editorState.read(() => {
-        const htmlContent = $generateHtmlFromNodes(editor, null);
+        const isEditorEmpty = $isRootTextContentEmpty();
+        const htmlContent = isEditorEmpty ? null : $generateHtmlFromNodes(editor, null);
         onChange(htmlContent);
       });
     });
```
mddilley, line 109: Could you add a dependency array to this useEffect and put `editor` in it like in the other plugin components in this file? Otherwise, this code will run on every render like it wasn't wrapped in useEffect at all.

mddilley (changes requested):
Thanks for adding the docs links and noting the conundrum about observing the existing note save success state! I added some notes about a possible path around it that I'd like to explore. I think that is an important detail to work out before getting this released.

Thanks for the continued work on this - it has been very cool to see you weave this into our existing functionality! 🪡 🧵

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: thanks for catching that!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,290 @@
...
+      })
+  };
+
+  const clearFormatting = () => {
+    const formatTypes = Object.keys(selectionMap);
+    formatTypes.forEach((formatType) => {
+      if (selectionMap[formatType]) {
+        onAction(formatType);
+      }
+    });
```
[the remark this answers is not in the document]
tillyw: this is an excellent solution! works perfectly and saves some lines of code :)

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: after chatting about this @mddilley and i decided to stick with this behavior for now and create a follow-up issue to add a modal popup for pasting and removing links to optimize usability of this feature.

tillyw:
thanks @mddilley for you additional comments, this is ready for your re-review and for everyone else to review!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
johnclary: sounds good—i'm definitely not a fan of the behavior that tries to read from the clipboard.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,287 @@
+import { useCallback, useEffect, useState } from "react";
+
+import {
+  Box,
+  ButtonGroup,
+  Button,
+} from "@mui/material";
+import {
+  Redo,
+  Undo,
```
johnclary, line 10: fwiw, i would not miss the undo/redo buttons if they weren't there, assuming the built-in browser/OS equivalent is working normally.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,287 @@
...
+    let isLink = false;
+    const selectedNodes = selection.getNodes();
+    selectedNodes.forEach((selectedNode) => {
+      const selectedNodeParent = selectedNode.getParent()
+      if ($isLinkNode(selectedNodeParent)) {
+        isLink = true
+      }
+    })
+    return isLink;
+  };
```
johnclary, line 129: i see some of these functions wrapped in useCallback and others not, so i'm curious what the distinction is. 

it looks like perhaps `checkLink` and  `checkListType` can be moved outside the component entirely to bring them out of the render cycle.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,287 @@
...
+        [richTextAction.bold]: selection.hasFormat("bold"),
+        [richTextAction.italics]: selection.hasFormat("italic"),
+        [richTextAction.underline]: selection.hasFormat("underline"),
+        [richTextAction.number]: checkListType(selection, "number"),
+        [richTextAction.bullet]: checkListType(selection, "bullet"),
+        [richTextAction.link]: checkLink(selection),
+      }
+      setSelectionMap(newSelectionMap)
+    }
+  }, [setSelectionMap])
```
johnclary, line 158: i'm surprised this isn't throwing an error about `checkListType`, `checkLink`, and `richTextAction` not being in the dependency array. i have doubts about the right way to set this up, but maybe an empty dependency array is actually what's desired here?

johnclary (changes requested):
this is working really well, Tilly! well done!

i left some feedback. i also want to ask, is it possible to make the text input area have multiple lines/rows? i know that the current production editor is just one line high, but something about this one having the input focus border effect makes it look extra cramped.

Screenshot 2024-09-23 at 11 58 43 AM

tillyw:
> this is working really well, Tilly! well done!
> 
> i left some feedback. i also want to ask, is it possible to make the text input area have multiple lines/rows? i know that the current production editor is just one line high, but something about this one having the input focus border effect makes it look extra cramped.
> 
> Screenshot 2024-09-23 at 11 58 43 AM

thanks, John! I'll defer to @mddilley on whether we should expand the default height of the text editor since we've had some discussion of it previously. lmk what y'all think!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: good call on this! i moved these out of the component and that is definitely cleaner :)

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: you're right, not sure why it wasn't throwing an error about those missing dependencies, but feeding in an empty dependency array does seem like the way to go! thanks

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: also to answer your first question i was receiving error messages on those particular functions stating they would continuously rerender so i had to wrap them in useCallbacks

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: i think they're pretty snazzy but if anyone else feels strongly about removing them we can consider it

moped-editor/src/views/projects/projectView/EditorTheme.css
chiaberry, line 1: Are there any custom styles here or is this a copy of the lexical theme?

tillyw: just a copy of the lexical theme!

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
mddilley: i just tested this updated behavior, and this looks great! Thanks Tilly! 🚀

mddilley: @tillyw I just accidentally saved a new Moped user's email to a link, and I think people will be caught off guard by this button interacting with their clipboard. I think that this should update to toggle between link/non-link when text is highlighted to prevent us from confusing users. It could even lead them to capture something sensitive that they don't into to into the app. I included a snippet of what I think it would take and a clip of the resulting behavior.

Could you update? Copying and pasting links like AMD often does works beautifully! ✨

```js
editor.update(() => {
   const isLinkApplied = selectionMap["link"];
   const selection = $getSelection();
   const textContent = selection.getTextContent();

   isLinkApplied
     ? editor.dispatchCommand(TOGGLE_LINK_COMMAND, null) // Remove link if already applied
     : editor.dispatchCommand(TOGGLE_LINK_COMMAND, textContent); // Make selected text into a link
 });
```

https://github.com/user-attachments/assets/1f54bbfa-ea35-4198-93f3-e68d74aa726f

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
mddilley: yes, i think that now that the `checkListType` and `checkLink` functions are moved out of the component, then an empty dependency array makes sense. Side note that I recently found that the new React docs don't mention state setter functions have stable references between renders but [the old ones do](https://legacy.reactjs.org/docs/hooks-reference.html#usestate).

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
mddilley: I think that John raises a good point about the keyboard shortcuts handling the functionality of these buttons. If we had a mobile user base, the buttons could provide more of an important role in the UI.

In light of the [ETL issues we faced this week](https://austininnovation.slack.com/archives/CHZE6BC6L/p1727122151675579), I think that we should dedicate more code to getting plain text out of this editor in future-scoped work, and keep things simple when we roll this out.

@tillyw Does removing the undo/redo buttons feel doable?

mddilley (changes requested):
Just a couple small changes, and I think that this one is over the line! 🙌

About the editor height, it does look short and wide to me now. Could you try 200px height and see how that looks? Also, could you run this code through Prettier? I forgot to note that last time around, and I'm seeing changes when I'm testing locally.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: this works well! will go with this for now

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
tillyw: went ahead and removed them!

tillyw:
> Just a couple small changes, and I think that this one is over the line! 🙌
> 
> About the editor height, it does look short and wide to me now. Could you try 200px height and see how that looks? Also, could you run this code through Prettier? I forgot to note that last time around, and I'm seeing changes when I'm testing locally.

200px looked a little too tall to me so i went with 150px

mddilley:
150px sounds excellent - thanks @tillyw! i will be looking at this shortly 👀

chiaberry:
I know there have been some changes, so the test steps as written are slightly out of date, but I want to ask about the link, should I be able to highlight text and apply a link to it, it seems like its just creating a link out of whatever is highlighted?
This is what is stored in the database: 

<p dir="ltr" class="EditorTheme__paragraph"><a class="EditorTheme__link" rel="noreferrer" href="can I do this"><span style="white-space: pre-wrap;">can I do this</span></a></p>

moped-editor/src/views/projects/projectView/EditorTheme.css
chiaberry, line 1: do we need css for styles the editor can't add? like the H1 H2 H3, the table ones?

mddilley: i think that Tilly knows better than me, but I think these styles cover content that could be pasted into the editor. Even if we don't offer a way to input them.

moped-editor/src/views/projects/projectView/EditorTheme.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+// import {EditorThemeClasses} from 'lexical';
```
mddilley, line 9: This comment caught my eye on this review and would be great to clean up before merge. 🙏

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -8,21 +8,27 @@ import {
...
-import { RichTextPlugin } from '@lexical/react/LexicalRichTextPlugin';
-import { ContentEditable } from '@lexical/react/LexicalContentEditable';
-import { HistoryPlugin } from '@lexical/react/LexicalHistoryPlugin';
-import { LexicalErrorBoundary } from '@lexical/react/LexicalErrorBoundary';
-import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext';
-import { $generateHtmlFromNodes, $generateNodesFromDOM } from '@lexical/html';
-import { $getRoot } from 'lexical';
-import { LinkPlugin } from '@lexical/react/LexicalLinkPlugin';
-import { LinkNode } from '@lexical/link';
+import { AutoFocusPlugin } from "@lexical/react/LexicalAutoFocusPlugin";
```
mddilley, line 17: I tested that the autofocus works. 🙏

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -219,8 +239,8 @@ const NoteInput = ({
               </div>
             )}
             <ProjectSaveButton
-              // disable save button if no text after removing html tags
-              disabled={!noteText.replace(/<[^>]*>/g, "")}
```
mddilley, line 223: nice that you were able to remove this regex! I confirmed that I was unable to save a note that looks empty in the editor but writes empty HTML tags into the database.

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
```diff
@@ -0,0 +1,233 @@
...
+      nodeParents.forEach((parent) => {
+        if (parent.__listType === listType) {
+          hasListType = true;
+        }
+      })
+    })
+    return hasListType;
+  };
+
+  const checkLink = (selection) => {
```
chiaberry, line 84: It would be nice for a brief comment explaining what this function does

mddilley (approved):
I created, edited, tested all the buttons, tested creating a link, and also pasting the contents of emails into the editor, and I'm seeing this behave exactly like we do in production - accounting for the removal of the strikethrough and the simplified link behavior. We'll see if they are missed, and, if not, less code to maintain! 🚢 🎉

Thanks for working through this and getting us to a place where we have an editor that is maintained. Also cool to see a path to get plain text output which we can explore to cover rich text that will end up displayed in other places like in AGOL.

Moving on to reviewing the next PR now - https://github.com/cityofaustin/atd-moped/pull/1377.

moped-editor/src/views/projects/projectView/EditorTheme.js
chiaberry: agreed!

moped-editor/src/views/projects/projectView/EditorTheme.css
chiaberry, line 1: Could be nice to include a link here to where this file came from, I am guessing the Lexical repo?

moped-editor/src/views/projects/projectView/ToolbarPlugin.js
johnclary: Mike, your comment sent me digging through the new docs out of curiosity, and there is a comment about "stable identity" in these [setState caveats](https://react.dev/reference/react/useState#setstate-caveats) 😅

johnclary (approved):
thanks @tillyw for those fixes. this is looking great! 🚢 ✅

johnclary:
one more minor thing—@tillyw i think you can remove `react-quill-new` from the `package.json`.

moped-editor/src/views/projects/projectView/EditorTheme.js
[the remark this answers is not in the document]
chiaberry: @tillyw same request as for the css file, a link to where this file came from

tillyw:
> one more minor thing—@tillyw i think you can remove `react-quill-new` from the `package.json`.

good catch - done!

moped-editor/src/views/projects/projectView/NoteInput.js
tillyw: yay thanks for confirming that! :)

tillyw:
> I created, edited, tested all the buttons, tested creating a link, and also pasting the contents of emails into the editor, and I'm seeing this behave exactly like we do in production - accounting for the removal of the strikethrough and the simplified link behavior. We'll see if they are missed, and, if not, less code to maintain! 🚢 🎉
> 
> Thanks for working through this and getting us to a place where we have an editor that is maintained. Also cool to see a path to get plain text output which we can explore to cover rich text that will end up displayed in other places like in AGOL.
> 
> Moving on to reviewing the next PR now - #1377.

yay thanks Mike!

moped-editor/src/views/projects/projectView/EditorTheme.css
tillyw: exactly!

PR #1406, opened 2024-09-03
Update more advanced search fields to use lookups with contains operator

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/18230

I removed some of the operators from the FiltersConfig and rearranged some to fix the default (first in list is default), but since I also refactored the Config from an array to an object, the diff is harder to follow. 

## Testing
**URL to test:** 

https://deploy-preview-1406--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

Test the advanced filters. 
- Inspectors, Designers, Types and Partners default to contains
- Created By, Inspectors and Designers now have an autocomplete option using the users table
- ~~If a field has a lookup table and autocomplete option *and* the default operator is "is", the contains operator now gives a freesolo option. These are project lead, project sponsor, created by and public process status.~~ dropping freesolo for now
- Lead and Sponsor default to "is"
- Because I refactored the Field Config, it is worth it to test out the chips again. 

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectsListView/ProjectsListViewQueryConf.js
```diff
@@ -2,7 +2,6 @@
  * The Query configuration for the ProjectsListView component
  * @constant
  * @type {object}
- * @augments PROJECT_LIST_VIEW_FILTERS_CONFIG
```
chiaberry, line 5: it does not

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -25,459 +23,454 @@ export const OPERATORS_WITHOUT_SEARCH_VALUES = [
...
+      "number_less_than",
+      "number_less_than_equal_to",
+    ],
+  },
+  type_name: {
+    name: "type_name",
+    label: "Type",
+    placeholder: "Enter Project type",
+    type: "string",
+    defaultOperator: "string_equals_case_insensitive",
```
roseeichelmann, line 102: i think this needs to be updated to `string_contains_case_insensitive` to get the default contains setting

roseeichelmann (changes requested):
Love the refinements we have been making to the advanced search recently! Type still needs to be updated to default to contains. Im also not seeing autocomplete options for Inspectors or Designers.Also confused about the free solo option - does that mean we should be able to search on those without actually clicking an option? Or what is the effect we see from that, i dont think im seeing it but maybe i dont know what im looking for

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -511,7 +502,12 @@ const Filters = ({
                     ) ? (
                       <Autocomplete
                         value={value || null}
-                        options={dedupedOptions}
+                        options={autocompleteOptions[fieldName]}
+                        // allow freeSolo option when the operator is contains and contains is not its only operator
+                        freeSolo={
+                          operator === "string_contains_case_insensitive" &&
+                          showFreeSolo
```
mddilley, line 509: i like Rose's comment about wanting more info about how this option is used. Moving this logic into a helper function could be helpful to doc and extend the logic in the future. I could see entering an arbitrary string helping with the `string_contains_not_case_insensitive` operator too.

mddilley:
@chiaberry I also ran into the issue that Rose described above. I'll hold off on review for now.

Also, thank you for tackling the config refactor!! It will be nice to get away from the array of fields shape that doesn't quite fit the way we handle the queries in the post-GraphQLAbstract code. 🙌

chiaberry:
I'm so sorry guys that I asked for review and hadnt pushed up my last commit 

Thanks for catching the default operator I missed updating @roseeichelmann , I am noticing now how the config is inconsistent with using that to define the default operator. I'll add that to the rest of the config entries. 

As for the fresolo option, let's talk about it tomorrow, I can pull it into a helper function but I am not sure how we could convey to users that that option is there....

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -25,459 +23,493 @@ export const OPERATORS_WITHOUT_SEARCH_VALUES = [
...
+  type_name: {
+    name: "type_name",
+    label: "Type",
+    placeholder: "Enter Project type",
+    type: "string",
+    defaultOperator: "string_contains_case_insensitive",
+    lookup: {
+      table_name: "moped_types",
+      getOptionLabel: (option) => option.type_name,
+      operators: ["string_contains_case_insensitive"],
```
mddilley, line 106: Forgive me for going back on the issue requirements but could you add `string_equals_case_insensitive` and `string_does_not_equal_case_insensitive` back in for `type_name` and `project_partners`? No change to default operator. Today's dev sync discussion made me realize this is actually taking away valid searches like "a project with one type that is _____".

mddilley (changes requested):
This works great!!! 🙌 The config refactor is a total win all around too. I thought more about the freeSolo option, and I agree with you that we should hit the eject button on it again. Maybe the third time around it will stick. I will create an issue for it so we can find the best way to incorporate it.

moped-editor/src/components/GridTable/helpers.js
```diff
@@ -99,9 +99,7 @@ export const handleApplyValidation = (filterParameters, filtersConfig) => {
   if (filterParameters) {
     filterParameters.forEach((filter) => {
       const { field: fieldName, value, operator } = filter;
-      const fieldConfig = filtersConfig.fields.find(
-        (field) => field.name === fieldName
-      );
+      const fieldConfig = filtersConfig[fieldName];
```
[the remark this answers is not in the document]
mddilley: https://github.com/cityofaustin/atd-data-tech/issues/18935

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
mddilley: thanks for updating! 🙏

mddilley (approved):
I tested through the updated defaults, new lookups, filter chips, and the rest of the filters too. All looks great to me! Thanks for the extra refactoring too - glad the time finally came to address the shape of the filter config! 🚢 😎

mddilley:
@roseeichelmann thanks again for flagging how confusing the freeSolo option could be! I captured your feedback and discussion from dev sync last week in https://github.com/cityofaustin/atd-data-tech/issues/18934. Re-requesting your review since Chia is out this morning.

tillyw (approved):
tested and working great! thanks for making these improvement Chia!

potential future pr, but i noticed during testing if you select an option in a dropdown and then change the field value the option remains selected even if it is not part of the new field's dropdown selections. it might be nice to clear the option value when the field value is changed.

roseeichelmann (approved):
this looks and works great!! kicking the can on the free solo feature to another pr sounds good to me :ship:

mddilley:
@tillyw just caught your comment - i will create an issue for that because it is also confusing to me to carry over a value from one search field to another unrelated one! 🙏

PR #1407, opened 2024-09-04
add `formatRelativeDate` to replace date with string

mateoclarke:
## Associated issues

Closes https://github.com/cityofaustin/atd-data-tech/issues/18671

Screenshot 2024-09-03 at 7 13 37 PM

## Testing
**URL to test:** https://deploy-preview-1407--atd-moped-main.netlify.app/moped/projects/1822?tab=funding

**Steps to test:**
- Go to URL above which is a test project's funding tab.
- Scroll down to the "Work activities" table.
- Scroll right in the table window to the "Updated at" column.
- The date should appear as a relative time string (ex: "Just now" or "1 day ago").
- Test this by clicking the "ADD WORK ACTIVITY" button on this or other projects.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/dateAndTime.js
+++ b/moped-editor/src/utils/dateAndTime.js
@@ -1,3 +1,5 @@
+import theme from "src/theme";
+
 /**
  * Format date from the DB in YYYY-MM-DD format into human-readable locale date
  * in MM/DD/YYYY format
@@ -116,15 +118,18 @@ export const formatRelativeDate = (timeStampTZString) => {
 };
 
 /**
- * A component that renders a formatted date string with primary display and optional secondary display
- * @param {Object} props - The component props
- * @param {string} props.date - The date to format, in a format parseable by new Date()
- * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} props.primary - The format to use for the primary display
- * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} [props.secondary] - The format to use for the secondary display, if any
- * @returns {JSX.Element} A span element containing the formatted date string(s)
+ * Formats a date according to the specified format
+ * @param {string|Date} date - The date to format, either as a string or Date object
+ * @param {string} format - The format to use. Can be one of:
+ *   - 'relative': Returns a relative time string (e.g., "1 year from now", "364 days from now", "Just now")
+ *   - 'absolute': Returns the date in local string format (e.g., "10/31/2024, 2:30:45 PM")
+ *   - 'expanded': Returns the date in US expanded form (e.g., "October 31, 2024")
+ *   - 'short': Returns the date in short format (e.g., "10/31/24")
+ *   - 'time': Returns the hour and minutes (e.g., "2:30 PM")
+ *   - 'fullTime': Returns the full time including seconds (e.g., "2:30:45 PM")
+ * @returns {string} The formatted date string
  */
-export const FormattedDateString = ({ date, primary, secondary }) => {
-  const formatDate = (format) => {
+  const formatDate = (date, format) => {
     switch (format) {
       case 'relative':
         return formatRelativeDate(date);
@@ -143,12 +148,21 @@ export const FormattedDateString = ({ date, primary, secondary }) => {
     }
   };
 
+/**
+ * A component that renders a formatted date string with primary display and optional secondary display
+ * @param {Object} props - The component props
+ * @param {string} props.date - The date to format, in a format parseable by new Date()
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} props.primary - The format to use for the primary display
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} [props.secondary] - The format to use for the secondary display, if any
+ * @returns {JSX.Element} A span element containing the formatted date string(s)
+ */
+export const FormattedDateString = ({ date, primary, secondary }) => {
   return (
     <span>
-      <div>{formatDate(primary)}</div>
+      <div>{formatDate(date, primary)}</div>
       {secondary && (
-        <div style={{ fontSize: '0.8em', color: 'gray' }}>
-          {formatDate(secondary)}
+        <div style={{ fontSize: '0.8em', color: theme.palette.text.secondary }}>
+          {formatDate(date, secondary)}
         </div>
       )}
     </span>


moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.js
```diff
@@ -118,8 +120,7 @@ const useColumns = ({ deleteInProgress, onDeleteActivity, setEditActivity }) =>
         field: "updated_at",
         width: 150,
         defaultVisible: true,
-        type: "date",
-        valueGetter: (field) => (field ? new Date(field) : null),
+        valueGetter: (field) => (field ? formatRelativeDate(field) : null),
```
mddilley, line 123: I think this is great for faster comprehension of when a record was updated, and we can now compare when records were updated within the same day which we couldn't do before. 🚀

Beyond this PR, this made me think even more about how we display timestamps elsewhere. We show the same level of info for status updates in the Project Summary page, but we show both this human-readable format and the actual date and time in the activity log for example.

Screenshot 2024-09-04 at 10 50 12 AM

Looking for feedback because I think a future issue to make a reusable DataGrid cell component for timestamps could be interesting. Another place would be the **Modified** column of the Project List view for example.

mddilley (approved):
Tested this on the deploy preview, and it looks great! Awesome to see same-day updates are now comparable. 🙏 I added a comment for some discussion on a future change to similar timestamps.

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.js
frankhereford: I was just looking at this, and came to the code looking for a place to share a very similar opinion to what you have here Mike. I am all-in on the relative date, but not at the expense of the absolute date. I would fully support enhancing this component to render the date as you have it in your screenshot, or at the very least, showing the absolute date on-mouseover of the relative date, perhaps with some light text decoration in the form of a dashed underline.

frankhereford (approved):
Thanks @mateoclarke - I'm so glad you're back! I think your work here fits the request of the issue to a tee, and the code looks great. I have a comment that lines up with Mike's assertion about including the absolute date -- I would support that strongly. 

But that's out of scope, and this PR is ready to roll. Thanks again!

roseeichelmann (approved):
looks good, i think the relative date is super useful for updates that happened in the past day or couple of days. i agree that having absolute date somewhere maybe on hover or underneath like in the screenshot mike shared would be nice. also i can see that for dates that are super far back, it is the absolute date that is formatted. how far back does a date have to be for it to get formatted as the absolute date?
image

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.js
mddilley: that sounds like a cool idea to me, Mateo! thanks for the discussion, y'all. There are other cells that use `renderCell` instead of `valueGetter` if you want to see some examples of cells using components. And the screenshot that I shared comes from the `ProjectActivityLog` component.

chiaberry: there are some differences with renderCell vs valueGetter and valueFormatter with how DataGrid uses them to sort, just throwing that out there
https://mui.com/x/react-data-grid/column-definition/#value-getter

We'd have to include a sortComparator in that case.

mddilley:
@roseeichelmann great question - i see in the code that the date format kicks in after 30 days which scores another point for showing them both to me.

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.js
johnclary: this all sounds great—i'm glad y'all are going to get this standardized across the app. fwiw, Github populates the full timestamp into the `title` attribute on their timestamp elements so that it appears on mouse-over via built-in browser behavior. it would be easy enough for us to do that, but i don't find it intuitive at all and i doubt our users would know to look for that behavior:

screenshot of this hover effect:
Screenshot 2024-09-05 at 11 19 32 AM

moped-editor/src/utils/dateAndTime.js
```diff
@@ -114,3 +114,43 @@ export const formatRelativeDate = (timeStampTZString) => {
...
+      default:
+        return new Date(date).toLocaleString();
+    }
+  };
+
+  return (
+    <span>
+      <div>{formatDate(primary)}</div>
+      {secondary && (
+        <div style={{ fontSize: '0.8em', color: 'gray' }}>
```
mddilley, line 150: @mateoclarke Could you update this to `theme.palette.text.secondary`? That way it matches our secondary text color throughout the app.

moped-editor/src/utils/dateAndTime.js
```diff
@@ -114,3 +114,43 @@ export const formatRelativeDate = (timeStampTZString) => {
...
+
+/**
+ * A component that renders a formatted date string with primary display and optional secondary display
+ * @param {Object} props - The component props
+ * @param {string} props.date - The date to format, in a format parseable by new Date()
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} props.primary - The format to use for the primary display
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} [props.secondary] - The format to use for the secondary display, if any
+ * @returns {JSX.Element} A span element containing the formatted date string(s)
+ */
+export const FormattedDateString = ({ date, primary, secondary }) => {
```
mddilley, line 126: Love how the primary/secondary hierarchy mirrors MUI patterns. 🚀

moped-editor/src/utils/dateAndTime.js
```diff
@@ -114,3 +114,43 @@ export const formatRelativeDate = (timeStampTZString) => {
...
+  const formatDate = (format) => {
```
mddilley, line 127: could you pull the definition of this function outside of this component so it isn't recreated on each render?

mddilley (changes requested):
This looks so great!! Love having this extra info - thanks for pushing to make a new component to handle timestamps. Going to be so handy to have it around. ✨✨✨ I requested a couple small changes, but this is 🚢-ready after those changes.

moped-editor/src/utils/dateAndTime.js
mddilley: this was actually a great reminder that we have a backlog issue to update the secondary color to be more accessible 🙏 https://github.com/cityofaustin/atd-data-tech/issues/17920

mddilley (approved):
🚢 🚢 🚢 Thanks for this improvement! 📆 🚀

mddilley:
Forgot to mention that I tested sorting this column too, and it worked as expected. 🚀

moped-editor/src/utils/dateAndTime.js
```diff
@@ -114,3 +116,55 @@ export const formatRelativeDate = (timeStampTZString) => {
...
+
+/**
+ * A component that renders a formatted date string with primary display and optional secondary display
+ * @param {Object} props - The component props
+ * @param {string} props.date - The date to format, in a format parseable by new Date()
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} props.primary - The format to use for the primary display
+ * @param {('relative'|'absolute'|'expanded'|'short'|'time'|'fullTime')} [props.secondary] - The format to use for the secondary display, if any
+ * @returns {JSX.Element} A span element containing the formatted date string(s)
+ */
+export const FormattedDateString = ({ date, primary, secondary }) => {
```
johnclary, line 159: i love this fancy new component—really nice. are there more places in the code where we can drop this in?

johnclary: awesome!

PR #1411, opened 2024-09-10
Update project development status and date fields in AGOL DB view

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18877
Closes https://github.com/cityofaustin/atd-data-tech/issues/18876
Closes https://github.com/cityofaustin/atd-data-tech/issues/18706

This PR updates the `project_development_status_date` to include the component-level phase completion dates, and it adds formats of this date that are useful for ATSD reporting.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start up your local stack
2. Run the query below to see the new formats of the existing `project_development_status` field
   - Important note: Fiscal year starts in October so the fiscal fields use an offset on the year and quarter (i.e. calendar Quarter 4 = Fiscal Quarter 1 and calendar year October 2023 = fiscal year 2024).
3. Notice that the `project_development_status_date` value matches the component phase completion date set on this component -> https://localhost:3000/moped/projects/1566?tab=map&project_component_id=2095

```sql
SELECT
	project_id,
	project_component_id,
	project_development_status_date,
	project_development_status_date_calendar_year,
	project_development_status_date_calendar_year_month,
	project_development_status_date_calendar_year_month_numeric,
	project_development_status_date_calendar_year_quarter,
	project_development_status_date_fiscal_year,
	project_development_status_date_fiscal_year_quarter
FROM
	component_arcgis_online_view
WHERE project_id = 1566 and project_component_id = 2095;
```

---
#### Ship list
- [x] Code reviewed 
- [x] Product manager checked [DB view dependencies](https://atd-dts.gitbook.io/moped-documentation/product-management/updating-the-arcgis-online-components-database-view) - let Charlie know about the incoming update


moped-database/views/component_arcgis_online_view.sql
```diff
@@ -225,4 +229,5 @@ LEFT JOIN moped_components mc ON mpc.component_id = mc.component_id
 LEFT JOIN related_projects rp ON mpc.project_id = rp.project_id
 LEFT JOIN latest_public_meeting_date lpmd ON mpc.project_id = lpmd.project_id
 LEFT JOIN earliest_active_or_construction_phase_date eaocpd ON mpc.project_id = eaocpd.project_id
+LEFT JOIN LATERAL (SELECT timezone('US/Central'::text, get_project_development_status_date(lpmd.latest::timestamp with time zone, eaocpd.earliest, coalesce(mpc.completion_date, plv.substantial_completion_date), plv.substantial_completion_date_estimated, coalesce(mph.phase_name_simple, current_phase.phase_name_simple))) AS result) project_development_status_date ON true
```
mddilley, line 232: Noting to myself that the SQLFluff GH action made a change here:
`LEFT JOIN LATERAL (SELECT get_project_development_status_date(lpmd.latest::timestamp with time zone, eaocpd.earliest, coalesce(mpc.completion_date, plv.substantial_completion_date), plv.substantial_completion_date_estimated, coalesce(mph.phase_name_simple, current_phase.phase_name_simple)) AT TIME ZONE 'US/Central' AS result) AS project_development_status_date ON true -- noqa: ST05`

mddilley: I tested the output of the pre- and post-linted version of this view, and I'd like for the version in the migration to match up with what shows up in our reference file so I decided to update the migration with the post-linted version.

mddilley: This was actually from Postgres since this GH action uses SQLFluff for linting, creates the view in the DB, and then reads what Postgres created and store out of the DB to create this output.

moped-etl/arcgis/settings.py
```diff
@@ -5,8 +5,6 @@
 COMPONENTS_QUERY_BY_LAST_UPDATE_DATE = """
 query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!) {
   component_arcgis_online_view(where: $where) {
-    completion_date
-    completion_end_date
```
johnclary, line 9: 👍 just a reminder to delete these columns from the AGOL layers as well

johnclary (approved):
I ran your test—everything looks good to me!

Nice to see these redundant completion date fields bite the dust 🚮 

🚢

moped-etl/arcgis/settings.py
mddilley: ah yes, thanks for helping me stay on top of these! 🙏 They are added to an issue to do a bulk removal. 🧹

moped-database/views/component_arcgis_online_view.sql
roseeichelmann: side note is it necessary for `project_development_status_date` to be a timestamp bc it seems like there is no customizing the time for that column its just selecting a date and time is always gonna be `00:00:00`?

roseeichelmann (approved):
this all looks and works great mike!! loving seeing the sqlfluff github action in play :ship: :sparkles:

moped-database/views/component_arcgis_online_view.sql
chiaberry: @roseeichelmann , Amenity asked the same question, and the answer is timezones!
https://austininnovation.slack.com/archives/CNUEPKLB1/p1725395804733349

chiaberry (approved):
Always amazes me how complicated time is and then when you add in the concept of a fiscal year -- you always make this look easy

mddilley:
thanks, y'all!

moped-database/views/component_arcgis_online_view.sql
johnclary: i will add that when in Moped v1.0 a lot of these columns were actually the postres `date` column type, which holds just a naive date string like `2020-01-01`. and we had timezone because all of those dates were stored from local time!

roseeichelmann: okay i was wondering if the timezone bugs were the reason, thanks guys! time is so :face_with_spiral_eyes:

PR #1423, opened 2024-09-19
Fix work types not showing as options with Project Extent - Generic

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/19133

## Testing
**URL to test:** 

Deploy preview

**Steps to test:**
1. Create a new component and choose the **Project Extent - Generic** component type
2. Try to select a work type and you should see three options - New, Maintenance / Repair, and Modification
3. Try the same in staging and you should see that no options show up in the work type dropdown
4. Check that subcomponents still work too since I updated that line to avoid this same type of bug

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Product manager checked [DB view dependencies](https://atd-dts.gitbook.io/moped-documentation/product-management/updating-the-arcgis-online-components-database-view)
- [ ] Product manager added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1425, opened 2024-09-19
Patch 2.19.2

mddilley:
## Associated issues
This patch includes the fix from https://github.com/cityofaustin/atd-moped/pull/1423 for the bug described in https://github.com/cityofaustin/atd-data-tech/issues/19133.


chiaberry (approved):
🩹 patch it!

PR #1426, opened 2024-09-20
Fix text overflowing outside of the Work Order Link column

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19047

Just needed to set the Link to display as `block` for the other styles to take effect. 

I also decided to move a function out of the Footer definition, just to try to cut down on the calls to get the github tag. 

## Testing
**URL to test:** 

https://deploy-preview-1426--atd-moped-main.netlify.app/moped/projects/2039?tab=funding

**Steps to test:**

I updated the project linked above to have the same long link as the project in the issue description, so compare above to https://mobility.austin.gov/moped/projects/3598?tab=funding

---
#### Ship list
- [x] Code reviewed 
   - Added test for version link

### DIFF
--- a/moped-editor/src/layouts/DashboardLayout/Footer.js
+++ b/moped-editor/src/layouts/DashboardLayout/Footer.js
@@ -15,17 +15,17 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 /**
- * Async function that updates the mostRecentTag state based on API call to Github "tag" endpoint
+ * Async function that retrieves most recent tag from an API call to Github "tag" endpoint
  */
-const getMostRecentGithubTags = async (setMostRecentTag) => {
+const getMostRecentGithubTags = async () => {
   const githubTagsApiEndpoint =
     "https://api.github.com/repos/cityofaustin/atd-moped/tags";
   const response = await fetch(githubTagsApiEndpoint);
   const data = await response.json();
   if (get(data, 0)) {
-    setMostRecentTag(data[0].name);
+    return data[0].name;
   } else {
-    setMostRecentTag(`v${pckg.version}`);
+    return `v${pckg.version}`;
   }
 };
 
@@ -34,8 +34,9 @@ const Footer = () => {
   const [mostRecentTag, setMostRecentTag] = useState(`v${pckg.version}`); // use default tag
 
   useEffect(() => {
-    getMostRecentGithubTags(setMostRecentTag);
-  }, [setMostRecentTag]);
+    const tag = getMostRecentGithubTags();
+    setMostRecentTag(tag);
+  }, []);
 
   return (
     <div className={classes.root}>
--- a/moped-editor/src/layouts/DashboardLayout/Footer.js
+++ b/moped-editor/src/layouts/DashboardLayout/Footer.js
@@ -1,8 +1,7 @@
-import React, { useEffect, useState } from "react";
+import React from "react";
 import { Typography } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import ExternalLink from "../../components/ExternalLink";
-import { get } from "lodash";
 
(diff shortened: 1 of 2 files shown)


mddilley (commented):
Looks great! I left one suggestion about the Footer update. Thank you for fitting that in here!

moped-editor/src/layouts/DashboardLayout/Footer.js
```diff
@@ -1,42 +1,41 @@
...
-      "https://api.github.com/repos/cityofaustin/atd-moped/tags";
-    const response = await fetch(githubTagsApiEndpoint);
-    const data = await response.json();
-    if (get(data, 0)) {
-      setMostRecentTag(data[0].name);
-    } else {
-      setMostRecentTag(`v${pckg.version}`)
-    }
-  };
+    getMostRecentGithubTags(setMostRecentTag);
```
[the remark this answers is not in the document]
chiaberry: thank you! i fixed it and pushed up my changes

moped-editor/src/layouts/DashboardLayout/Footer.js
```diff
@@ -1,42 +1,42 @@
...
+const getMostRecentGithubTags = async () => {
+  const githubTagsApiEndpoint =
+    "https://api.github.com/repos/cityofaustin/atd-moped/tags";
+  const response = await fetch(githubTagsApiEndpoint);
+  const data = await response.json();
+  if (get(data, 0)) {
+    return data[0].name;
+  } else {
+    return `v${pckg.version}`;
+  }
```
johnclary, line 29: my brain is moving slow right now... so we link to the latest release version, regardless of whether or not that release tag matches the package version? i don't mean to open this can of worms...

one more question, should we handle when the `response` is not `ok`? iirc we have an unhandled error when the github call fails.

mddilley: worms welcome! i have also wondered about whether our good release habits could eliminate this api call altogether. am i thinking about this right? we need two things:

- current full version number (aka `2.19.2`) which we can get from `package.json`
- a link to the latest tags which we can reach with https://github.com/cityofaustin/atd-moped/releases/latest  <-- this was a TIL

chiaberry: oh i didnt realize that link redirect existed! I want to get rid of the api call 

I scope creeped these worms and I have no regrets

moped-editor/src/layouts/DashboardLayout/Footer.js
```diff
@@ -17,34 +15,14 @@ const useStyles = makeStyles(theme => ({
...
-  };
 
   return (
     <div className={classes.root}>
       <Typography variant="caption" color="textSecondary">
         Moped{" "}
         <ExternalLink
           text={`v${pckg.version}`}
-          url={`https://github.com/cityofaustin/atd-moped/releases/tag/${mostRecentTag}`}
+          url="https://github.com/cityofaustin/atd-moped/releases/latest"
```
chiaberry, line 25: thank you Mike for finding out that we can just use this link all the time

johnclary (approved):
it's not scope creep if you end up deleting more lines than you added 👍 🚢

moped-editor/src/layouts/DashboardLayout/Footer.js
mddilley: Same - it was a nice surprise! So cool to have this feature covered in new and less GH API-dependent way!! 🚀

mddilley (approved):
Thanks for fixing up these url styles. Looks great! And thanks for improving the version link while you were at it! ✨ 🚢

roseeichelmann (approved):
:sparkles: lovely

PR #1427, opened 2024-09-23
Add audit and council district triggers to feature_school_beacons table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19105

## Testing

local because of db changes, use a copy of production

**Steps to test:**
1. To test update audit fields with dynamic parent table name, find a project updated a while back. Add a school beacon to the map. Save. Check that your project is now top of the list on the project list view, or run this to confirm the updated at time has changed

```sql
select project_id, project_name, updated_at from project_list_view plv where plv.project_id = [your id]
```

2. To test `set_updated_at()`, open the beacon you added in step one, and update the description. Look at the school beacons table, and you will see your beacon has an updated_at timestamp. Since there are so few beacons, I find it enough to just check the table:
```sql
select * from feature_school_beacons
```

3. To test council district, put a beacon in a district that you recognize. (the beacon at 1903 Ohlen road for example is in district 4). You will not see the council district updated on the summary page, that will be handled in a [future PR.](https://github.com/cityofaustin/atd-data-tech/issues/19111) But you can see the council district here: 

```sql
select fcd.id, council_district_id, fsb.beacon_name, mpc.project_id from features_council_districts fcd 
left join feature_school_beacons fsb on fcd.feature_id  = fsb.id
left join moped_proj_components mpc on mpc.project_component_id = fsb.component_id
where mpc.project_id = [your project]
```

5. After you have done your changes above, you can see the activity log event invocations in the console:
http://localhost:9695/console/events/data/activity_log_feature_school_beacons/logs
The entries on the activity tab are sourced from the moped proj components table.

---
#### Ship list
- [ ] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectComponents/utils/makeComponentData.js
```diff
@@ -178,7 +178,7 @@ export const getFeatureChangesFromComponentForm = (
   } else if (newSchoolBeaconKnackId) {
     if (
       previousSchoolBeacon &&
-      newSchoolBeaconKnackId !== previousSchoolBeacon.id
```
mddilley, line 181: good catch! 🙏 this is making me question why we didn't go with Knack id for signals too, but I think that `signal_id` has shown to be reliable for them.

mddilley (approved):
I tested the migration up and down, and I was able to see the audit fields populate and update as expected in the feature, project component, and project levels. 🚢 🚀😎

johnclary (approved):
yep—it works great!

in order to test the `updated_at` on the beacon feature, i had to edit the component and pick a different beacon. this caused the original beacon feature to become deleted, thus triggering the update trigger. i don't think there's any other way to edit a beacon feature besides that 🔍 

i love seeing these trigger functions get attached to a new table—nice!! 🚢 🚢 🚢 🚢

PR #1428, opened 2024-09-24
Patch Moped AGOL ETL to handle invalid HTML (into `main`)

mddilley:
## Associated issues
n/a

See https://austininnovation.slack.com/archives/CHZE6BC6L/p1727122151675579

I'll keep running the ETL manually tomorrow to give us time to review. Thanks y'all!

## Testing
**URL to test:** 

n/a

**Steps to test:**
1. Start the local stack and get a fresh snapshot
2. Go to `atd-moped/moped-etl/arcgis` and follow the readme to set up your `env_file`
3. Run the ETL (the date arg in the command captures the project status update to project 3601 that surfaced the bug):
```bash
$ docker compose run arcgis -d 2024-09-23T19:18:34.543Z
```
4. You should see the script run and log one instance of the message `Invalid HTML tag found in project_id: 3601. Escaping HTML...` for project 3601. Other than the message, you'll see the script deleting updated features from the 4 AGOL layers and replacing the features for the updated projects.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-etl/arcgis/components_to_agol.py
+++ b/moped-etl/arcgis/components_to_agol.py
@@ -4,7 +4,6 @@
 import argparse
 import logging
 import json
-import html
 from datetime import datetime, timezone, timedelta
 
 from bs4 import BeautifulSoup
@@ -74,11 +73,10 @@ def handle_status_updates(features):
         if status_update != None and has_html_tags(status_update):
             if not is_valid_HTML_tag(status_update):
                 logger.info(
-                    f"Invalid HTML tag found in project_id: {id}. Escaping HTML..."
+                    f"Invalid HTML tag found in project_id: {id}. Getting HTML text..."
                 )
-                record["attributes"]["project_status_update"] = html.escape(
-                    status_update
-                )
+                html_text = BeautifulSoup(status_update, "html.parser").get_text()
+                record["attributes"]["project_status_update"] = html_text
 
     return features
 
@@ -248,7 +246,6 @@ def main(args):
         for feature_type in ["points", "lines", "combined", "exploded"]:
             logger.info(f"Processing {feature_type} features...")
             features_of_type = all_features[feature_type]
-
             features = handle_status_updates(features_of_type)
 
             logger.info("Deleting all existing features...")
@@ -272,6 +269,9 @@ def main(args):
         # Delete outdated feature from AGOL and add updated features
         for feature_type in ["points", "lines", "combined", "exploded"]:
             logger.info(f"Processing {feature_type} features...")
+            features_of_type = all_features[feature_type]
+            features = handle_status_updates(features_of_type)
+
             logger.info(
                 f"Deleting all {len(all_features[feature_type])} existing features in {feature_type} layer for updated projects in chunks of {UPLOAD_CHUNK_SIZE}..."
             )
@@ -281,10 +281,6 @@ def main(args):
                 joined_project_ids = ", ".join(str(x) for x in delete_chunk)
                 logger.info(f"Deleting features with project ids {joined_project_ids}")
                 delete_features_by_project_ids(feature_type, joined_project_ids)
-
-            features_of_type = all_features[feature_type]
-
-            features = handle_status_updates(features_of_type)
 
             logger.info(
                 f"Uploading {len(features)} features in chunks of {UPLOAD_CHUNK_SIZE}..."
--- a/moped-etl/arcgis/components_to_agol.py
+++ b/moped-etl/arcgis/components_to_agol.py
@@ -254,21 +254,20 @@ def main(args):
 
             features = all_features[feature_type]
 
-            # Check project status updates for valid or invalid HTML; escape html if needed
+            # Check project status updates for valid or invalid HTML; escape HTML if needed
             for record in features:
                 id = record["attributes"]["project_id"]
 
                 status_update = record["attributes"]["project_status_update"]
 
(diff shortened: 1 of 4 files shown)


chiaberry (approved):
Excellent sleuthing and excellent fix -- I got this message after the one about escaping HTML, another clue to why agol says ❌ to this html?
`MarkupResemblesLocatorWarning: The input looks more like a filename than markup. You may want to open this file and pass the filehandle into Beautiful Soup.`

moped-etl/arcgis/components_to_agol.py
```diff
@@ -222,6 +254,21 @@ def main(args):
 
             features = all_features[feature_type]
 
+            # Check project status updates for valid or invalid HTML; escape HTML if needed
```
johnclary, line 257: @mddilley it looks like this new html check is only applied when using `args.date` and not `args.full`. i'm still reviewing but wanted to give you that heads-up.

mddilley: 🤯 thanks for catching that John! i'm fixing right now

mddilley:
@chiaberry thanks for mentioning that new message. I'm looking at the project IDs logged to see which note is causing that warning. 🙏

moped-etl/arcgis/components_to_agol.py
```diff
@@ -220,7 +282,9 @@ def main(args):
                 logger.info(f"Deleting features with project ids {joined_project_ids}")
                 delete_features_by_project_ids(feature_type, joined_project_ids)
 
-            features = all_features[feature_type]
+            features_of_type = all_features[feature_type]
+
+            features = handle_status_updates(features_of_type)
```
johnclary, line 287: i like this new function—you might want to move it up before `delete_features_by_project_ids`. that way, if it happens to error out you won't be left with partially deleted features.

moped-etl/arcgis/components_to_agol.py
```diff
@@ -23,6 +26,63 @@
...
+    See: https://pypi.org/project/beautifulsoup4/
+
+    Args:
+        html_string_to_check (str): A string that contains HTML to test for valid HTML.
+
+    Returns:
+        bool: True if the string contains valid HTML, False otherwise
+    """
+    soup = BeautifulSoup(html_string_to_check, "html.parser")
+    return html_string_to_check == str(soup)
```
johnclary, line 55: this is definitely catching the case of the busted HTML, but i'm not totally understanding why. when i compare the before and after strings, all i see is that the `rel` and `href` tags are swapped. bs4 is obviously doing some linting, and it makes me wonder if this check is going to sweep up valid html from time to time.

i'm in favor of getting this patched up and running again asap, and maybe it'd be worth revisiting this there's no sense of urgency?

moped-etl/arcgis/components_to_agol.py
```diff
@@ -23,6 +26,63 @@
...
+        status_update = record["attributes"]["project_status_update"]
+
+        if status_update != None and has_html_tags(status_update):
+            if not is_valid_HTML_tag(status_update):
+                logger.info(
+                    f"Invalid HTML tag found in project_id: {id}. Escaping HTML..."
+                )
+                record["attributes"]["project_status_update"] = html.escape(
+                    status_update
+                )
```
johnclary, line 81: TIL about the built-in `html` library 👍 

since this ends up loading html tags as text into the status update, i wanted to offer one alternative, which would be to use bs4 to grab the text content instead.

```shell
soup.get_text()
```

in the case of our busted html that spits out:
```
'Bluebeam session for design review by PDD PM (by Friday, 9/20)Cesar Chavez Blank Out SignsSession ID: 187-616-135Session URL: https://studio.bluebeam.com/hyperlink.html?link=studio.bluebeam.com/sessions/187-616-13 
```

i'm not sure how reliable this—maybe worth punting it to another issue.

johnclary (approved):
Mike, this is looking good and for sure fixing the problem. I offered some feedback that might be useful in a follow-up issue.

Thanks for sorting this out—there have been lots of lessons learned through this! 🚢 🚢

moped-etl/arcgis/components_to_agol.py
mddilley: i'm updating this so that incremental mode has the same order of function calls in the full mode. thanks for flagging! 🙏

moped-etl/arcgis/components_to_agol.py
mddilley: i like this idea a lot. Feels better than outputting HTML tags and will look nicer. Thanks for finding this method!

moped-etl/arcgis/components_to_agol.py
mddilley: @johnclary i should have include more info about this change. At the end of the status update, there is a break tag that is `<br>` and the str() method converts that to be self-closing `<br/>`. 

I can't explain the switch of the rel and href attributes. 🤷

johnclary: ah, thanks for walking me through that. i stared at this for a while and must have missed it. is there any hope that replacing the status update with the output of `str(soup)` makes it valid HTML that can be uploaded?

mddilley: john and i chatted a bit about this, and i really liked the idea but the AGOL API did not. 🙃 Sticking to text for now.

PR #1429, opened 2024-09-24
Patch Moped AGOL ETL to handle invalid HTML (into `production`) 

mddilley:
## Associated issues
Patching the same changes in https://github.com/cityofaustin/atd-moped/pull/1428 in production.


johnclary (approved):
✅

PR #1436, opened 2024-09-30
Create a generic point component type

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19132

## Testing
**URL to test:** 

Local

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/migrations/1727471090930_update_generic_proj_ext/down.sql
+++ b/moped-database/migrations/1727471090930_update_generic_proj_ext/down.sql
@@ -1,0 +1,5 @@
+DELETE FROM moped_components WHERE component_name = 'Project Extent - Generic' AND line_representation = false;
+
+UPDATE moped_components
+SET component_name = 'Project Extent - Generic'
+WHERE component_name = 'Project Extent - Generic (linear)' AND line_representation = true;
--- a/moped-database/migrations/1727471090930_update_generic_proj_ext/down.sql
+++ b/moped-database/migrations/1727471090930_update_generic_proj_ext/down.sql
@@ -1,4 +1,4 @@
-DELETE FROM moped_components WHERE component_name = 'Project Extent - Generic' AND line_representation = false;
+UPDATE moped_components SET is_deleted = true WHERE component_name = 'Project Extent - Generic' AND line_representation = false;
 
 UPDATE moped_components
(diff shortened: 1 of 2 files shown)


chiaberry (commented):
The work types do not include replacement, but include maintenance / repair. I think the test instructions are incorrect, because this does match what is in production for generic type now. 
Screenshot 2024-10-04 at 9 21 50 AM

As for restoring the point type component if it was soft deleted, this would only happen if the down migration was run at some point and then we needed to run the up migration again, right?

I'll mark this approved after its confirmed that we didnt need to update the work types for the generic project extent line.

moped-database/migrations/1727471090930_update_generic_proj_ext/down.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE moped_components SET is_deleted = true WHERE component_name = 'Project Extent - Generic' AND line_representation = false;
```
johnclary, line 1: in the spirit of having the down migrations restore the DB to the previous state, i think the right move here would be to actually hard delete the component that the up migration created. obviously, that's going to fail if there are any `moped_proj_components` that reference the component, so here are a few ideas:
- just acknowledge that the down migration will not work without manual intervention by the dev to deal with the extant `moped_proj_components` that reference it
- write the down migration to also migrate the `moped_proj_components` to some other component type (like intersection improvement generic)
- write the down migration to hard delete `moped_proj_components` which use this component type, yolo style

in any case, i don't think this is a big deal so i'm going to approve your PR! 🤷

johnclary (approved):
i approve but also +1 to Chia's comments about the work types 🙏 🚢

mddilley:
@chiaberry Thank you for catching that detail in the test instructions. 🙏 I just corrected them to include Maintenance / Repair and not Replacement. I think that I copypasta'd that from a different component type.

chiaberry (approved):
🚢  generic 👉🏼

tillyw (approved):
tested locally and the new generic component and naming conventions look great! can add and edit work types as expected. ran the down migration and the records i created reverted without issue. thanks for making this update! ship ship ship

mddilley:
thanks y'all - merging!

PR #1437, opened 2024-09-30
Update MUI secondary color

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17920

Screenshot 2024-09-30 at 12 34 52 PM

Screenshot 2024-09-30 at 12 35 47 PM

## Testing
**URL to test:** 
https://deploy-preview-1437--atd-moped-main.netlify.app/moped/dashboard

**Steps to test:**
 - Open up a page with secondary text, for example, the welcome text on the dashboard. 
 - Confirm it is the updated shade of grey, `#666666`

Screenshot 2024-09-30 at 6 01 32 PM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/theme/index.js
+++ b/moped-editor/src/theme/index.js
@@ -42,6 +42,16 @@ const theme = createTheme(adaptV4Theme({
   },
   shadows,
   typography,
+  overrides: {
+    MuiInputBase: {
+      input: {
+        "&::placeholder": {
+          color: "#666666",
+          opacity: 1,
+        },
+      },
+    },
+  },
 }));
 
 export default theme;


moped-editor/src/theme/index.js
```diff
@@ -26,7 +26,7 @@ const theme = createTheme(adaptV4Theme({
     },
     text: {
       primary: "#212121",
-      secondary: "#848484",
+      secondary: "#666666",
       header: "#848484",
```
mddilley, line 30: This is outside the scope of the issue, but I noticed that we use `text.header` in only three places in the app where we have tabs: the Dashboard view, the main app header, and in the project view. For some reason, the app header respects `text.header` but the other two do not.

This is making me feel like the header style isn't necessary since it isn't working as expected. What do you think about removing this style and those three places that it is used? All three would then default to the new secondary color that you updated.

mddilley (commented):
Looks great! Much easier to read the small text. I checked every view that references this color and every part of the app. I have one small request before merge that I could use some feedback on. Thanks for updating! 🙏

moped-editor/src/theme/index.js
mddilley: sounds great - thanks!

mddilley (approved):
Looks amazing! Thanks for making this improvement. 🙏 🚢

roseeichelmann (approved):
looks good, yay for better accessibility :)

chiaberry (approved):
Easier on the eyes!

moped-editor/src/theme/index.js
```diff
@@ -43,6 +42,16 @@ const theme = createTheme(adaptV4Theme({
...
+  overrides: {
+    MuiInputBase: {
+      input: {
+        "&::placeholder": {
+          color: "#666666",
+          opacity: 1,
+        },
+      },
+    },
+  },
```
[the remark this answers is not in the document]
mddilley: thanks @mateoclarke ! Mateo and I talked about this in our catchup today, and I agree with this belonging in the global override to make it clear what were are doing across the app. I looked at the code closer as well, and we are not concerned with styling input placeholders and input labels differently so this feels right to me!

PR #1438, opened 2024-10-01
Migrate Files table to data grid

chiaberry:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFiles.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFiles.js
@@ -6,7 +6,6 @@ import {
   CardContent,
   CircularProgress,
   Link,
-  TextField,
   Typography,
   FormControl,
   Select,
@@ -15,7 +14,6 @@ import {
 } from "@mui/material";
 
 import makeStyles from "@mui/styles/makeStyles";
-import typography from "../../../theme/typography";
 import {
   DeleteOutline as DeleteOutlineIcon,
   EditOutlined as EditOutlinedIcon,
@@ -100,16 +98,7 @@ const useColumns = ({
           return { ...params.props, error: hasError };
         },
         renderEditCell: (props) => (
-          <DataGridTextField
-            helperText="Required"
-            {...props}
-            // variant="standard"
-            // id="file_name"
-            // name="file_name"
-            // value={props.value}
-            // onChange={(e) => props.onChange(e.target.value.trim())}
-            // helperText="Required"
-          />
+          <DataGridTextField helperText="Required" {...props} />
         ),
       },
       {
@@ -120,7 +109,7 @@ const useColumns = ({
         // validate input
         preProcessEditCellProps: (params) => {
           const hasError = params.props.value.trim().length < 1;
-          return {...params.props, error:hasError}
+          return { ...params.props, error: hasError };
         },
         renderCell: ({ row }) => {
           if (row.file_key) {
@@ -146,25 +135,16 @@ const useColumns = ({
             </Typography>
           );
         },
-        renderEditCell: ({ row }, props) =>
+        renderEditCell: (props) =>
           // users cannot edit the file_key, since its provided by the FilePond upload interface
-          row.file_key ? (
-            <Typography>{cleanUpFileKey(row.file_key)}</Typography>
+          props.row.file_key ? (
+            <Typography>{cleanUpFileKey(props.row.file_key)}</Typography>
           ) : (
             <DataGridTextField
               helperText="Required"
-              disabled={!!row.file_key}
+              disabled={!!props.row.file_key}
               {...props}
             />
-            // <TextField
-            //   variant="standard"
-            //   id="file_path"
-            //   name="file_path"
-            //   value={row.file_url}
-            //   // onChange={(e) => props.onChange(e.target.value.trim())}
-            //   helperText="Required"
-            //   disabled={!!row.file_key}
-            // />
           ),
       },
       {
@@ -194,16 +174,7 @@ const useColumns = ({
         field: "file_description",
         editable: true,
         width: 200,
-        renderEditCell: (props) => (
-          <DataGridTextField
-            {...props}
-            // variant="standard"
-            // id="file_description"
-            // name="file_description"
-            // value={props?.value ?? ""}
-            // onChange={(e) => props.onChange(e.target.value)}
-          />
-        ),
+        renderEditCell: (props) => <DataGridTextField {...props} />,
       },
       {
         headerName: "Uploaded by",
@@ -435,8 +406,6 @@ const ProjectFiles = () => {
 
   // saves row update, either editing an existing row or saving a new row
   const processRowUpdate = (updatedRow, originalRow) => {
-    const updateProjectFileData = updatedRow;
-
     const hasRowChanged = !isEqual(updatedRow, originalRow);
 
     if (!hasRowChanged) {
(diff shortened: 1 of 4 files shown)


moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -56,33 +51,225 @@ const useStyles = makeStyles((theme) => ({
   },
   downloadLink: {
     cursor: "pointer",
+    overflow: "hidden",
+    display: "block",
+    textOverflow: "ellipsis",
```
johnclary, line 56: i'm not sure if this is working quite right—here's a long file link in project 437
Screenshot 2024-10-04 at 11 48 33 AM

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -56,33 +51,225 @@ const useStyles = makeStyles((theme) => ({
...
+      {
+        headerName: "Name",
+        field: "file_name",
+        width: 175,
+        editable: true,
+        // validate input
+        preProcessEditCellProps: (params) => {
+          const hasError = params.props.value.trim().length < 1;
+          return { ...params.props, error: hasError };
+        },
```
johnclary, line 94: i'm not sure if this is source of the bug, but i'm hitting this error when i edit any column 💥 
Screenshot 2024-10-04 at 11 47 27 AM

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -56,33 +51,225 @@ const useStyles = makeStyles((theme) => ({
...
+          ),
+      },
+      {
+        headerName: "Type",
+        field: "file_type",
+        editable: true,
+        width: 150,
+        valueGetter: (value) => fileTypes[value],
+        renderEditCell: (props) => <ProjectFilesTypeSelect {...props} />,
+      },
```
johnclary, line 152: This input is a little squished (like half the width of the column) in edit mode. Maybe there are some tweaks needed in the Select component?
Screenshot 2024-10-04 at 11 47 44 AM

I am also getting the dreaded autocomplete warning when i switch into edit mode

Screenshot 2024-10-04 at 11 49 01 AM

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -56,33 +51,225 @@ const useStyles = makeStyles((theme) => ({
...
+    ];
+  }, [
+    classes,
+    token,
+    rowModesModel,
+    handleSaveClick,
+    handleCancelClick,
+    handleEditClick,
+    handleDeleteOpen,
+  ]);
```
johnclary, line 252: love the memoized columns—looks great!!

johnclary (changes requested):
Thank you for knocking out another Material Table instance. I'm running into some bugginess that I'm sure will be a quick fix 🙏 🙌

moped-editor/src/views/projects/projectView/ProjectFiles.js
chiaberry: ah because this isnt a link, this is just text. If it's truncated or overflow hidden, then users wouldn't be able to read the entire path

moped-editor/src/views/projects/projectView/ProjectFiles.js
chiaberry: it is the source, thanks for catching it -- the weird thing is that I see the bug locally, but I am not seeing it on the netlify deploy link, which is confusing the heck out of me

moped-editor/src/views/projects/projectView/ProjectFiles.js
chiaberry: I broke this when I added the valueGetter to make sorting work ☹️ . I am going to see how to fix this but also maintain ability to sort by the words and not the numbers

chiaberry: I broke the select when I used valueGetter, valueFormatter changes the value only for rendering and doesn't affect the underlying value. which means you still cant sort. valueGetter affects the underlying value, which means we can sort alphabetically vs 1,2,3,4 but it was then passing the string to the select component instead of the file_type number. 

So now its valueFormatter to render the file_type strings, and passing those strings into the data grid sort comparator, while keeping the original value so the select works.

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -134,278 +334,161 @@ const ProjectFiles = (props) => {
...
+          console.error(error);
+        });
     },
-  ];
+    [rows, deleteProjectFileAttachment, refetch]
+  );
+
+  // saves row update after editing an existing row
+  const processRowUpdate = (updatedRow, originalRow) => {
+    const hasRowChanged = !isEqual(updatedRow, originalRow);
```
mddilley, line 401: thanks for preventing the empty activity log rows! 🙏 I did notice an existing bug with file activities again. I moved [the issue for it into the backlog](https://github.com/cityofaustin/atd-data-tech/issues/17597) and made it a snackoo. 🍫

mddilley (changes requested):
Yay - this is looking great! I tested locally, and I was able to create linked files and edit from within the table rows. I'm requesting one change based on what I was seeing with long network locations for files that are links.

moped-editor/src/views/projects/projectView/ProjectFiles.js
johnclary: ditto—great to have these details ironed out for the next MaterialTable that gets replaced ✅

johnclary (approved):
Nice—it's working great! Two minor things that are not blocking my approval:

When I edit a row, sometimes the sort order changes after I save. That was tripping me up a bit because i thought my edits were not getting saved correctly. I'm not sure if this is a bug in prod, so I'll let y'all decide on it 👍 

Also, I noticed the activity logs for my edits were missing a field or two—possibly audit fields? Anyway, it's probably not a new bug.

Screenshot 2024-10-08 at 9 24 59 AM


🚢

moped-editor/src/views/projects/projectView/ProjectFiles.js
mddilley: this is now patched in this pr. 🚀 🙌

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -56,33 +52,232 @@ const useStyles = makeStyles((theme) => ({
...
   downloadLink: {
     cursor: "pointer",
+    overflow: "hidden",
+    display: "block",
+    textOverflow: "ellipsis",
   },
   codeStyle: {
     backgroundColor: "#eee",
     fontFamily: "monospace",
     display: "inline-block",
```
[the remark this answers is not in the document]
mddilley: tested and this looks great! 🙏

mddilley (approved):
Looks amazing! Thanks for updating the sort and handling the network locations wrapping too. So excited to have another table converted to DataGrid, and there are nice improvements in here too! It is going to make the next one even more straightforward. 🙌 🚢

tillyw (approved):
tested locally and looks and functions great! thank you for upgrading this table with DataGrid and adding the increased functionality, validation and modal! being able to inline edit is so nice ✨ 

Screenshot 2024-10-09 at 2 46 43 PM

a couple minor things we could maybe address in the future:
- is it possible to disable editing on the 'file' field when it is an actual file (as opposed to a url or file path) since you can't actually edit that field in that case?
- i know we've come across this before, but when i double click on the 'file' field when it is a url it tries to open the url rather than allowing me to edit the field

as far as column widths, i think we could shrink the 'file size' field by at least half and give that extra space to the 'file' field, since file size will probably only ever be around 5-6 characters

chiaberry:
@tillyw thank you for pointing out the file size column width could be smaller! 
and thanks everyone for the reviews

PR #1439, opened 2024-10-03
adds components dropdown and contains not operator

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/19052

## Testing
**URL to test:** 

https://deploy-preview-1439--atd-moped-main.netlify.app/

**Steps to test:**
- note the number of records in the project view with no filters applied
- add an advanced filter for 'components' 'contains', select an option and note the number of records
- replace the filter with the same query but using the 'does not contain' operator, and note whether the number of records makes up the difference. also check to make sure records with no value in the 'component' field appear.
- repeat the steps with the 'is' and 'is not' operators for a different field (such as 'type')
- confirm that fields with the 'is' and 'contains' operators also include their opposites

---
#### Ship list
- [x] Code reviewed 
   - Added to the existing "contains" test to check the sum of "contains" and "does not contain" to make sure it equals the total number of projects

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
+++ b/moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
@@ -80,7 +80,7 @@ const makeAdvancedSearchWhereFilters = (filters) =>
         }
       }
       // If we would like to return results that do not contain a particular string, we should return null values as well
-      if (gqlOperator.includes("_nilike") && envelope) {
+      if (gqlOperator.includes("_nilike")) {
         whereString = `_or: [ { ${whereString} }, { ${field}: { ${`_is_null`}: true } }]`
       }
       return whereString;
--- a/moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
+++ b/moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
@@ -79,6 +79,10 @@ const makeAdvancedSearchWhereFilters = (filters) =>
           whereString = `${field}: { ${"_gte"}: ${nextDay} }`;
         }
       }
+      // If we would like to return results that do not contain a particular string, we should return null values as well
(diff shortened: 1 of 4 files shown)


chiaberry (commented):
The test instructions say to repeat the steps with "is" and "is not" but I don't see them in the components drop down options 
Screenshot 2024-10-09 at 2 14 17 PM
I dont see that requirement in the issue, so I dont know if its okay its not there or if its an oversight

moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -63,6 +63,11 @@ const makeAdvancedSearchWhereFilters = (filters) =>
         }
       }
       let whereString = `${field}: { ${gqlOperator}: ${value} }`;
+      // If we would like to return results that do not contain a particular string ('_nilike'), we should return null values as well
+      if (gqlOperator.includes("_nilike")) {
+        const whereStringWithNullValues = `_or: [ { ${whereString} }, { ${field}: { ${`_is_null`}: true } }]`;
```
roseeichelmann, line 68: nice touch :ok_hand:

roseeichelmann (approved):
looks and works great! i noticed what chia pointed out too but im approving on the assumption that you meant to write "is blank" and "is not blank" in the test instructions since adding is and is not isn't in the issue requirements

tillyw:
oops sorry @chiaberry, i meant to check one of the other fields that has 'is' and 'is not' options (such as 'type') since this pr also corrects an issue where records with null values were not being returned under the 'is not' condition

chiaberry (approved):
Thanks for the clarification Tilly! I tested with Lead is not COA ATD and got all the results, whereas in staging you only get 179 results

🚢 🚢

mddilley (approved):
Such an improvement! I tested using every type of field search (date, string, string that is a concatenated list, and number) and all operators with and without dropdown options, and I see this working exactly as expected. 🚢 🚀 🙌

tillyw:
thanks y'all! gonna go ahead and merge this one :)

PR #1441, opened 2024-10-07
Fix missed project and component deletes in AGOL ETL incremental runs

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19233

## Testing
**URL to test:** 

Local

**Steps to test:**

**Note: I added a test mode (dry run) that does not update the AGOL dataset to make it easier to test this fix.**
1. Generate an ISO string timestamp using this JS or however you'd like
```js
const dateObj = new Date();
dateObj.toISOString();
```
2. Start your local Moped stack and set up an `env_file` for this script using the instructions in the readme if needed
3. Create a new project and add one point type component (like Bike Box) to **delete** in a later step
4. Create another new project and add one line type component (like Bike Lane) to **edit** in a later step
5. Run the script in test mode from `atd-moped/moped-etl/arcgis` with your ISO string timestamp:
```shell
docker compose run arcgis -d 2024-10-07T21:18:42.586Z -t
```
6. You should see `Deleting features with project ids <your two new project ids>...` logged four times and also see the counts 1, 1, 2, and 1 logged for each type associated with the 4 AGOL layers. This is showing that 1 point feature, 1 line feature, 2 combined (point and line), and one exploded point feature would be transferred if this was not a test/dry run.
7. Now, get a fresh timestamp and then add a project lead to the project with the bike lane. This tests a project edit and the resulting AGOL dataset update. Its features will be deleted from the AGOL dataset and the updated feature rows are transferred.
8. Run the script again in test mode with your newest timestamp
9. Now you should see `Deleting features with project ids <your bike lane project id>...` logged four times and also see the counts 0, 1, 1, 0 logged this time (1 line type and 1 combined type).
10. Get a new timestamp and then delete your bike box project. This tests what should happen when a project is deleted. Its features will be deleted from the AGOL dataset and no new feature rows are transferred.
11. Run the script again in test mode with your newest timestamp
12. Now you should see `Deleting features with project ids <your bike box project id>...` logged four times and see 0, 0, 0, 0 logged for each type since we don't have any features to transfer for the deleted project.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-etl/arcgis/README.md
+++ b/moped-etl/arcgis/README.md
@@ -37,4 +37,4 @@ The fourth layer is sourced from a derivative view, `exploded_component_arcgis_o
 To run the script without making changes to the AGOL dataset, use the `-t` flag (`--test`) to fetch the project IDs to delete and the component data to use for AGOL record replacement. This is useful to observe what projects have updated and what component data will be transferred without updating the production AGOL dataset.
 
 1. Run the script with the test flag and any of the available options above:
-   - `docker compose run arcgis -d` to start the script in test mode with the default interval of changes over the last week.
+   - `docker compose run arcgis -d <timestamptz> -t` to start the script in test mode with the default interval of changes over the last week.
--- a/moped-etl/arcgis/README.md
+++ b/moped-etl/arcgis/README.md
@@ -31,3 +31,10 @@ The fourth layer is sourced from a derivative view, `exploded_component_arcgis_o
    - `docker compose run arcgis -f` to start the script with a full refresh.
    - `docker compose run arcgis -d <timestamptz>` to start the script with a refresh since the given timestamp.
    - `docker compose run --entrypoint /bin/bash arcgis` to start a shell inside the container.
+
+## Testing the Script
+
+To run the script without making changes to the AGOL dataset, use the `-t` flag (`--test`) to fetch the project IDs to delete and the component data to use for AGOL record replacement. This is useful to observe what projects have updated and what component data will be transferred without updating the production AGOL dataset.
+
+1. Run the script with the test flag and any of the available options above:
+   - `docker compose run arcgis -d` to start the script in test mode with the default interval of changes over the last week.
--- a/moped-etl/arcgis/README.md
+++ b/moped-etl/arcgis/README.md
@@ -37,4 +37,4 @@ The fourth layer is sourced from a derivative view, `exploded_component_arcgis_o
 To run the script without making changes to the AGOL dataset, use the `-t` flag (`--test`) to fetch the project IDs to delete and the component data to use for AGOL record replacement. This is useful to observe what projects have updated and what component data will be transferred without updating the production AGOL dataset.
 
 1. Run the script with the test flag and any of the available options above:
-   - `docker compose run arcgis -d <timestamptz> -t` to start the script in test mode with the default interval of changes over the last week.
+   - `docker compose run arcgis -d <timestamptz> -t` to start the script in test mode with a refresh since the given timestamp.
--- a/moped-etl/arcgis/cli.py
+++ b/moped-etl/arcgis/cli.py
@@ -1,0 +1,37 @@
+import argparse
+from datetime import datetime, timezone, timedelta
+
+
+def get_cli_args():
+    """Create the CLI and parse args
+
+    Returns:
+        argparse.Namespace: The CLI namespace
+    """
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "-d",
+        "--date",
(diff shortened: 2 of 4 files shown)


moped-etl/arcgis/cli.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    parser.add_argument(
+        "-f",
+        "--full",
+        action="store_true",
+        help="Delete and replace all project components.",
+    )
+
+    parser.add_argument(
```
mddilley, line 30: I moved arg parsing into its own module since it was taking a lot of space in the main script. This is the new arg to run in test mode (dry run).

moped-etl/arcgis/settings.py
```diff
@@ -3,8 +3,11 @@
 LAYER_IDS = {"points": 0, "lines": 1, "combined": 2, "exploded": 3}
 
 COMPONENTS_QUERY_BY_LAST_UPDATE_DATE = """
-query GetProjectsComponents($where: component_arcgis_online_view_bool_exp!) {
-  component_arcgis_online_view(where: $where) {
+query GetProjectsComponents($project_where: moped_project_bool_exp!, $component_where: component_arcgis_online_view_bool_exp!) {
+  moped_project(where: $project_where) {
+    project_id
```
mddilley, line 8: I thought about using a [skip directive](https://hasura.io/docs/2.0/queries/postgres/variables-aliases-fragments-directives/#skipif-boolean) here in the query since we don't need them for the full replace mode. But, adding another variable and more logic in this query felt like a lot to prevent getting these ids in rare cases when we run the full replace.

moped-etl/arcgis/README.md
```diff
@@ -31,3 +31,10 @@ The fourth layer is sourced from a derivative view, `exploded_component_arcgis_o
    - `docker compose run arcgis -f` to start the script with a full refresh.
    - `docker compose run arcgis -d <timestamptz>` to start the script with a refresh since the given timestamp.
    - `docker compose run --entrypoint /bin/bash arcgis` to start a shell inside the container.
+
+## Testing the Script
+
+To run the script without making changes to the AGOL dataset, use the `-t` flag (`--test`) to fetch the project IDs to delete and the component data to use for AGOL record replacement. This is useful to observe what projects have updated and what component data will be transferred without updating the production AGOL dataset.
+
+1. Run the script with the test flag and any of the available options above:
+   - `docker compose run arcgis -d <timestamptz> -t` to start the script in test mode with the default interval of changes over the last week.
```
chiaberry, line 40: This is such a nice addition!

mddilley: glad to hear! and you just helped me catch a typo! 🙏

moped-etl/arcgis/components_to_agol.py
```diff
@@ -249,22 +254,19 @@ def main(args):
...
+                    add_features(feature_type, feature_chunk)
     else:
-        # Get unique project IDs that need to have features deleted & replaced
-        project_ids = []
-
-        for component in data:
-            project_ids.append(component["project_id"])
-
-        project_ids_for_feature_delete = list(set(project_ids))
+        # Get project IDs have been updated (including soft-deleted projects) for deletes
```
chiaberry, line 268: missing a word between IDs and have, "that" or "which" maybe?

chiaberry (approved):
I really like the `-t` option so I dont have to worry about messing up real data while testing

Great testing instructions, and the logging helps me understand whats going on as its running. 

🚢

moped-etl/arcgis/components_to_agol.py
```diff
@@ -249,22 +254,19 @@ def main(args):
...
+        project_ids_for_delete = [project["project_id"] for project in projects_data]
```
johnclary, line 269: cool—i see that the this will delete all components in the AGOL dataset that are associated with any project that has been updated recently, and it works around the issue of deleted components being excluded from the `component_arcgis_online_view` — nice!

johnclary (approved):
I ran this with the fancy new `-t` flag—works great. nice change! Glad to have this tricky bug sorted out 🚢 🚢 🚢 🚢

tillyw (approved):
this was a fun one to test - i felt like i was solving a puzzle! my console logged all the right counts and project ids, so everything worked as expected! thanks for adding the additional documentation and the new testing flag! made me feel a lot less anxious hitting enter >.<

moped-etl/arcgis/components_to_agol.py
mddilley: ooo thanks - i'll fix! 🙏

PR #1453, opened 2024-10-11
shows substantial completion date plus advanced search

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/19257

## Testing
**URL to test:** 

https://deploy-preview-1453--atd-moped-main.netlify.app/

**Steps to test:**
- go to projects list view, 'Substantial completion date' should be visible by default right before 'Start date'
- sort the projects by that field, try ascending, descending and unsorted
- export the table and confirm the column is visible and populated in the csv
- add an advanced search filter for the new field and try different operators and values

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -67,6 +67,9 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
   project_status_update: {
     label: "Status Update",
   },
+  substantial_completion_date: {
+    label: "Substantial completion date",
+  },
   construction_start_date: {
     label: "Construction start",
     filter: formatTimeStampTZType,
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -69,6 +69,7 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
   },
   substantial_completion_date: {
     label: "Substantial completion date",
+    filter: formatTimeStampTZType,
   },
   construction_start_date: {
     label: "Construction start",
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
@@ -292,7 +292,7 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
       getOptionLabel: (option) => option.funding_source_name,
       operators: [
         "string_contains_case_insensitive",
-        "string_contains_not_case_insensitive"
+        "string_contains_not_case_insensitive",
       ],
     },
     defaultOperator: "string_contains_case_insensitive",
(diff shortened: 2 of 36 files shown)


chiaberry (commented):
The test steps work out and it handles what the issue laid out!

However I noticed the export config wasn't updated, so I tested by downloading the visible columns while having substantial completion visible, and it gets stuck in an infinite loop. I didn't dig into it but I think its upset it cant find that column.

moped-editor/src/views/projects/projectsListView/helpers.js
```diff
@@ -312,6 +312,12 @@ export const useColumns = ({ hiddenColumns }) => {
         },
         width: COLUMN_WIDTHS.xlarge,
       },
+      {
+        headerName: "Substantial completion date",
+        field: "substantial_completion_date",
+        valueFormatter: (value) => value && formatDateType(value),
+        width: COLUMN_WIDTHS.small,
```
mddilley, line 319: Part of me feels like this could be updated to `COLUMN_WIDTHS.medium` so that the entire column header is visible by default. Counter point is that we don't do this for other longer column headers like **Public process status** or **Interim MPD (Access) ID**.

What you you think?

mddilley (changes requested):
Looks so great! I replicated the export issue that Chia found earlier. Could you update the export configuration to include this new column? 🙏 🙏 Apologies for missing that in the scope of issue

moped-editor/src/views/projects/projectsListView/helpers.js
tillyw: i had that thought as well! since this is a requested feature and i think will be highly utilized i think it's fair to give it better visibility. i think it looks better at medium so i will push that but i am open to changing ti back!

tillyw:
> The test steps work out and it handles what the issue laid out!
> 
> However I noticed the export config wasn't updated, so I tested by downloading the visible columns while having substantial completion visible, and it gets stuck in an infinite loop. I didn't dig into it but I think its upset it cant find that column.

thank you for catching that @chiaberry ! i definitely overlooked the export config. just pushed up a change that should fix that! will update test steps as well

chiaberry (approved):
🚢 !!

moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js
```diff
@@ -337,6 +337,21 @@ export const PROJECT_LIST_VIEW_FILTERS_CONFIG = {
...
+    placeholder: "Substantial completion date",
+    type: "date",
+    defaultOperator: "date_equals",
+    operators: [
+      "date_equals",
+      "date_greater_than",
+      "date_greater_than_equal_to",
+      "date_less_than",
+      "date_less_than_equal_to",
+      "date_is_null",
```
mddilley, line 352: beyond in the scope of this PR, but reading through this code made me realize we don't have an "is not blank" operator for dates. I created [an issue](https://github.com/cityofaustin/atd-data-tech/issues/19470) but it isn't a priority.

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
```diff
@@ -67,6 +67,9 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
   project_status_update: {
     label: "Status Update",
   },
+  substantial_completion_date: {
+    label: "Substantial completion date",
+  },
   construction_start_date: {
     label: "Construction start",
     filter: formatTimeStampTZType,
```
mddilley, line 75: Last change - I swear! @tillyw could you update the config for this field to use this same filter so it shows as a date (not a timestamp) in the export file?

mddilley (commented):
Looks and works great! Thanks for making the adjustments. I requested one last change for the export config to use a helper to format this field as a date in the export file.

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
tillyw: yes, thanks for pointing that out!

mddilley (approved):
Yay!!! Thanks for updating the export settings. 🚢 🙌

PR #1458, opened 2024-10-17
Remove funding_source_name from AGOL database view and ETL

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19023

This PR removes a column from the AGOL database view and the ETL that transfers its value to the AGOL dataset since it is no longer used in the ATSD AGOL app.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack with or without production data
2. Explore your DB with your favorite client and see that the `component_arcgis_online_view` DB view no longer selects `funding_source_name` from the joined `project_list_view`
3. Run the down migration and see that `funding_source_name` is back

**Note**: I also removed `funding_source_name` from the GraphQL query used by the AGOL ETL. This will not break the ETL, and, after the release next Thursday, the ETL will no longer populate this field in the AGOL dataset. I will then request for this field to be removed from the AGOL dataset in https://github.com/cityofaustin/atd-data-tech/issues/17971. Doing this out of order would break the ETL since AGOL's API would complain about not having a matching field in the AGOL dataset during the data transfer.

---
#### Ship list
- [x] Code reviewed 


moped-database/views/component_arcgis_online_view.sql
```diff
@@ -191,7 +191,6 @@ SELECT
     plv.interim_project_id,
     plv.project_partners,
     plv.task_order_names,
-    plv.funding_source_name,
```
roseeichelmann, line 194: still loving how helpful it is to get these diffs on views

roseeichelmann (approved):
test steps worked, this looks good to go :ship:

moped-database/views/component_arcgis_online_view.sql
johnclary: sameee

johnclary (approved):
it all checks out—nice work! 🚢

tillyw (approved):
tested up and down, works as expected!

mddilley:
fyi i just updated this migration to be last before merging and tested the up and down migrations again locally. The bot updated our pointers to the correct migrations for the latest changes to the `project_list_view`, `component_arcgis_online_view`, and `exploded_component_arcgis_online_view` as well. Good bot. 🤖

chiaberry (approved):
Thanks for the extra context on the removal of the field and how it works with the agol dataset!

🚢

PR #1459, opened 2024-10-21
Patch 2.21.1

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19538

This PR updates the Lexical `OnEditPlugin` to handle status updates created through the input in the `ProjectPhaseForm` which is not saved as rich text. **This patch will save non-rich text status updates as rich text even though they were not originally saved that way** which I think is fine since we are already primarily creating notes this way. I'll create an issue to see what adding the Lexical editor to this form would look like though wrapping it in Reach Hook Form could get bumpy.

## Testing
**URL to test:** 

https://deploy-preview-1459--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**
1. Go to a project timeline in staging ([like this one](https://moped.austinmobility.io/moped/projects/2045?tab=timeline)) and create or edit a phase and add a status update before saving. This will create a new status update that you can see in the **Notes** tab.
Screenshot 2024-10-21 at 12 31 34 PM

2. Go to the **Notes** tab and edit the status update that you just created. Notice that the editor opens up but does not show text.
3. Now, do the same in this deploy preview. Notice that the content of the status update does display as expected when opening in the **Notes** tab.

---
#### Ship list
- [x] Code reviewed 
   - Updated existing test to check if Phase status updates can be edited in the **Notes** tab

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectNotes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectNotes.js
@@ -354,8 +354,8 @@ const ProjectNotes = (props) => {
                     /**
                      * Only allow the user who wrote the status to edit it
                      */
-                    const editableNote = true;
-                    // userSessionData.user_id === item.created_by_user_id;
+                    const editableNote =
+                      userSessionData.user_id === item.created_by_user_id;
                     return (
                       <React.Fragment key={item.project_note_id}>
                         <ListItem alignItems="flex-start">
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectNotes.js
mddilley, line 1: these changes came from Prettier

johnclary (approved):
👍 i tested this through the deploy preview and the code change makes sense to me. this looks good to go 🚢

chiaberry (approved):
im glad you were able to find a fix so quickly 

🚤

mddilley:
Thank y'all!!! I'm going to go ahead and merge this and then move these changes into `main` too.

PR #1467, opened 2024-10-24
updates phases table to use mui confirmation modal

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/19436

## Testing
**URL to test:** 

https://deploy-preview-1467--atd-moped-main.netlify.app/

**Steps to test:**
- go to a project's timeline tab and click the delete icon next to a phase
- confirm that an mui confirmation modal appears asking 'Are you sure you want to delete this phase?'
- click 'delete' and confirm the correct phase was deleted

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectPhases.js
+++ b/moped-editor/src/views/projects/projectView/ProjectPhases.js
@@ -248,7 +248,7 @@ const ProjectPhases = ({ projectId, data, refetch }) => {
         refetch={refetch}
       />
       <DeleteConfirmationModal
-        type={"project phase"}
+        type={"phase"}
         submitDelete={handleDeleteClick(deleteConfirmationId)}
         isDeleteConfirmationOpen={isDeleteConfirmationOpen}
         setIsDeleteConfirmationOpen={setIsDeleteConfirmationOpen}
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Works great - thanks for updating so these all work the same now! 🙌 🚢

roseeichelmann (approved):
test steps check out :rocket:  i wonder if theres a way to fix the bug where the table resizes for a second on delete before going back? but not a big deal and this is good to go!

chiaberry (approved):
Nice to keep knocking out these UX improvements 

Naming things is hard, I agree with Mateo that we could do a better job naming these (I'm afraid to check and see that it was me who named these functions). but this does follow the pattern we have everywhere else, so if we do rename this it should be across all our confirmation modals

tillyw:
> Nice to keep knocking out these UX improvements
> 
> Naming things is hard, I agree with Mateo that we could do a better job naming these (I'm afraid to check and see that it was me who named these functions). but this does follow the pattern we have everywhere else, so if we do rename this it should be across all our confirmation modals

thanks for pointing that out, @chiaberry! you're right, i was following the existing patterns and i totally agree about keeping it consistent. so i'm not sure if i should go ahead and rename these functions in other components in this pr or stick with the naming convention we have now and then perhaps we can come up with guidelines for a new naming convention and follow up with that as tech debt. @mddilley do you have any leanings?

mddilley:
@tillyw Yeah, renaming them all sounds good to me to keep the code consistent. If you don't mind updating in this PR, then go for it! I like `handleDeleteModalOpen` that Mateo suggested above. I can also spin off an issue if you want to move on from this PR.

PR #1468, opened 2024-10-25
Populate "geometry_type" in AGOL dataset

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19301

**This should not be released until this field is added in https://github.com/cityofaustin/atd-data-tech/issues/17971** ✅

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack and the inspect the `component_arcgis_online_view` database view
2. You should see that a new column called `geometry_type` is added and that it contains "Point" or "Line" as its value. You can check the geometry in the `geometry` to make sure it matches up - Multipoint -> "Point" and MultiLineString -> "Line". 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/views/exploded_component_arcgis_online_view.sql
+++ b/moped-database/views/exploded_component_arcgis_online_view.sql
@@ -1,4 +1,4 @@
--- Most recent migration: moped-database/migrations/1729197757694_remove_funding_source/up.sql
+-- Most recent migration: moped-database/migrations/1729197757695_add_agol_geometry_type/up.sql
 
 CREATE OR REPLACE VIEW exploded_component_arcgis_online_view AS SELECT
     component_arcgis_online_view.project_id,
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -29,6 +29,7 @@
           - feature_ids
           - funding_sources
           - geometry
+          - geometry_type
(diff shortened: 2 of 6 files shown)


moped-database/views/exploded_component_arcgis_online_view.sql
```diff
@@ -1,4 +1,4 @@
--- Most recent migration: moped-database/migrations/1729197757694_remove_funding_source/up.sql
+-- Most recent migration: moped-database/migrations/1729197757695_add_agol_geometry_type/up.sql
```
roseeichelmann, line 1: why does this needs to be dropped and recreated if nothing else was changed?

roseeichelmann (approved):
looks great :100:

moped-database/views/exploded_component_arcgis_online_view.sql
mddilley: great question and it is because this view selects from `component_arcgis_online_view` so it depends on it.

the full chain is `project_list_view` -> `component_arcgis_online_view` -> `exploded_component_arcgis_online_view` so these diffs can definitely get big when editing some of these views!!

chiaberry (approved):
`select geometry_type, geometry from component_arcgis_online_view` checks out!!

PR #1479, opened 2024-11-08
Show component work types in component list items

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19560

## Testing
**URL to test:** 
https://deploy-preview-1479--atd-moped-main.netlify.app/moped/projects/2049?tab=map&project_component_id=2686

**Steps to test:**

Expand components and see the work type in a gray chip

Screenshot 2024-11-11 at 8 45 30 AM

---
#### Ship list
- [x] Code reviewed 
   - Added to component focus test - "You should see chips with the component works types in the expanded list item."


moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -33,6 +34,19 @@ const useStyles = makeStyles((theme) => ({
...
+    "& .MuiChip-label": {
+      display: "block",
+      whiteSpace: "normal",
+    },
+    fontWeight: "500",
+    fontSize: "12px",
+    borderRadius: "2rem",
+    height: "1.75rem",
+  },
+  workTypeGroup: {
```
mddilley, line 47: ~I think this could use some marginTop (0.5?) to keep the hover state background from showing right against the top of the chips. Could you try that out?~ I think that this hover state background running into the chips can also be helped with wrapping the `Grid` in `ListItem` since it applies padding  by default. My other comment is related.

Screenshot 2024-11-11 at 3 18 26 PM

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -107,6 +121,16 @@ export default function ComponentListItem({
                 />
               </ListItem>
             )}
+            <Grid container spacing={0.5} className={classes.workTypeGroup}>
```
mddilley, line 124: I was testing related component list items, and I see that the work type chips run into the next list item if there is no description and also because there is no row of action buttons below related components. I think that wrapping the grid of chips with `ListItem` will help space this out. It might also remove the need for a classname for this grid.

Screenshot 2024-11-11 at 3 34 23 PM

mddilley (changes requested):
I requested a couple tiny spacing adjustments, and then this is 🚢. Looks great, and it is good to see these component details in the expanded item!

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
```diff
@@ -107,6 +121,18 @@ export default function ComponentListItem({
                 />
               </ListItem>
             )}
+            <ListItem className={classes.workTypeGroup}>
```
mddilley, line 124: Small detail, but this class is now a duplicate of the existing `nested` one.

moped-editor/src/views/projects/projectView/ProjectComponents/ComponentListItem.js
mddilley: New commit looks great!

mddilley (approved):
Nice!!! Good to see these work types in the list items. 🚀 🙌

johnclary (commented):
I tested on the deploy preview, and it looks great. I love chips!

One minor detail is that clicking related-project components now causes them to expand ever so slightly:

Screenshot 2024-11-12 at 1 32 33 PM

See [john's ironweed project (edited)](https://deploy-preview-1479--atd-moped-main.netlify.app/moped/projects/1935?tab=map) for example.

chiaberry:
@johnclary ah because those are components that were either migrated or added before we had the requirement to include a work type

@mddilley should I conditionally show the ListItem if there is a component work type? Since its required, i assumed it would always show up, but theres probably a number of components with no type

mddilley:
@chiaberry Yes, I think that we should conditionally render like we do for the description below it. Thanks for catching that @johnclary! 🙏

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

tillyw (approved):
the code looks clean and the chips look great! thanks for adding this :)

PR #1481, opened 2024-11-14
add an action for the release drafter

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/19588

## Testing

https://github.com/marketplace/actions/release-drafter

I think the only way to test is see that its working as PRs are merged to main.

There are many ways to [configure the drafter](https://github.com/marketplace/actions/release-drafter#configuration-options), some to note:
- specify which paths to include in release notes. I dont think this applies to moped, but it could be useful in the future
- if you feel like reworking how the changes are formatted, you can `change-template: '- $TITLE @$AUTHOR (#$NUMBER)'`

I'm torn about adding the ability to categorize changes, I dont think its that much effort to sort them on release day and it would mean needing labels on PRs. 

---
#### Ship list
- [ ] Code reviewed 


README.md
```diff
@@ -14,7 +14,6 @@
 
 - Production: [https://mobility.austin.gov/moped/](https://mobility.austin.gov/moped/)
 - Staging: [https://moped.austinmobility.io/moped/](https://moped.austinmobility.io/moped/)
-- Test: [https://moped-test.austinmobility.io/moped/](https://moped-test.austinmobility.io/moped/)
```
mddilley, line 17: Thanks for cleaning up this dead link! 💐 🧹✨

mddilley (approved):
The action makes sense to me! Thanks for checking out the config for how to format the output more. I'm definitely down to explore more if we get to a point where release notes are a lot of work. And, I agree that the best way to test is merge it. 🙌 🚢 🚀

PR #1486, opened 2024-11-18
Update project fund sources to reflect reorganization

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/19853

This PR updates the funding source lookup table to reflect the merge of ATD and PWD into TPW.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack and replicate from the production read replica. Then, use the two test queries below to see that there are now no project funding records that use the soft-deleted `Austin Transportation` and `Public Works` rows. There should be around 66 rows associated with the new TPW funding source.
2. Run the same queries on the production read replica to see that the opposite is true.
3. Check a project like ids 399, 3491, or 553 to see that the project funding records in the Funding table for these projects now show `Austin Transportation and Public Works` as the funding source.
4. Try to add a new funding row in the table, and see that only `Austin Transportation and Public Works` is shown as a selectable source and not `Austin Transportation` and `Public Works`.
5. Check the activity log, and you should see an activity updating the funding source that was done by `Data and Tech Admin` 🤖 You can check the history in the activity log too and see that the original creates and updates reflect the pre-merger names.
6. You can also check the funding columns in the `project_list_view` and `component_arcgis_online_view` database views to see that these changes work downstream as well.

**Check for project funding rows with either ATD or PWD as the funding source**
```sql
WITH funding_source_todos AS (
    SELECT funding_source_id AS ids
    FROM
        moped_fund_sources
    WHERE
        funding_source_name IN (
            'Austin Transportation',
            'Public Works'
        )
)
SELECT * FROM moped_proj_funding WHERE funding_source_id IN(SELECT ids FROM funding_source_todos);
```

**Check for project funding rows with either TPW as the funding source**
```sql
WITH funding_source_todos AS (
    SELECT funding_source_id AS ids
    FROM
        moped_fund_sources
    WHERE
        funding_source_name = 'Austin Transportation and Public Works'
)
SELECT * FROM moped_proj_funding WHERE funding_source_id IN(SELECT ids FROM funding_source_todos);
```

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
+++ b/moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
@@ -3,3 +3,31 @@ INSERT INTO "public"."moped_fund_sources" ("funding_source_name") VALUES
 ('Austin Transportation and Public Works');
 
 UPDATE moped_fund_sources SET is_deleted = true WHERE funding_source_name IN ('Austin Transportation', 'Public Works');
+
+-- Find existing project funding records that are associated with the funding sources that are merging
+-- and update them to the new funding source called Austin Transportation and Public Works
+WITH funding_source_todos AS (
+    SELECT funding_source_id AS ids
+    FROM
+        moped_fund_sources
+    WHERE
+        funding_source_name IN (
+            'Austin Transportation',
+            'Public Works'
+        )
+),
+
+new_funding_source_row AS (
+    SELECT funding_source_id AS id
+    FROM
+        moped_fund_sources
+    WHERE
+        funding_source_name = 'Austin Transportation and Public Works'
+)
+
+UPDATE
+moped_proj_funding
+SET
+    funding_source_id = (SELECT id FROM new_funding_source_row)
+WHERE
+    funding_source_id IN (SELECT ids FROM funding_source_todos);
--- a/moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
+++ b/moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
@@ -1,0 +1,5 @@
+-- Insert new funding source and soft-delete existing funding sources that are merging into the new one
+INSERT INTO "public"."moped_fund_sources" ("funding_source_name") VALUES
+('Austin Transportation and Public Works');
+
+UPDATE moped_fund_sources SET is_deleted = true WHERE funding_source_name IN ('Austin Transportation', 'Public Works');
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
```diff
@@ -0,0 +1,33 @@
+-- Insert new funding source and soft-delete existing funding sources that are merging into the new one
+INSERT INTO "public"."moped_fund_sources" ("funding_source_name") VALUES
+('Austin Transportation and Public Works');
+
+UPDATE moped_fund_sources SET is_deleted = true WHERE funding_source_name IN ('Austin Transportation', 'Public Works');
+
+-- Find existing project funding records that are associated with the funding sources that are merging
+-- and update them to the new funding source called Austin Transportation and Public Works
+WITH funding_source_todos AS (
```
mddilley, line 9: I'm not entirely convinced these CTEs make this more readable since there are still sub-queries in the UPDATE statement below. Open to updating if anyone feels strongly about it!

johnclary: i like them! i wish i could bookmark this migration for the next time we need to do this kind of operation.

i suppose i can bookmark it... 🤔

johnclary (approved):
Test steps check out, and everything looks good in the UI.

This is great, but fwiw, if you wanted to cut down on the noise in the activity log, I think it would be ok to rename the **Austin Transportation** funding source instead of creating a new one. But I appreciate that the way you have it preserves the history of the org change.

In any case, I'm glad to see you addressing this naming tech debt!  🚢 🚢 🚢 🚢 🚢

moped-database/migrations/1730931541884_reorg_funding_sources/up.sql
mddilley: cool, and, yes, this builds off some of the fancy SQL you've written in the past that I bookmarked in my browser. Maybe a SQL recipes folder with examples like this would be handy! 🧑‍🍳 I often find myself searching through old migrations for some copypasta.

chiaberry (approved):
Thanks for tackling this change, and i agree i would love a sql recipe aka where i can look for the good patterns

tillyw (approved):
thank you for making this update! i like this elegant sql query and agree this type of migration could be useful in the future!

PR #1502, opened 2024-12-03
Add component work types to project_list_view DB view

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20016

The PR adds a new column to the `project_list_view` that creates a de-duped list of all work types added to all components in a project.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start up your local stack and then create a new non-signal project
2. Filter the `project_list_view` database view to the project id that you just created
3. You will see that the value of the `component_work_type_names` is `NULL` since there are no components in the project yet
4. Go to the project map and create a component that has a work type of "New"
5. You should see that the value of the view column updates to `New`
6. Now, create another component with a work type of "New"
7. You should see that the value of the view column is still only `New` since we are looking for distinct values only
8. Now, try creating another component OR adding more work types to the existing components
9. You should see that the value of this column contains all of the work types across all components added with no duplicate values
10. Last, delete one of your components and you should see the associated work types removed from the view column
11. Test the down migration and you will see the new column removed from the view

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1733184698645_add_comp_work_types/down.sql
+++ b/moped-database/migrations/1733184698645_add_comp_work_types/down.sql
@@ -1,0 +1,542 @@
+DROP VIEW IF EXISTS exploded_component_arcgis_online_view;
+DROP VIEW IF EXISTS component_arcgis_online_view;
+DROP VIEW IF EXISTS project_list_view;
+
+CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
+    SELECT
+        mpp.project_id,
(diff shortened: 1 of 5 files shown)


johnclary (approved):
Looks great! All of my component work types are showing up in the list view—in alphabetical order! I don't see a noticeable hit to the query performance, either 😎 

✅ 🚢

chiaberry (approved):
tested up and down migrations. once again im glad we have the views copied in different files so we can see the diffs more easily

mddilley:
thanks y'all! h/t to @frankhereford for the alpha order that i copied from the funding sources column he added earlier in the year ✨

PR #1505, opened 2024-12-10
Disable save button if required fields are not populated

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20084

This actually ended up being kinda weird to implement, i had to really dig thru github and stack overflow before i found out how to get the edited values while still in edit mode and cause a re-render to enable the save button. Finally I found this github issue where someone had the same problem and a rly helpful [comment](https://github.com/mui/mui-x/issues/11218#issuecomment-1832762046) describing how to handle our use case with [useGridSelector](https://mui.com/x/react-data-grid/state/#access-the-state)

## Testing
**URL to test:** 

https://deploy-preview-1505--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. On the project summary page, test adding a new subproject. The save button should be disabled until you select a project. Make sure the cancel and delete buttons work as well.
2. Test all the other datagrid tables too, Milestones, Funding, and Files. If you dont have the required fields filled out then the save should be disabled. BTW Funding has no required fields.
3. Test editing and deleting required fields, the save button should get disabled again.

---
#### Ship list
- [x] Code reviewed 
   - Added `Try to save a row without the Milestone column populated. You should see the Required helper text and input highlighted in red. The save button should also be disabled.` as a Milestones table test.


moped-editor/src/components/DataGridPro/DataGridActions.js
roseeichelmann, line 1: created a reuseable component to be used in all our data grid tables for the action buttons

moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -212,43 +205,17 @@ const useColumns = ({
...
-              label="Delete"
-              onClick={() => handleDeleteOpen(id)}
-              color="inherit"
-            />,
-          ];
-        },
+        renderCell: ({ id }) => (
+          <DataGridActions
+            id={id}
+            requiredFields={["file_name", "file_url", "file_type"]}
```
roseeichelmann, line 211: do yall like this or do you want to define requiredFields elsewhere in a var somewhere?

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -227,43 +220,17 @@ const useColumns = ({
...
-              label="Delete"
-              onClick={() => handleDeleteOpen(id)}
-              color="inherit"
-            />,
-          ];
-        },
+        renderCell: ({ id }) => (
+          <DataGridActions
+            id={id}
+            requiredFields={[]}
```
mddilley, line 226: Could you update to either handle `requiredFields` having a null or undefined value? Or maybe a default value for the prop in `DataGridActions` could work. idk i could be convinced that an empty array equals "no required fields" too 🤔

moped-editor/src/components/DataGridPro/DataGridActions.js
mddilley, line 1: This reusable component is so nice! ✨ Really cool to this common part of these tables extracted. 🧱

Also, thanks for matching this path up with the spot that Mateo began stashing DataGrid-related components. I know there are a lot of moving parts right now. 🙏

moped-editor/src/views/projects/projectView/ProjectFiles.js
mddilley: this feels good to me. If I were to start a new table that needed an inline form, I could imagine this prop making me stop to consider what fields need are required as soon as i added this component.

I briefly thought about a way to tie to the column config but that feels like over-optimization at this point imo. None of these inline forms have a long list of required fields since we've tried to minimize side-scrolling.

moped-editor/src/views/projects/projectView/ProjectMilestones.js
mddilley, line 1: @roseeichelmann totally unrelated to this PR, but could you sneak in a small fix to the "Required" text in the `MilestoneAutocompleteComponent` component? 

The red error highlighting isn't working because the helper text is rendered outside the `TextField` component (`<FormHelperText>Required</FormHelperText>`).
Removing that and adding the `helperText="Required"` prop to the `TextField` would fix this. Thank you!!!!

moped-editor/src/components/DataGridPro/DataGridActions.js
```diff
@@ -0,0 +1,96 @@
...
+        label="Edit"
+        className="textPrimary"
+        onClick={handleEditClick(id)}
+        color="inherit"
+      />
+    ),
+    <GridActionsCellItem
+      icon={<DeleteOutlineIcon sx={defaultEditColumnIconStyle} />}
+      label="Delete"
+      onClick={() => handleDeleteOpen(id)}
```
mddilley, line 90: This is not related to code introduced in this PR, but seeing this anonymous function for this delete handler right next to function calls of `handleCancelClick(id)` and `handleEditClick(id)` makes me feel like we should update the definition of the delete handlers in the table components to match up with the [MUI code example](https://mui.com/x/react-data-grid/editing/#full-featured-crud) syntax. [This thread](https://austininnovation.slack.com/archives/CHZE6BC6L/p1730494043545059) is also related for context. like:

```js
const handleDeleteOpen = (id) => () => {
    // do something with id
};
```

Would you up for making this change in this PR if it feels small to you? I'm happy to make a snackoo too since its performance impact isn't immediate.

mddilley (approved):
This looks awesome! I tested each table, and I don't see any regressions. I also see that saving on click-away is blocked when a required field is missing. 👍 

Thank you for the level up on introspecting the grid state too. I added two questions, and requested one change outside the scope of this issue. his is ready to 🚢!

frankhereford (approved):
This is great - works perfect. Code looks good, but some of it is def above my pay grade. 🚢🚢🚢

--- 

I have one request -- and this is embarrassing -- I'm not sure how to nail down exactly where in the app this change needs to be made, but I figure y'all can real easy. Also - I'm pretty sure this behavior is not a result of this PR, so maybe this is a spin off thing, if y'all agree.

When I click on the cell, it gets this "focus" like behavior, and the enclosing cell gets this one pixel blue stroke around it that is out of place, imho. Like this:

image

This is the element, and it's this element that I'm struggling to point to in the application. 

image

If you give that element a `outline: none;` CSS property, it's fixed. I'm curious what you and @mddilley think. Thanks!!

mddilley:
@frankhereford thanks for the feedback! This is default styling and a feature of DataGridPro as seen in the [CRUD example](https://mui.com/x/react-data-grid/editing/#full-featured-crud) so I'd like to stick with the focus highlighting unless users have negative feedback about it.

frankhereford:
@mddilley wrote:

> This is default styling and a feature of DataGridPro

Sounds great to me. A last thought and forgive me, this is at the risk of feeling like i'm overly pushing back, but I don't mind the example you linked at all. 

I think it's the fact that we have a stroke around the cell, and a stroke around what's in the cell, and we have no padding separating the two. It becomes unusual looking to me because the hierarchy of what contains what is unclear. I guess my take is that like borders shouldn't abut onto other borders of the same style.

In the screenshot above, I see a figure-8 of thin stroke, and is that a pair of cells next to each other or one containing the other? 

Again - i'm super cool with sticking with it -- i just wanted to better explain my observation. Thank you for the super quick response!!

moped-editor/src/views/projects/projectView/ProjectMilestones.js
johnclary: Was just coming here to mention that! 🙏 

milestone

moped-editor/src/views/projects/projectView/ProjectFiles.js
johnclary: obligatory comment about non-stable object references: i think this is slightest of anti-patterns versus declaring  a `const` outside of the component 😇

moped-editor/src/components/DataGridPro/DataGridActions.js
```diff
@@ -0,0 +1,96 @@
...
+   */
+  const hasRequiredFields = useGridSelector(apiRef, () => {
+    const editState = apiRef.current.state.editRows;
+    for (const field of requiredFields) {
+      if (!editState[id]?.[field]?.value) {
+        return false;
+      }
+    }
+    return true;
+  });
```
johnclary, line 54: very cool to see how this works and nice docstring 🙏

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
johnclary: I agree—I think allowing the prop to be undefined and then conditionally using an empty array as part of the component's internals is a classy way to go 👍

johnclary (approved):
Good stuff—I couldn't get anything done without filling in those required fields. Very nice UX upgrade here! 🚢 🚢 🚢 🚢

mddilley:
@frankhereford thanks for expanding about the focus outline colliding with the input, and now I'm seeing it! 🔍 🙏

We have inconsistent spacing from table to table so I captured making it uniform in https://github.com/cityofaustin/atd-data-tech/issues/20171 and https://github.com/cityofaustin/atd-data-tech/issues/20213 for when we consolidate and make reusable components to use across the tables.

moped-editor/src/views/projects/projectView/ProjectFiles.js
mddilley: ahh yeah that is a great point, John! could you update @roseeichelmann?

moped-editor/src/views/projects/projectView/ProjectMilestones.js
roseeichelmann: fixed :)

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
roseeichelmann: fixed prop can be undefined and then theres a default val of an empty array for it in `DataGridActions`

roseeichelmann:
@mddilley this is ready for re-review :))

moped-editor/src/components/DataGridPro/DataGridActions.js
mddilley: thanks for squeezing this in! 🙌

mddilley (approved):
New commits look great, and I retested with the latest changes. Thanks for updating!! 🚀🚀🚀

PR #1511, opened 2024-12-17
Add component work types summary field to project summary view

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20019

This PR updates the project summary to display the list of component work types. There are also some Prettier changes that came along as I was looking at our components that render the summary items.

## Testing
**URL to test:** 

Local because of the Hasura metadata update

**Steps to test:**
1. Start your local stack and create a new, non-signal project
2. Go to the summary view and see the new item called `Component work types`
3. Since the new project has no components, the UI should show `-`
4. Now, add a component with a few works types to the project
5. Back in the summary view, you should see each component work types in a list with each type on a new line. The list should not highlight when hovered since it is not editable in the summary view.
6. You can try adding more components and work types to see the list update

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryComponentWorkTypes.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryComponentWorkTypes.js
@@ -10,6 +10,11 @@ import ProjectSummaryLabel from "src/views/projects/projectView/ProjectSummary/P
  * @constructor
  */
 const ProjectSummaryComponentWorkTypes = ({ classes, data }) => {
+  const componentWorkTypes =
+    data?.project_list_view?.[0]?.component_work_type_names;
+  /* component_work_type_names is a comma-separated list in the database view */
+  const componentWorkTypesArray = componentWorkTypes?.split(", ") ?? [];
+
   return (
     <Grid item xs={12} className={classes.fieldGridItem}>
       <Typography className={classes.fieldLabel}>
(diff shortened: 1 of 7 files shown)


moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryComponentWorkTypes.js
```diff
@@ -0,0 +1,37 @@
...
+      <Typography className={classes.fieldLabel}>
+        Component work types
+      </Typography>
+      <Box
+        display="flex"
+        justifyContent="flex-start"
+        className={classes.fieldBox}
+      >
+        <ProjectSummaryLabel
+          text={componentWorkTypesArray}
```
mddilley, line 29: i was really tempted to refactor this component since this `text` prop can accept either strings or arrays of strings, but it was going to require a large diff. Rather focus on getting this shipped for now.

frankhereford (approved):
🚢 🚢 🚢  This worked just like you described and your code is fantastic, as always.

I have the most tiny suggestion, and it couldn't be less of a thing.

When you're picking a work type out when you're adding a component, you get the choices in an intuitive (at least to me) order:

image

Do you think it'd be a good add to use that same order when displaying the work types on the summary page? Something about "new" to "mods" to "repair" feels like the lifecycle of a thing to me.

image

tillyw (approved):
tested locally and looks and works great! thanks for adding this!!

mddilley:
@frankhereford thanks for those screenshots. 📷 🙏  I followed up on the code in those two places, and we display the work types in the component form with no order specified, and they are in alphabetical order in the DB view. I'm curious to see if users notice or prefer either, and we can make a change at that time.

frankhereford:
> I'm curious to see if users notice or prefer either, and we can make a change at that time.

No one is going to think to say anything.

PR #1512, opened 2024-12-17
Add component work types to project list view and make exportable/searchable

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20020

## Testing
**URL to test:** 

you will need to test locally because of metadata until Mike's PR gets merged
use prod data if you can

**Steps to test:**

1. from the projects page, click on the columns menu in the upper left hand corner of the table and toggle on the  "Component work types" column
2. Try sorting on that column
3. With the column visible, download records. confirm the column is in your csv
4. now try to advanced search on work types. contains, is, is blank etc

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -116,6 +116,6 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
     label: "Council districts",
   },
   component_work_type_names: {
-    label: "Work types",
+    label: "Component work types",
   },
 };
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -115,4 +115,7 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
   project_and_child_project_council_districts: {
     label: "Council districts",
   },
+  component_work_type_names: {
+    label: "Work Types",
+  },
 };
--- a/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
+++ b/moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
@@ -116,6 +116,6 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
     label: "Council districts",
   },
   component_work_type_names: {
-    label: "Work Types",
+    label: "Work types",
   },
 };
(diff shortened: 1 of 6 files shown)


frankhereford (approved):
Tested locally with production data -- works like a dream. Your code is 💯 -- this is ready to ship!

Thank you @chiaberry!

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
```diff
@@ -115,4 +115,7 @@ export const PROJECT_LIST_VIEW_EXPORT_CONFIG = {
   project_and_child_project_council_districts: {
     label: "Council districts",
   },
+  component_work_type_names: {
+    label: "Work types",
```
mddilley, line 119: @chiaberry Could you update the labels in this PR to "Component work types"? That is what I went with in the summary view because I almost added a tooltip but then I thought adding the "component" part to the name would be enough to point to where these values are coming from.

Open to other ideas here too!

mddilley (commented):
Awesome! I tested this ever which way, and I can search and export perfectly. I requested one change to make our naming of this field consistent in the UI, and I'm totally open to more thoughts on it. Thank you!

moped-editor/src/views/projects/projectsListView/ProjectsListViewExportConf.js
chiaberry: absolutely!

mddilley (approved):
Thanks for updating! Works great! 🚢 🚀🚢

tillyw:
tested locally and this looks and works great! there were a couple filter operators that gave me somewhat unexpected results, specifically the 'is', 'is not', and 'ends with' (and i think 'begins with' as well).

Screenshot 2024-12-18 at 10 24 45 PM
Screenshot 2024-12-18 at 10 25 19 PM
Screenshot 2024-12-18 at 10 26 19 PM

since these operators are better suited to strings than lists (which i think is often what people will be searching on here) and i think the 'contains' and 'does not contain' does what these operators would be doing, how would y'all feel about removing the four i listed above? maybe there are some use cases i'm not thinking of, so if i'm overlooking something please disregard this!

mddilley:
@tillyw Thanks for catching that! 🙏 I ran with the existing operators for `Type` in the requirements, but you are totally right that some of these don't make sense for lists of strings. I coordinated with Chia on changes for this PR, and I made this spin-off to check back in on other lists. https://github.com/cityofaustin/atd-data-tech/issues/20351

mddilley:
@chiaberry Just retested and it looks 🚢-able to me!

PR #1513, opened 2024-12-18
Add project phase status badges to the Data dictionary

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20325

## Testing
**URL to test:** 

https://deploy-preview-1513--atd-moped-main.netlify.app/moped/dev/lookups#moped-phases

**Steps to test:**

There are now badges on the table instead of just the phase name. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/dev/LookupsView/RecordTable.js
+++ b/moped-editor/src/views/dev/LookupsView/RecordTable.js
@@ -14,7 +14,7 @@ import CircularProgress from "@mui/material/CircularProgress";
  * @returns { JSX } an MUI TableCell component with the row/column value
  */
 const WrappedTableCell = ({ row, column: { key, handler } }) => {
-  return <TableCell>{handler ? handler(row[key]) : row[key]}</TableCell>;
+  return <TableCell>{handler ? handler(row[key], row) : row[key]}</TableCell>;
 };
 
 /**
--- a/moped-editor/src/views/dev/LookupsView/settings.js
+++ b/moped-editor/src/views/dev/LookupsView/settings.js
@@ -42,12 +42,19 @@ const workTypeHandler = (workTypes) =>
     <div key={workType.moped_work_type.id}>{workType.moped_work_type.name}</div>
   ));
 
-const statusBadgeHandler = (phaseName) => {
-  const phaseKey = phaseName.toLowerCase().replace(/[\s/-]/g, "_");
-  return (
-    <ProjectStatusBadge phaseKey={phaseKey} phaseName={phaseName} condensed />
-  );
-};
+/**
+ * Uses phase name and phase key from row object to render status badge
+ * @param {string} phaseName - phase name
+ * @param {Object} row a single Moped record object as returned from Hasura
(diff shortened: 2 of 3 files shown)


moped-editor/src/views/dev/LookupsView/RecordTable.js
```diff
@@ -14,7 +14,7 @@ import CircularProgress from "@mui/material/CircularProgress";
  * @returns { JSX } an MUI TableCell component with the row/column value
  */
 const WrappedTableCell = ({ row, column: { key, handler } }) => {
-  return <TableCell>{handler ? handler(row[key]) : row[key]}</TableCell>;
+  return <TableCell>{handler ? handler(row[key], row) : row[key]}</TableCell>;
```
chiaberry, line 17: I added the row object into the handler call, but only one of our handler functions uses `row`, the one I just wrote. 

I can't decide if i should just leave the rest of the handler function definitions as is, or if i should update all of them to make explicit that they are getting an unneeded row object.

moped-editor/src/views/dev/LookupsView/settings.js
```diff
@@ -40,6 +42,20 @@ const workTypeHandler = (workTypes) =>
     <div key={workType.moped_work_type.id}>{workType.moped_work_type.name}</div>
   ));
```
chiaberry, line 43: for example, should this be defined instead (expand to see line 37)
`const workTypeHandler = (workTypes, row) => `

moped-editor/src/views/dev/LookupsView/RecordTable.js
mddilley: hmmm, i'm going to read about this some. I think part of why this feels off is that some of the `handler` functions are returning JSX so they are function components in themselves. 🤔

frankhereford (approved):
I took a look at the deploy preview's data dictionary, and seeing the badges on the table drives home that this was a really good idea -- that is quite helpful.

Your code looks great to me, and I don't really have any idea about what the best route is re your question about the handler. I'll be real interested to see what you decide. 

All that said -- this is ready to be anchors-up, shipping-out.

moped-editor/src/views/dev/LookupsView/RecordTable.js
johnclary: this looks great to me! nbd to omit the `row` prop from the other handlers that do not access it. i think that is preferable to adding, for example, the `row` arg to the `workTypeHandler` function signature 🚢

mddilley: yep, i was getting side-tracked with code beyond this diff. thanks for reining this back in, John! 🙏

mddilley (approved):
Great to see these badges! Thanks! 🚢 😎

tillyw (approved):
tested and the badges look great! i'm glad y'all came to a decision about `row` prop, i think that makes sense. thank you!

johnclary:
This such a nice lil upgrade—great work!

PR #1516, opened 2024-12-26
Add validation of required fields to Team table and combine DataGridTextField components

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20357
Closes https://github.com/cityofaustin/atd-data-tech/issues/20213

This PR updates the Team table to validate required fields, and it combines duplicate `DataGridTextField` components. It also extends the existing validation to use error states that we can get from [preProcessEditCellProps](https://mui.com/x/react-data-grid/editing/?srsltid=AfmBOorAjy4Egs9sAYiz-XwgY_B6xVw3agN4JORVCgeFvxCLA5VKK3q5#validation) in addition to the existing logic that checks for values in the form. So, we can now validate with custom logic on a field-specific level in addition to checking whether there is a value or not. This was needed to handle validating the **Role** field which is an array of values.

## Testing
**URL to test:** 

Local or deploy preview

**Steps to test:**
1. Test the Team table and see if you can get it to save without having a user and at least one role chosen. You should see the check icon greyed otherwise. You will also see red highlighted inputs and helper text on required field once you start editing the form.
2. Test the Milestones table to makes sure the validation on the **Milestone** input still works. Also, test the **Description** input to make sure it still works.
3. Test the Files table to make sure the **Name**, **File** (for a link since this isn't editable for uploaded files), and **Description** inputs still works.
4. Test the Funding table to make sure the **Description** inputs still works.
5. I also made some small tweaks to the styles to keep inputs from butting up against the cell borders and to move some old styles to use the theme spacing. If anything looks off, let me know, and I'll fix or restore if I made things worse.

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/components/DataGridPro/DataGridActions.js
```diff
@@ -45,8 +45,12 @@ const DataGridActions = ({
    */
   const hasRequiredFields = useGridSelector(apiRef, () => {
     const editState = apiRef.current.state.editRows;
+
     for (const field of requiredFields) {
-      if (!editState[id]?.[field]?.value) {
+      const hasError = editState[id]?.[field]?.error;
```
mddilley, line 50: this `error` state comes from its definition in [preProcessEditCellProps](https://mui.com/x/react-data-grid/editing/?srsltid=AfmBOorAjy4Egs9sAYiz-XwgY_B6xVw3agN4JORVCgeFvxCLA5VKK3q5#validation) in the column config coming out of the `useColumns` hook.

Further down in the diff, you can see the check for the team role value having an array length greater than 0.

moped-editor/src/components/DataGridPro/DataGridActions.js
```diff
@@ -58,6 +62,7 @@ const DataGridActions = ({
       <GridActionsCellItem
         icon={<CheckIcon sx={defaultEditColumnIconStyle} />}
         label="Save"
+        key="save"
```
mddilley, line 65: DataGrid expects an array of components for columns with `actions` type in the column config. I noticed it wasn't adding keys to this abstracted component the way it would if these components were defined directly in the config which is odd. It was causing console warnings, so, I added keys based on the labels since these components never change order in the component tree.

moped-editor/src/views/projects/projectView/ProjectTeam/TeamAutocompleteComponent.js
```diff
@@ -65,10 +70,7 @@ const TeamAutocompleteComponent = ({
   };
 
   return (
-    <FormControl
-      variant="standard"
-      sx={{ width: "100%", mx: theme.spacing(1) }}
-    >
+    <FormControl variant="standard" sx={{ width: "100%", mx: 1 }}>
```
johnclary, line 73: Would it make sense to add a css class or `const` to store these styles?

johnclary (approved):
Nice upgrades here—I followed along in the code and tested locally. Everything is looking great! 🚢 🚢 🚢 🚢

mddilley:
I just pushed one last commit because a VZ PR reminded me of part of https://github.com/cityofaustin/atd-data-tech/issues/20213 that I left out of this PR. 🙃

moped-editor/src/views/projects/projectView/ProjectTeam/TeamAutocompleteComponent.js
mddilley: idk about right now. i'm hoping that https://github.com/cityofaustin/atd-data-tech/issues/20171 will result in us combining all these different autocompletes into one reusable component that contains the styles in one place.

mddilley:
@mateoclarke Thanks for catching that bug! I'm seeing the same behavior in staging. A good example is https://moped.austinmobility.io/moped/projects/2055?tab=files when I try to update the row with named "test file upload." I'll fix in this PR 🙏

mddilley:
I just pushed a change to make the `file_url` field not required. The wrinkle is that the required fields are dependent on whether a file is an uploaded file which doesn't need the `file_url` field or if it is a file link which does require the `file_url` to be populated. This compromise leaves us where the checkbox is not greyed out when `file_url` is cleared but the red validation text shows and the form cannot be saved.

We would need to make the validation more sophisticated so that it could determine which type of file record that we are editing so I decided to punt. I'll write up an issue to capture this along with pulling the `hasValue` validation up into `preProcessEditCellProps` callback.

moped-editor/src/components/DataGridPro/DataGridActions.js
roseeichelmann: +1 nice catch

moped-editor/src/components/DataGridPro/DataGridActions.js
```diff
@@ -45,8 +45,12 @@ const DataGridActions = ({
    */
   const hasRequiredFields = useGridSelector(apiRef, () => {
     const editState = apiRef.current.state.editRows;
+
     for (const field of requiredFields) {
-      if (!editState[id]?.[field]?.value) {
+      const hasError = Boolean(editState[id]?.[field]?.error);
+      const hasValue = Boolean(editState[id]?.[field]?.value);
+
+      if (hasError || !hasValue) {
```
roseeichelmann, line 53: i like how you are handling the errors too!

roseeichelmann (approved):
thanks for adding all this validation, its great! i just tested with your most recent update and i see what you mean about the save not disabling anymore when you delete the url, looking forward to seeing how the fix will work!

chiaberry (approved):
🚢 followed the test steps 👍🏼 👍🏼

PR #1517, opened 2024-12-30
Migrate subproject funding modal to DataGridPro

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18179

This PR migrates the eCapris subproject funding source table in the "FROM ECAPRIS" modal to use DataGridPro.

## Testing
**URL to test:** 

Local or [deploy preview](https://deploy-preview-1517--atd-moped-main.netlify.app/moped/)

**Steps to test:**

This table fetches rows from an open dataset by the eCapris subproject ID set on a project. Some IDs match many results while other only match a few so you can use both **11280.006** which will show 18 rows in the table and **7333.001** which will show 4 rows in the table.

You can find other eCapris subproject IDs to test using [this query on the open dataset](https://data.austintexas.gov/Transportation-and-Mobility/Transportation-Capital-Project-Financial-Codes/jega-nqf6/explore/query/SELECT%20%60sp_number_txt%60%2C%20count%28DISTINCT%20%60fdu%60%29%20AS%20%60count_distinct_fdu%60%0AWHERE%20caseless_one_of%28%60dept_unit_status%60%2C%20%22Active%22%29%0AGROUP%20BY%20%60sp_number_txt%60%0AORDER%20BY%20%60count_distinct_fdu%60%20DESC%20NULL%20LAST/page/filter) - you just have to remove the comma in the `SP_NUMBER_TXT` column.

1. Create a new project and set `11280.006` as the eCapris subproject ID in the project summary view or in the Funding tab.
2. In the Funding tab, add a new funding source using the "from eCapris" option.
3. A modal will open up to show the table, and you should be able to select one, some, or all FDU rows. Notice that the "Add Funding Source" button is only enabled if a row(s) is selected. The columns are also sortable.
4. Add one or more rows, and you should see new funding source rows populated with:
   - Status is "Confirmed"
   - Fund and Dept-unit
   - Amount is $0
5. Also notice that once a funding source is added from the modal, then you can no longer add it. Delete some funding sources from the table and notice that you can added them again from the modal.
6. Try these steps again with the subproject ID set to `7333.001` to see how the table hides pagination if there are 10 or less rows in the table.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -408,8 +408,7 @@ const ProjectFundingTable = () => {
             proj_funding_id: id,
           },
         })
-          .then(() => refetch())
-          .then(() => setIsDeleteConfirmationOpen(false))
+          .then(() => refetch().then(() => setIsDeleteConfirmationOpen(false)))
           .catch((error) => {
             setSnackbarState({
               open: true,
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -24,7 +24,6 @@ import {
   DELETE_PROJECT_FUNDING,
 } from "../../../../queries/funding";
 
-import { getDatabaseId, useUser } from "../../../../auth/user";
 import FundingDeptUnitAutocomplete from "./FundingDeptUnitAutocomplete";
 import DollarAmountIntegerField from "./DollarAmountIntegerField";
 import DataGridTextField from "../DataGridTextField";
@@ -245,12 +244,6 @@ const ProjectFundingTable = () => {
   const apiRef = useGridApiRef();
   const classes = useStyles();
 
-  /**
-   * User Hook
-   * @type {object} CognitoUserSession
-   */
-  const { user } = useUser();
-
   /** Params Hook
    * @type {integer} projectId
    * */
@@ -335,8 +328,6 @@ const ProjectFundingTable = () => {
     },
     [apiRef]
   );
-
-  const userId = getDatabaseId(user);
 
   /**
    * Wrapper around snackbar state setter
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -408,7 +408,8 @@ const ProjectFundingTable = () => {
             proj_funding_id: id,
           },
         })
-          .then(() => refetch().then(() => setIsDeleteConfirmationOpen(false)))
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -245,12 +244,6 @@ const ProjectFundingTable = () => {
   const apiRef = useGridApiRef();
   const classes = useStyles();
 
-  /**
-   * User Hook
-   * @type {object} CognitoUserSession
-   */
-  const { user } = useUser();
```
mddilley, line 252: This user data was retrieved to be passed to the SubprojectFundingModal component where it was no longer used.

moped-editor/src/views/projects/projectView/ProjectFunding/SubprojectFundingModal.js
```diff
@@ -118,29 +140,26 @@ const SubprojectFundingModal = ({
...
-            tableLayout: "fixed",
-            selection: true,
-            rowStyle: typographyStyle,
-            pageSize: 10,
-            showSelectAllCheckbox: false,
-          }}
-          onSelectionChange={(rows) => setSelectedFdus(rows)}
+          pagination
+          pageSizeOptions={[PAGE_SIZE]}
+          hideFooter={filteredData.length < PAGE_SIZE}
```
mddilley, line 159: The previous table hid the pagination if there were 10 or less rows so I kept that around, and I'm on the fence. Here is what it looks like both ways.

**With pagination hidden when there are 10 or less results:**
Screenshot 2024-12-30 at 11 01 10 AM

**With pagination always showing:**
Screenshot 2024-12-30 at 11 01 25 AM

johnclary: yeah, I also think i prefer the pagination visible all the time. in addition to the record count i like when it's disabled on-screen because it's a little hint that sometimes there may be a lot of rows in the table.

lately i have been feeling down about hidden functionality in general—i feel like it leads to a more consistent UX and maybe fewer bugs.

johnclary (approved):
Another one bites the dust—looks great! 🙌

moped-editor/src/views/projects/projectView/ProjectFunding/SubprojectFundingModal.js
roseeichelmann: i agree with john and mateo!

roseeichelmann (approved):
this looks and works great! :rocket:

moped-editor/src/views/projects/projectView/ProjectFunding/SubprojectFundingModal.js
mddilley: awesome y'all - thanks! I removed the conditional render. 🗑️ 👋

chiaberry (approved):
welcome to the datagrid pro world!

PR #1520, opened 2025-01-08
Create database view of project funding data

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20009

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack from a production snapshot
2. Browse your local DB and see that there is a new view called `project_funding_view` that includes project funding rows that are not soft-deleted from projects that are not soft-deleted, and this view includes source, program, and status lookup values joined in by their foreign key ids.
3. Verify that the view is tracked in Hasura but only available through the Hasura admin role and not the Moped admin, editor, or viewer roles. This view will be accessed through the Power BI dataflow only so it could be argued that we don't need to track at all, but I think it could be helpful to see in the Hasura console since some prefer it for browsing the DB.

---
#### Ship list
- [x] Code reviewed 


chiaberry (approved):
I tested adding and deleting funding sources to projects and deleting projects and confirmed soft deleted records don't appear on the view. 

I like that the table is added to the metadata, for the times when im using the console. 

🚢

moped-database/migrations/1736291471358_add_proj_fund_view/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    mpf.updated_at,
+    mfs.funding_source_name,
+    mfp.funding_program_name,
+    mfst.funding_status_name
+FROM
+    moped_project AS mp
+LEFT JOIN moped_proj_funding AS mpf ON mp.project_id = mpf.project_id
+LEFT JOIN moped_fund_sources AS mfs ON mpf.funding_source_id = mfs.funding_source_id
+LEFT JOIN moped_fund_programs AS mfp ON mpf.funding_program_id = mfp.funding_program_id
+LEFT JOIN moped_fund_status AS mfst ON mpf.funding_status_id = mfst.funding_status_id
```
frankhereford, line 18: So - this style of abbreviating / aliasing tables with the acronym of the name is common, and I want to acknowledge that up front. In fact, DBeaver will offer them up as autocomplete. 

That said, I believe that it's a habit left over from the 70s and 80s, when SQL was being developed and being programmed on actual TTY terminals, [like this](https://preview.redd.it/afl76c7kl5r51.jpg?width=1080&crop=smart&auto=webp&s=3149bfc6d04523cf9e04b1db28cb5cbb4c11ce0c), that were really 80 characters wide before the text wrapped, no ifs-ands-or-buts.

I find it particularly hard to read. Quick, without looking, what is the relationship between `mpf` and `mfp`? Which tables are they again? 

I find myself having to reference back and forth between the `FROM` and `JOIN` clauses to just read the `SELECT` clause, especially given our repetitious table naming pattern when expressed in these acronyms. 

Here, I might have done my alias like this:

```sql
FROM
    moped_project AS projects
LEFT JOIN moped_proj_funding AS funding ON projects.project_id = funding.project_id
LEFT JOIN moped_fund_sources AS fund_sources ON funding.funding_source_id = fund_sources.funding_source_id
LEFT JOIN moped_fund_programs AS fund_programs ON funding.funding_program_id = fund_programs.funding_program_id
LEFT JOIN moped_fund_status AS fund_statuses ON funding.funding_status_id = fund_statuses.funding_status_id
```

This is obviously a personal preference, and I am grateful for even just y'all considering it when you're writing queries.

moped-database/metadata/tables.yaml
```diff
@@ -5493,6 +5493,9 @@
           - workgroup_name
         filter: {}
         allow_aggregations: true
+- table:
+    name: project_funding_view
+    schema: public
```
frankhereford, line 5498: Yea - I dig this going into the metadata, and not because I like to use the console as a client -- y'all know it's not my fav -- but I like it rather because this is now an explicit declaration that we don't offer this data up over the graphql endpoint.

image

frankhereford (approved):
Looks great Mike - thanks for this! It's 🚢🚢🚢🚢

moped-database/migrations/1736291471358_add_proj_fund_view/up.sql
chiaberry: I cant without looking say what the relationship between `mpf` and `mfp` are, but if its a shorter chunk of sql like this one, i can find the definitions with a glance. when its much longer though (like the project list view) its easier for me to get lost. 

i dont have a huge preference one way or another, im cool with more readable

mddilley: cool, thanks! side note that this thread motivated me to change to no abbreviations [in what i'm working on right now](https://github.com/cityofaustin/atd-moped/commit/f9dd75bbe81c7699df902d10ec438eda7b641347)

johnclary: fun discussion here! 🙏

moped-database/views/project_funding_view.sql
johnclary, line 1: i know we don't conventionally add `COMMENT`s for views, because they tend to get lost over time, but it might be nice to stick one here that mentions this is a view that is useful for powerbi since this view seems rather unlikely to change 🤷

johnclary (approved):
🚢🚢🚢

moped-database/views/project_funding_view.sql
mddilley: iirc we ran into an issue with view comments being lost over time through the SQL bot 🤖 I agree that it would be a nice enhancement to find a way to retain them in the future though in here or if we do something similar in VZ!

i have aded a release step to document this view in the Moped PM docs as one available through the Power BI integration which has been working okay so far.

PR #1533, opened 2025-02-07
Add length validation to description inputs that are sent to the components AGOL feature service

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20421

## Testing
**URL to test:** 

Local or deploy preview

**Steps to test:**
TBD

```sql
SELECT
	moped_proj_components.project_id,
	LENGTH(moped_proj_components.description) AS component_description_length,
	LENGTH(moped_proj_components.location_description) AS component_location_description_length,
	LENGTH(moped_project.project_description) AS project_description_length
FROM
	moped_proj_components
	LEFT JOIN moped_project ON moped_project.project_id = moped_proj_components.project_id
WHERE
	TRUE
	AND moped_proj_components.description IS NOT NULL
	AND moped_proj_components.location_description IS NOT NULL
	AND moped_project.project_description IS NOT NULL;
```

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/constants/projects.js
+++ b/moped-editor/src/constants/projects.js
@@ -1,2 +1,10 @@
 export const substantialCompletionDateTooltipText =
   "The earliest confirmed date of any Complete or Post-construction phase";
+
+/**
+ * Number of characters that falls below max set in the project and component description fields
+ * sent to AGOL from component_arcgis_online_view database view
+ * Names: project_description, component_description, component_location_description
+ * See Field settings at https://austin.maps.arcgis.com/home/item.html?id=1c084c8756a84e6db7e2796c98c850a2#data
+ */
+export const agolDescriptionCharacterMax = 5000;
--- a/moped-editor/src/constants/projects.js
+++ b/moped-editor/src/constants/projects.js
@@ -2,15 +2,13 @@ export const substantialCompletionDateTooltipText =
   "The earliest confirmed date of any Complete or Post-construction phase";
 
 /**
- * Number of characters that falls below max set in the project and component description fields
- * sent to AGOL from component_arcgis_online_view database view
- * Names: project_description, component_description, component_location_description
+ * Number of string characters that fall below maximum set in fields sent to AGOL from
+ * component_arcgis_online_view database view.
+ * Key example: component_location_description (AGOL) -> componentLocationDescription
  * See Field settings at https://austin.maps.arcgis.com/home/item.html?id=1c084c8756a84e6db7e2796c98c850a2#data
  */
-export const agolDescriptionCharacterMax = 5000;
-
-// TODO: Should these all different limits
-// Based of max char queries in prod:
-// 2000 for project description
-// 5000 for component description
-// 500 for location description
+export const agolFieldCharMax = {
+  projectDescription: 2000,
+  componentDescription: 5000,
+  componentLocationDescription: 500,
+};
--- a/moped-editor/src/constants/projects.js
+++ b/moped-editor/src/constants/projects.js
@@ -8,3 +8,9 @@ export const substantialCompletionDateTooltipText =
  * See Field settings at https://austin.maps.arcgis.com/home/item.html?id=1c084c8756a84e6db7e2796c98c850a2#data
  */
 export const agolDescriptionCharacterMax = 5000;
+
+// TODO: Should these all different limits
+// Based of max char queries in prod:
+// 2000 for project description
+// 5000 for component description
+// 500 for location description
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -80,6 +80,7 @@ const DefineProjectForm = ({
       </Grid>
       <Grid container spacing={3} style={{ margin: 20 }}>
         <Grid item xs={6}>
+          {/* TODO: Add limit here too */}
           <TextField
             variant="standard"
             required
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { TextField, Grid, InputLabel, Switch, Box } from "@mui/material";
-import SignalAutocomplete from "./SignalAutocomplete";
+import SignalAutocomplete from "src/views/projects/newProjectView/SignalAutocomplete";
 
 const DefineProjectForm = ({
   projectDetails,
@@ -80,7 +80,6 @@ const DefineProjectForm = ({
       </Grid>
       <Grid container spacing={3} style={{ margin: 20 }}>
         <Grid item xs={6}>
-          {/* TODO: Add limit here too */}
           <TextField
             variant="standard"
             required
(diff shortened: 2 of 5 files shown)


moped-editor/src/constants/projects.js
```diff
@@ -1,2 +1,14 @@
...
+ * Number of string characters that fall below maximum set in fields sent to AGOL from
+ * component_arcgis_online_view database view.
+ * Key example: component_location_description (AGOL) -> componentLocationDescription
+ * See Field settings at https://austin.maps.arcgis.com/home/item.html?id=1c084c8756a84e6db7e2796c98c850a2#data
+ */
+export const agolFieldCharMax = {
+  projectDescription: 2000,
+  componentDescription: 5000,
+  componentLocationDescription: 500,
+};
```
mddilley, line 14: I'm looking for feedback on starting this pattern for setting length limits in the UI. My hope is that we can keep the AGOL-defined character limits in mind if we ever update these again. I'm open for this to live elsewhere too.

As for the "why?" behind the different numbers, I based them off of the longest strings in each of these columns in the database. I included a SQL query to see them in the description of this PR.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -25,109 +34,117 @@ import { useMutation } from "@apollo/client";
...
    * Saves the new project description...
    */
-  const handleProjectDescriptionSave = () => {
-    if (description.trim().length > 0) {
-      updateProjectDescription({
-        variables: {
-          projectId: projectId,
-          description: description,
-        },
-        refetchQueries: [{ query: listViewQuery }, "ProjectSummary"],
```
mddilley, line 58: The refetch of the project summary data used to be handled here, but I moved it into the save handler so the refetch happens before the edit mode switches back to the view mode.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -25,109 +34,117 @@ import { useMutation } from "@apollo/client";
...
               minRows={4}
-              error={description.length < 1}
-              id="moped-project-description"
-              label={null}
-              onChange={handleProjectDescriptionChange}
-              value={description}
-              helperText={
-                description.length < 1 ? "Description cannot be blank" : ""
-              }
+              maxRows={10}
```
mddilley, line 118: I added this maximum so that long strings don't endlessly grow in this input. You can compare this behavior to staging where you can paste super long strings and they endlessly grow the page.

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
```diff
@@ -89,8 +89,8 @@ const DefineProjectForm = ({
             type="text"
             fullWidth
             value={projectDetails.project_description}
-            error={descriptionError}
-            helperText="Required"
+            error={descriptionError !== null}
+            helperText={descriptionError?.message || "Required"}
```
roseeichelmann, line 93: i dont see any reason to not have this fallback but just to be sure, if the description is length 0 then `descriptionError.message` would already be "Required" from the logic set in `NewProjectView`?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
```diff
@@ -1,16 +1,25 @@
...
-  TextField,
-  Typography,
-  IconButton,
-} from "@mui/material";
-import ProjectSummaryLabel from "./ProjectSummaryLabel";
+import { Box, Grid, Icon, Typography, IconButton } from "@mui/material";
+import ProjectSummaryLabel from "src/views/projects/projectView/ProjectSummary/ProjectSummaryLabel";
 
-import { PROJECT_UPDATE_DESCRIPTION } from "../../../../queries/project";
+import { PROJECT_UPDATE_DESCRIPTION } from "src/queries/project";
```
roseeichelmann, line 5: lovin all the absolute paths!

roseeichelmann (approved):
i noticed that if you put too long of an input in the project description and get the error, then cancel, and then edit it again, that long input is still there along with the error instead of the input getting reverted to the actual saved value. just checking that thats the behavior we want. 

but this works great and is so clean, all these little details really make the app feel :sparkles:

moped-editor/src/constants/projects.js
johnclary: nice—i think basing the magic numbers on existing data is reasonable. for the sake of keeping things simple in the AGOL layer config, it might be nice to use the same limit across all fields. especially if there's not a UI or business reason for having different values 🤷

johnclary (approved):
Mike, this looks great! And i'm excited to hear you're planning to lean into RFH some more in the future.

one observation is that there are other text fields in the app which have no length constraint, e.g. the project title, component SRTS ID string, project notes. i don't think we need to go out of our way to design for adversarial users, but it would be kinda nice to have the guardrails up universally so as to future proof the AGOL integration for when we add more fields.  

in any case, you've laid out a great pattern here that shouldn't be hard to carry forward across other parts of the app. 🚢 🚢 🚢 🚢 🚢 🍕

chiaberry (approved):
this is so good and I'm glad we are moving towards more input validation!

moped-editor/src/constants/projects.js
mddilley: when you say the same limit across all fields do you mean here in this Moped code config? I thought about this too - making them all 5000 for example. 

Honestly, I might have also arrived at these numbers to mitigate the massive growing blob of text in the UI before adding the `maxRows` prop.

Just want to make sure that I understand before merging, because I was concerned about having a bunch of different arbitrary numbers in here.

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
mddilley: good catch! i'll update. thank you, Rose!!! 🙏

mddilley:
@johnclary Agreed about the other fields. I agree about continuing this pattern now that we've clamped down on these multiline inputs that are easier to run wild with character count inside. I asked once question above that i forgot to tag you in btw.

Definitely took inspiration from all the RHF that I was seeing in VZE ✨ Makes me happy! 😄

moped-editor/src/constants/projects.js
johnclary: I was thinking both—just pick one arbitrary number instead of 3 and use it everywhere. But it's not a strongly held opinion 🤷 

But if you wanted to have a more fine tuned approach here but keep it simple with one standard max length in the AGOL feature service, that would be cool too.

mddilley: makes sense to me! i thought about this overnight, and I'll update these numbers to fall into two categories: 2000 (our current limit that no fields besides descriptions have surpassed) and 10000 (the update limit recently made to the descriptions). The AGOL field length updates require columns to be removed and then re-defined by the Geo team so I'm hesitant to request updates across the board.

This way we'll have matches between the lengths in the AGOL feature service and our code.

mddilley:
@roseeichelmann just caught your comment about the old comment sticking around when canceling an edit. Thank you for catching that I just fixed. 🙏 🙏

PR #1534, opened 2025-02-11
Fix how we hide the project view feedback snackbar

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20772

## Testing
**URL to test:** 

Local or deploy preview

**Steps to test:**
1. Project summary view
   - Test updated a project summary input like Lead or Description
   - Make sure you see the success feedback when saving
   - Throttle your connection speed in browser to "offline" and make sure you still see error feedback when you try to save a change
2. Dashboard
   - Test updating phases, milestones, and notes through the dashboard widgets
   - Test when "offline" to make sure error feedback works too.
3. Last, I left any enhancements that concern handling multiple snackbars at once out of scope here. There are some [cool solutions to this in the MUI docs](https://mui.com/material-ui/react-snackbar/#common-examples), but, for now, I just avoided naming our custom hook as `useSnackbar` since it conflicts with a dependency used in one of these examples.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/FeedbackSnackbar.js
+++ b/moped-editor/src/components/FeedbackSnackbar.js
@@ -51,8 +51,13 @@ export const useSnackbar = () => {
   /**
    * Callback to reset the snackbar state on snackbar close
    */
-  const handleSnackbarClose = () =>
+  const handleSnackbarClose = (_, reason) => {
+    if (reason === "clickaway") {
+      return;
+    }
+
     setSnackbarState({ open: false, message: "", severity: "" });
+  };
 
   return { snackbarState, handleSnackbar, handleSnackbarClose };
 };
--- a/moped-editor/src/components/FeedbackSnackbar.js
+++ b/moped-editor/src/components/FeedbackSnackbar.js
@@ -1,6 +1,43 @@
-import React from "react";
+import React, { useCallback, useState } from "react";
 import { Alert } from "@mui/material";
 import { Snackbar } from "@mui/material";
+
+export const useSnackbar = () => {
+  const [snackbarState, setSnackbarState] = useState(false);
+
+  /**
+   * Wrapper around snackbar state setter
+   * @param {boolean} open - The new state of open
+   * @param {String} message - The message for the snackbar
+   * @param {String} severity - The severity color of the snackbar
+   * @param {Object} error - The error to be displayed and logged
+   */
+  const handleSnackbar = useCallback(
+    (open, message, severity, error) => {
+      // if there is an error, render error message,
+      // otherwise, render success message
+      if (error) {
+        setSnackbarState({
+          open: open,
+          message: `${message}. Refresh the page to try again.`,
+          severity: severity,
+        });
+        console.error(error);
+      } else {
+        setSnackbarState({
+          open: open,
+          message: message,
+          severity: severity,
+        });
+      }
+    },
+    [setSnackbarState]
+  );
+
+  const handleSnackbarClose = () => handleSnackbar(false, "", "");
+
+  return { snackbarState, handleSnackbar, handleSnackbarClose };
+};
 
 /**
  * A project summary snackbar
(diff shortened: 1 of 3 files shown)


moped-editor/src/components/FeedbackSnackbar.js
```diff
@@ -1,33 +1,87 @@
...
+ * @param {Number} autoHideDuration - Duration in milliseconds (default: 5000 milliseconds--for five seconds)
  * @returns {JSX.Element}
  * @constructor
  */
 const FeedbackSnackbar = ({
   snackbarState,
-  handleSnackbar,
-  timeOut = 7000,
+  handleSnackbarClose,
+  autoHideDuration = 5000,
```
mddilley, line 76: 5 seconds feels long enough to me. It also matched what we already had in the docstrings so less edits needed! 😅

moped-editor/src/components/FeedbackSnackbar.js
```diff
@@ -1,33 +1,87 @@
...
 
 /**
- * A project summary snackbar
- * @param {Object} snackbarState - The message you would like to see
- * @param {function} handleSnackbar - The function to update state
- * @param {Number} timeOut - Time out in milliseconds (default: 5000 milliseconds--for five seconds)
+ * Custom hook to manage snackbar state
+ * @returns {Object} - snackbarState, handleSnackbar, handleSnackbarClose
+ */
+export const useFeedbackSnackbar = () => {
```
johnclary, line 9: ✨

johnclary (approved):
@mddilley this looks very great! the snackbar is working very smoothly.

for what it's worth, i have barely ever noticed this snackbar because of its low contrast.

i tried this with the `filled` Alert variant and kinda liked it. although the dual shades of green in this screenshot are tad awkward:

Screenshot 2025-02-12 at 2 13 12 PM

🚢 🚢 🚢 🚢 🚢 🚢

moped-editor/src/components/FeedbackSnackbar.js
```diff
@@ -1,33 +1,87 @@
...
+        });
+      }
+    },
+    [setSnackbarState]
+  );
+
+  /**
+   * Callback to reset the snackbar state on snackbar close
+   */
+  const handleSnackbarClose = (_, reason) => {
```
chiaberry, line 54: I was unfamiliar with this syntax and there are some places where i have definitely included variables I was ignoring (I am thinking `event` for example) and this seems much cleaner

chiaberry (approved):
Five seconds feels a bit long, but I think thats because I immediately looked for the snackbar and watched it until it went away. 

you almost get the same award I get for net negative code added, but you missed out because of your good documentation

tillyw (approved):
this looks great! thank you for consolidating the state and handlers into one custom hook, it's a lot cleaner! everything worked as expected, although there was one behavior that i think may be unexpected.

Screenshot 2025-02-12 at 5 26 17 PM

to recreate, go to the summary page of a project, throttle to offline, and then attempt to update a status. click status update, you will get the expected error messages, but when you click out of the modal you are redirected to the fallback error page. not a major issue, just wanted to point this out!

moped-editor/src/components/FeedbackSnackbar.js
mddilley: cool, i'm on the fence really. I can totally see the value of including `event` so we know it is available in the future. We don't enforce a lint rule about it either so either way flys! 😄

mddilley:
@johnclary @mateoclarke Agreed that it is hard to catch the snackbar. I'll create a follow-up issue about this and capture suggestions from this PR. Also curious to see if this comes up in the ACC design students' project.

mddilley:
@tillyw thanks for catching that error. I will keep that on the radar! 🙏

PR #1540, opened 2025-02-18
Create reusable DataGrid lookup autocompletes

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20171

I am branching off this branch to tackle the MilestoneAutocomplete and the Team Member Autocomplete. Both of those update other fields in their tables and the Milestone Table has some other special behavior (like field names not matching the column render etc)

I have updated the LookupAutocompleteComponent and moved it out of the Project Funding folder. It has default   getOptionLabel and  isOptionEqualToValue functions, but also can accept custom versions of those functions. 

I updated the autocompletes for Subproject Lookup, Fund Autocomplete and Fund Dept Autocomplete. The Lookup Autocomplete component was already being used in the Program and Source fields in the Funding Table. 

## Testing
**URL to test:** 

**Steps to test:**

1. Test all the CRUD operations for the funding table: Source, Program, Fund and Dept-Unit
2. Test adding a subproject from the main project summary page. 

---
#### Ship list
- [x] Code reviewed 


.github/release-drafter.yml
```diff
@@ -1,5 +1,5 @@
 template: |
-  $PREVIOUS_TAG
+  previous tag: $PREVIOUS_TAG
```
chiaberry, line 2: snuck this in here because I get confused on moped release days

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -89,6 +88,12 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+/*
+ * Transportation Project Financial Codes
+ */
+const SOCRATA_ENDPOINT =
+  "https://data.austintexas.gov/resource/bgrt-2m2z.json?$limit=9999";
```
chiaberry, line 95: this used to be in the FundingDeptUnitAutocomplete, it is called by the custom hook `useSocrataJson`. I think its better for it be in this parent component instead of the data being requested *every time* the field goes into edit state (as it does now in prod), but doing like this means its also requested even when users only visit this table without editing

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -426,6 +452,14 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
...
         : updateProjectFundingData.funding_description;
 
+    updateProjectFundingData.funding_source_id =
+      updateProjectFundingData.moped_fund_source?.funding_source_id;
+    delete updateProjectFundingData.moped_fund_source;
+
+    updateProjectFundingData.funding_program_id =
+      updateProjectFundingData.moped_fund_program?.funding_program_id;
+    delete updateProjectFundingData.moped_fund_program;
+
```
chiaberry, line 462: the mutations are expecting the ids and not the objects, so I'm adjusting the payload here.

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -0,0 +1,87 @@
...
+    value[`${name}_id`] === option[`${name}_id`];
+
+  return (
+    <Autocomplete
+      sx={{ width: "100%", mx: 1, alignContent: "center" }}
+      value={value?.[`${name}_id`] ? value : null}
+      // use customized popper component so menu expands to fullwidth
+      PopperComponent={fullWidthPopper && CustomPopper}
+      id={name}
+      filterOptions={filterOptions}
```
chiaberry, line 65: this limits the number of results in the options to 40 at a time, we were previously using it only in the fundAutocompleteComponent, but I think it doesnt hurt using it in all the autocompletes. It is definitely speeding up the subproject one, since that autocomplete was previously trying to render *all* of the projects in its options list

roseeichelmann (approved):
this looks so so good, awesome that this is a reusable component now. i did notice that this brought back some flashing values - when you edit a field and then save it, you see the previous value flashed for a second. other than that this is chefs kiss :man_cook: :kiss:

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -185,7 +183,18 @@ const useColumns = ({
         valueFormatter: (value) =>
           !!value?.fund_name ? `${value?.fund_id} | ${value?.fund_name}` : "",
         renderEditCell: (props) => (
-          <FundAutocompleteComponent {...props} data={data.moped_funds} />
+          <LookupAutocompleteComponent
+            {...props}
+            name={"fund"}
```
mddilley, line 188: I noticed that this is the only autocomplete without the `fullWidthPopper` prop, and I can't see a reason not to add here to keep these consistent in this table. Could you update? Thanks!

I tested adding to the subproject table, and it can make the dropdown really wide with longer project full names so I think that one is ok to leave without the full width dropdown.

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -0,0 +1,87 @@
...
+
+  const defaultIsOptionEqualToValue = (value, option) =>
+    value[`${name}_id`] === option[`${name}_id`];
+
+  return (
+    <Autocomplete
+      sx={{ width: "100%", mx: 1, alignContent: "center" }}
+      value={value?.[`${name}_id`] ? value : null}
+      // use customized popper component so menu expands to fullwidth
+      PopperComponent={fullWidthPopper && CustomPopper}
```
mddilley, line 63: I know this is old code, but could be worth renaming `CustomPopper` to `FullWidthPopper` or something like that since that is what makes it custom and also matches with the prop that is toggling it.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
mddilley: Thanks for calling out this change! I think this is ok too since this dataset only updates once per day. 

If we run into issues in the future, we could always create a wrapper component to render in the edit cell that handles the API call and feeds them to your nice, shiny, and new reusable lookup autocomplete! ✨

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -526,6 +557,7 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
           rowModesModel={rowModesModel}
           onRowModesModelChange={handleRowModesModelChange}
           processRowUpdate={processRowUpdate}
+          onProcessRowUpdateError={(error) => console.error(error)}
```
mddilley, line 560: 🙏 thanks for catching this one and the ones below!

mddilley (changes requested):
Awesome!!! I requested one tiny change for a small regression and added some suggestions. Thank you for slashing so much code and working through these autocompletes. So many tiny details covered! Thanks! 🙌

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -0,0 +1,87 @@
...
+  value,
+  field,
+  hasFocus,
+  name,
+  lookupTable,
+  getOptionLabel,
+  isOptionEqualToValue,
+  fullWidthPopper,
+  error = "false",
+  textFieldHelperText,
```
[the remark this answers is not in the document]
mddilley: adding this since we talked in dev sync about this today:

I was thinking something like:
```js
<LookupAutocomplete
   textFieldProps={{
      helperText: textFieldHelperText,
      ...any other props that will be used in the TextField component
   }}
   autocompleteProps={{
      getOptionLabel: () => {
         // the function code
      },
      ...any other props that will be used in the Autocomplete component
   }}
```

Curious about other thoughts on this too - especially if anyone thinks this would make it more confusing.

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
chiaberry: i like this idea, done!

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -185,7 +183,18 @@ const useColumns = ({
         valueFormatter: (value) =>
           !!value?.fund_name ? `${value?.fund_id} | ${value?.fund_name}` : "",
         renderEditCell: (props) => (
-          <FundAutocompleteComponent {...props} data={data.moped_funds} />
+          <LookupAutocompleteComponent
+            {...props}
+            name={"fund"}
+            lookupTable={data.moped_funds}
```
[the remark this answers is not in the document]
chiaberry: much better prop name, thank you

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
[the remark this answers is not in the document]
chiaberry: I am going to rework the props a little bit to mimic this example. I just noticed that the `textFieldHelperText` prop I added isnt being used here, its used in the milestone autocomplete that I punted to another branch. But I will still keep it in here. 

in order to pass props that are not explicitly declared here as props but also use our internal defaults, it would look something like this? 

```jsx
<Autocomplete
     {...autocompleteProps}
    options={options}
    getOptionLabel={autocompleteProps.getOptionLabel ? autocompleteProps.getOptionLabel : defaultGetOptionLabel}
 />
```

right?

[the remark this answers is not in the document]
chiaberry: actually both of the TextField props i was passing is the punted milestone autocomplete, i am going to remove them from this PR to keep it cleaner

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -109,44 +114,37 @@ const useColumns = ({
...
   handleCancelClick,
   handleEditClick,
+  deptUnitData,
 }) =>
   useMemo(() => {
     return [
       {
         headerName: "Source",
-        field: "funding_source_id",
+        field: "moped_fund_source",
```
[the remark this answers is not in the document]
mddilley: thanks for fixing!!! 🎉

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -109,44 +114,39 @@ const useColumns = ({
...
-          ),
+        valueFormatter: (value) => value?.funding_source_name,
         renderEditCell: (props) => (
           <LookupAutocompleteComponent
             {...props}
             name={"funding_source"}
-            lookupTable={data["moped_fund_sources"]}
+            options={data["moped_fund_sources"]}
+            fullWidthPopper={true}
+            autocompleteProps={{}}
```
mddilley, line 133: Another way to handle this could be a default value for the prop in the `LookupAutocompleteComponent` or use optional chaining when reach the prop values in the component code. Caught my eye and had me thinking about how to avoid passing an empty object.

mddilley (approved):
Thanks for fixing the autofocus! I ran through the test steps again, and all is working just as before. So nice to have another reusable table input component. 🙌 🚢 🚀

I added one last comment about the autocomplete prop.

tillyw (approved):
tested locally and working flawlessly! i love seeing all the deleted files replaced by one reusable component, so nice and clean!

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -0,0 +1,80 @@
...
+    apiRef.current.setEditCellValue({
+      id,
+      field,
+      value: newValue,
+    });
+  };
+
+  const defaultGetOptionLabel = (option) => option[`${name}_name`];
+
+  const defaultIsOptionEqualToValue = (value, option) =>
```
johnclary, line 49: this function and the one above it look like utilities that could get declared/imported elsewhere 😇

moped-editor/src/queries/funding.js
```diff
@@ -14,7 +14,15 @@ export const FUNDING_QUERY = gql`
...
       funding_program_id
+      moped_fund_program {
+        funding_program_id
+        funding_program_name
+      }
       funding_source_id
+      moped_fund_source {
+        funding_source_id
+        funding_source_name
+      }
```
johnclary, line 25: always nice to see our graphql API being better leveraged!

moped-editor/src/utils/autocompleteHelpers.js
```diff
@@ -4,7 +4,7 @@
 export const filterOptions = (options, { inputValue, getOptionLabel }) => {
   // limits options to ensure fast rendering
   const limit = 40;
-  // applies the default autcomplete matching behavior plus our limit filter
+  // applies the default autocomplete matching behavior plus our limit filter
```
johnclary, line 7: this has my fingerprints all over it 😳

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
johnclary: agree with what's been said. this is preferred. 

as an aside it probably wouldn't hurt to add another two 9's to that limit param to make it forever proof

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -185,7 +183,19 @@ const useColumns = ({
...
+            options={data.moped_funds}
+            fullWidthPopper={true}
+            autocompleteProps={{
+              getOptionLabel: (option) =>
+                option.fund_id ? `${option.fund_id} | ${option.fund_name}` : "",
+              isOptionEqualToValue: (value, option) =>
+                value.fund_id === option.fund_id &&
+                value.fund_name === option.fund_name,
+            }}
+          />
```
johnclary, line 198: another spot where declaring an object out of the component scope would put my mind at ease about performance degradation 😇

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -199,7 +209,20 @@ const useColumns = ({
...
+            name={"dept_unit"}
+            options={deptUnitData}
+            autocompleteProps={{
+              getOptionLabel: (option) =>
+                !!option.dept
+                  ? `${option.dept} | ${option.unit} | ${option.unit_long_name} `
+                  : "",
+              isOptionEqualToValue: (value, option) =>
+                value.unit_long_name === option.unit_long_name,
+            }}
```
johnclary, line 223: ditto here. i know i'm probably wringing my hands over nothing, but out of scope variables don't cost a thing.

moped-editor/src/views/projects/projectView/ProjectSummary/SubprojectsTable.js
```diff
@@ -50,7 +50,15 @@ const useColumns = ({
...
         renderEditCell: (props) => (
-          <SubprojectLookupComponent {...props} data={data} />
+          <LookupAutocompleteComponent
+            {...props}
+            name="project"
+            options={data?.subprojectOptions}
+            autocompleteProps={{
+              getOptionLabel: (option) =>
+                `${option.project_id} - ${option.project_name_full}`,
+            }}
```
johnclary, line 60: last one where the props could be declared out of scope so as to provide a stable object reference to the component 👍

johnclary (approved):
So much red! Really nice work! 🚢

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
chiaberry: So you propose that we declare the entire `autocompleteProps` out of the scope of the useColumns hook, or the individual functions?

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
chiaberry: I started trying to declare them in the utils folder but then got tripped up with trying to pass `name` to the function since Autocomplete is expecting only `options` https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-getOptionLabel 

Chatted with Mike, I am going to wrap them in a useCallback

PR #1541, opened 2025-02-18
Update data type of council district lists to improve performance

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/21151

This fix updates the data type of the columns called `project_council_districts` and `project_and_child_project_council_districts` to be jsonb which seems to perform better with Hasura's generated query that relies on the `@>` operator for searching if a list of districts contains a specific district number

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Spin up your local stack
2. Test a filter on the project list that includes `Council district contains <one of the dropdown options>` like [this one](https://localhost:3000/moped/projects?filters=%5B%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22Canceled%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22Potential%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22On+hold%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_is_not_blank%22%2C%22value%22%3Anull%7D%2C%7B%22field%22%3A%22project_and_child_project_council_districts%22%2C%22operator%22%3A%22council_districts_array_contains%22%2C%22value%22%3A%229%22%7D%5D&isOr=false&map=false)
3. Test a filter on the project list that includes `Council district is <one of the dropdown options>` like [this one](https://localhost:3000/moped/projects?filters=%5B%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22Canceled%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22Potential%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_does_not_equal_case_insensitive%22%2C%22value%22%3A%22On+hold%22%7D%2C%7B%22field%22%3A%22current_phase%22%2C%22operator%22%3A%22string_is_not_blank%22%2C%22value%22%3Anull%7D%2C%7B%22field%22%3A%22project_and_child_project_council_districts%22%2C%22operator%22%3A%22council_districts_array_is%22%2C%22value%22%3A%226%22%7D%5D&isOr=false&map=false)
4. Now, checkout `main` branch and repeat the tests. You should see they perform worse. I am seeing about 5 seconds when using contains locally.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/views/project_list_view.sql
+++ b/moped-database/views/project_list_view.sql
@@ -1,4 +1,4 @@
--- Most recent migration: moped-database/migrations/1737754622032_deprecate_type_tables/up.sql
+-- Most recent migration: moped-database/migrations/1739832264644_update_districts_to_jsonb/up.sql
 
 CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
     SELECT
@@ -105,8 +105,8 @@ parent_child_project_map AS (
 
     SELECT
         projects.project_id,
-        array_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null) AS project_council_districts,
-        array_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null) AS project_and_child_project_council_districts
+        jsonb_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null) AS project_council_districts,
+        jsonb_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null) AS project_and_child_project_council_districts
     FROM parent_child_project_map projects
     LEFT JOIN project_council_district_map project_and_children_districts ON projects.self_and_children_project_ids = project_and_children_districts.project_id
     LEFT JOIN project_council_district_map project_districts ON projects.project_id = project_districts.project_id
@@ -173,8 +173,8 @@ project_component_work_types AS (
             ORDER BY mwt.name
         ) AS component_work_type_names
     FROM moped_proj_components mpc
-    LEFT JOIN moped_proj_component_work_types mpcwt ON mpcwt.project_component_id = mpc.project_component_id
-    LEFT JOIN moped_work_types mwt ON mwt.id = mpcwt.work_type_id
+    LEFT JOIN moped_proj_component_work_types mpcwt ON mpc.project_component_id = mpcwt.project_component_id
+    LEFT JOIN moped_work_types mwt ON mpcwt.work_type_id = mwt.id
     WHERE true AND mpc.is_deleted = false AND mpcwt.is_deleted = false
     GROUP BY mpc.project_id
 )
--- a/moped-database/migrations/1739832264644_update_districts_to_jsonb/down.sql
+++ b/moped-database/migrations/1739832264644_update_districts_to_jsonb/down.sql
@@ -1,0 +1,553 @@
+DROP VIEW IF EXISTS exploded_component_arcgis_online_view;
+DROP VIEW IF EXISTS component_arcgis_online_view;
+DROP VIEW IF EXISTS project_list_view;
+
+CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
+    SELECT
+        mpp.project_id,
+        string_agg(DISTINCT concat(mu.first_name, ' ', mu.last_name, ':', mpr.project_role_name), ','::text) AS project_team_members
+    FROM moped_proj_personnel mpp
+    JOIN moped_users mu ON mpp.user_id = mu.user_id
+    JOIN moped_proj_personnel_roles mppr ON mpp.project_personnel_id = mppr.project_personnel_id
+    JOIN moped_project_roles mpr ON mppr.project_role_id = mpr.project_role_id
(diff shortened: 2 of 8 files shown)


moped-database/views/project_list_view.sql
```diff
@@ -105,8 +105,8 @@ parent_child_project_map AS (
 
     SELECT
         projects.project_id,
-        array_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null) AS project_council_districts,
-        array_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null) AS project_and_child_project_council_districts
+        jsonb_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null) AS project_council_districts,
+        jsonb_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null) AS project_and_child_project_council_districts
```
mddilley, line 109: this is the change - updating these columns to be jsonb which is performing better with Hasura's generated query.

the two columns don't have any downstream dependencies so this change feels safe to me.

mddilley:
@Charlie-Henry I requested your review too since this `project_list_view` database view that is being update in this PR is used in the Power BI dataflow.

chiaberry (approved):
yeesh, yes, its so slow in main right now. I dont know the nuances between jsonb_agg and array_agg, but I agree that since they dont have downstream dependencies, it feels safe

moped-database/views/project_list_view.sql
frankhereford: Mike, I'm baffled -- and i'm curious what you think. I was wondering how this change doesn't break [the project list column](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/views/projects/projectsListView/ProjectsListViewFiltersConf.js#L562-L582) & [search](https://github.com/cityofaustin/atd-moped/blob/5f6fd414d3ae877af147470666563aad6fa97941/moped-editor/src/components/GridTable/FiltersCommonOperators.js#L192-L206), but by any account, they still seem to work fine. Do you think that it's `graphql-engine` representing the native postgres array as JSON because that is its nature, and it representing the native JSON as JSON because, like what else could it do?

Could it be the fact that both of those distinct data types in postgres are presented the same way by `graphql-engine` that this is a non-breaking change?

frankhereford (approved):
This works good on my machine, plenty quick. I have one question / thought, but it's also neither here nor there, so this is 🚀

moped-database/views/project_list_view.sql
mddilley: thanks again for catching this, Frank! It looks like the only place that we use the Hasura `_contains` operator in the editor code is for this council district search. https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/components/GridTable/FiltersCommonOperators.js#L192-L198

I'll look more at this after MUG today, but this is so helpful to make sure that our code lines up with the data types that we are operating on.

Charlie-Henry: Hmm this will probably break a few things on Power BI as we're processing this list of integers into a separate table for council districts. Not saying this is something we can't fix but it will likely break two dashboards for me, not entirely sure about Nathan.

mddilley: @Charlie-Henry thanks for the heads up! I'll spend more time this afternoon to see if we have other options.

johnclary (approved):
LGTM + TIL 🚢

moped-database/views/project_list_view.sql
mddilley: Charlie and I chatted more about this in slack, and this looks safe to merge as-is after all. 🙌 I'll handle anything unexpected results but there shouldn't be any at all.🛡️

mddilley:
@frankhereford @chiaberry @johnclary I pushed on last update based on the code that Frank pointed out. I think this makes it clearer that we are handling a different data type.

I'll do some last testing the morning but will target a patch tomorrow if all looks good.

johnclary:
@mddilley sounds good—ship away!

fwiw, i think that `array` was slightly preferable because, in the eyes of our frontend client, the array and jsonb types are one and the same—a JSON array. it's not really a concern here what the underlying DB types are. just my two cents. please proceed 🚢 🚢 🚢

mddilley:
@johnclary that is a good point. 

@chiaberry @frankhereford have thoughts? I'm too lost in the sauce, and `jsonb` does make it seem like it could be any blob when it really is just an array.

I'm also still annoyed that there is no straightforward way to make Hasura choose the `ANY` operator which we've proven works better on the original integer[]. Gonna have to get over that. 😅🙃

chiaberry (commented):
I too am lost in the sauce, but my approval still stands ✅

mddilley:
Thanks Chia! I now have coffee on my side ☕, and I reverted the front-end code changes:
- to keep the code using front-end speak since it is not concerned with non-array json
- to keep the advanced filter code from becoming more complex

PR #1547, opened 2025-02-26
Add column to components database view to show if a component is mapped or not

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20901

**DO NOT MERGE UNTIL https://github.com/cityofaustin/atd-data-tech/issues/20825 is complete.**

This PR adds a column that can be used to surface components that need mapping in Moped from an AGOL map like https://austin.maps.arcgis.com/apps/webappviewer/index.html?id=404d31d56b57491abe53ccfd718fcaee.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Run the SQL below and make sure your results match up with the comments

```sql
-- Should return 0 rows
SELECT * 
FROM component_arcgis_online_view
WHERE is_mapped = 'Yes' AND geometry IS NULL;

-- Note this count as the total
SELECT count(*) 
FROM component_arcgis_online_view;

-- Note this count to add to the next count
SELECT count(*) 
FROM component_arcgis_online_view
WHERE is_mapped = 'Yes';

-- This count plus the previous count should add up to the total number of rows in the view
SELECT count(*) 
FROM component_arcgis_online_view
WHERE is_mapped = 'No';
```

2. Test the down migration

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/views/component_arcgis_online_view.sql
+++ b/moped-database/views/component_arcgis_online_view.sql
@@ -1,4 +1,4 @@
--- Most recent migration: moped-database/migrations/1739832264645_fix_agol_view_bug/up.sql
+-- Most recent migration: moped-database/migrations/1739832264646_agol_view_is_mapped/up.sql
 
 CREATE OR REPLACE VIEW component_arcgis_online_view AS WITH work_types AS (
     SELECT
@@ -164,6 +164,10 @@ SELECT
         WHEN mc.line_representation = true THEN 'Line'::text
         ELSE 'Point'::text
     END AS geometry_type,
+    CASE
+        WHEN comp_geography.geometry IS null THEN 'No'::text
+        ELSE 'Yes'::text
+    END AS is_mapped,
     subcomponents.subcomponents AS component_subcomponents,
     work_types.work_types AS component_work_types,
     component_tags.component_tags,
--- a/moped-database/metadata/tables.yaml
+++ b/moped-database/metadata/tables.yaml
@@ -76,6 +76,7 @@
           - geometry_type
           - interim_project_component_id
           - interim_project_id
+          - is_mapped
           - is_within_city_limits
           - knack_data_tracker_project_record_id
           - length_feet_total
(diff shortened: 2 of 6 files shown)


roseeichelmann (approved):
ran the SQL commands and down migration, this looks good!

chiaberry (approved):
tested up and down migration! again thank you for the test steps and explanation

moped-database/views/component_arcgis_online_view.sql
```diff
@@ -164,6 +164,10 @@ SELECT
         WHEN mc.line_representation = true THEN 'Line'::text
         ELSE 'Point'::text
     END AS geometry_type,
+    CASE
+        WHEN comp_geography.geometry IS null THEN 'No'::text
+        ELSE 'Yes'::text
```
johnclary, line 169: i would feel remiss if i didn't ask: why not reach for a `boolean` here? i do appreciate that it's good to have the view spitting out what we want to render in a UI—such as an AGOL tooltip, perhaps? it's just one of those things we might regret if we want to pull this data into another system that respects data types 🤷

johnclary (approved):
looks good and test steps check out 🚢

moped-database/views/component_arcgis_online_view.sql
mddilley: i thought about this too but I decided to adhere to the request of having human-readable values instead of needing to map a boolean to them elsewhere.

If it backfires on me, we have the expected data type of String set up in the AGOL dataset like we have for other booleans like `is_within_city_limits` that populate as `0` or `1`. So, an adjustment shouldn't be too bad.

johnclary: cool—thanks for the context and sorry if you already wrote that in issue itself!

tillyw (approved):
the numbers add up as they should! down migration looks good 👍

moped-database/views/component_arcgis_online_view.sql
mddilley: nah i didn't do a great job of capturing that in the issue itself. thanks for keeping an eye on these things! 🙏

mddilley:
@mateoclarke That sounds like an awesome Disco Day project! 👩‍🚀 And pgTap is a TIL for me. 👀

mddilley:
thanks y'all - merging!

PR #1548, opened 2025-02-27
Update `is_mapped` to be a real actual boolean

mddilley:
## Associated issues
n/a

This is an adjustment made to changes made in https://github.com/cityofaustin/atd-moped/pull/1547 based off new info.

## Testing
**URL to test:** 

Local

**Steps to test:**
Read the small difference between this diff and the one in https://github.com/cityofaustin/atd-moped/pull/1547 and/or test the SQL below. Another similar column to compare to is `is_within_city_limits`.

```sql
-- Should return 0 rows
SELECT * 
FROM component_arcgis_online_view
WHERE is_mapped = true AND geometry IS NULL;

-- Note this count as the total
SELECT count(*) 
FROM component_arcgis_online_view;

-- Note this count to add to the next count
SELECT count(*) 
FROM component_arcgis_online_view
WHERE is_mapped = true;

-- This count plus the previous count should add up to the total number of rows in the view
SELECT count(*) 
FROM component_arcgis_online_view
WHERE is_mapped = false;
```

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1739832264647_agol_view_is_mapped_bool/down.sql
+++ b/moped-database/migrations/1739832264647_agol_view_is_mapped_bool/down.sql
@@ -1,0 +1,266 @@
+DROP VIEW IF EXISTS exploded_component_arcgis_online_view;
+DROP VIEW IF EXISTS component_arcgis_online_view;
+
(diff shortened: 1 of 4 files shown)


johnclary (approved):
looks good to me! 🚢 🚢 🚢 🚢 🚢

PR #1549, opened 2025-02-27
Fix funding table to not save on clickaway when there is no change

mddilley:
## Associated issues
n/a

## Testing
**URL to test:** 

Local or deploy preview

**Steps to test:**
These came straight from the QA sheet. 🍝 

1. Use "Add funding source" button to add a blank funding source to the funding table.
2. Fill out all fields and save the row - verify that you can't use a funding amount with a decimal or non-numeric characters
3. Double-click a cell to edit. Click away to cancel without editing. This should not create a new edit event in the activity log. Now, edit a value this way. This should save the edit and create an activity log row.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -459,11 +459,11 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
         : updateProjectFundingData.funding_description;
 
     updateProjectFundingData.funding_source_id =
-      updateProjectFundingData.moped_fund_source?.funding_source_id;
+      updateProjectFundingData.moped_fund_source?.funding_source_id || null;
     delete updateProjectFundingData.moped_fund_source;
 
     updateProjectFundingData.funding_program_id =
-      updateProjectFundingData.moped_fund_program?.funding_program_id;
+      updateProjectFundingData.moped_fund_program?.funding_program_id || null;
     delete updateProjectFundingData.moped_fund_program;
 
     if (updatedRow.isNew) {
@@ -504,6 +504,8 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
     } else {
       // Remove __typename since we removed it from updatedRow and check if the row has changed
       delete originalRow.__typename;
+      delete originalRow.moped_fund_source;
+      delete originalRow.moped_fund_program;
       const hasRowChanged = !isEqual(updatedRow, originalRow);
 
       if (!hasRowChanged) {


moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -459,11 +459,11 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
         : updateProjectFundingData.funding_description;
 
     updateProjectFundingData.funding_source_id =
-      updateProjectFundingData.moped_fund_source?.funding_source_id;
+      updateProjectFundingData.moped_fund_source?.funding_source_id || null;
```
mddilley, line 462: the optional chaining was working but the value was falling back to `undefined` which was being compared for equality against `null` which seemed like an edit.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -504,6 +504,8 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
     } else {
       // Remove __typename since we removed it from updatedRow and check if the row has changed
       delete originalRow.__typename;
+      delete originalRow.moped_fund_source;
+      delete originalRow.moped_fund_program;
```
mddilley, line 508: a couple of extra keys were being interpreted by our code as an edit.

chiaberry (approved):
Thanks for fixing the regression I introduced -- I notice that if you add new row, dont edit anything but click away it saves the "empty" row. I put empty in quotes because it does have a Tentative status. I checked out the production branch locally and I am seeing it saves there too, so its not a regression.

johnclary (approved):
Regression definitely fixed 🙏 

The click-away save behavior is a bit surprising to me!

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
roseeichelmann: pesky `undefined`'s!

chiaberry:
surprising how? its a feature!!

roseeichelmann (approved):
test steps worked, awesome you found such a quick fix! and also awesome that the moped test instructions are so thorough to catch stuff like this!

mddilley:
thanks for the 👀 👀 👀 y'all!!!

mddilley:
@johnclary yeah it is a feature that can be disabled. the goal was to reduce clicks but maybe we'll get some feedback from the UX students. 🤷 Haven't heard anything from MUG 🦗

@chiaberry thanks for noting that. i see that behavior too in prod and this branch 🙏

PR #1551, opened 2025-02-28
Update team table so that duplicate team members cannot be added

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21199

Even though theres minimal code changes i found this to be way more complicated to untangle/implement than i thought it would! 

## Testing
**URL to test:** 
https://deploy-preview-1551--atd-moped-main.netlify.app/moped/

**Steps to test:**
1. Test adding new members to a team table. You shouldnt be able to add members that are already on the team, they shouldnt be an option
2. Edit a team member row. You shouldnt see options for members that are already on the team, but you should be able to click on a diff member and then click back to the original member of the row before you edited it and save that, therefore not really updating anything.
3. Test deleting members, updating, adding fields, etc try to break something!
4. See that the [object Object] on the workgroup field is no longer showing when saving new team member edits

---
#### Ship list
- [x] Code reviewed 
   - Updated existing test: `Verify that you cannot add a team member without selecting a team member name and at least one role. Also verify that you can't add the same person to one project more than once.`

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
@@ -73,15 +73,18 @@ const useColumns = ({
           return user ? `${user.first_name} ${user.last_name}` : "";
         },
         renderEditCell: (props) => {
+          // the team member object for the current row
           const currentRowMember =
             data?.moped_project_by_pk?.moped_proj_personnel.filter(
               (user) => user.project_personnel_id === props.id
             );
+          // the existing team members on this project
           const existingTeamMembers =
             data?.moped_project_by_pk?.moped_proj_personnel.map(
               (option) => option.moped_user.email
             );
           // filter out existing team members from list of options unless they are the current row member
+          // that way the current member remains an option when editing a row
           const teamMembersWithoutDuplicates = data?.moped_users.filter(
             (user) => {
               return (
--- a/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
@@ -73,15 +73,22 @@ const useColumns = ({
           return user ? `${user.first_name} ${user.last_name}` : "";
         },
         renderEditCell: (props) => {
-          const currentTeamMembers =
+          const currentRowMember =
+            data?.moped_project_by_pk?.moped_proj_personnel.filter(
+              (user) => user.project_personnel_id === props.id
+            );
+          const existingTeamMembers =
             data?.moped_project_by_pk?.moped_proj_personnel.map(
               (option) => option.moped_user.email
             );
-          // filter out current team members from list of moped users unless they are the row.moped_user.email
+          // filter out existing team members from list of options unless they are the current row member
           const teamMembersWithoutDuplicates = data?.moped_users.filter(
-            (user) =>
-              !currentTeamMembers.includes(user.email) ||
-              user.email === props.row.moped_user.email
+            (user) => {
+              return (
+                !existingTeamMembers.includes(user.email) ||
+                user.email === currentRowMember?.[0]?.moped_user.email
+              );
+            }
           );
           return (
             <TeamAutocompleteComponent
@@ -90,6 +97,7 @@ const useColumns = ({
               value={props.row.moped_user}
               options={teamMembersWithoutDuplicates}
               error={props.error}
+              workgroupLookup={workgroupLookup}
             />
           );
         },
--- a/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
@@ -73,12 +73,22 @@ const useColumns = ({
           return user ? `${user.first_name} ${user.last_name}` : "";
         },
         renderEditCell: (props) => {
+          const currentTeamMembers =
+            data?.moped_project_by_pk?.moped_proj_personnel.map(
+              (option) => option.moped_user.email
+            );
+          // filter out current team members from list of moped users unless they are the row.moped_user.email
+          const teamMembersWithoutDuplicates = data?.moped_users.filter(
+            (user) =>
+              !currentTeamMembers.includes(user.email) ||
+              user.email === props.row.moped_user.email
+          );
           return (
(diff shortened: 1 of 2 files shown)


moped-editor/src/views/projects/projectView/ProjectTeam/TeamAutocompleteComponent.js
```diff
@@ -49,7 +50,10 @@ const TeamAutocompleteComponent = ({
     apiRef.current.setEditCellValue({
       id,
       field: "moped_workgroup",
-      value: { workgroup_id: newValue?.workgroup_id },
+      value: {
+        workgroup_id: newValue?.workgroup_id,
+        workgroup_name: workgroupLookup[newValue?.workgroup_id],
```
roseeichelmann, line 55: this fixed a bug thats currently in prod where if you edit a row and click a new user and then in the same edit session click the old user again and save that, you would see [object Object] in the workgroup name field. also in other cases where you are actually saving a new user, [object Object] would just flash on the screen in the time between clicking save and [this hook](https://github.com/cityofaustin/atd-moped/blob/37752cc60f54d49fb19ec077d23c2165b5e28915/moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js#L219) running which manually adds the `workgroup_name` to the row data

mddilley: Nice! Thank you for catching AND fixing this bug. 🚀 I was able to see the native object peek out when I adjusted my connection speed. 🙏

FYI @chiaberry this probably affects your work on reusable autocompletes that have dependent fields

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
```diff
@@ -73,13 +73,34 @@ const useColumns = ({
...
+            data?.moped_project_by_pk?.moped_proj_personnel.filter(
+              (user) => user.project_personnel_id === props.id
+            );
+          // the existing team members on this project
+          const existingTeamMembers =
+            data?.moped_project_by_pk?.moped_proj_personnel.map(
+              (option) => option.moped_user.email
+            );
+          // filter out existing team members from list of options unless they are the current row member
+          // that way the current member remains an option when editing a row
```
mddilley, line 87: this feels like it could be moved up in scope and wrapped with `useMemo`. Similar to `useWorkgroupLookup` in this same file.

I don't think there will be long lists of team members in projects so performance is debatable, but I think it would be an easy win to move up and pass into `useColumns` since it is used by all rows.

also, what do you think about sticking to using user IDs to find objects in this new code? I might be missing a limitation that makes it necessary to use email in some places.

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
```diff
@@ -73,13 +73,34 @@ const useColumns = ({
           return user ? `${user.first_name} ${user.last_name}` : "";
         },
         renderEditCell: (props) => {
+          // the team member object for the current row
+          const currentRowMember =
+            data?.moped_project_by_pk?.moped_proj_personnel.filter(
```
mddilley, line 78: small thing but if you use `.find()` here then you will return the object instead of the object wrapped in an array.

mddilley (commented):
Looks great! I tested adding, editing, and deleting team members, and I can no longer add the same person over and over. Thanks for bundling in a bug fix too. Great find! 🙏

I left some comments about code that I'd like to get your feedback on, but this is 🚢 based on functionality. 🙌

roseeichelmann:
thanks for your feedback mike, i pushed up some updates lmk what you think!

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
roseeichelmann: ooo til, thank u!

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
roseeichelmann: i updated to move `existingTeamMembers` up in scope and pass to `useColumns`, ty for that rec! also no reason to use email i think i picked that bc it was easier to debug with and then never went back to reevaluate my choice, `user_id` is definitely the better option here!

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
```diff
@@ -73,13 +82,29 @@ const useColumns = ({
...
         },
         renderEditCell: (props) => {
+          // the team member object for the current row
+          const currentRowMember =
+            data?.moped_project_by_pk?.moped_proj_personnel.find(
+              (user) => user.project_personnel_id === props.id
+            );
+          // filter out existing team members from list of options unless they are the current row member
+          // that way the current member remains an option when editing a row
+          const teamMembersWithoutDuplicates = data?.moped_users.filter(
```
chiaberry, line 92: I'm not 100% sold on this variable name. Because our existing list of users already does not have duplicates in it. This is a list of team members that aren't already on the project, so its *preventing* duplicates. 

maybe `unassignedTeamMembers`?

moped-editor/src/views/projects/projectView/ProjectTeam/TeamAutocompleteComponent.js
chiaberry: Thanks for addressing this bug! I am going to check the milestones table to see if I see that same bug there.

chiaberry (approved):
🚢  Good fix! and thank you for fixing that bug. 
I am not sold on the variable name, but its nothing preventing this from being merged

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
roseeichelmann: cool i like that, just updated!

mddilley (approved):
Works just as expected, and a bug squashed in the process! 🙏 🚢 🚀

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
mddilley: cool cool thank you for updating 🙏

PR #1552, opened 2025-03-05
Edit LookupAutocompleteComponent to update related field

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/21358

This shouldn't be merged until Rose's PR is merged https://github.com/cityofaustin/atd-moped/pull/1551. I pulled her branch into this one since she fixed a bug in the Team Autocomplete. I tried to see if the Milestone table had the same bug, but I couldnt reproduce. I think its because the related field in the milestone table is displaying something from the same object as the moped_milestone. 

## Testing
**URL to test:** 

https://deploy-preview-1552--atd-moped-main.netlify.app/moped/projects/

**Steps to test:**

This PR removes the Milestone Autocomplete and the Team Member Autocomplete and replaces them with the LookupAutocomplete from https://github.com/cityofaustin/atd-moped/pull/1540. 

Team table 
- Test adding a new user. Confirm the workgroup field is also updated as you select the new user. Save the user.
- Test editing an existing user.

Milestones table (on Timeline tab)
- Test adding a new Milestone. Related phase should also update as you select the milestone. Save new row. 
- Test editing an existing milestone. 

Test another LookupAutocompleteField that doesn't update a dependent, to make sure there are no regressions. Any of the autocompletes on the Funding table are good candidates. 

---
#### Ship list
- [x] Code reviewed 


mddilley:
@chiaberry is it possible to update the base branch of this PR to Rose's branch since you pulled her work in here? Not sure if you already tried or if the history works out to clean up the diff, but it could help with the merge order dependency.

mddilley (approved):
I tested the milestones and team tables, and I see the dependent fields working just as before. I also don't see any regressions in the other tables when I tested locally. 🙌 🚢

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -42,6 +48,13 @@ const LookupAutocompleteComponent = ({
       field,
       value: newValue,
     });
+    if (dependentFieldName) {
+      apiRef.current.setEditCellValue({
+        id,
+        field: dependentFieldName,
+        value: dependentFieldValue(newValue),
```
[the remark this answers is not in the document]
chiaberry: i agree with this opinion!

mddilley:
@chiaberry Would you be up for fixing this bug that I found in the funding table in this PR? I can demo at dev sync today or whenever you are available. I think this is a one-liner fix since it has to do with a bug in the state of the table and a mutation isn't clearing the data.

To reproduce:
1. Start editing an existing cell but don't make any changes
2. Click away and see that the table clears data from the source and program columns from its internal state
3. Refresh the page and see that the mutation didn't save the change which is what we want

https://github.com/user-attachments/assets/4b70b96a-1756-4a37-852a-da9bf78b0962

roseeichelmann (approved):
followed the test steps, this looks great! the bug in my pr had to do with `workgroup_name` not being set, so it was specific to the team table and makes sense you couldnt reproduce. yayy for prs with more red than green :partying_face:

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -445,7 +445,7 @@ const ProjectFundingTable = ({ handleSnackbar }) => {
 
   // saves row update, either editing an existing row or saving a new row
   const processRowUpdate = (updatedRow, originalRow) => {
-    const updateProjectFundingData = updatedRow;
+    const updateProjectFundingData = { ...updatedRow };
```
chiaberry, line 448: @mddilley I did this to address the bug you found -- now `updatedRow` isn't being touched so if the row hasn't changed, then when the Promise resolves, it keeps the same data

mddilley: awesome! thank you for tracking this down and fixing! I retested and the bug is gone. 🙏 🙌

PR #1555, opened 2025-03-07
Fix files table to only remove row from state after mutation succeeds

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20881

## Testing
**URL to test:** 

https://deploy-preview-1555--atd-moped-main.netlify.app/moped/

**Steps to test:**

1. Go to a project's files table with files (like project [290](https://localhost:3000/moped/projects/290?tab=files) in a production snapshot) or add some files
2. Use your browser dev tools to set your connection to "Offline"
3. Delete a file and see that an error snackbar appears and the row does not change. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFiles.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFiles.js
@@ -357,9 +357,6 @@ const ProjectFiles = ({ handleSnackbar }) => {
   // handles row delete
   const handleDeleteClick = useCallback(
     (id) => () => {
-      // remove row from rows in state
-      setRows(rows.filter((row) => row.project_file_id !== id));
-
       deleteProjectFileAttachment({
         variables: {
           fileId: id,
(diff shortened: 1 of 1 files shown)


moped-editor/src/views/projects/projectView/ProjectFiles.js
```diff
@@ -357,16 +357,15 @@ const ProjectFiles = ({ handleSnackbar }) => {
...
-
       deleteProjectFileAttachment({
         variables: {
           fileId: id,
         },
       })
         .then(() => refetch())
         .then(() => {
+          // remove row from rows in state
+          setRows(rows.filter((row) => row.project_file_id !== id));
```
mddilley, line 368: Work and looks great! I noticed that in some places we do:

```js
refetch().then(() => {
// closes snackbar and other things
});
```
What do you think about updating so the UI is frozen until the table data refetch is complete? I'm also open to creating an issue to find a way to iron these out across the board. It is feeling a little like Whac-a-Mole to me but I could be fretting too much.

mddilley (commented):
Works great! I left one comment with a question.

moped-editor/src/views/projects/projectView/ProjectFiles.js
mddilley: nvm we talked about this in dev sync, and then I proved to myself that this doesn't matter. Thanks for bearing with my over-thinking. Approving! 🚀

mddilley (approved):
Thanks - looks great! 🚢

PR #1561, opened 2025-03-11
Update component phases and completion dates by SRTS ID using current SRTS statuses

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/21460

This PR adds a helper script to:
1. Find project components that:
   - Match SRTS ID rows to a source .csv file that has current statuses and completion dates for those SRTS IDs
   - Don't already have a substantial completion date according to the `component_arcgis_online_view` DB view
      - Querying this view excludes soft-deleted components which we want to do
2. Update components that fit these filters with the "Complete" phase and the completion date in the .csv file
3. Write csvs to track what updates were made and which SRTS IDs had no component matches
4. You can check a project like https://localhost:3000/moped/projects/2667?tab=map&project_component_id=4352 to see a component that had a Complete phase and completion date added. This one should now have **Complete** phase and a completion date of **06/08/2022** since it matched. You can check the phase and complete date [in the spreadsheet](https://docs.google.com/spreadsheets/d/1M2dwi-GqheZLmy-GJwmrf6LAJpu07aoiD3aTEOo7IUQ/edit?gid=1500102334#gid=1500102334) under the SRTS ID of **2C - 602**.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start up your local stack and then follow the readme in this PR
2. You should see the script succeed and see two new .csvs in the `/data/` folder.

---
#### Ship list
- [x] Code reviewed 


moped-toolbox/backfill_complete_srts_ids/secrets_template.py
```diff
@@ -0,0 +1,12 @@
+HASURA = {
+    "HASURA_ENDPOINT": {
+        "local": "http://localhost:8080/v1/graphql",
+        "staging": "",
+        "prod": "",
+    },
+    HASURA_ADMIN_SECRET: {
```
roseeichelmann, line 7: i think you need to surround `HASURA_ADMIN_SECRET` in quotes as well for this to work

moped-toolbox/backfill_complete_srts_ids/backfill_complete_srts_ids.py
```diff
@@ -0,0 +1,229 @@
...
+
+def convert_to_utc_timestamp(date_str):
+    date_format = "%m/%d/%y"
+    naive_datetime = datetime.strptime(date_str, date_format)
+    # Set the time to midnight and convert to UTC
+    naive_datetime = naive_datetime.replace(hour=0, minute=0, second=0, microsecond=0)
+    utc_datetime = naive_datetime.astimezone(timezone.utc)
+    # Convert to ISO 8601 format
+    utc_timestamp = utc_datetime.isoformat()
+    return utc_timestamp
```
johnclary, line 56: @mddilley — i am just eyeballing this, but this appears to be a timezone bug if the input `date_str` is in US/Central time. 

this has fortunately gotten much easier with the new `zoneinfo` 
 library. this example might be helpful of how we do this for the EMS import: https://github.com/cityofaustin/vision-zero/blob/a8c636e4be6b8767b93e27a4616a661becefc841/etl/afd_ems_import/incident_import.py#L250-L252

sorry if i'm wrong about that being bug! just wanted to get you that quick feedback in case.

roseeichelmann: could you explain more what part of this code would make this a bug?

roseeichelmann (approved):
tested with production data, this seems to be working great to me! curious about what john pointed out about a potential timezone bug though

moped-toolbox/backfill_complete_srts_ids/.gitignore
```diff
@@ -0,0 +1 @@
+/data/*
```
chiaberry, line 1: wanna include `secrets.py` here jussssst in case?

moped-toolbox/backfill_complete_srts_ids/backfill_complete_srts_ids.py
johnclary: @roseeichelmann when i run this code for "1/1/25", the ISO timestamp is '2025-01-01T05:00:00+00:00', which is 11pm on 12/31/24 US/Central time.

This is a tricky gotcha with `astimezone` assuming the naive timestamp you provide it is in the local timezone of the machine running the code.

moped-toolbox/backfill_complete_srts_ids/.gitignore
chiaberry: oh nevermind, i see its covered by the .gitignore one level up

johnclary:
[like]  Clary, John reacted to your message:
________________________________

frankhereford (changes requested):
Thanks for this Mike. My request changes is for the missing quotes in the secrets template. Rose observed it as well -- I think the fix is worth it to do before it gets incorporated. 

As for the `/data/` comment -- not as important, but worth thinking about. 

Finally - I looked at the function that John observed with the timezones, and I think what he says makes a lot of sense -- thank you John for that observation!

frankhereford (approved):
That was fast - thanks Mike!

moped-toolbox/backfill_complete_srts_ids/.gitignore
mddilley: i went ahead and changed this to be careful - i definitely had a short moment of panic when I push this branch yesterday 😬 😅

moped-toolbox/backfill_complete_srts_ids/README.md
```diff
@@ -0,0 +1,32 @@
...
+with existing substantial completion dates are excluded.
+
+## Running locally
+
+1. Start the Hasura cluster with production data
+
+1. Use `secrets_template.py` as a template to create `secrets.py`
+
+1. Export https://docs.google.com/spreadsheets/d/1M2dwi-GqheZLmy-GJwmrf6LAJpu07aoiD3aTEOo7IUQ/edit?gid=1500102334#gid=1500102334
+   as a csv file and place in the /data/ directory
```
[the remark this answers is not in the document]
mddilley: awesome - i made this change since it did cause confusion. thanks!

moped-toolbox/backfill_complete_srts_ids/backfill_complete_srts_ids.py
```diff
@@ -0,0 +1,229 @@
...
+import json
+import requests
+from datetime import datetime, timezone
+from time import sleep
+from secrets import HASURA
+
+csv_filename = "ATSD Moped Issue Tracking - SRTS Complete IDs.csv"
+
+# Query for project components AGOL view with SRTS ID. This query captures components that
+# have not been substantially completed yet and match a givent Safe Routes to School ID.
```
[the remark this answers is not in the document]
mddilley: 🙏 ✅

moped-toolbox/backfill_complete_srts_ids/secrets_template.py
mddilley: thanks y'all - fixed! 🛠️

moped-toolbox/backfill_complete_srts_ids/backfill_complete_srts_ids.py
mddilley: thanks for that test case as I was struggling to wrap my head around this yesterday. 🙏 

I was able to replicate the bug using John's example. I just updated the code using the EMS script as inspiration and tested again. There are some test cases in the source data that can be explored with:

```sql
-- components that have a completion date of 1/1/18 added to them
SELECT * FROM moped_proj_components WHERE srts_id = '4F - 627';
```

chiaberry (approved):
Time is confusing and daylight savings too. 

Anyway, ran through tests, I like the option for -v. 🚢

PR #1567, opened 2025-03-20
18175 migrate dashboard tables to datagridpro

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/18175

## Testing
**URL to test:** 

https://deploy-preview-1567--atd-moped-main.netlify.app/

**Steps to test:**
1) make sure you are following and assigned to at least two or three projects and visit the dashboard
2) test the 'my projects' and 'following' tabs by clicking the full name to make sure the link directs you to the correct project, clicking the status field to update the current status, adding a status update, and adding or setting milestones as completed
3) for each change, the view should automatically refetch the updated field
4) test the sorting on each column to make sure it sorts as expected, pinning columns, filtering, and hiding and managing columns

---
#### Ship list
- [x] Code reviewed 
   - Updated existing test - "Look at the milestones % complete for a project on the dashboard. Add a milestone to the project, set it to complete, and revisit dashboard to ensure the % in the milestone column changes. You should see the edited cell highlighted in the table."


moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -125,47 +211,66 @@ const DashboardView = () => {
...
+      setRows(selectedData);
+    }
+  }, [data, activeTab]);
+
+  const handleRowModesModelChange = (newRowModesModel) => {
+    setRowModesModel(newRowModesModel);
+  };
 
   /** Build custom user greeting
    */
```
chiaberry, line 276: What do y'all think about pulling this custom user greeting code outside of the component definition? its currently being recreated on every render I believe

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -125,47 +211,66 @@ const DashboardView = () => {
...
+    if (TABS[activeTab].label === "Following" && !!data) {
+      selectedData = data.moped_user_followed_projects;
+    } else if (TABS[activeTab].label === "My projects" && !!data) {
+      selectedData = data.moped_proj_personnel;
+    }
+
+    if (selectedData) {
       /**
-       * Get percentage of milestones completed
+       * Build data needed in Dashboard Material Table
```
chiaberry, line 232: Is there a way to memoize this, or just build once and not every time a tab is clicked? When I use prod data locally, moped thinks I'm Ivonne and as a result I have a ton of records in the My Projects tab. Clicking back and forth between the tabs, it feels a tiny bit sluggish.

chiaberry (commented):
Congrats on getting rid of what I think is the last table that is using material table! I tested and functionality works great, I did leave some comments about possible performance gains, wondering if you think they are feasible?

moped-editor/src/views/dashboard/DashboardView.js
tillyw: this is definitely doable and worth it in my opinion! i'll think about how to do this as simply as possible, although if it's more complex it might be best to spin it off into a separate issue as an enhancement for the subsequent release

moped-editor/src/views/dashboard/DashboardView.js
tillyw: i think this is a great idea!

moped-editor/src/views/dashboard/DashboardView.js
chiaberry: I was thinking about this, and how its not a regression since its currently behaving like this in prod. this is an improvement because of the useEffect, so spinoff makes sense so we arent rushed

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -100,18 +99,105 @@ const TABS = [
...
+            projectName={row.project_name_full}
+            currentPhaseId={row.current_phase_id}
+            modalParent="dashboard"
+            statusUpdate={row.status_update}
+            queryRefetch={refetch}
+            handleSnackbar={handleSnackbar}
+            classes={classes}
+          >
+            {parse(String(row.status_update))}
+          </DashboardStatusModal>
```
frankhereford, line 168: I'm not sure if you want to consider this `DashboardStatusModal` component in scope or not, but there are couple of small layout things that maybe should be worked on if it is. 

* If the status is longer than two lines, there is not a clear indication that there's more content that's hidden. If this state can be detected, then adding an elipsese may make that more obvious -- or of course, the expanding the rows height to include all the content. 
* The `⊕` character is rendering at the very top of the row and may benefit from being vertically aligned in the center.
* Similar for the Milestone gauge -- it is also touching the top of the row.

image

In this picture, the bulleted list of cat colors / patterns has a 3rd and 4th entry.

frankhereford (approved):
Thanks for this Tilly -- I've gone over your code and the site on the netlify deploy, and I think you've checked all the boxes on this issue and done it really well. I have an observation about some rendering things that I am not sure if you consider in scope or not -- but they are minor. This looks great.

moped-editor/src/views/dashboard/DashboardView.js
tillyw: i think these would be valuable improvements! in the interest of shipping this update, i'm going to add these to the follow-up issue @chiaberry and i discussed!

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -125,170 +241,89 @@ const DashboardView = () => {
...
-      case curHr >= 12 && curHr < 18:
-        return "afternoon";
-      default:
-        return "evening";
+        // project status update equivalent to most recent project note
+        // html is parsed before being rendered in the DashboardStatusModal component
+        project["status_update"] = "";
+        if (project?.project?.moped_proj_notes?.length) {
+          const note = project.project.moped_proj_notes[0]["project_note"];
+          project["status_update"] = note ? note : "";
```
johnclary, line 292: i'm struggling a bit with this block. is note ever null/undefined, and why is an empty string preferred?

johnclary (commented):
The grid is working really well! I flagged a few potential concerns, a few of which may be picking nits 🙏

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -28,15 +27,19 @@ import FeedbackSnackbar, {
...
 
 import { DASHBOARD_QUERY } from "../../queries/dashboard";
 
 import { getSessionDatabaseData } from "../../auth/user";
 
 import parse from "html-react-parser";
 
+const userSessionData = getSessionDatabaseData();
+const userId = userSessionData?.user_id;
+const userName = userSessionData?.first_name;
```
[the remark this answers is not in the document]
tillyw: good call, i'm moving this back into the component just to be safe!

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -100,19 +102,142 @@ const TABS = [
...
+          <Link
+            href={`/moped/projects/${row.project.project_id}`}
+            sx={{
+              overflow: "hidden",
+              textOverflow: "ellipsis",
+              display: "block",
+            }}
+          >
+            {row.project.project_name_full}
+          </Link>
```
mddilley, line 149: There is a small regression here since the MUI Link is using browser redirect instead of client-side navigation with React Router. Could you update the `Link` component to use the `component` and `to` props like in [this project list view column](https://github.com/cityofaustin/atd-moped/blob/c79b0ce8476e2720e985e9ecca521f753851e76f/moped-editor/src/views/projects/projectsListView/helpers.js#L162-L163)? 🙏

mddilley (changes requested):
So much code clean up ✨ - thanks for addressing the suggested improvements while you were in here. I requested a few changes to polish before launch, but this is working just as it should.

moped-editor/src/components/DataGridPro/utils/helpers.js
mddilley, line 1: could you move these helpers to [dateAndTime.js](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/utils/dateAndTime.js) in the interest of making them more discoverable in the future? Thanks for pulling these out of this component btw!

tillyw: yes, that is a better place for them, thank you!

moped-editor/src/views/dashboard/DashboardView.js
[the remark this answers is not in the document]
tillyw: thanks for pointing that out!

tillyw:
i've made some enhancements and reworked the vertical spacing of the column content and it's ready for rereview!

moped-editor/src/components/DataGridPro/utils/helpers.js
```diff
@@ -12,3 +14,25 @@ export const getLookupValueByID = (lookupTable, attribute, id) => {
     `${attribute}_name`
   ];
 };
+
+/**
+ * Get the time of day to display to users on the dashboard
+ * @returns {string} - Time of day phrased as a salutation
+ */
+export const getTimeOfDay = (date) => {
+  const curHr = format(date, "HH");
```
johnclary, line 23: i believe you can do this with plane ole javascript: `date.getHours()`

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -249,52 +258,13 @@ const DashboardView = () => {
...
-
-        // project status update equivalent to most recent project note
-        // html is parsed before being rendered in the DashboardStatusModal component
-        project["status_update"] = "";
-        if (project?.project?.moped_proj_notes?.length) {
-          const note = project.project.moped_proj_notes[0]["project_note"];
-          project["status_update"] = note ? note : "";
-        }
-      });
-    }
```
johnclary: nice!

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -315,32 +320,22 @@ const DashboardView = () => {
...
-                          paging: false,
-                        }),
-                        pageSize: 50,
-                        pageSizeOptions: [10, 50, 100],
-                        idSynonym: "project_id",
+                    <DataGridPro
+                      sx={dataGridProStyleOverrides}
+                      columns={dataGridColumns}
+                      rows={rows}
+                      autoHeight
```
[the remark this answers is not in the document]
johnclary: I assume this is related—just wanted to mention the note text is not vertically centered in the same way as other tables

Screenshot 2025-04-21 at 11 36 57 AM

johnclary (commented):
Looking good, Tilly! Thanks for making this updates.

One issue I'm noticing is that sorting does not seem to have any affect. I am testing with the **Followed** projects

Screenshot 2025-04-21 at 11 41 11 AM

mddilley (changes requested):
I requested a few smaller changes, and I'd like the sorting to work as well. I think that Rose and Chia are good resources since they've worked with sorting in other tables recently. I'm happy to rubber duck on getting the sorting working as well!

tillyw:
i fixed the sorting issue! it was due to the rows containing nested project data, so i mapped out the nested project data before setting the rows in DataGridPro.

note: i disabled sorting for the 'status update' column because those fields contain nested html tags that get pretty complicated, and i didn't see a use case for sorting by most recent note. happy to reconsider this if y'all think it would be helpful to have!

chiaberry (approved):
all the updates look good and the DashboardView code is in a much better place now than what's in prod. (sounds like I think its dead, but you know what i really mean)

I agree that I dont think its necessary to be able to sort the project status update alphabetically. Sorting it on timestamp of when last updated could be a helpful future enhancement, but we'd need to surface the timestamp in this table too.

When I sorted by project name, I noticed that project 17 and project 3601 both have leading empty spaces in their project name. @mddilley do you think we should clean up project names leading spaces? dont know if its affecting sorting on agol too

moped-editor/src/views/dashboard/DashboardView.js
```diff
@@ -100,19 +106,154 @@ const TABS = [
...
+            </Link>
+          </div>
+        ),
+        flex: 4,
+      },
+      {
+        headerName: "Status",
+        field: "moped_proj_phases",
+        editable: false,
+        valueGetter: (value) => value[0]?.moped_phase.phase_name,
```
roseeichelmann, line 166: "Unknown" technically isnt sorting alphabetically correctly here, this is how i handled this in a diff pr i worked on migrating the subprojects table to datagrid pro!  https://github.com/cityofaustin/atd-moped/commit/175e779ddef40ae6546fb760a22d88b076b20094

roseeichelmann (approved):
this looks great tilly!! so exciting to be done with the datagridpro migration!! its been a long time coming. thanks for getting us over the finish line! other than the one line fix i mentioned in a comment to make sure the status sorts alphabetically correctly, this is good to go!

moped-editor/src/views/dashboard/DashboardView.js
tillyw: thanks for catching that!! i've implemented your fix :)

mddilley (approved):
Looks amazing! Thanks for adding these final touches. 🙏 Can't wait to have this final table out in the wild! 🚀 🚢  🙌

Forgot to mention that a hidden improvement in this code is that it is now much easier to see which status or milestone was updated after editing in the modals since DataGrid Pro highlights cells. This has been a note in the QA spreadsheet for a while that we get for free. 🙌
 

https://github.com/user-attachments/assets/666e3454-576f-4ed0-9385-91799ad0f0dd

tillyw:
> Looks amazing! Thanks for adding these final touches. 🙏 Can't wait to have this final table out in the wild! 🚀 🚢 🙌
> 
> Forgot to mention that a hidden improvement in this code is that it is now much easier to see which status or milestone was updated after editing in the modals since DataGrid Pro highlights cells. This has been a note in the QA spreadsheet for a while that we get for free. 🙌
> 
>  Screen.Recording.2025-04-23.at.5.29.33.PM.mov

yesss i can see how that would be really helpful for keeping track of which project you're editing!

mddilley:
Merging this one to continue testing. Thanks @tillyw for helping test this fix! 🙏

For everyone else, see [this thread](https://austininnovation.slack.com/archives/CNUEPKLB1/p1745513893713209)

PR #1570, opened 2025-03-25
Fix sorting in team, milestone, and funding tables

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21366

The problem is that in any column where we are using a function to render the cell without actually changing the internal value of the row (such as `valueFormatter`, `renderCell`), the datagrid internal sort is trying to sort on the fields internal value, not what is rendering in the cell. Oftentimes that value is an Object so the sorting just doesnt work, or in cases like the funding status, it was sorting the under the hood value, not what users see in the column on the front end. 

This means we either need to use a custom sort function on these columns or use `valueGetter` instead which actually sets the value, but in order to implement that we would have to refactor a lot, like the `LookupAutocompleteComponent` and `LookupSelectComponent` components, and even then im not 100% how feasible that is. So for the purpose of fixing these bugs already i opted for the straightforward solution using the custom sorts.

## Testing
**URL to test:** 

[Netlify](https://deploy-preview-1570--atd-moped-main.netlify.app/moped/session/signin)

**Steps to test:**
1. Test out all the sorting in the Teams, Funding, and Milestones tables.
2. Also test adding new rows to these tables, editing, and deleting

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -142,6 +142,8 @@ const useColumns = ({
         width: 200,
         editable: true,
         valueFormatter: (value) => value?.funding_source_name,
+        sortComparator: (v1, v2) =>
+          v1.funding_source_name.localeCompare(v2.funding_source_name),
         renderEditCell: (props) => (
           <LookupAutocompleteComponent
             {...props}
@@ -157,6 +159,8 @@ const useColumns = ({
         width: 200,
         editable: true,
         valueFormatter: (value) => value?.funding_program_name,
+        sortComparator: (v1, v2) =>
+          v1.funding_program_name.localeCompare(v2.funding_program_name),
         renderEditCell: (props) => (
           <LookupAutocompleteComponent
             {...props}
@@ -178,11 +182,19 @@ const useColumns = ({
         field: "funding_status_id",
         editable: true,
         width: 200,
-        renderCell: ({ value }) =>
+        valueFormatter: (value) =>
           getLookupValueByID(
             data["moped_fund_status"],
             "funding_status",
             value
+          ),
+        sortComparator: (v1, v2) =>
+          getLookupValueByID(
+            data["moped_fund_status"],
+            "funding_status",
+            v1
+          ).localeCompare(
+            getLookupValueByID(data["moped_fund_status"], "funding_status", v2)
           ),
         renderEditCell: (props) => (
           <LookupSelectComponent
@@ -200,6 +212,10 @@ const useColumns = ({
         editable: true,
         valueFormatter: (value) =>
           !!value?.fund_name ? `${value?.fund_id} | ${value?.fund_name}` : "",
+        sortComparator: (v1, v2) =>
+          `${v1?.fund_id} | ${v1?.fund_name}`.localeCompare(
+            `${v2?.fund_id} | ${v2?.fund_name}`
+          ),
         renderEditCell: (props) => (
           <LookupAutocompleteComponent
             {...props}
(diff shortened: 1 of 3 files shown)


moped-editor/src/views/projects/projectView/ProjectMilestones.js
```diff
@@ -92,8 +94,8 @@ const useColumns = ({
         field: "moped_milestone_related_phase",
         editable: true, // this is to be able to use the renderEditCell option to update the related phase
         // during editing -- the input field is always disabled
-        renderCell: (props) =>
-          phaseNameLookup[props.row.moped_milestone?.related_phase_id] ?? "",
+        valueGetter: (value, row) =>
+          phaseNameLookup[row.moped_milestone?.related_phase_id] ?? "",
```
roseeichelmann, line 98: bc of the fact that this is a view only field and the `ViewOnlyTextField` below is based off the value of another field, `props.row.moped_milestone`, i was able to just update this to `valueGetter` which actually sets the value for the field and so sorting will just work without having to add a custom sort function

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
```diff
@@ -144,6 +150,7 @@ const useColumns = ({
         field: "moped_proj_personnel_roles",
         width: 200,
         editable: true,
+        sortable: false,
```
roseeichelmann, line 153: this is a cop out but with such a complicated renderCell function below, it just doesnt seem feasible to implement a custom sort here. do all columns really need to be sorted?

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -178,12 +182,20 @@ const useColumns = ({
...
             "funding_status",
             value
           ),
+        sortComparator: (v1, v2) =>
+          getLookupValueByID(
+            data["moped_fund_status"],
+            "funding_status",
+            v1
+          ).localeCompare(
+            getLookupValueByID(data["moped_fund_status"], "funding_status", v2)
```
roseeichelmann, line 197: the funding status wasnt actually sorting correctly, bc previously it was sorting by this fields `value` which is the `id` of the status, so it was numerically sorting the ids and not alphabetically sorting the lookup values users see on the front end

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
mddilley: not a cop out to me. thanks for stopping where it became complex. I am okay with this since the UI isn't showing a broken sort feature anymore. Thanks!

chiaberry (approved):
tested sorting on the tables and all the columns you updated work. I got thrown off at first because I forgot that the sort works by toggling between sort ascending, descending and then unsort. 

I completely forgot about this until I worked on that table, but in ProjectFiles we import `gridStringOrNumberComparator` from data grid to handle the custom sorting, but I think your localeCompare solution is more straightforward.

moped-editor/src/views/projects/projectView/ProjectTeam/ProjectTeamTable.js
tillyw: i think this is fair!

tillyw (approved):
tested all the columns and all the enabled ones work as expected! thanks for sorting this out xD

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -142,6 +142,8 @@ const useColumns = ({
         width: 200,
         editable: true,
         valueFormatter: (value) => value?.funding_source_name,
+        sortComparator: (v1, v2) =>
+          v1.funding_source_name.localeCompare(v2.funding_source_name),
```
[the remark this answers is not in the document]
roseeichelmann: honestly copied it from the [datagridpro docs](https://github.com/cityofaustin/vision-zero/pull/1742), i interpreted it as value 1 and value 2. but i like vampire too xD :vampire:

mddilley (approved):
Looks and functions great! Thanks for making these work - lots of TILs for me in here about sorting in these tables! 🚢  🚀

PR #1575, opened 2025-04-03
Add length validation to project status update input

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21263

This PR adds yup schema validation to status update entry to keep us under the length limit in the AGOL feature service. I also add some docs and updated some variable names and assignments as I was thinking through the code.

## Testing
**URL to test:** 

https://deploy-preview-1575--atd-moped-main.netlify.app/moped/ or local

**Steps to test:**
1. Find your favorite ipsum generator (https://pizzaipsum.com/, https://baconipsum.com/)
2. Go to a project summary and click into the Status update area to create a new status update
3. Enter a long update, and you should see a validation error message in red and the save button should be disabled when it becomes too long
4. Change the text to be very short and save to make sure you can still create a new status update
5. Now go to the **Notes** tab and try creating a really long internal note. You should see no error message and the save button should never disable
6. Now, use the radio toggle to switch to **Status update** note type
7. You should see an immediate validation error and the save button disable. You can then make the note longer or shorter to see the validation state change
8. Save a valid status update and see it display below the rich text editor
9. Click the pencil to edit your status update, and test the same steps of changes the length and type. It should behave exactly as in the other steps.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/utils/validation.js
+++ b/moped-editor/src/utils/validation.js
@@ -1,7 +1,7 @@
 /**
  * Wrapper function for Yup.validateSync that returns formatted errors
  * See https://github.com/jquense/yup?tab=readme-ov-file#schemavalidatesyncvalue-any-options-object-infertypeschema
- * @param {object} value - Key-value pairs of the field to validate
+ * @param {object} value - Key-value pairs of the fields to validate
  * @param {object} yupValidationSchema - The Yup validation schema to use
  * @returns {object|null} - Returns an object with errors or null if no errors
  */
--- a/moped-editor/src/views/projects/projectView/NoteInput.js
+++ b/moped-editor/src/views/projects/projectView/NoteInput.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from "react";
+import React, { useEffect, useState } from "react";
 import {
   Box,
   Button,
@@ -12,7 +12,7 @@ import {
   FormHelperText,
 } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
-import ProjectSaveButton from "../newProjectView/ProjectSaveButton";
+import ProjectSaveButton from "src/views/projects/newProjectView/ProjectSaveButton";
 import ToolbarPlugin from "./ToolbarPlugin";
 
 import { AutoFocusPlugin } from "@lexical/react/LexicalAutoFocusPlugin";
@@ -29,7 +29,7 @@ import { LinkPlugin } from "@lexical/react/LexicalLinkPlugin";
 import { LinkNode } from "@lexical/link";
 import { ListPlugin } from "@lexical/react/LexicalListPlugin";
 import { ListNode, ListItemNode } from "@lexical/list";
-import EditorTheme from "./EditorTheme";
+import EditorTheme from "src/views/projects/projectView/EditorTheme.js";
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -213,22 +213,17 @@ const NoteInput = ({
 
   const [validationErrors, setValidationErrors] = useState(null);
 
-  const validateContent = useCallback(
-    (htmlContent) => {
-      const errors = validator ? validator(htmlContent) : null;
-      if (errors) {
-        setValidationErrors(errors);
-      } else {
-        setValidationErrors(null);
-      }
-    },
-    [validator]
-  );
-
   // Validate content when the note text or note type changes
   useEffect(() => {
-    validateContent(noteText);
-  }, [noteText, newNoteType, editingNoteType, validateContent]);
+    const errors = validator
+      ? validator({ projectStatusUpdate: noteText })
+      : null;
+    if (errors) {
+      setValidationErrors(errors);
+    } else {
+      setValidationErrors(null);
+    }
+  }, [noteText, newNoteType, editingNoteType, validator]);
 
   const onChange = (htmlContent) => {
     setNoteText(htmlContent);
--- a/moped-editor/src/views/projects/projectView/NoteInput.js
+++ b/moped-editor/src/views/projects/projectView/NoteInput.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import {
   Box,
   Button,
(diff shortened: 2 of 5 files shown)


moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -173,23 +175,57 @@ const initialConfig = {
...
+    if (validator === null) return;
+
+    const errors = validator({ projectStatusUpdate: noteText });
+
+    if (errors) {
+      setValidationErrors(errors);
+    } else {
+      setValidationErrors(null);
+    }
+  }, [noteText, newNoteType, editingNoteType, validator]);
```
mddilley, line 227: I thought about handling this in the `onChange` handlers, but I felt like this was more centralized and would push code further down into the Lexical plugins.

moped-editor/src/views/projects/projectView/NoteInput.js
```diff
@@ -173,23 +175,57 @@ const initialConfig = {
...
+    const errors = validator
+      ? validator({ projectStatusUpdate: noteText })
+      : null;
+
+    if (errors) {
+      setValidationErrors(errors);
+    } else {
+      setValidationErrors(null);
+    }
+  }, [noteText, newNoteType, editingNoteType, validator]);
```
mddilley, line 227: I thought about handling this in the onChange handlers, but I felt like this was more centralized and would push code further down into the Lexical plugins.

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -329,21 +355,22 @@ const ProjectNotes = (props) => {
...
                 setNewNoteType={setNewNoteType}
-                editingNote={editingNote}
+                isEditingNote={isEditingNote}
                 noteAddLoading={noteAddLoading}
                 noteAddSuccess={noteAddSuccess}
                 submitNewNote={submitNewNote}
                 submitEditNote={submitEditNote}
                 cancelNoteEdit={cancelNoteEdit}
                 isStatusEditModal={isStatusEditModal}
+                validator={isStatusUpdate ? validator : null}
```
mddilley, line 373: This felt a little awkward, but the validation needs to only apply when entering status updates. The NoteInput only knows enough to use the validator function if there is one passed in this prop.

moped-editor/src/utils/validation.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Wrapper function for Yup.validateSync that returns formatted errors
+ * See https://github.com/jquense/yup?tab=readme-ov-file#schemavalidatesyncvalue-any-options-object-infertypeschema
+ * @param {object} value - Key-value pairs of the fields to validate
+ * @param {object} yupValidationSchema - The Yup validation schema to use
+ * @returns {object|null} - Returns an object with errors or null if no errors
+ */
+export const yupValidator = (value, yupValidationSchema) => {
```
mddilley, line 8: I added this helper in case we ever have a need for validation outside of React Hook Form. Not sure if that day will come or not.

moped-editor/src/views/projects/projectView/NoteInput.js
johnclary: cool—this works great. i noticed zero lagginess, which would be my only concern as well. `useEffect` works fine here, but wanted to mention that assigning `validationErrors ` from a `useMemo` hook might be ever so slightly more optimal, since you get the stable object reference for free and don't need to use setState. but it is probably a 100% wash performance-wise.

johnclary (approved):
I tried very hard to write a long status update, and I failed. It's awesome to see Yup doing it's thing in another part of the app. We love Yup!

This could be a snackoo that goes deep in the backlog, but I noticed that the validation error does not trigger the red input border like it does in other places. I figured this might be complicated by the Lexical component enough to leaving it out of scope.

Screenshot 2025-04-04 at 2 56 02 PM

🚢

tillyw:
this works great! i did came across potential snag. under the notes tab, if i save a valid status update, click the edit icon, put in a really long note, and switch the toggle button to 'internal note,' should i be able to save that? right now it is not allowing me to.

Screenshot 2025-04-10 at 2 05 50 PM

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -112,7 +136,7 @@ const ProjectNotes = (props) => {
   const [editingNoteType, setEditingNoteType] = useState(null);
   const [noteAddLoading, setNoteAddLoading] = useState(false);
   const [noteAddSuccess, setNoteAddSuccess] = useState(false);
-  const [editingNote, setEditingNote] = useState(false);
+  const [isEditingNote, setIsEditingNote] = useState(false);
```
chiaberry, line 139: I independently came to this same edit in the branch I am working on! the state var names were confusing to me

chiaberry:
I was able to reproduce what Tilly described I figured out why its happening, its here [on line 149](https://github.com/cityofaustin/atd-moped/blob/mike/21263_status_update_validation/moped-editor/src/views/projects/projectView/ProjectNotes.js#L149) in the ProjectNotes. In order for the validation in the edit notes input to be null, `isStatusUpdate` has to be false which happens when *both* the `newNoteType` and the `editingNoteType` have to be internal note type. 

If you toggle the new note to be "internal note" and then go to edit and then follow Tilly's steps, the validation works as expected.

moped-editor/src/views/projects/projectView/NoteInput.js
mddilley: i left this one as-is, because, interestingly enough, useMemo warns about unnecessary extra dependencies since it could cause unneeded recalculations.

useEffect does not warn about unnecessary dependencies - only missing ones. This was a TIL for me!

chiaberry (approved):
bug fixed!! 🚢

mddilley:
Thanks y'all! Merging. 🚀

moped-editor/src/views/projects/projectView/NoteInput.js
johnclary: 👍 TIL for me as well!

PR #1578, opened 2025-04-11
Limit one of the same team member to a single project with DB constraing

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21292

Screenshot 2025-04-10 at 10 57 56 PM

## Testing
Local

**Steps to test:**

- Replicate a copy of prod data (It might make sense to do this on `main` and then switch to this branch so the new migration doesn't run automatically with the `replicate` command)
- Run this query provided by Mike:
```
SELECT
    project_id,
    user_id,
    COUNT(*) as duplicate_count
FROM
    moped_proj_personnel
WHERE is_deleted = FALSE   
GROUP BY
    project_id,
    user_id
HAVING
    COUNT(*) > 1;
```
- Note the project ids with duplicate moped_proj_personnel instances. In my case, 4 records in prod had duplicates of the same team member on a single project.
  - Bonus: Look at the project's Team tab and confirm the duplicate team member entries in the UI.
- Screenshot 2025-04-10 at 10 52 58 PM
- Run the migration in this PR with `hasura migrate apply`
- Rerun the query above not confirm that all the duplicates are removed. 
  - Bonus: Look at the record in the Moped Editor to see duplicate team member entry removed. 
- In a DB GUI or Hasura Console, try to insert a duplicate record (same project_id and user_id) to verify the unique constraint is being enforced.
  - ex:
```
  INSERT INTO moped_proj_personnel (
    project_id,
    user_id
) VALUES (
    2095,  -- project_id
    192
);
```
- Test the down migration runs successfully. 
  - Bonus: Try to insert a duplicate record (same project_id and user_id) to verify the constraint is gone after the down migration runs.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1744343940108_add_unique_constraint_to_moped_proj_personnel/up.sql
+++ b/moped-database/migrations/1744343940108_add_unique_constraint_to_moped_proj_personnel/up.sql
@@ -1,0 +1,70 @@
+-- Create a temporary table to store the roles from duplicate records
+CREATE TEMP TABLE temp_duplicate_roles AS
+WITH duplicates AS (
+    SELECT 
+        project_id,
+        user_id,
+        COUNT(*) as duplicate_count
+    FROM
+        moped_proj_personnel
+    WHERE is_deleted = FALSE   
+    GROUP BY
+        project_id,
+        user_id
+    HAVING
+        COUNT(*) > 1
+)
+SELECT 
+    p.project_personnel_id,
+    p.project_id,
+    p.user_id,
+    r.project_role_id
+FROM moped_proj_personnel p
+JOIN duplicates d ON p.project_id = d.project_id AND p.user_id = d.user_id
+JOIN moped_proj_personnel_roles r ON p.project_personnel_id = r.project_personnel_id
+WHERE p.is_deleted = FALSE;
+
+-- Keep the first record for each project/user pair and mark others as deleted
+WITH first_records AS (
+    SELECT DISTINCT ON (project_id, user_id)
+        project_personnel_id
+    FROM moped_proj_personnel
+    WHERE is_deleted = FALSE
+    ORDER BY project_id, user_id, project_personnel_id
+)
+UPDATE moped_proj_personnel
+SET is_deleted = TRUE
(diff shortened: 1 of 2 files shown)


moped-database/migrations/1744343940108_add_unique_constraint_to_moped_proj_personnel/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    AND pr.is_deleted = FALSE
+);
+
+-- Clean up temporary table
+DROP TABLE temp_duplicate_roles;
+
+-- Add unique constraint to prevent duplicate project personnel entries using a partial index
+CREATE UNIQUE INDEX moped_proj_personnel_project_id_user_id_key 
+ON moped_proj_personnel (project_id, user_id) 
+WHERE is_deleted = FALSE;
```
johnclary, line 70: neat— i did not know you can use a `where` clause in indexes like this. this so handy. TIL 🙏

johnclary (approved):
🚢

chiaberry (approved):
thank you for the steps to get bonus points! i have the bonus points 

both files are missing the blank line at EOF, but other than that they shine!

moped-database/migrations/1744343940108_add_unique_constraint_to_moped_proj_personnel/up.sql
mddilley: +1 - very cool to know!

moped-database/migrations/1744343940108_add_unique_constraint_to_moped_proj_personnel/up.sql
mddilley, line 1: thanks for the comments in here that explain each step. So valuable when we find ourselves going back to reference previous migrations! 🙏

mddilley (approved):
Work just as described. Thanks for the thorough test steps! I was able to: replicate the dupes, remove them, roll back the change, add more, and then remove them again.

There is some pretty rad SQL in here too! 😎 🚢

mddilley:
thanks Mateo! Merging this one to get closer to starting QA testing.

PR #1584, opened 2025-04-24
Change default DB name in env template

frankhereford:
We are renaming the production moped database when we transfer it today, and this PR reflects that in the environment template.

@mddilley & @chiaberry: I am so surprised that this is the extent of what I think needs to be renamed in the codebase. Do you all agree?

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/docker-compose.yml
+++ b/moped-database/docker-compose.yml
@@ -1,7 +1,7 @@
 # Architecture Independent Docker Stack Components for Moped Development Environment
 services:
     hasura:
-        image: hasura/graphql-engine:v2.4.0
+        image: hasura/graphql-engine:v2.47.0
         depends_on:
             - moped-pgsql
         expose:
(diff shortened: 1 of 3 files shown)


mddilley (approved):
I searched the repo, and I am also surprised. I cannot find any other references to the soon-to-be-outdated DB name! 🙏 🚢

chiaberry (commented):
Agreed this looks like the only place to rename the db. We need to update this line too, right?

https://github.com/cityofaustin/atd-moped/blob/main/moped-database/hasura-cluster#L30

frankhereford:
@chiaberry _wisely_ wrote:

> We need to update this line too, right?

oh snap great catch yes; that would have messed up replication

mddilley:
thanks @chiaberry! @frankhereford - i guess now that we will be on a shared RDS with common PG versions, then https://github.com/cityofaustin/atd-data-tech/issues/22091 applies to both repos, right? i missed that discussion in slack

PR #1588, opened 2025-04-29
Add/edit tags to track decomissioned components and add protection type subcomponents to Access Control components

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/22248, Closes https://github.com/cityofaustin/atd-data-tech/issues/22232

This PR makes changes to component tags to make it easier to mark and filter out decommissioned components and also to be able to add protection type subcomponents like flex posts to some of the Access Control component types.

## Testing
**URL to test:** 

Local 🙏

**Steps to test:**
1. Start up you local stack and then go to the data dictionary https://localhost:3000/moped/dev/lookups
2. Load the staging data dictionary to compare https://moped.austinmobility.io/moped/dev/lookups
3. Navigate to the component tags section and notice that the tag with type "Bikeways - Performance Measure" and name "Upgrades Existing" now only has name "Upgrades Existing Component"
4. Note that there is also a new component tag with type "Decommissioned / Upgraded Component"
5. Now, go to the components section of the dictionary and notice that these four components now list all protection type subcomponents as available:
   - Access Control - Median
   - Access Control - Driveway Closure
   - Access Control - Driveway Modification
   - Access Control - Hardened Centerline
6. Try adding these component tags and creating an Access Control component listed above with a protection type subcomponents

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
+++ b/moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
@@ -1,0 +1,48 @@
+-- Insert new tag to mark decommissioned components
+INSERT INTO moped_tags (name, type, slug) VALUES
+('', 'Decommissioned / Upgraded Component', 'decommissioned_upgraded_component');
+
+-- Update existing component tag name to include 'component' in name
+
+-- Associate protection type subcomponents with Access Control components
+WITH inserts_todo AS (
+    SELECT
+        work_type_id,
+        'Trail' AS component_name,
+        'Shared Use Path (Paved Dual Trail)' AS component_subtype
+    FROM
+        moped_component_work_types AS mcwt
+    LEFT JOIN moped_work_types AS mwt ON mcwt.work_type_id = mwt.id
+    LEFT JOIN moped_components AS mc ON mcwt.component_id = mc.component_id
+    WHERE
+        mc.component_subtype LIKE 'Shared Use Path (Paved)'
+)
+
+INSERT INTO moped_component_work_types (component_id, work_type_id)
+SELECT
+    mc.component_id,
+    inserts_todo.work_type_id
+FROM
+    inserts_todo
+    -- gets the component id of the new component we created
+LEFT JOIN moped_components AS mc ON inserts_todo.component_name = mc.component_name
+    AND inserts_todo.component_subtype = mc.component_subtype;
+
+-- Components
+-- Access Control - Median
+-- Access Control - Driveway Closure
+-- Access Control - Driveway Modification
+-- Access Control - Hardened Centerline
+-- Subcomponents
+-- Protection Type - Car free
+-- Protection Type - Concrete
+-- Protection Type - Concrete and on-street parking
+-- Protection Type - Concrete traffic bump
+-- Protection Type - Concrete with landscape
+-- Protection Type - Flex posts
+-- Protection Type - Flex posts (City Post)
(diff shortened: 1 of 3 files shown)


moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
```diff
@@ -0,0 +1,42 @@
+-- Insert new tag to mark decommissioned components
+INSERT INTO moped_component_tags (name, type, slug) VALUES
+('', 'Decommissioned / Upgraded Component', 'decommissioned_upgraded_component');
+
+-- Update existing component tag name to include 'component' in name
```
chiaberry, line 5: is this comment misplaced?

chiaberry (approved):
Test steps check out, and I was able to add components to the map with the new tags and subcomponents

tillyw (approved):
tested and was able to add the new component and subcomponent types!

moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
```diff
@@ -0,0 +1,44 @@
...
+        )
+)
+
+INSERT INTO
+moped_components_subcomponents (component_id, subcomponent_id)
+SELECT
+    inserts_todo.component_id,
+    inserts_todo.subcomponent_id
+FROM
+    inserts_todo;
```
johnclary, line 44: Looks good. I noticed we don't have a unique constraint on `(component_id, subcomponent_id)` on this table—what do you think about adding one?

moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
```diff
@@ -0,0 +1,44 @@
+-- Insert new tag to mark decommissioned components
+INSERT INTO moped_component_tags (name, type, slug) VALUES
+('', 'Decommissioned / Upgraded Component', 'decommissioned_upgraded_component');
```
johnclary, line 3: I know this is an existing pattern for this table, but inserting an empty feels very icky. If null values are acceptable, can we relax the `not null` constraint?

johnclary (approved):
Test steps check out ✅

moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
mddilley: great catch - thanks John! I will add in this PR. 🙏

moped-database/migrations/1745881623802_decom_and_new_subcomponents/up.sql
mddilley: this does indeed feel blah - thanks for spotting this! I will update this and some of the existing tags with empty strings while I'm at it (and makes sure they render as expected). 🧹✨🙏

mddilley: I ran into a snag here so I created https://github.com/cityofaustin/atd-data-tech/issues/22462 to address and fix a bug in the AGOL view where `null` names cause the tags to not list at all and empty strings are causing unwanted dashes. We'll get there!

johnclary: Complications! Thanks for following-up on that 👍

PR #1592, opened 2025-05-07
Add database constraint to prevent multiple current phases per project

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C21629

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Go to a project that has an existing current phase, add a new phase and set it as current. The current phase badge should switch to the new phase you added
2. Also test updating existing phases to be current
3. Delete a phase that is not current
4. In your hasura console or db manager test updating that deleted phase to be current, you should be able to do this without error. This is just testing to make sure our partial index is doing what we want
5. Test the down migration

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/queries/project.js
```diff
@@ -401,6 +406,12 @@ export const UPDATE_PROJECT_PHASE = gql`
     $object: moped_proj_phases_set_input!
```
roseeichelmann, line 406: sidenote but i cant find `UPDATE_PROJECT_PHASE` to be used anywhere in our codebase, do we still need to keep it around?

moped-editor/src/queries/project.js
```diff
@@ -337,11 +337,22 @@ export const TIMELINE_QUERY = gql`
...
+ * For all the following project phase mutations, the mutation to set
+ * current phases to false must run first or unique constraint could be violated
+ */
+
 export const ADD_PROJECT_PHASE = gql`
   mutation AddProjectPhase(
     $objects: [moped_proj_phases_insert_input!]!
     $current_phase_ids_to_clear: [Int!] = []
   ) {
+    update_moped_proj_phases(
```
roseeichelmann, line 350: moving this operation to the top ensures that we are first updating the phase that is already current to not be current before we update the new phase to be current, so that we dont violate the unique index

roseeichelmann:
just realize that this migration is not working w production data, so it looks like there is a violating... looking into it now

moped-database/migrations/1746640451690_run_sql_migration/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Ensures that projects can only have one current phase
+REATE UNIQUE INDEX moped_proj_phases_partial_constraint ON moped_proj_phases (project_id)
+    WHERE (is_current_phase = TRUE AND is_deleted = FALSE);
```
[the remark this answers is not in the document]
mddilley: Could you add a comment on this partial index to capture the business logic of it? I can see index comments when using TablePlus so hopefully that is true of other GUIs and would be helpful to our future selves or others.

Something like "Ensures each project can only have one active current phase"? 🙏

moped-editor/src/queries/project.js
mddilley: glad this worked out - thanks for verifying! 🙌

moped-editor/src/queries/project.js
mddilley: i'm cool with removing since it probably took you time to figure out if it was used or not. It would be nice to reduce confusion, and I don't see our phase mutation needs changing soon. thanks for catching!

mddilley (approved):
I tested this over and over, tried and failed violating the constraint in the database manually, and even checked the previous mutation order to see how any errors might show in the UI. All looks great!

I requested one comment on the new index, but this is 🚢! Thanks for helping keep our AGOL process stable. 🚀

tillyw:
tested all these steps locally, and i was able to update `is_current_phase` of a recently deleted phase to `TRUE` manually in TablePlus without issue!

tillyw (approved):

moped-editor/src/queries/project.js
chiaberry: I agree with removing it if we arent using it anywhere

moped-database/migrations/1746640451690_run_sql_migration/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Ensures that projects can only have one current phase
+CREATE UNIQUE INDEX moped_proj_phases_partial_constraint ON moped_proj_phases (project_id)
+    WHERE (is_current_phase = TRUE AND is_deleted = FALSE);
+
```
chiaberry, line 4: it thinks there isnt a blank line at the end of the file because this line has empty spaces in it. I ran it through a linter and it removed the blank lines

moped-database/migrations/1746640451690_run_sql_migration/down.sql
chiaberry, line 1: my personal preference is for the migration folders to be named based on whats in them like 1746640451690_prevent_multiple_current_phases, but obviously this works fine as is

chiaberry (approved):
Thanks for getting this done before today's release! I left a few comments about the formatting of the migrations, nothing related to execution.

moped-database/migrations/1746640451690_run_sql_migration/down.sql
roseeichelmann: oops i always update the names if i generate them thru the hasura console but i totally forgot to this time, thanks for pointing that out!

johnclary (approved):
Good stuff! 🔒 🚢

PR #1595, opened 2025-05-09
adds ability to save simple search as view

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/22452

## Testing
**URL to test:** 

https://deploy-preview-1595--atd-moped-main.netlify.app/

**Steps to test:**
- go to project list view and execute a simple search
- the "save view" button should appear, click on it and note the description field is autopopulated with the search term
- save the view and confirm it appears in the saved views table and the link works
- go to project list view and execute a simple search, then add a filter (or vice versa)
- the "save view" button should appear, click on it and note the description field is autopopulated with the search term plus the filters applied
- save the view and confirm it appears in the saved views table and the link works

try out different combinations, adding a simple search and filters in different orders, removing one or the other, and saving the view. all combinations and orders should work!

---
#### Ship list
- [x] Code reviewed 
   - Added new test, "Now, filter using only the simple search and make sure the "Save view" button appears so you can save a simple search term with or without advanced filters."

### DIFF
--- a/moped-editor/src/components/GridTable/Search.js
+++ b/moped-editor/src/components/GridTable/Search.js
@@ -184,6 +184,7 @@ const Search = ({
                 filtersConfig={filtersConfig}
                 resetSimpleSearch={resetSimpleSearch}
                 setIsOr={setIsOr}
+                searchTerm={searchTerm}
                 setSearchParams={setSearchParams}
                 handleSnackbar={handleSnackbar}
               />
--- a/moped-editor/src/components/GridTable/SearchBar.js
+++ b/moped-editor/src/components/GridTable/SearchBar.js
@@ -69,7 +69,8 @@ const SearchBar = ({
   resetSimpleSearch,
   setSearchParams,
   setIsOr,
-  handleSnackbar
+  handleSnackbar,
+  searchTerm,
 }) => {
   const classes = useStyles();
 
@@ -165,7 +166,7 @@ const SearchBar = ({
         variant="outlined"
         value={searchFieldValue}
       />
-      {filterStateActive && !advancedSearchAnchor && (
+      {(filterStateActive || searchTerm) && !advancedSearchAnchor && (
         <FiltersChips
           filters={filters}
           setFilters={setFilters}
@@ -174,6 +175,7 @@ const SearchBar = ({
           isOr={isOr}
           setSearchParams={setSearchParams}
           handleSnackbar={handleSnackbar}
+          searchTerm={searchTerm}
         />
       )}
     </>
--- a/moped-editor/src/views/projects/projectsListView/components/SaveUserViewModal.js
+++ b/moped-editor/src/views/projects/projectsListView/components/SaveUserViewModal.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useMemo } from "react";
 import { useMutation } from "@apollo/client";
 import { Dialog, DialogTitle, DialogContent, IconButton } from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
@@ -19,7 +19,7 @@ const SaveUserViewModal = ({
 
   let { pathname, search } = useLocation();
 
-  const getDefaultDescription = () => {
+  const defaultDescription = useMemo(() => {
     const filtersDescription = filtersLabels
       .map(
         (filter) =>
@@ -38,7 +38,7 @@ const SaveUserViewModal = ({
     } else {
       return "";
     }
-  };
+  }, [filtersLabels, searchTerm]);
 
   const onClose = () => {
     setIsSaveViewModalOpen(false);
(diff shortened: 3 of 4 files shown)


moped-editor/src/components/GridTable/SearchBar.js
```diff
@@ -165,7 +166,7 @@ const SearchBar = ({
         variant="outlined"
         value={searchFieldValue}
       />
-      {filterStateActive && !advancedSearchAnchor && (
+      {(filterStateActive || searchTerm) && !advancedSearchAnchor && (
```
tillyw, line 169: this works fine since there will be no filters to display if the filter state is not active. it may be worth considering breaking out the 'save view' button into its own component separate from the `<FiltersChips>` but I wasn't sure if that would fit the scope of this issue

moped-editor/src/views/projects/projectsListView/components/SaveUserViewModal.js
```diff
@@ -13,17 +13,32 @@ const SaveUserViewModal = ({
...
 
   let { pathname, search } = useLocation();
 
-  const defaultDescription = filtersLabels
-    .map(
-      (filter) =>
-        `${filter.filterLabel} ${filter.operatorLabel} ${filter.filterValue}`
-    )
-    .join(", ");
+  const getDefaultDescription = () => {
```
tillyw, line 22: sets the default description based on what the user has selected

mddilley: @tillyw I think this could be a good place to use `useMemo`.

Maybe something like:
```js
const defaultDescription = useMemo(() => {
// all the same default name prep in your function
}, [the dependencies here])
```

frankhereford (approved):
This looks great Tilly - thank you for this. I have reviewed your code, as well as tested the behavior of the button, modal and search bar and how they interplay with each other. I think this is ready to ship. Thanks again!

roseeichelmann:
image

This is the description i got in the dashboard when i typed in "test" as a simple search and saved "test" as the description in the modal?

chiaberry (commented):
I found an edge case where if I put something in the simple search, save the view and then select the search term and enter a new search term, the "save view" button is still disabled. In other words, I searched a new term without fully deleting the first one, just replacing it.

mddilley (changes requested):
This is working well, and it is nice to be able to save the simple search on its own! I requested a couple of small changes.

johnclary:
> I found an edge case where if I put something in the simple search, save the view and then select the search term and enter a new search term, the "save view" button is still disabled.

The save view button is re-enabled for me once I modify the search term and press the **Enter** key, which seems consistent since one must press **Enter** in order to initiate text search anyway 🤔

moped-editor/src/components/GridTable/Search.js
```diff
@@ -184,6 +184,7 @@ const Search = ({
                 filtersConfig={filtersConfig}
                 resetSimpleSearch={resetSimpleSearch}
                 setIsOr={setIsOr}
+                searchTerm={searchTerm}
```
johnclary, line 187: I noticed that leading/trailing whitespace are preserved in the search term. Might be nice to trim that when the user executes the search?

johnclary (approved):
Looks + works great. Thank you for enabling saving of simple searches!

🚢 🚢 🚢 🚢 

Not related to changes in this PR, I wanted to mention the tiniest bit of awkward behavior when modifying the search input at the same time as modifying the advanced filters.

If I do a simple search, then modify my search input, then add an advanced filter and press **Apply**, the modified search input is not included in the search, and it's not included when I save the view.

I also noticed that If I start modifying the advanced search filters and then, with the filter menu still open, I modify the simple search input, the filter menu closes and i lose those changes. It would be nice if you could use the simple search and advanced filters in tandem and expect the **Apply** button to apply the updated search input and the updated advanced filters.

chiaberry (approved):
looks great! thanks for fixing the edge case

moped-editor/src/components/GridTable/Search.js
mddilley: @tillyw could you sneak this change to trim in before merge? i noticed that the extra whitespace can make its way into the saved view descriptions as well 🙏

Screenshot 2025-05-21 at 11 50 42 AM

mddilley (approved):
Looks and works so great! Thanks for updating to re-enable the button! 🚢 🚀

mddilley:
@mateoclarke I've also been seeing that error on branches that don't have your metadata update in them yet since the `./hasura-cluster setenv local` command was required to have a `config.yml` at all. Now that you git-tracked it we should see this error gone completely! 🙏

I'm also glad to see you expecting that behavior for the saved view button. Tilly had the same feedback, and I'm expecting to hear that from users if we make the saved views more and more accessible like in the ACC designs. I think it would be a really nice improvement! https://github.com/cityofaustin/atd-data-tech/issues/22411

PR #1598, opened 2025-05-20
Update AGOL DB view and component tag table to handle missing (null) names

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/22462

This PR fixes a bug where the calculation of the full component tag names in the `component_arcgis_online_view` DB view wasn't handling null values of component names or types as expected. It also makes both name and type nullable and adds a generated column called `full_name` to handle nulls in those columns.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack from a production snapshot
1. Explore the `moped_component_tags` table and notice that the `name` and `type` columns are now nullable and that some names are null instead of empty strings like in production
1. Also, see there is a new generated column called `full_name` that handles null names and types and also concatenates them with ` - ` if they are both not null
1. See that the column called `component_tags` in `component_arcgis_online_view` is now defined by selecting the new generated column instead of calculating in the view
1. Now, start the local Moped app and test viewing, adding, and removing component tags in the Map tab in the project view. You should see that all component dropdown options and chips display the name and type joined by ` - ` or just the type if there is no name like `MAP Signals`

---
#### Ship list
- [x] Code reviewed 
   - Added new data dictionary test, "Scan the tables to make sure all data is populated"


moped-editor/src/queries/tableLookups.js
```diff
@@ -67,9 +67,8 @@ export const TABLE_LOOKUPS_QUERY = gql`
       order_by: { type: asc, name: asc }
     ) {
       id
-      type
-      name
       slug
+      full_name
```
johnclary, line 71: The `type` and `name` columns are still rendering in the data dictionary table. Looks like you just update the settings to use this new column.

Screenshot 2025-05-21 at 10 18 46 AM

johnclary (commented):
@mddilley this looks really great. I love this example of replacing frontend code with a DB solution. The use of a generated column for this is really nice.

Once that data dictionary table is fixed this looks shippable to me 🚢 🚢 🚢

chiaberry (approved):
a much cleaner way to handle it!
and thank you John for catching the dictionary update

PR #1599, opened 2025-05-20
Refactor NavigationSearchInput to not use GraphQLAbstract and remove abstract

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19720

## Testing
**URL to test:** 

https://deploy-preview-1599--atd-moped-main.netlify.app/moped/projects

**Steps to test:**

Test using the search input in the navigation bar
1. use a search term that returns a lot of results, "test" is a good one. Click on "more results" at the bottom of the dropdown and confirm it takes you to the project list view with the search term populated. 
2. use a search term with a few results. "sidewalk" in staging will give you one result, click to see project. 
3. Try searching again, confirm the results dont persist after closing the dropdown. 
4. try to break it. 

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/layouts/DashboardLayout/NavBar/useNavigationSearchQuery.js
```diff
@@ -0,0 +1,47 @@
...
+            : searchValue;
+
+          return `{ ${column}: {${operator}: ${graphqlSearchValue}} }`;
+        })
+      : [];
+    const searchWhereString = searchOperatorsOrValues.join(", ");
+    const queryString = `query NavigationProjectsList {
+      project_list_view (
+          limit: 10
+          order_by: { updated_at: desc },
```
chiaberry, line 29: wondering if i should keep these in the config, so they are easier to find?

chiaberry: also I am limiting to 10 arbitrarily, we used to have a limit of 250, but the navigation search results will never show anything over 5

moped-editor/src/layouts/DashboardLayout/NavBar/useNavigationSearchQuery.js
mddilley, line 1: It would be interesting to see one day if we could combine this query hook with the ones used in the project list view. We are wayyyy closer to that now!

moped-editor/src/layouts/DashboardLayout/NavBar/useNavigationSearchQuery.js
mddilley: love the update because there is no way we need 250 results. 🙏 I think that having these set in the query config could be nice to align with how there are set in the existing `PROJECT_LIST_VIEW_QUERY_CONFIG` default limit and order.

mddilley (approved):
Works perfectly! Thanks for fixing the quirk where unfiltered results made their way into the dropdown. I tried to break this but I couldn't. I added a few comments instead. Looks amazing! 🚢

moped-editor/src/layouts/DashboardLayout/NavBar/NavigationSearchInput.js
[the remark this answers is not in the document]
chiaberry: fixed and fixed!

moped-editor/src/layouts/DashboardLayout/NavBar/useNavigationSearchQuery.js
chiaberry: i moved it to the config 🚚

moped-editor/src/libs/GQLAbstract.js
johnclary, line 1: 👋 You had a good run, Abstract

mddilley:
love that suggestion, Mateo! i am a keyboard navigation fan and it feels like a missing piece https://github.com/cityofaustin/atd-data-tech/issues/22720

PR #1600, opened 2025-05-21
Fix dropdown selection bug

mateoclarke:
## Associated issues
Closes: https://github.com/cityofaustin/atd-data-tech/issues/20826

## Testing
**URL to test:** 
https://deploy-preview-1600--atd-moped-main.netlify.app/

**Steps to test:**
1. Go to the Teams tab and Project Teams table
2. Use keyboard navigation, add a new team member, select a name from the dropdown, roles, and note, then save the row
3. Try to edit an existing team member row using keyboard navigation.

In both new and edit scenarios, you should be able to use up and down arrows to navigate through dropdowns, the Enter key should confirm a selection without saving the row, and the Tab and Shift + Tab should move you left and right through a row. Pressing Enter when the checkbox is selected should save and same for hitting Enter to cancel.

---
#### Ship list
- [x] Code reviewed 
   - Added to existing test: "As you fill out fields, tab through them to make sure they are keyboard accessible and that choosing a dropdown option by pressing enter does not save the whole row."


mddilley (approved):
I tested this locally and all the steps work out. The QA tests steps work with no regressions too. In addition to squashing the bug, this is feeling more intuitive than the existing "save on enter" behavior when I'm using the form. Thanks for diving into this one! It taught me more about how this library works. 🙌

I offered up a few suggestions that don't affect the functionality.

mddilley (approved):
Retested locally, and this works so great - thanks again! 🚢 🚀 Moving on to reviewing the follow-up PR now. 🙌

mddilley:
@mateoclarke Sorry that I forgot about the unused import in your upstream branch that I merged down into this one. I made a small change to get Netlify to stop complaining. 🤫

PR #1604, opened 2025-05-29
Use an atddocker hosted postgresql + postgis image

frankhereford:
## Associated issues

This PR is part of the solution for https://github.com/cityofaustin/atd-data-tech/issues/22091.

## Testing

* Replicate from a production snapshot
* If that worked, 👍 
* Feel free to test any other feature of the app you'd like. Everything is in-bounds.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-database/docker-compose.yml
+++ b/moped-database/docker-compose.yml
@@ -25,7 +25,7 @@ services:
 
 
     moped-pgsql:
-        image: "frankinaustin/postgis:16-3.5"
+        image: "atddocker/postgis:unified-db"
         volumes:
             - moped-db-vol:/var/lib/postgresql/data
         expose:
--- a/moped-database/hasura-cluster
+++ b/moped-database/hasura-cluster
@@ -27,7 +27,7 @@ function __get_platform() {
 
 function __set_postgresql_image() {
   __get_platform;
-  HASURA_CLUSTER_POSTGRES_IMAGE="frankinaustin/postgis:16-3.5";
+  HASURA_CLUSTER_POSTGRES_IMAGE="atddocker/postgis:unified-db";
   echo "Database Docker Image: ${HASURA_CLUSTER_POSTGRES_IMAGE}"
 }
 
--- a/moped-etl/arcgis/env
+++ b/moped-etl/arcgis/env
@@ -1,0 +1,4 @@
+AGOL_USERNAME=ATD_Publisher_Scripts
+AGOL_PASSWORD=DTS@8700cameron
+HASURA_ENDPOINT=http://host.docker.internal:8080/v1/graphql
+HASURA_ADMIN_SECRET=hasurapassword


moped-database/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
 
 
     moped-pgsql:
-        image: "frankinaustin/postgis:16-3.5"
+        image: "atddocker/postgis:unified-db"
```
mddilley, line 28: at first, i was questioning not having the versions in the tag, but then I reminded myself that they have no real connection to the versions in the image and just as well be `latest-db-version`. So, I like the idea of not needing to PR in these changes every update.

I am curious for other thoughts too. It seems straightforward enough to check the local DB for the postgres and postgis versions but maybe I am missing something.

moped-database/docker-compose.yml
mddilley, line 1: @frankhereford I wanted to check in on the order of updating since I saw the RDS is on postgres 16.3 and this new image is showing 16.9 for me. Makes sense to update the prod RDS after merging these but wanted to check for  the sake of the DevOps maintenance template since I rely heavily on that template 😅 😆

mddilley (approved):
I spun up the local stack successfully from a fresh snapshot, and I did a quick tour of Moped as well. Everything looks in order to me! 🙏 🚢

moped-database/docker-compose.yml
frankhereford: Great questions Mike, thanks!

## Re which minor/patch version we use

I am building the latest version of whatever major version we're running. Well, really, what I'm building is the version that the postgis folks picked when they wrote their dockerfiles, which I think is actually pretty valuable.

Since the major versions should have no breaking changes between them, I think we're well down the route of "diminishing returns" trying to compile postgres and postgis so they match exactly what's in RDS. As is, I'm able to use the postgres published image and a debian package of postgis to build (not compile) our postgis images, and if we want to exact-match what is in RDS, we're going to have to compile both from source. 

Certainly possible, but we loose the benefits of running the same binaries as a ton of other users and we get to avoid whatever bugfixes the found in their minor and patch releases. 

## Re the order of updates

1) Decide what major pg AWS version you're going to target and look up what postgis version it provides [here](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html). 
1) Build new PG image if needed of the major version you want to target with the right postgis version. Use these Dockerfile as-is, or modify them as needed to get the combination you need. 
  * For example, I had to pick a hash of the postgis repo from Q3 last year to put in the dockerfile as found today so that I could build 16-3.4.
1) Put that new image into service in our dev local stacks by getting it into our postgis [repo](https://hub.docker.com/repository/docker/atddocker/postgis) here and tagging it as 'unified-db' 
1) devs `docker compose pull` in the VZ and moped checkouts so they start using it
1) Provided that's working well, proceed to update RDS to bring it up to, but not ahead, of where the local dev stack is.

johnclary: I am curious what y'all think about calling this image `"atddocker/postgis-multiarch:<postgis-version-number>` — ? Since the image itself can be used for whatever postgis needs we have, not simply related to the unified db. and then we get postgis version baked in as well.

frankhereford: The [images are also tagged](https://hub.docker.com/r/atddocker/postgis/tags) in that way, if that's helpful. The idea is to support versioned tagging in the style of other projects we use like `graphql-engine` but also give us a way to take advantage of actually controlling the upstream tagging and being able to avoid having to PR a pinned change in two repos. 

The update frequency of our unified database's version is going to be real slow of course -- I do recognize this "time saving tag" will save literally 10s of minutes every few months, at absolute most.

frankhereford: > Which leads to one more follow-up comment. Where are we thinking the docs for this image should live? Is there an actual dockerfile that can be committed to a standalone repo with instructions for the build process?

I vote gitbook. The real short story is that we use these [dockerfiles](https://github.com/postgis/docker-postgis) from the postGIS crew, and for anything that isn't a current version (like `git pull` current), you do what it takes to compile what you want. 

For example, I had to fish around to get a commit to plug into the `16-3.5` Dockerfile to turn it into a 3.4 build. Trying to document anything more than the `buildx` command that defines the architecture targets, IMHO, is near impossible because you just never know what it's going to take to make software build. It's a thing you practice and get better at over time.

chiaberry (approved):
ran locally 👍🏼 
thanks for everyone's thoughtful discussion on versioning and and updates!

PR #1607, opened 2025-06-03
Create view to combine Moped internal notes, Moped status updates, and eCapris status updates

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/22700

This PR creates a new view to collect all types of notes that can be displayed in the Notes tab into one single timeline. The view handles the mapping of eCapris columns that we want to show in the UI to conform to the existing schema of Moped notes and status updates. 

It also updates the `project_list_view` to consider whether a project has an eCapris subproject ID and if syncing is enabled. This will prevent eCapris statuses from showing on projects where the PMs have opted out of eCapris statuses (toggle is off in Notes tab).

These are the places that we are concerned about seeing status updates (Moped or eCapris):
- Project notes tab: query from new view and filter by 3 note types
- Project summary view status update: query from new view filter to 2 status update types
- Dashboard (My Projects and Following tabs): query from new view filter to 2 status update types
- Project list: query from updated `project_list_view` which handles whether to include eCapris statuses or not
- AGOL dataset: selects from updated `project_list_view` which handles whether to include eCapris statuses or not

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start your local stack and follow this SQL:
```sql
-- Query notes from project 331 that has eCapris subproject ID 13240.027 set on it and has syncing enabled
SELECT project_id, ecapris_subproject_id, project_status_update FROM project_list_view WHERE project_id = 331;

-- Now, soft-delete the latest Moped note so that we can see that the view will use the latest eCapris status
UPDATE moped_proj_notes SET is_deleted = TRUE WHERE project_note_id = 11307;

-- Repeat the query to see that the view is now showing the latest status update that is sourced from eCapris
SELECT project_id, ecapris_subproject_id, project_status_update FROM project_list_view WHERE project_id = 331;

-- Check the AGOL view to see the latest project status rolled up into the component rows
SELECT project_id, project_component_id, project_status_update FROM component_arcgis_online_view WHERE project_id = 331;

-- Now, disable syncing so that our view will not used eCapris statuses when finding the latest status update
UPDATE moped_project SET should_sync_ecapris_statuses = FALSE WHERE project_id = 331;

-- Repeat the query to see that the view will show the next newest Moped status update with eCapris syncing now disabled
SELECT project_id, ecapris_subproject_id, project_status_update FROM project_list_view WHERE project_id = 331;
SELECT project_id, project_component_id, project_status_update FROM component_arcgis_online_view WHERE project_id = 331;
```

---
#### Ship list
- [ ] Code reviewed 


moped-database/migrations/default/1748534889272_create_notes_view/up.sql
```diff
@@ -0,0 +1,631 @@
...
+    moped_proj_notes.created_by_user_id,
+    moped_proj_notes.project_id,
+    (moped_users.first_name || ' ' || moped_users.last_name) AS author,
+    moped_note_types.name AS note_type_name,
+    moped_note_types.slug AS note_type_slug,
+    moped_phases.phase_name AS phase_name,
+    moped_phases.phase_key AS phase_key,
+    moped_proj_notes.is_deleted,
+    moped_proj_notes.phase_id,
+    TRUE AS is_editable,
```
mddilley, line 28: The intent of `is_editable` in this view is to help us know when to edit components like the pencil and trash icons. This would work along with the existing business logic that note authors can only edit their own notes and statuses.

moped-database/migrations/default/1748534889272_create_notes_view/up.sql
```diff
@@ -0,0 +1,631 @@
...
+    ecapris_subproject_statuses.created_by_user_id,
+    NULL AS project_id,
+    COALESCE(
+        (moped_users.first_name || ' ' || moped_users.last_name),
+        CASE
+            WHEN ecapris_subproject_statuses.reviewed_by_name LIKE '%,%' 
+                THEN TRIM(SPLIT_PART(ecapris_subproject_statuses.reviewed_by_name, ',', 2)) || ' ' || TRIM(SPLIT_PART(ecapris_subproject_statuses.reviewed_by_name, ',', 1))
+            ELSE LOWER(ecapris_subproject_statuses.reviewed_by_email)
+        END
+    ) AS author,
```
mddilley, line 51: I was originally going to use `reviewed_by_email` as a fallback to eCapris statuses that didn't match an existing Moped user, but this column is not always populated in the eCapris database. 

I could not find any statuses without `reviewed_by_name` or statuses where the format was not `Last, First` but I added a check just in case we see something different in the future.

moped-database/migrations/default/1748534889272_create_notes_view/up.sql
```diff
@@ -0,0 +1,631 @@
...
+    moped_note_types.name AS note_type_name,
+    moped_note_types.slug AS note_type_slug,
+    NULL AS phase_name,
+    NULL AS phase_key,
```
mddilley, line 55: eCapris statuses do have phases with their own lookup values. No stakeholders have asked for us to attempt to map them, and it doesn't seem like a great idea to me to put effort into this. 

Either way, our React code doesn't render a status badge when these are null which is what we want to happen.

roseeichelmann (approved):
tested and reviewed the code, the new view looks great to me!!

moped-database/views/combined_project_notes.sql
```diff
@@ -0,0 +1,49 @@
+-- Most recent migration: moped-database/migrations/default/1748534889272_create_notes_view/up.sql
+
+CREATE OR REPLACE VIEW combined_project_notes AS SELECT
+    'M'::text || moped_proj_notes.project_note_id AS id,
```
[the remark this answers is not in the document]
mddilley: yes, that is the intent, and I think that adding the full names is a great call. I'll update 🙏

moped-database/views/combined_project_notes.sql
```diff
@@ -0,0 +1,49 @@
...
+    ecapris_subproject_statuses.review_timestamp AS created_at,
+    ecapris_subproject_statuses.created_by_user_id,
+    null::integer AS project_id,
+    COALESCE(
+        (moped_users.first_name || ' '::text) || moped_users.last_name,
+        CASE
+            WHEN ecapris_subproject_statuses.reviewed_by_name ~~ '%,%'::text THEN (TRIM(BOTH FROM SPLIT_PART(ecapris_subproject_statuses.reviewed_by_name, ','::text, 2)) || ' '::text) || TRIM(BOTH FROM SPLIT_PART(ecapris_subproject_statuses.reviewed_by_name, ','::text, 1))
+            ELSE LOWER(ecapris_subproject_statuses.reviewed_by_email)
+        END
+    ) AS author,
```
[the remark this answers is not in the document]
mddilley: I can't remember if comments persist inside of these views or not so I'll give it a go, and we'll find out. 🚀

moped-database/views/combined_project_notes.sql
```diff
@@ -0,0 +1,49 @@
...
+    moped_note_types.name AS note_type_name,
+    moped_note_types.slug AS note_type_slug,
+    null::text AS phase_name,
+    null::text AS phase_key,
+    false AS is_deleted,
```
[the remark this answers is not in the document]
mddilley: Not sure if I mentioned or not but they are not deleted upstream. We should be covered handling edits only.

moped-database/views/project_list_view.sql
```diff
@@ -284,11 +284,11 @@ LEFT JOIN min_estimated_phase_dates mepd ON mp.project_id = mepd.project_id
...
-        mpn.project_note,
-        mpn.created_at AS date_created
-    FROM moped_proj_notes mpn
-    WHERE mpn.project_id = mp.project_id AND mpn.project_note_type = 2 AND mpn.is_deleted = false
-    ORDER BY mpn.created_at DESC
+        combined_project_notes.project_note,
+        combined_project_notes.created_at AS date_created
+    FROM combined_project_notes
+    WHERE (combined_project_notes.project_id = mp.project_id OR mp.should_sync_ecapris_statuses = true AND mp.ecapris_subproject_id IS NOT null AND combined_project_notes.ecapris_subproject_id = mp.ecapris_subproject_id) AND (combined_project_notes.note_type_name = any(ARRAY['Status Update'::text, 'eCapris Status Update'::text])) AND combined_project_notes.is_deleted = false
+    ORDER BY combined_project_notes.created_at DESC
```
johnclary, line 291: seeing this bit

```sql
 (combined_project_notes.note_type_name = any(ARRAY['Status Update'::text, 'eCapris Status Update'::text]))
```

makes me wonder if it would be worth adding an `is_status_update` boolean column to your view. it might provide for a tiny performance improvement and push a bit more business logic into the view. but i think it's a minor consideration.

and fwiw the `is_deleted` filter here seems not needed because your view is already excluding deleted notes.

moped-database/views/project_list_view.sql
johnclary, line 1: do you think it would be wort adding the note source to the project list view (maybe the `note_type_name` column from your new view), so that you can tell where the status update came from? or future enhancement if no obvious feature depends on that currently.

johnclary (approved):
Looks great!  🚢 🚢 🚢 🚢 🚢 🚢

moped-database/views/combined_project_notes.sql
```diff
@@ -0,0 +1,49 @@
+-- Most recent migration: moped-database/migrations/default/1748534889272_create_notes_view/up.sql
+
+CREATE OR REPLACE VIEW combined_project_notes AS SELECT
```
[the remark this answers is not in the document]
chiaberry: I +1 this suggestion

moped-database/views/project_list_view.sql
```diff
@@ -284,11 +284,11 @@ LEFT JOIN min_estimated_phase_dates mepd ON mp.project_id = mepd.project_id
...
     SELECT
-        mpn.project_note,
-        mpn.created_at AS date_created
-    FROM moped_proj_notes mpn
-    WHERE mpn.project_id = mp.project_id AND mpn.project_note_type = 2 AND mpn.is_deleted = false
-    ORDER BY mpn.created_at DESC
+        combined_project_notes.project_note,
+        combined_project_notes.created_at AS date_created
+    FROM combined_project_notes
+    WHERE (combined_project_notes.project_id = mp.project_id OR mp.should_sync_ecapris_statuses = true AND mp.ecapris_subproject_id IS NOT null AND combined_project_notes.ecapris_subproject_id = mp.ecapris_subproject_id) AND (combined_project_notes.note_type_name = any(ARRAY['Status Update'::text, 'eCapris Status Update'::text])) AND combined_project_notes.is_deleted = false
```
chiaberry, line 290: why is it `OR mp.should_sync_ecapris_statuses` and not `AND` ?

moped-database/views/combined_project_notes.sql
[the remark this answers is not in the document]
mddilley: thanks for catching, y'all! i will added this shortly. 🙏

moped-database/views/project_list_view.sql
mddilley: I like this idea a lot. Now that we have the additional status update type, I agree that handling that upstream in the new view would be an improvement and make what we are checking in this project_list_view more obvious too. 👍 

I'll remove the redundant filtering too - thanks for catching that! 🙏

moped-database/views/project_list_view.sql
mddilley: Yes, I think that taking advantage of the join here is a good call. I've already found a place in the code used to display notes that would benefit from having the `id` from `moped_note_types` too so I'll bundle this in the view while I'm at it.

moped-database/views/project_list_view.sql
mddilley: This would be another great place for a comment if we could get them to stick between the bot processing. This is saying that we want to select from Moped status updates OR (if eCAPRIS syncing is enabled and a project has an eCAPRIS subproject ID) eCAPRIS status updates.

I'm also noticing that the parentheses that I used to group the logic here didn't stick around when the Postgres-created view was read into this file. That makes this harder to follow imo. I'll double check that the view is following what I described above.

chiaberry: thank you for helping me parse it ❤️

PR #1609, opened 2025-06-06
Add toggle for eCapris statuses sync in Notes tab

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C21793

## Testing
**URL to test:** 

https://deploy-preview-1609--atd-moped-main.netlify.app/moped/dashboard

**Steps to test:**
1. Go to a project without an ecapris subproject. In the notes page, the eCapris switch should be disabled with a tooltip saying "Add eCapris subproject ID to sync from eCapris"
2. On the project summary page, add an ecapris id to the project.
3. Return to the notes page. The ecapris switch should not be disabled and it should have defaulted to being on. The tooltip now says "Statuses are synced from eCapris every 30 minutes". Test toggling on and off.
4. Now go to the project summary page and clear the ecapris id. Return to the notes page, the ecapris switch should be disabled again.

---
#### Ship list
- [x] Code reviewed 
   - Added note about new test in collector branch ship list

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -26,6 +26,7 @@ import {
   advancedSearchFilterParamName,
   advancedSearchIsOrParamName,
 } from "src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch";
+import { simpleSearchParamName } from "src/views/projects/projectsListView/useProjectListViewQuery/useSearch";
 import {
   areAllFiltersComplete,
   checkIsValidInput,
@@ -96,6 +97,9 @@ const useStyles = makeStyles((theme) => ({
  * @param {Function} resetSimpleSearch - Function to reset the simple search
  * @param {Object} searchParams - The URL search params
  * @param {Function} setSearchParams - Function to set the URL search params
+ * @param {string} searchFieldValue - The current search field value
+ * @param {Function} setSearchFieldValue - Function to set the search field value
+ * @param {Function} setSearchTerm - Function to set the search term
  * @return {JSX.Element}
  * @constructor
  */
@@ -108,6 +112,9 @@ const Filters = ({
   setIsOr,
   searchParams,
   setSearchParams,
+  searchFieldValue,
+  setSearchFieldValue,
+  setSearchTerm,
 }) => {
   /**
    * The styling of the search bar
@@ -278,6 +285,10 @@ const Filters = ({
    * Applies the current local state and updates the parent's state
    */
   const handleApplyButtonClick = () => {
+    const trimmedSearchFieldValue = searchFieldValue.trim();
+    setSearchFieldValue(trimmedSearchFieldValue);
+    setSearchTerm(trimmedSearchFieldValue);
+
     if (filterParameters.length > 0) {
       /* If we have advanced filters, set query state values and update search params */
       setSearchParams((prevSearchParams) => {
@@ -285,6 +296,11 @@ const Filters = ({
 
         prevSearchParams.set(advancedSearchFilterParamName, jsonParamString);
         prevSearchParams.set(advancedSearchIsOrParamName, isOrToggleValue);
+        if (trimmedSearchFieldValue) {
+          prevSearchParams.set(simpleSearchParamName, trimmedSearchFieldValue);
+        } else {
+          prevSearchParams.delete(simpleSearchParamName);
+        }
         return prevSearchParams;
       });
 
@@ -295,7 +311,11 @@ const Filters = ({
       setSearchParams((prevSearchParams) => {
         prevSearchParams.delete(advancedSearchFilterParamName);
         prevSearchParams.delete(advancedSearchIsOrParamName);
-
+        if (trimmedSearchFieldValue) {
+          prevSearchParams.set(simpleSearchParamName, trimmedSearchFieldValue);
+        } else {
+          prevSearchParams.delete(simpleSearchParamName);
+        }
         return prevSearchParams;
       });
       /* If we have no advanced filters, reset query state */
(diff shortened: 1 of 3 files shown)


roseeichelmann:
@mddilley i noticed that currently projects without ecapris subproject ids have `should_sync_ecapris_statuses` to true, so the ecapris switch is disabled but it is checked. is this what is wanted, or do we need to backfill?

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -316,6 +325,16 @@ const ProjectNotes = ({
     setDeleteConfirmationId(id);
   };
 
+  const handleECaprisSwitch = () => {
+    updateShouldSyncECapris({
+      variables: {
+        projectId: noteProjectId,
+        shouldSync: !shouldSyncFromECapris,
+      },
+    });
```
mddilley, line 334: Didn't see that `handleSnackbar` was readily available in this file so it could be nice to handle success and errors here like in `submitEditNote`. What do you think about adding?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
mddilley, line 1: When I was scoping this issue, I assumed that the summary and funding tabs having this component in common would be enough to have them behave the same, but it looks like they are receiving different `refetch` functions.

Could you take a look at updating the `refetch` that is passed to the eCapris component in the funding table so it refetches the project summary and funding table data? That way we would see the sync switch updates to the eCAPRIS ID are made in the funding tab too.

mddilley (changes requested):
This is working great! I requested a few small changes. Let me know how I can help and thank you! 🙏

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -170,6 +174,16 @@ const ProjectNotes = ({
   const [isDeleteConfirmationOpen, setIsDeleteConfirmationOpen] =
     useState(false);
   const [deleteConfirmationId, setDeleteConfirmationId] = useState(null);
+  const [shouldSyncFromECapris, setShouldSyncFromECapris] = useState(
+    projectData.moped_project[0].should_sync_ecapris_statuses
+  );
```
johnclary, line 179: Sort of along the lines of Mike's question about refetching, I'm wondering if we need to use a state variable here, or if we can let the live data control the UI state. I suppose that would add a slight delay in UI feedback while the mutation and refetch happen...

moped-editor/src/views/projects/projectView/ProjectNotes.js
```diff
@@ -316,6 +332,27 @@ const ProjectNotes = ({
...
+    updateShouldSyncECapris({
+      variables: {
+        projectId: noteProjectId,
+        shouldSync: !shouldSyncFromECapris,
+      },
+    })
+      .then(() => {
+        handleSnackbar(true, "ECAPRIS sync status updated", "success");
+      })
+      .catch((error) =>
```
johnclary, line 345: It would probably be good to add a state setter here to flip the UI toggle back to it's previous position if the update fails.

johnclary (approved):
Looks good! 🚢

moped-editor/src/views/projects/projectView/ProjectNotes.js
roseeichelmann: actually the delay isnt that bad....i throttled my network and theres a delay but maybe its fine? @mddilley i pushed up an update implementing johns suggestion here, what do you think?

moped-editor/src/views/projects/projectView/ProjectView.js
```diff
@@ -227,7 +227,7 @@ const ProjectView = () => {
           state: locationState, // Preserve the location state when changing tabs
         }
       );
-      if (newTab === 0) refetch();
+      refetch();
```
mddilley, line 230: I think this works well to get fresh data in the Notes tab when editing from the Funding tab, but I'm seeing a small difference in the way the eCAPRIS switch hydrates with its boolean value when adding a new ID or removing a new ID.

When the eCAPRIS ID is updated in the project summary view, the refetch happens right when the ID is added or removed so the updated data is ready before clicking the Notes tab. When updating in the Funding table, we rely on the refetch happening here on tab change which causes the toggle to animate in view which I think would be great to avoid. Hard to describe so let me know if you want a demo or screengrab!

I can create a spin-off for this, but I think we would need to drill the refetch for the project summary data from this file `TabComponent` -> `ProjectFunding` - `ProjectFundingTable` and bundle with the Funding table refetch. Something like that.

Anyways, I'm approving this because this is working great!

mddilley (approved):
Looks amazing! I tested with the QA steps and your steps, and I can't find an occasion where this breaks. Thanks for the extra error handling and feedback that you added. 🙌 🚀 🚢

mddilley:
@mateoclarke It looks like the existing activity log code is trying to handle the `should_sync_ecapris_statuses` in the project records. I wanted to mention since it is related to https://github.com/cityofaustin/atd-data-tech/issues/22431, and this branch could be a good starting point for that one (or [collector branch](https://github.com/cityofaustin/atd-moped/pull/1607) once this one eventually merges into it).

Screenshot 2025-06-13 at 5 06 32 PM

moped-editor/src/views/projects/projectView/ProjectNotes.js
mddilley: Left this comment hanging out after my last review - sorry! I think that removing the state is a great call, and I think that the snackbars provide good feedback when there is either a successful update or an error. We also have seen sneaky bugs in the past with de-synced state so that is nice to avoid as well. Thanks!

mddilley (approved):
Retested, and this works great! Thanks for working through this one with me. I love where everything landed. 🙏 🚢

PR #1613, opened 2025-06-17
Update should_sync_ecapris_statuses to false for projects without eCAPRIS IDs

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/23041

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Before applying migrations, start your local DB and find a project in `moped_projects` table without an `ecapris_subproject_id`. See that the `should_sync_ecapris_statuses = true`. Copy that project id to look up later
2. Apply migrations, now look up that project id and see that it now has `should_sync_ecapris_statuses = false`
3. New projects will default to `should_sync_ecapris_statuses = false`

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/down.sql
+++ b/moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/down.sql
@@ -1,0 +1,1 @@
+alter table "public"."moped_project" alter column "should_sync_ecapris_statuses" set default 'true';
--- a/moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/up.sql
+++ b/moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/up.sql
@@ -1,4 +1,10 @@
+-- use replica mode to disable triggers
+set session_replication_role = replica;
+
 -- should_sync_ecapris_statuses will default to false for projects without an eCAPRIS subproject id
 update moped_project set should_sync_ecapris_statuses = false where ecapris_subproject_id is null;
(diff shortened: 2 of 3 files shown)


mddilley (approved):
Tested this with some SQL queries, and there are no projects without eCAPRIS ids that have syncing turned on. Also tested creating a new project, and I no longer see that awkward state with the disabled toggle that is switched on. 

Thanks again for catching - and fixing! 🙏 🚢

moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/down.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."moped_project" alter column "should_sync_ecapris_statuses" set default 'true';
```
johnclary, line 1: If desired we could suppress  triggers activity logs using the old `SET session_replication_role = replica;` trick

johnclary (approved):
Looks good to me! 
🚢

moped-database/migrations/default/1750186599182_set_should_sync_ecapris_false/down.sql
mddilley: i say let's go for it! i often forget about this trick so this will help me keep it in mind. 🙏

roseeichelmann: done!

PR #1055, opened 2023-05-23
Rename refuge island to crossing island

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/12339

## Testing
**URL to test:** 

run locally

**Steps to test:**

Refuge Islands are now Crossing Islands
Screenshot 2023-05-23 at 1 21 36 PM

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Added to [QA test script](https://docs.google.com/spreadsheets/d/1n_O6MLh9cwwPf57HUM394Ea-z9uuoEV1-QW4axNZXLE/edit#) if applicable

PR #1619, opened 2025-06-27
23235 add lane conversion work type to various components

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/23235

## Testing
**URL to test:** 

local due to database changes

**Steps to test:**
- apply the migration and click to add a component to a project
- check the following component types for the work type 'Lane Conversion'
  - Project Extent - Generic (linear)
  - Transit - Lane
  - Transit - Managed Lane Access Point
  - Transit - Transit/Bike Lane
  - Transit - Transit Queue Jump
- run the down migration and check to make sure those work types are no longer present

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/migrations/default/1751058211875_add_lane_conversion_worktype_to_various_components/down.sql
+++ b/moped-database/migrations/default/1751058211875_add_lane_conversion_worktype_to_various_components/down.sql
@@ -1,2 +1,9 @@
-DELETE FROM public.moped_component_work_types WHERE component_id = 0 AND work_type_id = 4;
-DELETE FROM public.moped_component_work_types WHERE component_id BETWEEN 52 AND 55 AND work_type_id = 4;
+DELETE FROM public.moped_component_work_types
+WHERE work_type_id = (
+        SELECT id FROM moped_work_types WHERE key = 'lane_conversion'
+    )
+    AND component_id IN (
+        SELECT component_id
+        FROM moped_components
+        WHERE component_name_full IN ('Project Extent - Generic (linear)', 'Transit - Lane', 'Transit - Managed Lane Access Point', 'Transit - Transit/Bike Lane', 'Transit - Transit Queue Jump')
+    );


moped-database/views/project_list_view.sql
tillyw, line 1: 🤖 did this

moped-database/migrations/default/1751058211875_add_lane_conversion_worktype_to_various_components/up.sql
mddilley, line 1: @tillyw This works to insert these by their ids, but another alternative would be selecting the ids by the component names and work type names in the off-chance that something unexpected happens like a row is deleted or an id changes. I might be overthinking since we shouldn't have different ids across environments since we make all staging and production changes through migrations. Here is some previous discussion about the same thing: https://github.com/cityofaustin/atd-moped/pull/1055#discussion_r1202969496 and https://github.com/cityofaustin/atd-moped/pull/1434#discussion_r1781625852.

Anyways, we have a recipe that is kind of similar [here](https://github.com/cityofaustin/atd-moped/blob/main/moped-database/recipes/copy_component_work_types_from_other_component.sql) but something like this could work too (with my more-than-likely wrong column and table names fixed up):

```sql
INSERT INTO moped_component_work_types (component_id, work_type_id)
SELECT moped_components.id, moped_work_types.id 
FROM moped_components, moped_work_types
WHERE moped_components.component_name_full IN ('Component A', 'Component B', 'Component C')
AND moped_work_types.name = 'Lane Conversion';
```

tillyw: good call, thanks for the guidance! i have updated both migrations to use `component_name_full` rather than `id` and i used `key` in the `moped_work_types` table since those values should never change

mddilley (approved):
Ran the migrations up and down, and I see the new work types in the component UI and the data dictionary too. Thanks for adding these! 🙏 🚀

roseeichelmann (approved):
ran the up and down migrations and this looks good! :rocket:

frankhereford (approved):
SQL looks great and applied cleanly in both directions. Thanks Tilly!

The `insert into` sourced from the `select` is really elegant. ✨

PR #1620, opened 2025-06-30
Disable eCAPRIS Notes Filter

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/22702

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Use prod replica data so you can actually see projects that have eCAPRIS statuses
2. Go to the Notes tab of a project that has an eCAPRIS id (For example: https://localhost:3000/moped/projects/3248?tab=notes). Use the toggle button to turn off eCAPRIS syncing
3. See that the "Synced from eCAPRIS" filter button gets disabled and has a tooltip when you hover
4. Turn the syncing toggle back on and see that you can now see and filter to eCAPRIS statuses
5. Test removing the eCAPRIS id from the summary or funding page, then return to the Notes tab and see that the button is disabled again

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectNotes/NoteTypeButton.js
+++ b/moped-editor/src/views/projects/projectView/ProjectNotes/NoteTypeButton.js
@@ -7,6 +7,8 @@ import { Button, Tooltip } from "@mui/material";
  * @param {function} setFilterNoteType - function to set the filter note type state
  * @param {string} filterNoteType - id for toggled note type
  * @param {string} label - button label
+ * @param {boolean} isDisabled - should the button be disabled
+ * @param {string} disabledMessage - message to render in tooltip if button is disabled
  * @return {JSX.Element}
  * @constructor
  */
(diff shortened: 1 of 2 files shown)


johnclary (approved):
I followed the test steps and they checked out. Code looks great 🚢 🚢 🚢 🚢

moped-editor/src/views/projects/projectView/ProjectNotes/NoteTypeButton.js
```diff
@@ -15,15 +17,24 @@ const NoteTypeButton = ({
...
-  >
-    {label}
-  </Button>
-);
+  isDisabled,
+  disabledMessage,
+}) => {
+  return (
+    <Tooltip title={isDisabled ? disabledMessage : ""}>
+      <span>
```
mddilley, line 25: This was a TIL for me that [tooltips won't show on disabled buttons since user interactions aren't trigger on them](https://v5.mui.com/material-ui/react-tooltip/#disabled-elements). Could you add a comment or link to the docs here to help us avoid an accidental regression? 🙏

mddilley (approved):
Works great! Thanks for taking care of this last piece of the puzzle. 🧩  🙏 🚢

moped-editor/src/views/projects/projectView/ProjectNotes/NoteTypeButton.js
chiaberry: +1, I could definitely see future me deleting this span that is silently doing so much work here

PR #1624, opened 2025-07-09
Add new subcomponents to Signal - Traffic component

mateoclarke:
## Associated issues

Closes https://github.com/cityofaustin/atd-data-tech/issues/23396

Screenshot 2025-07-09 at 4 08 41 PM

## Testing
Local testing required

**Steps to test:**
- Run the migration: `hasura migrate apply`
- Go to a project's Map tab
- Edit a component, in the modal form, change the Component Type to "Signal - Traffic"
- Click the subcomponents dropdown, you should be able to select the two new options: "Bicycle signal (may use pedestrian
  signal sign)" and "Bicycle signal (permissive conflicts)"
- Test the down migration and confirm those options are removed

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-database/views/project_list_view.sql
+++ b/moped-database/views/project_list_view.sql
@@ -273,15 +273,13 @@ LEFT JOIN project_district_association districts ON mp.project_id = districts.pr
 LEFT JOIN min_confirmed_phase_dates mcpd ON mp.project_id = mcpd.project_id
 LEFT JOIN min_estimated_phase_dates mepd ON mp.project_id = mepd.project_id
 LEFT JOIN project_component_work_types pcwt ON mp.project_id = pcwt.project_id
-LEFT JOIN LATERAL (
-    SELECT
-        combined_project_notes_view.project_note,
-        combined_project_notes_view.created_at AS date_created,
-        combined_project_notes_view.author
-    FROM combined_project_notes_view
-    WHERE (combined_project_notes_view.project_id = mp.project_id OR mp.should_sync_ecapris_statuses = true AND mp.ecapris_subproject_id IS NOT null AND combined_project_notes_view.ecapris_subproject_id = mp.ecapris_subproject_id) AND combined_project_notes_view.is_status_update = true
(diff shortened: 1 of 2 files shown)


tillyw (approved):
tested and confirmed the two new subcomponents are available! and the down migration removes them. 

in a previous pr it was suggested i make selections using a name or key column rather than id, so that may be the case here as well. the comments are helpful and the migrations look good aside from the warnings about missing line breaks at the end.

thank you for adding these!

moped-database/migrations/default/1751995587779_add_bicycle_signal_subcomponents/up.sql
```diff
@@ -0,0 +1,15 @@
+-- Add two new bicycle signal subcomponents for Signal - Traffic component
+INSERT INTO moped_subcomponents (subcomponent_name) VALUES 
+    ('Bicycle signal (may use pedestrian signal sign)'),
+    ('Bicycle signal (permissive conflicts)');
```
mddilley, line 4: Nathan reached out to me about a small change here for better clarity. Could you update to `Bicycle signal face (permissive conflicts)` before merging? Thank you!

johnclary (approved):
I see the new subcomponents!

Screenshot 2025-07-14 at 1 02 41 PM

I did not realize it was illegal for a cyclist to use the walk sign in Austin. They should but these signs up at every intersection that has an LPI 😈

chiaberry (approved):
yaay for bike signals!

>I did not realize it was illegal for a cyclist to use the walk sign in Austin.
is it really??? also what is an LPI @johnclary ?

johnclary:
> is it really? also what is an LPI 

Leading pedestrian interval—when the walk signs lights up a few seconds before the bulbs turn green. Based on searching I don't _think_ it's legal to ride during this interval in Austin unless the sign is present. NYC legalized this fairly recently but the police still issue tickets for it 👍

PR #1645, opened 2025-08-07
23840 update search and filter components to sx prop

tillyw:
## Associated issues
closes https://github.com/cityofaustin/atd-data-tech/issues/23840

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/components/GridTable/Filters.js
+++ b/moped-editor/src/components/GridTable/Filters.js
@@ -1,5 +1,4 @@
 import React, { useState } from "react";
-import PropTypes from "prop-types";
 import { useQuery } from "@apollo/client";
 
 import {
@@ -8,7 +7,6 @@ import {
   FormControl,
   Grid,
   Hidden,
-  Icon,
   IconButton,
   Grow,
   Typography,
@@ -16,8 +14,7 @@ import {
 import RadioGroup from "@mui/material/RadioGroup";
 import Radio from "@mui/material/Radio";
 import FormControlLabel from "@mui/material/FormControlLabel";
-
-import makeStyles from "@mui/styles/makeStyles";
+import { DeleteOutline, Close, Search, PlaylistAdd } from "@mui/icons-material";
 
 import { Autocomplete } from "@mui/material";
 import BackspaceOutlinedIcon from "@mui/icons-material/BackspaceOutlined";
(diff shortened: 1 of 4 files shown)


moped-editor/src/components/GridTable/Filters.js
```diff
@@ -8,16 +7,14 @@ import {
...
   IconButton,
   Grow,
   Typography,
 } from "@mui/material";
 import RadioGroup from "@mui/material/RadioGroup";
 import Radio from "@mui/material/Radio";
 import FormControlLabel from "@mui/material/FormControlLabel";
-
-import makeStyles from "@mui/styles/makeStyles";
+import { DeleteOutline, Close, Search, PlaylistAdd } from "@mui/icons-material";
```
tillyw, line 17: for some reason the `sx` prop does not seem to work on font icons so I replaced all of these icons imported directly from MUI. see https://stackoverflow.com/questions/74068619/react-change-size-of-icon-received-as-props

johnclary (commented):
Looks good to me. I spy the the tiniest difference in the `X` button in the filter menu—it looks a tiny bit smaller in this PR versus prod. 🤷‍♂️ 

**This PR**
Image

**Prod**
Image

johnclary (approved):

tillyw:
@mddilley this should be ready to ship, however I haven't been able to get the margins on the top row of the advanced filter container to align with what is in staging. the values should all be the same, but I tried making adjustments as well and did not get any results :/

local
Screenshot 2025-08-13 at 9 54 41 PM
<p />
staging
Screenshot 2025-08-13 at 9 54 49 PM

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -334,13 +275,56 @@ const Filters = ({
     setIsOrToggleValue(isOr);
   };
 
+  // Shared style classes
+  const formControlStyle = {
+    margin: theme.spacing(1),
+    minWidth: 120,
+  };
```
mddilley, line 282: @tillyw i looked at this again, and I think that these can be moved back outside of the component definition so they aren't recreated on each render. Either by using the callbacks on the object level `const style = (theme) => {...needsTheme}`, the attribute level `const style = {...otherStyles, (theme) => {...needsTheme}}, or shorthand `const style = { margin: 1 }`.

I know there are a million and one ways to implement this 😵‍💫, but these ones would work around the need for these objects to live inside the component.

moped-editor/src/components/GridTable/Filters.js
```diff
@@ -372,22 +356,22 @@ const Filters = ({
         <Grid
           item
           xs={6}
-          className={classes.filtersContainer}
+          sx={{ filtersContainerStyle }}
```
mddilley, line 359: @tillyw I had to do a triple take on why these styles weren't applying, and then I saw an extra set of curly braces in here that is preventing sx from accessing the key/values. There are a couple of other ones in here too. could you update? 🙏

mddilley (changes requested):
Requesting a couple more changes to help the any/all radio styles apply. Thanks for continuing the work on this one! 🙏

moped-editor/src/components/GridTable/Filters.js
tillyw: ok done! i'll default to doing it this way (at the object level for consistency) from now on so we can use imports only when necessary

mddilley (approved):
Looks great - thanks Tilly! 🚢 🚀 🚢

PR #1648, opened 2025-08-12
Updates to handle decreased lifespan of Cognito ID tokens

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/24016

This PR updates our Apollo Client instance to fetch a fresh token if needed before making GraphQL requests and mutations. The `useUserApi` hook is also updated the same way to ensure successful calls to the Moped API. It also moves us one more version closer to [AWS Amplify v6](https://github.com/cityofaustin/atd-data-tech/issues/24261) since [there are only small differences between v4 and v5](https://aws.amazon.com/blogs/mobile/announcing-aws-amplify-javascript-library-version-5/).

**Old flow**
Cognito id token expires -> sometimes a JWT error from Apollo Client was fed to ApolloErrorHandler -> this component forced a browser refresh to get a fresh id token using the valid refresh token -> Sometimes success, sometimes fallback component error, sometimes silent error which would require a browser refresh

**New flow**
Cognito id token expires -> before any GraphQL or Moped API requests, we check the session and get a fresh id token if needed -> Success

## Testing
**URL to test:** 

https://deploy-preview-1648--atd-moped-main.netlify.app/

**Steps to test:**
1. If you are logged in the deploy preview, sign out and then sign in using the username and password in our **Moped Test Admin (local, staging, prod)** entry
2. After signing in, check the local storage entry called `atd_moped_user_db_data`. It should contain an object with user data that was fetched from the GraphQL endpoint when first logging in. You can check your network tab for a query named GetUserProfile.
3. I added logs to show the current token's expiration so having the JS console open will be helpful to track when to proceed from one step to another. I'll remove logging before merge.
4. Now, go to https://deploy-preview-1648--atd-moped-main.netlify.app/moped/projects/, and wait for 5 minutes. ⏳
5. After waiting, click on a project link to go to a project details page. You should notice a brief delay while a fresh token is being retrieved before Apollo Client makes the GraphQL request. This tests new behavior where our Apollo Client `authLink` now checks the current session and retrieves a fresh token if the current one is expired. **This will cover all GraphQL requests and mutations even if they are not wrapped with our previous refresh mechanism in the ApolloErrorHandler component.**
6. You can also watch for network requests to `https://cognito-idp.us-east-1.amazonaws.com/` every time a new id token is fetched in these steps.
7. Now, go to https://deploy-preview-1648--atd-moped-main.netlify.app/moped/staff/edit/8 and wait 5 minutes. ⏳
8. After waiting, make an edit to my first name and save the change. Again there will be a delay, but the change should save, and you will see a fresh token expiration logged. This tests that our user management requests handled by the Moped API are still working as expected.
9. Last, go to https://deploy-preview-1648--atd-moped-main.netlify.app/moped/projects/2004?tab=files and then click "Add File". Wait 5 minutes. ⏳
10. Drag and drop a file or click "Browse" to select one to upload. Another delay but then the file should upload successfully, and you can save it. This tests that the file upload requests through our API are still working as expected.
11. You can try waiting 5 more minutes ⏳ to download the file from the table if you'd like. You can also test the global search or any other part of the app with data fetching or mutations if you want more 5 minute breaks.
12. Wait another 5 minutes ⏳ and then refresh the page using your browser. You should see a loading spinner briefly while an async call for the current session is made. You should arrive back in the app where you were before.
13. Sign out and try to visit https://deploy-preview-1648--atd-moped-main.netlify.app/moped/projects/. You should be redirected to the login page.
14. Log in using the username/password in the **Moped Test Editor (local, staging, prod)** entry, go to https://deploy-preview-1648--atd-moped-main.netlify.app/moped/staff and make sure that you cannot see a button that says "+ Add Staff" in the top right above the table or edit icons on the left side of the table rows. This tests that role-based access is still working since only admin roles can add or edit staff.
15. Signing in with SSO (COA) is not supported in deploy previews or local so we will have to test that in staging and patch if needed
 
---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -57,20 +57,6 @@ export const getSessionDatabaseData = () => {
   }
 };
 
-function epochToCentralTime(epochTimestamp) {
-  const date = new Date(epochTimestamp * 1000);
-  return date.toLocaleString("en-US", {
-    timeZone: "America/Chicago",
-    year: "numeric",
-    month: "2-digit",
-    day: "2-digit",
-    hour: "2-digit",
-    minute: "2-digit",
-    second: "2-digit",
-    hour12: true,
-  });
-}
-
 /** Get the Cognito ID JWT from a Cognito session.
  *
  * @param {CognitoUserSession} session - The Cognito user session.
@@ -114,32 +100,41 @@ export const initializeUserDBObject = async (session) => {
   const sessionDataFromLocalStorage = getSessionDatabaseData();
 
   // If the session is valid and there is no existing data...
-  if (session && sessionDataFromLocalStorage === null) {
+  if (sessionDataFromLocalStorage === null) {
     // Fetch the data from Hasura
-    fetch(config.env.APP_HASURA_ENDPOINT, {
-      method: "POST",
-      headers: {
-        "Content-Type": "application/json",
-        Authorization: `Bearer ${token}`,
-        "X-Hasura-Role": `${getHighestRole(session)}`,
-      },
-      body: JSON.stringify({
-        query: ACCOUNT_USER_PROFILE_GET_PLAIN,
-        variables: {
-          userId: getDatabaseId(session),
+    try {
+      const res = await fetch(config.env.APP_HASURA_ENDPOINT, {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Authorization: `Bearer ${token}`,
+          "X-Hasura-Role": `${getHighestRole(session)}`,
         },
-      }),
-    }).then((res) => {
-      // Then we parse the response
-      res.json().then((resData) => {
-        if (resData?.errors) {
-          console.error(resData.errors);
-        }
-        if (resData?.data?.moped_users) {
-          setSessionDatabaseData(resData.data.moped_users[0]);
-        }
+        body: JSON.stringify({
+          query: ACCOUNT_USER_PROFILE_GET_PLAIN,
+          variables: {
+            userId: getDatabaseId(session),
+          },
+        }),
       });
-    });
+
+      const resData = await res.json();
+
+      if (resData?.errors) {
+        // Show error feedback in sign in form
+        throw new Error("Error fetching user data");
+      }
+
+      if (resData?.data?.moped_users) {
+        const userData = resData.data.moped_users[0];
+        return userData;
+      }
+
+      throw new Error("No user data found");
+    } catch (error) {
+      console.error("Failed to fetch user data:", error);
+      throw error;
+    }
   }
 };
 
@@ -172,12 +167,13 @@ export const UserProvider = ({ children }) => {
       await Auth.signIn(usernameOrEmail, password);
 
       const session = await Auth.currentSession();
-      await initializeUserDBObject(session);
+      const userDBData = await initializeUserDBObject(session);
+      setSessionDatabaseData(userDBData);
       setUser(session);
 
       setIsLoginLoading(false);
     } catch (err) {
-      if (err.code === "UserNotFoundException") {
+      if (err?.code === "UserNotFoundException") {
         err.message = "Invalid username or password";
       }
 
@@ -197,7 +193,8 @@ export const UserProvider = ({ children }) => {
       await Auth.federatedSignIn({ provider: "AzureAD" });
 
       const session = await Auth.currentSession();
-      await initializeUserDBObject(session);
+      const userDBData = await initializeUserDBObject(session);
+      setSessionDatabaseData(userDBData);
       setUser(session);
 
       setIsLoginLoading(false);
@@ -231,19 +228,16 @@ export const UserProvider = ({ children }) => {
     try {
       const session = await Auth.currentSession();
 
-      console.log("User session refreshed:", session);
-      console.log(
-        "Token expires:",
-        epochToCentralTime(session?.idToken?.payload.exp)
-      );
-
       return session;
     } catch (err) {
+      // Log out user if a Cognito session cannot be retrieved to force a fresh login.
+      // This can happen when a refresh token is expired or invalid for example.
+      await logout();
       console.error("Error getting Cognito session: ", err);
 
       return null;
     }
-  }, []);
+  }, [logout]);
 
   /**
    * This effect runs when the component mounts and checks if Cognito has a valid session.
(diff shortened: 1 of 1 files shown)


moped-editor/src/auth/useAuthentication.js
```diff
@@ -8,6 +8,20 @@ import React, {
 import { Auth, Hub } from "aws-amplify";
 import { signInButton, signInButtonContent } from "@aws-amplify/ui";
 
+function epochToCentralTime(epochTimestamp) {
```
mddilley, line 11: this code is not intended to merge. just a quick returned from an LLM to make token expiration easier to watch

moped-editor/src/auth/user.js
```diff
@@ -26,13 +27,6 @@ export const UserContext = createContext(null);
  */
 export const atdColorKeyName = "atd_moped_user_color";
 
-/**
- * This is a constant string key that holds the profile for a user.
- * @type {string}
- * @constant
- */
-export const atdSessionKeyName = "atd_moped_user_context";
```
mddilley, line 34: This PR also removes the a local storage entry called `atd_moped_user_context` which was another copy of the user session that can be retrieved using the Amplify library. 

It took some figuring out, but this value was keeping users at their current route (like `/projects/1392?tab=notes`) when the browser refreshed since it was the initial value of the `user` state. The app would not redirect to the sign in screen as long as the `user` state was populated. 

The initial value of the `user` state is now set using Amplify, and the redirect to previous route is handled by React Router state that I'll comment on later in this PR.

moped-editor/src/components/ApolloErrorHandler.js
```diff
@@ -1,51 +1,15 @@
...
-  const errorString = error
-    ? JSON.stringify(error, Object.getOwnPropertyNames(error))
-    : "";
-  const jwtError = errorString.includes("JWT") || errorString.includes("token");
-
-  if (jwtError) {
-    setTimeout(() => {
-      window.location.reload();
-    }, 5000);
-  }
```
mddilley, line 27: This is the previous mechanism to refresh the id token that is no longer needed. Since we are now prevent Apollo jwt errors in `App.js`, this component now only handles other Apollo errors by displaying them with the `FallbackComponent`.

moped-editor/src/components/FileUpload/FileUpload.js
```diff
@@ -24,12 +24,11 @@ registerPlugin(
 
 const FileUpload = (props) => {
   const theme = useTheme();
+  const { getCognitoSession } = useUser();
```
mddilley, line 27: i didn't want to make an even larger and more confusing diff, but do y'all think that `useAuthentication` would be a better name for this hook? It does more than get the user session.

moped-editor/src/auth/user.js
```diff
@@ -69,6 +70,28 @@ export const getSessionDatabaseData = () => {
   }
 };
 
+function epochToCentralTime(epochTimestamp) {
```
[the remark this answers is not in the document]
chiaberry: it does make it easier for testing!

moped-editor/src/auth/user.js
```diff
@@ -136,86 +147,130 @@ export const initializeUserDBObject = (userObject) => {
...
+      console.error("Error getting Cognito session: ", err);
+
+      return null;
+    }
+  }, []);
+
+  /**
+   * This effect runs when the component mounts and checks if Cognito has a valid session.
+   * If there is no user state, it retrieves the current session and sets it to prevent logout
+   * when browser is refreshed. If there is no current session, getCognitoSession will return null.
```
roseeichelmann, line 251: and what happens when `getCognitoSession` returns null?

roseeichelmann (approved):
this was a really hefty PR! i followed all the test steps and they all worked as expected. this is a big improvement on the token expiration flow :sparkles:

moped-editor/src/components/FileUpload/FileUpload.js
mddilley: thanks! i am going to punt on this and put it in https://github.com/cityofaustin/atd-data-tech/issues/24261 for later refinement.

moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
```diff
@@ -49,7 +55,7 @@ const DashboardLayout = () => {
       </div>
     </div>
   ) : (
-    <Navigate to="/moped/session/signin" />
+    <Navigate to="/moped/session/signin" state={{ from: location }} replace />
```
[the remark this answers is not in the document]
mddilley: glad to hear! i think this happened before but it required a couple 5 seconds waits, and I never really figured out how the route was being restored before.

johnclary (commented):
😍 Thanks for all this amazing work. It's so great to see the auth code get cleaned up—there are so many moving parts! 

I managed to out-wait the refresh token lifespan while testing this PR, and as a result I bumped into a few `TypeError`'s. I assume this all stems from the refresh token functionality that you've PR'd separately.

1. Sign in
2. Wait 60+ minutes while reviewing your PR code
3. Navigate to the dashboard page, then projects page, then staff pages—all of them throw an error


Dashboard page:

Image

Staff page
Image

moped-editor/src/auth/user.js
```diff
@@ -136,86 +147,130 @@ export const initializeUserDBObject = (userObject) => {
...
-    return JSON.parse(localStorage.getItem(atdSessionKeyName)) || null;
-  };
+  const login = useCallback(async (usernameOrEmail, password) => {
+    setIsLoginLoading(true);
+
+    try {
+      await Auth.signIn(usernameOrEmail, password);
+
+      const session = await Auth.currentSession();
+      await initializeUserDBObject(session);
```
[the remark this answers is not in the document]
mddilley: i like this a lot so `initializeUserDBObject` isn't pulling double-duty and makes a change that isn't super obvious by reading the name. I'll make this change in https://github.com/cityofaustin/atd-moped/pull/1656/files before I merge it since I updated this function to async/await. 🙏

moped-editor/src/auth/user.js
```diff
@@ -136,86 +147,130 @@ export const initializeUserDBObject = (userObject) => {
...
+      setUser(session);
+
+      setIsLoginLoading(false);
+    } catch (err) {
+      if (err.code === "UserNotFoundException") {
```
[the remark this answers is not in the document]
mddilley: I'll update this in https://github.com/cityofaustin/atd-moped/pull/1656/files to use optional chaining to be safe. It was tough to pick and choose what old code to trust and what to focus on improvements. I also think it is entirely possible that we never ran into some of the error handling like this since our previous code would force a browser refresh so they are good details to catch - thanks!

PR #1654, opened 2025-08-25
Update Apollo Client error handling to use errorLink

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/24260

**Heads up:** Most of this diff is the result of removing the ApolloErrorHandler wrapper from components and the resulting shift in formatting.

This PR moves our Apollo GraphQL error and network error handling up into the Apollo Client instance. Previously, we wrapped components that used `useQuery` and returned `error` from it with `ApolloErrorHandler` which would display any GraphQL errors and also JWT expiration network errors passed to it. That component now wraps the whole app and receives error state from the Apollo Client hook that returns GraphQL or network errors that occur on all requests and mutations to our GraphQL endpoint.

[Previous work](https://github.com/cityofaustin/atd-moped/pull/1648) enhances the app to handle id token refresh which means that we should not see JWT expiration errors. The new code will handle them though and any other unexpected network errors while continuing to show us GraphQL errors related to column name mismatches or permissions issues.

## Testing
**URL to test:** 

Local so that you can trigger a query or mutation error we might see in development and update permissions in the local Hasura console

**Steps to test:**
These steps will catch a few types of errors and make sure our existing error handling is working the same. These types of errors should be infrequent and caught in development, and we would want to learn about the first two types in production fast so we could patch them since there is no way to recover from them through the UI.

**Before starting these steps, start your local stack and open the Hasura console so we can updates permissions on a column**
1. Catching unexpected local query or mutation errors (invalid column in the code)
   - Update `SUMMARY_QUERY` with a non-existent column in the `moped_project` part of the query like `project_pizzas` (line 24 in `project.js`)
      - Go to a project page and see an error that says `field 'project_sponsors' not found in type: 'moped_project'`
      - Click one of the tabs in the top navigation bar to test that you can navigate away from the error
      - Remove `project_pizzas` from the query so you can continue testing
2. Catching Hasura permissions issues
   - Start the Hasura console and go to [the moped_project table permissions](http://localhost:9695/console/data/default/schema/public/tables/moped_project/permissions). Remove the permission for the role that you are logged in for so that you can't read the `project_name` column
   - Go to a project page and see an error that says `field 'project_name' not found in type: 'moped_project'` 
   - Restore the permission so that you can continue testing
4. Network errors which should only happen when a real connectivity error happens (offline mode or internet connection issues)
   - Go to the [project list table](https://localhost:3000/moped/projects)
   - In your browser dev tools, set yourself connection to offline
   - Click on a project link in the table
   - See that the error `Failed to fetch` is shown. I offered up an alternative in this PR.
6. Handling errors from invalid project list view filter config that come from outdated config. This code does not change this error catching that is handled by the `FallbackComponent` but this proves that it still works.
   -  Go to [a project list filter link with invalid column name](https://localhost:3000/moped/projects?filters=%5B%7B%22field%22%3A%22project_name_fulls%22%2C%22operator%22%3A%22string_contains_case_insensitive%22%2C%22value%22%3A%22test%22%7D%5D&isOr=false)
   - Go to [a project list filter link with invalid operator name](https://localhost:3000/moped/projects?filters=%5B%7B%22field%22%3A%22project_name_full%22%2C%22operator%22%3A%22string_contains_case_insensitives%22%2C%22value%22%3A%22test%22%7D%5D&isOr=false)

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
+++ b/moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
@@ -16,10 +16,10 @@ const DashboardLayout = () => {
   const location = useLocation();
   const { apolloError, setApolloError } = useApolloErrorContext();
 
-  // Clear Apollo error when navigating to a new page
+  // Clear Apollo error when navigating to a new route so ApolloErrorHandler doesn't block UI
   useEffect(() => {
     setApolloError(null);
-  }, [location, setApolloError]);
+  }, [location.pathname, setApolloError]);
 
   /* If user is not authenticated, redirect to sign-in page
    * and preserve the current location so they can be redirected back
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectDescription.js
@@ -74,7 +74,7 @@ const ProjectSummaryProjectDescription = ({
     updateProjectDescription({
       variables: {
         projectId: projectId,
-        description: 1,
+        description: description,
       },
       refetchQueries: [{ query: listViewQuery }],
     })


moped-editor/src/App.js
mddilley, line 1: This is where the new code lives. The existing custom hook that manages the Apollo Client instance now sets up an errorLink to handle request and network errors. It also has state to collect the error to display in `ApolloErrorHandler`.

One important change that I'm looking for feedback on is that `ApolloErrorHandler` now lives outside of our routing and doesn't get the nice navbar and navigation menu that is composed in our routes. I don't see this as a significant downgrade in UX since these error will only show when something is so wrong that users couldn't recover however they tried. We would want a service request to come in so we could fix the bug in our code.

moped-editor/src/App.js
```diff
@@ -51,9 +55,25 @@ const useClient = () => {
...
+
+      if (networkError) {
+        setError(networkError);
+      }
+    });
+
     return new ApolloClient({
       // Join authLink and httpLink to handle auth in each request
-      link: authLink.concat(httpLink),
+      link: from([errorLink, authLink, httpLink]),
```
mddilley, line 76: `from` does the same thing as `.concat`. We just have more than two so we need to use `from` to chain the links together. `errorLink` handles errors, `authLink` handles our auth logic needed to make successful requests, and `httpLink` sets up our connection to the Hasura graphql endpoint.

moped-editor/src/App.js
chiaberry: Could a user encounter an error in the dashboard tab, but still be able to navigate to other tabs and still do work?
I do agree that being able to log out and back would be good to have, especially after the bugs VZ encountered yesterday

chiaberry (commented):
this seems like a good way to make sure we handle errors everywhere and not just in components that we remember to wrap in the ApolloErrorHandler and return error

I am holding off on approving only because you said you are still thinking about some aspects of it

moped-editor/src/layouts/DashboardLayout/DashboardLayout.js
mddilley, line 1: Summary of changes in this file:
- Use a new error context to provide the error and error setter that comes out of our errorLink in `App.js`
- Wrap the `Outlet` component that renders the protected routes with the `ApolloErrorHandler` so that all routes will render our nice, existing error component if there is a Apollo error
- Use useEffect to clear the error so we can navigate away from the error and not have the UI be blocked by `ApolloErrorHandler`

moped-editor/src/utils/errorHandling.js
```diff
@@ -6,3 +8,28 @@
...
+/**
+ * Custom hook to access Apollo error state from anywhere in the component tree.
+ * Must be used within an ApolloErrorProvider (set up in App.js).
+ * @returns {JSX.Element} - The context provider wrapping the children
+ */
+export const useApolloErrorContext = () => {
+  const context = useContext(ApolloErrorContext);
+  if (!context) {
+    throw new Error(
+      "useApolloErrorContext must be used within an ApolloErrorProvider"
```
johnclary, line 26: nice touch 👍

johnclary (approved):
I have read the code and I have caused many errors, all of them handled quite well by the app. It's great to see the fancy Apollo features being fully leveraged here—really nice little refactor here 🚀 🚢 🚢 🚢 🚢 🚢

frankhereford (approved):
Thanks for this Mike. Tested and behaved out just like I would have expected, and I think the code looks right on the money. This is a quality refactor! 🚢

PR #1656, opened 2025-08-26
Handle expired refresh tokens

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/24016

This PR handles an expired refresh token which lasts 30 days in production by forcing the user to log out and then log back in. I also made a change to await the fetching of the user's database data before proceeding to the previous view since the dashboard view and project summary view have queries that rely on the user id being populated.

## Testing
**URL to test:** 

**Steps to test:**
1. Go to the deploy preview. If you are already signed in, sign out and then log in with one of the **Moped Test** users from our vault.
2. Go to the project list view, make sure you have a project link in view that you will click later, and then go work on something else for 60 minutes. ⏳
3. After an hour, come back to your tab that is on the project list view and click on a project link.
4. You should be logged out since the refresh token expiration is set to 60 minutes right now in our staging user pool.
5. Sign in again with the same test user and notice that you are being directed to the project details page of the link that you clicked on previously. This also tests that the user database local storage entry is populated since the summary query requires it to be defined to show whether you are following this project or not.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -135,8 +135,8 @@ export const initializeUserDBObject = async (session) => {
       const resData = await res.json();
 
       if (resData?.errors) {
-        console.error(resData.errors);
-        throw new Error("GraphQL errors occurred");
+        // Show error feedback in sign in form
+        throw new Error("Error fetching user data");
       }
 
       if (resData?.data?.moped_users) {
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -114,32 +114,44 @@ export const initializeUserDBObject = async (session) => {
   const sessionDataFromLocalStorage = getSessionDatabaseData();
 
   // If the session is valid and there is no existing data...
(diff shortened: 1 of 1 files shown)


moped-editor/src/auth/user.js
```diff
@@ -114,32 +114,44 @@ export const initializeUserDBObject = async (session) => {
...
-        "Content-Type": "application/json",
-        Authorization: `Bearer ${token}`,
-        "X-Hasura-Role": `${getHighestRole(session)}`,
-      },
-      body: JSON.stringify({
-        query: ACCOUNT_USER_PROFILE_GET_PLAIN,
-        variables: {
-          userId: getDatabaseId(session),
+    try {
+      const res = await fetch(config.env.APP_HASURA_ENDPOINT, {
```
mddilley, line 120: i ended up converting this to async/await since I'm finding it easier to follow then the promise chains.

PR #1659, opened 2025-08-28
Fix missed update to new way initializeUserDBObject works

mddilley:
## Associated issues
This fixes a third place where the user DB row needs to be set. It also updates how we handle setting user state and local storage after the SSO redirect.

## Testing
**URL to test:** 

Local if you want?

**Steps to test:**
1. Check this file to see that the other two `initializeUserDbObject` calls work this same way. They return the user data and then we set it in local storage. I just forgot to update this one.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/moped-editor/src/auth/user.js
+++ b/moped-editor/src/auth/user.js
@@ -260,7 +260,8 @@ export const UserProvider = ({ children }) => {
             // We must populate userDatabaseData if it's null otherwise the query will fail
             // when users are redirected to their last route after a forced logout
             // (see MainLayout.js and DashboardLayout.js for previous route restoration handling).
-            await initializeUserDBObject(session);
+            const userDBData = await initializeUserDBObject(session);
+            setSessionDatabaseData(userDBData);
           }
 
           setUser(session);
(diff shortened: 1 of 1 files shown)


moped-editor/src/auth/user.js
```diff
@@ -190,17 +190,9 @@ export const UserProvider = ({ children }) => {
...
       await Auth.federatedSignIn({ provider: "AzureAD" });
-
-      const session = await Auth.currentSession();
-      console.log("Logged in with SSO:", session);
-      const userDBData = await initializeUserDBObject(session);
-      console.log("User DB Data:", userDBData);
-      setSessionDatabaseData(userDBData);
-      setUser(session);
-
-      setIsLoginLoading(false);
```
mddilley, line 203: Since the SSO Amplify call initiates a redirect, this code was never reached.

moped-editor/src/auth/user.js
```diff
@@ -260,7 +252,8 @@ export const UserProvider = ({ children }) => {
             // We must populate userDatabaseData if it's null otherwise the query will fail
             // when users are redirected to their last route after a forced logout
             // (see MainLayout.js and DashboardLayout.js for previous route restoration handling).
-            await initializeUserDBObject(session);
```
mddilley, line 263: This function used to fetch the user row and set it. It now only fetches it, and I forgot to make the update to set it in this part of the code.

mddilley:
thanks y'all! merging 🚀

PR #1660, opened 2025-08-28
Fix undefined local storage user row loop

mddilley:
## Associated issues
We were getting caught in a loop where `initializeUserDBObject` only returned user details if it fetched the data but was returned `undefined` if it didn't. This was causing links to Moped pages opened in new tabs to initialize with `undefined` as the value. Then, queries that need the user DB row id would fail because they expected an int user id.

## Testing
**URL to test:** 

[Deploy preview](https://deploy-preview-1660--atd-moped-main.netlify.app/)

**Steps to test:**
1. Test adding a subproject to a project and then opening the link in a new tab. You should arrive at the subproject page in the other tab without an error.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/package.json
+++ b/moped-editor/package.json
@@ -2,7 +2,7 @@
   "name": "atd-moped-editor",
   "author": "ATD Data & Technology Services",
   "license": "CC0-1.0",
-  "version": "2.41.0",
+  "version": "2.42.0",
   "homepage": "/moped",
   "private": false,
   "repository": {


johnclary (approved):
Looks good to me 🚢

moped-editor/src/auth/user.js
```diff
@@ -136,6 +137,8 @@ export const initializeUserDBObject = async (session) => {
       throw error;
     }
   }
+
+  return sessionDataFromLocalStorage;
```
mddilley, line 141: I didn't update this function to return the user data if it already existed in the local storage entry. So, we were overwriting the valid user data object with `undefined`.

tillyw:
i'm getting an endless spinner when i try to log in to the deploy preview 🤔 

Screenshot 2025-08-28 at 1 16 59 PM

mddilley:
thanks @tillyw!! looking into this now 🔍

moped-editor/src/auth/user.js
```diff
@@ -246,23 +249,21 @@ export const UserProvider = ({ children }) => {
...
-          setUser(session);
-          setIsLoginLoading(false);
 
-          if (session === null) {
+            setUser(session);
+          } else {
             destroyProfileColor();
             deleteSessionDatabaseData();
           }
+          setIsLoginLoading(false);
```
mddilley, line 266: Need to turn off the loading spinner whether there is a session or not. Otherwise, we get stuck at the login screen with the spinner blocking any interactions to log back in.

thanks @tillyw for catching this!! 🙏

frankhereford (approved):
I had the same deal as Tilly at first, but with your fix, this tested out perfectly. Thanks!

tillyw (approved):
thanks for the fix, @mddilley ! tested and it's working smoothly for me now! opened a project in a new tab and tested dashboard queries

PR #1766, opened 2026-02-05
Update inline funding form to populate eCAPRIS fields when saving an FDU

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/26703

The eCAPRIS fields we are capturing now include the Source, the Program and the amount. In order to show the users what is being updated, I updated the lookup autocomplete to populate the three fields when someone selects an FDU. 

## Testing
**URL to test:** 

Test locally because there are metadata updates.

**Steps to test:**

Go to a project's funding table. 
Add a record manually. The FDU field is autofocused.
You can select an FDU, but some of the FDUs dont have sources or programs. To test that all the expected columns are being populated, try `820B 2507 D157` (you can type in D157 and the autocomplete will find it). The unit name, program, source and amount are automatically filled in. Don't edit any of the fields and save. 

When you click the record again to edit, you should get the override modal. Check that the "Override eCAPRIS appropriated amount" toggle is off since you did not edit the amount. 

Now try adding another record. try `820B 2507 D482` is one that has a source and an amount. Now edit the amount and any other field you feel like. Save and click to edit. 
The "Override eCAPRIS appropriated amount" should be on, since you did in fact override the amount. 

---
#### Ship list
- [x] Code reviewed 
   - Added new test: "Select an FDU like 820B 2507 D157 and see that source, program, and amount from eCAPRIS populate in the form inputs"

### DIFF
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -11,11 +11,11 @@ import { filterOptions } from "src/utils/autocompleteHelpers";
  * @param {string} field - name of Field
  * @param {Boolean} hasFocus - does field have focus in table
  * @param {String} name - name of lookup table relationship
- * @param {Array|Objects} options - the lookup table data
+ * @param {Array<Object>} options - the lookup table data
  * @param {Boolean} fullWidthPopper - should component use custom Popper component
  * @param {Object} autocompleteProps - props passed to the MUI Autocomplete Component
  * @param {Object} textFieldProps - props passed to the renderInput TextField
- * @param {Array|Objects} dependentFieldsArray - optional, array of objects {fieldName: String, setFieldValue: function
+ * @param {Array<Object>} dependentFieldsArray - optional, array of objects {fieldName: String, setFieldValue: function
  * that takes newValue as input and returns the dependent fields change}
  * @param {function} refetch - optional, function to refetch lookup table data when dropdown is opened
  *
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -16,7 +16,7 @@ import { filterOptions } from "src/utils/autocompleteHelpers";
  * @param {Object} autocompleteProps - props passed to the MUI Autocomplete Component
  * @param {Object} textFieldProps - props passed to the renderInput TextField
  * @param {Array|Objects} dependentFieldsArray - optional, array of objects {fieldName: String, setFieldValue: function
- * that newValue as input and returns the dependent fields change}
+ * that takes newValue as input and returns the dependent fields change}
  * @param {function} refetch - optional, function to refetch lookup table data when dropdown is opened
  *
  * @returns {React component}
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -33,6 +33,7 @@ const LookupAutocompleteComponent = ({
   textFieldProps,
   dependentFieldName,
   setDependentFieldValue,
+  dependentFieldsArray,
   refetch,
 }) => {
   const apiRef = useGridApiContext();
@@ -59,11 +60,14 @@ const LookupAutocompleteComponent = ({
       field,
       value: newValue,
     });
-    if (dependentFieldName) {
-      apiRef.current.setEditCellValue({
-        id,
-        field: dependentFieldName,
-        value: setDependentFieldValue(newValue),
+
+    if (dependentFieldsArray && dependentFieldsArray.length > 0) {
+      dependentFieldsArray.forEach((field) => {
+        apiRef.current.setEditCellValue({
+          id,
+          field: field.fieldName,
+          value: field.setFieldValue(newValue),
+        });
       });
     }
   };
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -15,8 +15,8 @@ import { filterOptions } from "src/utils/autocompleteHelpers";
  * @param {Boolean} fullWidthPopper - should component use custom Popper component
  * @param {Object} autocompleteProps - props passed to the MUI Autocomplete Component
  * @param {Object} textFieldProps - props passed to the renderInput TextField
- * @param {string} dependentFieldName - optional, if another field should be updated on change, name of field
- * @param {function} setDependentFieldValue - optional, takes newValue as input and returns the dependent fields change
+ * @param {Array|Objects} dependentFieldsArray - optional, array of objects {fieldName: String, setFieldValue: function
+ * that newValue as input and returns the dependent fields change}
  * @param {function} refetch - optional, function to refetch lookup table data when dropdown is opened
  *
  * @returns {React component}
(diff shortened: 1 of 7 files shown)


moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -73,6 +92,58 @@ const useColumns = ({
 }) =>
   useMemo(() => {
     return [
+      {
```
chiaberry: moved the fdu column to be the first in the table

johnclary (approved):
The updated form works great. 

I am still getting a little confused about the **SYNCED FROM ECAPRIS** distinction, when I should expect to edit my funding source in a modal, and what counts as an override. 

Some rambling thoughts on this—offered with the understanding that my mind is holding on to a legacy way of working with funding and that I don't have the larger context/business knowledge:
- When adding an FDU manually, I noticed that I can edit the Status while adding the row inline, but I cannot edit it from the modal.
- It seems that editing any of the Source, Program and/or Status also count as an "override", since editing any of those fields creates a manual funding record and removes the **SYNCED FROM ECAPRIS** label
- So perhaps we could make all fields in the modal disabled until the "Override" toggle is checked. Ie it's all or nothing when it comes to overides.
- It would be good to move the **Description** field in the modal form to be after **Program**, so as to mirror the order of the fields in the table. Ditto for the amount column, which would pair well with making the override toggle more generic and leaving it at the top of the form.
- I realized it was throwing me off a bit that I couldn't edit an FDU after I manually added a funding source. Maybe render that field in the modal form as disabled as well? I dunno!
- Could we do away with inline editing entirely, since inline editing is really only an option when creating a new FDU and/or editing one that doesn't have an FDU?
- Alternatively or additionally, it could be nice to some visual indicator in the table when a row or column is an override

And lastly, when testing locally I noticed that when editing the FDU I added manually, the modal form was taking quite a while to submit, like more than a second. Curious if others have noticed a bit of lag as well.

roseeichelmann (approved):
test steps all worked as expected!

i do find the behavior with the toggle button confusing. its weird to me that you can toggle on the "Override amount" button but not actually change the amount and then save and exit, then when you open the modal again the override toggle is on insinuating it has been overrided but you still have the OG ecapris funding amount 🤔

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
mddilley: glad to have more feedback on this 🙏 Nathan felt the same way now that FDUs drive a lot of the data that we get in this table.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -59,6 +59,25 @@ const fduAutocompleteProps = {
     value?.ecapris_funding_id === option?.ecapris_funding_id,
 };
 
+const fduAutocompleteDependentFields = [
```
mddilley: I am going to check in on the **Status** field in our next MUG because it wasn't a big topic in discussions over that last few months.

It will be interesting to check if the importance of this field has changed at all over time or in relation to the new features. I don't think that Nathan uses it in his dashboard.

Anyways, what I'm really trying to say is that this looks 100% ready for any future changes. 🙏

Charlie-Henry (commented):
I'm not seeing the source and program, maybe it's because I'm using an old snapshot?
Image

mddilley (approved):
Looks great, Chia! I tested the team and milestones tables too, and all looks great. 🚢 ✨✨✨ 

Thanks for iterating on this one with me to cover the "FDUs from other subproject ids w/ eCAPRIS amount" scenario surfaced in last MUG. I added a launch checklist item to add info about it to the funding user guide.

mddilley:
i appreciate the ramblings which are very coherent to me, @johnclary! 🙏

> When adding an FDU manually, I noticed that I can edit the Status while adding the row inline, but I cannot edit it from the modal.

I noticed this when testing too, and I'm hoping that this helps surface discussion about this status field in general. It is awkward sorting through some of the conflicts between the "old way" and "new way". I've been wondering if auto-populating "Set up" would be a way forward too since we essentially do that in the combined funding DB view for eCAPRIS records.

> So perhaps we could make all fields in the modal disabled until the "Override" toggle is checked. Ie it's all or nothing when it comes to overides.

This gets tricky because there is a need to be able to set source and program while also potentially splitting FDU amounts between Moped projects until we can make the program matching more robust. Pedestrian crossings is an example where there is overlap with AMD funding that needs to be manually split out in Moped for now.

> It would be good to move the Description field in the modal form to be after Program, so as to mirror the order of the fields in the table. Ditto for the amount column, which would pair well with making the override toggle more generic and leaving it at the top of the form.

🙏 I created https://github.com/cityofaustin/atd-data-tech/issues/26863 for this one

> Could we do away with inline editing entirely, since inline editing is really only an option when creating a new FDU and/or editing one that doesn't have an FDU?

I'm curious to see where this lands over time too. Backwards-compatibility was a big topic for legacy funding in older projects leading up to launching this. But, that is changing now that folks are seeing how we can convert legacy FDUs to work in the new way and even freshen up stale data in the process.

> Alternatively or additionally, it could be nice to some visual indicator in the table when a row or column is an override

Nathan and I were discussing this the other day too. I have a MUG agenda item to check in on this. Maybe we could move to badging like phases that makes it really clear whenever anything is eCAPRIS-sourced or overridden.

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
```diff
@@ -80,6 +80,15 @@ export const transformGridToDatabase = (gridRecord) => {
...
     : null;
 
+  const ecapris_subproject_id = gridRecord.fdu
+    ? gridRecord.fdu.ecapris_subproject_id
+    : null;
+
+  const fdu_record_amount = gridRecord.fdu ? gridRecord.fdu.amount : null;
+  // if the amount on the fdu matches what we are saving, its not an override
+  const should_use_ecapris_amount =
+    fdu_record_amount === Number(gridRecord.funding_amount);
```
mateoclarke: I wonder if there's any edge case where this strict equality could fail because of JS floating-point weirdness. It might be overkill but you could normalize both values to two decimal place/currency format to avoid that.

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -15,8 +15,8 @@ import { filterOptions } from "src/utils/autocompleteHelpers";
  * @param {Boolean} fullWidthPopper - should component use custom Popper component
  * @param {Object} autocompleteProps - props passed to the MUI Autocomplete Component
  * @param {Object} textFieldProps - props passed to the renderInput TextField
- * @param {string} dependentFieldName - optional, if another field should be updated on change, name of field
- * @param {function} setDependentFieldValue - optional, takes newValue as input and returns the dependent fields change
+ * @param {Array|Objects} dependentFieldsArray - optional, array of objects {fieldName: String, setFieldValue: function
```
mateoclarke: I got confused when I first read this because I interpreted the `|` to mean "Or" but from the details, `dependentFieldsArray` is specifically an array of objects. NOT either an array OR an object. 

So I think the correct syntax for the JSDoc param would be `Object[]` or `Array.<Object>`. Or `Array` would be fine and the simplest too. Had to poke around Stack Overflow, so its a TIL for me: https://stackoverflow.com/questions/14611995/how-to-specify-an-array-of-objects-as-a-parameter-or-return-value-in-jsdoc

mateoclarke (approved):
Good discussions on this PR. Tested locally and all works as expected from testing instructions. This FDU stuff is starting to make more sense as the UI gets refined. 👏

PR #1783, opened 2026-02-24
Update order of funding override dialog inputs to mirror inline form order and add status input to dialog

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/26863

## Testing
**URL to test:** 

You will have to test locally until the ecapris funding is in staging.

**Steps to test:**

Go to a record with an eCAPRIS ID. Toggle on sync from eCAPRIS if necessary. Here is a sample project
https://localhost:3000/moped/projects/3401?tab=funding

Click on a synced from eCAPRIS record to edit and open the override modal. 
Check that the order of the inputs now match top to bottom the left to right order of the table. 

Test updating the status and save your work. 

---
#### Ship list
- [x] Code reviewed 
   - Updated existing test for testing override form: "Click the edit button and switch on the override amount. Try clearing the status and see that it defaults to "Tentative". Enter a custom amount and save. See that the table reflects your custom amount."
   - Updated existing test for testing inline (manual) form: "Fill out all fields except status and save the row - verify that you can't use a funding amount with a decimal or non-numeric characters and that you can typeahead FDUs using any part of the number and the unit name and that the status defaulted to "Tentative""

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/OverrideFundingForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/OverrideFundingForm.js
@@ -144,74 +144,6 @@ const OverrideFundingForm = ({
     <form onSubmit={handleSubmit(onSubmit)} autoComplete="off">
       <Grid container spacing={2}>
         <Grid item xs={12}>
-          <Controller
-            name="should_use_ecapris_amount"
-            control={control}
-            render={({ field: { onChange, value } }) => (
-              <FormControlLabel
-                control={
-                  <Switch
-                    // Invert value for switch so that "on" means we are overriding and NOT using eCAPRIS amount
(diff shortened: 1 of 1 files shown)


johnclary (approved):
Looks good!! 🙌 🚢

frankhereford (approved):
This makes much more sense -- having them ordered in the same way. Thank you @chiaberry, this is great. The code makes sense and the app looks just like it should. 🙏 🚢 🚀

mddilley (changes requested):
Looks and works great! Thank you, Chia! 🙌  I have one small requested change around the fallback value for the status input. 🙏

mddilley (approved):
Thanks Chia!! 🚢 🙏

PR #1822, opened 2026-04-16
adds subproject name to project summary ecapris funding dropdown

### LINKED ISSUE #27565: Add subproject name to Moped eCAPRIS funding table
In the Oracle DB view, we have the subproject name and subproject number in each FDU row. We should add to our ETL and database so that we could make this name available in the eCAPRIS subproject id dropdown or other future places to make it even easier to get quick context on which subproject is being chosen.

**In Scope**
- Update ETL to move `SP_NAME` into a new column called `subproject_name` (text type and not nullable) in the `ecapris_subproject_funding` table
- Also add to the list of fields in the ETL's upsert mutation
- Update `ProjectSummaryProjectECapris` to display after the subproject id like `1152.04 - ATD Speed Management Program`

**Not in scope**
- Running the staging or production ETL. We can do this during the launch party to backfill this field

tillyw:
## Associated issues
closes https://github.com/cityofaustin/atd-data-tech/issues/27565

## Testing
**URL to test:** 

local

**Steps to test:**
- run migrations
- run ETL script updates in `moped-etl\ecapris-funding`
- go to a project summary page, open the eCapris Subproject ID dropbox and note that the autocomplete now shows the eCapris Subproject ID followed by a dash and the subproject name
- select a new option and save to confirm the mutation runs properly

---
#### Ship list
- [x] Code reviewed 
   - Updated existing test: "Select a eCAPRIS subproject id from the dropdown **that should show both id and subproject name in the options**, and check that you can copy and paste a link to the eCAPRIS subproject page in another browser tab."

### DIFF
--- a/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
+++ b/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
@@ -1,2 +1,1 @@
-alter table "public"."ecapris_subproject_funding" add column "subproject_name" text
- null;
+alter table "public"."ecapris_subproject_funding" add column "subproject_name" text not null;
--- a/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
+++ b/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
@@ -1,0 +1,2 @@
+alter table "public"."ecapris_subproject_funding" add column "subproject_name" text
+ null;
--- a/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
+++ b/moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
@@ -1,1 +1,2 @@
-alter table "public"."ecapris_subproject_funding" add column "subproject_name" text not null;
+alter table "public"."ecapris_subproject_funding" add column "subproject_name" text
+ null;
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
@@ -111,7 +111,11 @@ const ProjectSummaryProjectECapris = ({
               value={selectedValue}
               sx={fieldSelectItem}
               options={options}
-              getOptionLabel={(e) => e?.["ecapris_subproject_id"] ?? ""}
+              getOptionLabel={(e) =>
+                e?.["ecapris_subproject_id"] && e?.["subproject_name"]
+                  ? `${e["ecapris_subproject_id"]} - ${e["subproject_name"]}`
+                  : ""
+              } // Display subproject name in dropdown to provide more context for user when selecting
               isOptionEqualToValue={(option, value) =>
                 option?.["ecapris_subproject_id"] ===
                 value?.["ecapris_subproject_id"]
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
@@ -134,7 +134,9 @@ const ProjectSummaryProjectECapris = ({
               value={selectedValue}
               sx={fieldSelectItem}
               options={options}
-              getOptionLabel={(e) => formatOptionLabel(e)}
+              getOptionLabel={(e) =>
+                e?.["ecapris_subproject_id"] ? formatOptionLabel(e) : ""
+              }
               isOptionEqualToValue={(option, value) =>
                 option?.["ecapris_subproject_id"] ===
                 value?.["ecapris_subproject_id"]
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
@@ -47,14 +47,13 @@ const ProjectSummaryProjectECapris = ({
   const userEmail = user?.idToken?.payload?.email;
 
   // Helper: find full option object by id and format option labels
-  const findOptionById = useCallback(
-    (id) => options?.find((o) => o?.ecapris_subproject_id === id),
-    [options]
-  );
-  const formatOptionLabel = (o) => {
-    if (!o) return "";
-    const id = o?.ecapris_subproject_id ?? "";
-    const name = o?.subproject_name ?? "";
+  const findOptionById = (id) =>
+    options?.find((option) => option?.ecapris_subproject_id === id);
+
+  const formatOptionLabel = (option) => {
+    if (!option) return "";
+    const id = option?.ecapris_subproject_id ?? "";
+    const name = option?.subproject_name ?? "";
     return name ? `${id} - ${name}` : id;
   };
 
@@ -76,23 +75,6 @@ const ProjectSummaryProjectECapris = ({
   const [clearECaprisId, { loading: clearLoading }] = useMutation(
     PROJECT_CLEAR_ECAPRIS_SUBPROJECT_ID
   );
-
-  // When entering edit mode, prefer the full option object from `options`
-  // so the Autocomplete can display both the id and the name if available.
-  useEffect(() => {
-    if (!editMode) return;
-
-    if (!eCaprisSubprojectId) {
-      setSelectedValue(null);
-      return;
-    }
-
-    setSelectedValue(
-      findOptionById(eCaprisSubprojectId) ?? {
-        ecapris_subproject_id: eCaprisSubprojectId,
-      }
-    );
-  }, [editMode, findOptionById, eCaprisSubprojectId]);
 
   const handleFieldClose = () => {
     setSelectedValue(initialValue);
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
@@ -46,10 +46,11 @@ const ProjectSummaryProjectECapris = ({
   const { user } = useUser();
   const userEmail = user?.idToken?.payload?.email;
 
-  // Helper: find full option object by id and format option labels
+  // Find full option object by id and format option labels
   const findOptionById = (id) =>
     options?.find((option) => option?.ecapris_subproject_id === id);
 
+  // Format option label to include subproject name if available to provide more context for user when selecting an eCAPRIS subproject ID
   const formatOptionLabel = (option) => {
     if (!option) return "";
     const id = option?.ecapris_subproject_id ?? "";
@@ -58,9 +59,7 @@ const ProjectSummaryProjectECapris = ({
   };
 
   const initialValue = eCaprisSubprojectId
-    ? (findOptionById(eCaprisSubprojectId) ?? {
-        ecapris_subproject_id: eCaprisSubprojectId,
-      })
+    ? findOptionById(eCaprisSubprojectId)
     : null;
 
   const [editMode, setEditMode] = useState(false);
@@ -122,7 +121,7 @@ const ProjectSummaryProjectECapris = ({
               value={selectedValue}
               sx={fieldSelectItem}
               options={options}
-              getOptionLabel={(e) => formatOptionLabel(e)} // Display subproject name in dropdown to provide more context for user when selecting
+              getOptionLabel={(e) => formatOptionLabel(e)}
               isOptionEqualToValue={(option, value) =>
                 option?.["ecapris_subproject_id"] ===
                 value?.["ecapris_subproject_id"]
--- a/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
+++ b/moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useCallback } from "react";
 import ProjectSummaryIconButtons from "src/views/projects/projectView/ProjectSummary/ProjectSummaryIconButtons";
 import ProjectSummaryLabel from "src/views/projects/projectView/ProjectSummary/ProjectSummaryLabel";
 import CopyTextButton from "src/components/CopyTextButton";
@@ -47,7 +47,10 @@ const ProjectSummaryProjectECapris = ({
   const userEmail = user?.idToken?.payload?.email;
 
   // Helper: find full option object by id and format option labels
-  const findOptionById = (id) => options?.find((o) => o?.ecapris_subproject_id === id);
+  const findOptionById = useCallback(
+    (id) => options?.find((o) => o?.ecapris_subproject_id === id),
+    [options]
+  );
   const formatOptionLabel = (o) => {
     if (!o) return "";
     const id = o?.ecapris_subproject_id ?? "";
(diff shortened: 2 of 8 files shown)


tillyw:
this may be a bit tricky to test because of the `not null` constraint for the new `subproject_name` field. i was able to implement this by allowing null values initially, then running the etl, then adding the null constraint. does anyone have a suggestion for how to get around this?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -46,10 +46,22 @@ const ProjectSummaryProjectECapris = ({
   const { user } = useUser();
   const userEmail = user?.idToken?.payload?.email;
 
+  // Helper: find full option object by id and format option labels
+  const findOptionById = useCallback(
+    (id) => options?.find((o) => o?.ecapris_subproject_id === id),
+    [options]
+  );
+  const formatOptionLabel = (o) => {
```
chiaberry: It would be easier to read if this o was spelled out as option, and could you add some comments for what this function is doing?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -65,6 +77,23 @@ const ProjectSummaryProjectECapris = ({
...
+      return;
+    }
+
+    setSelectedValue(
+      findOptionById(eCaprisSubprojectId) ?? {
+        ecapris_subproject_id: eCaprisSubprojectId,
+      }
+    );
+  }, [editMode, findOptionById, eCaprisSubprojectId]);
+
```
chiaberry: I don't think this useEffect is doing anything, I tested my hunch by putting in some console logging and I only see it being called with editMode changes. I thought it was going enter in a race condition with the Autocomplete's onChange, but it looks like the state is handled internally to the autocomplete and changing the selection isn't affecting the `eCaprisSubprojectId` until saving

chiaberry (changes requested):
I left a comment on the useEffect, but I dont think it is doing anything. Could you check me and let me know what I am missing? 

It would be helpful to update the test instructions to edit the migration and remove the not null constraint. (Or remove the constraint and it is something added later?) I was fussing with the errors until I read your comments and saw that the up migrations aren't going to work as written.

mateoclarke:
I see the `WIP` label on this. Is it ready for review or still being worked on?

tillyw:
@chiaberry @mateoclarke apologies for the confusion! i had the wip tag on there because i wasn't sure how to approach the not null constraint issue but we discussed it in dev sync and decided it would be best to just remove it for now in order to not block this pr and add the constraint in once those fields have been populated through the etl if necessary.

also @chiaberry i'll see if i can get rid of that useEffect!

tl;dr please go ahead with reviews!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -46,10 +46,21 @@ const ProjectSummaryProjectECapris = ({
...
+  const formatOptionLabel = (option) => {
+    if (!option) return "";
+    const id = option?.ecapris_subproject_id ?? "";
+    const name = option?.subproject_name ?? "";
+    return name ? `${id} - ${name}` : id;
+  };
+
+  // Display subproject name in summary view if available
   const initialValue = eCaprisSubprojectId
-    ? {
```
tillyw, line 50: i don't think we need the fallback option to display just the id, it will either be the id + subproject name or null afaik, but i can add it back if we want to be safe!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: done!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: thanks for catching that, this was an artifact that was no longer serving its purpose!

tillyw:
> I see the `WIP` label on this. Is it ready for review or still being worked on?

it's ready for review! i removed the label

tillyw:
> I left a comment on the useEffect, but I dont think it is doing anything. Could you check me and let me know what I am missing?
> 
> It would be helpful to update the test instructions to edit the migration and remove the not null constraint. (Or remove the constraint and it is something added later?) I was fussing with the errors until I read your comments and saw that the up migrations aren't going to work as written.

sorry for the confusion! i went ahead and removed the `not null` constrain for the time being. @mddilley are you okay with leaving it as a nullable field? otherwise, can we add the null constraint after we've run the etl?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
chiaberry: Kinda confused by this comment here, which fallback option is removed?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -46,10 +46,21 @@ const ProjectSummaryProjectECapris = ({
...
+
+  // Format option label to include subproject name if available
+  const formatOptionLabel = (option) => {
+    if (!option) return "";
+    const id = option?.ecapris_subproject_id ?? "";
+    const name = option?.subproject_name ?? "";
+    return name ? `${id} - ${name}` : id;
+  };
+
+  // Display subproject name in summary view if available
```
chiaberry: should this comment go on line 182 instead?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: the one that would display just the subproject id!

chiaberry: but doesnt it still exist in the formatOptionLabel function? This sets the initialValue and isnt handling the display, thats the other function right?

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -46,10 +46,21 @@ const ProjectSummaryProjectECapris = ({
...
+  const findOptionById = (id) =>
+    options?.find((option) => option?.ecapris_subproject_id === id);
+
+  // Format option label to include subproject name if available
+  const formatOptionLabel = (option) => {
+    if (!option) return "";
+    const id = option?.ecapris_subproject_id ?? "";
+    const name = option?.subproject_name ?? "";
+    return name ? `${id} - ${name}` : id;
+  };
```
chiaberry: I think this should be pulled out of the component and memoized, it keeps reprocessing everytime i scroll through the options. I am trying to find a similar example in other parts of our code.

chiaberry: this isnt quite the same, but it is what i am thinking of https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/views/projects/projectView/ProjectMilestones.js#L34

chiaberry (commented):
good to see the useEffect could be removed! I left some more comments and also I tested the field on the funding table, and I dont see the dropdown options there. I think the query is missing the new field for that one, but I didnt confirm.

tillyw:
> good to see the useEffect could be removed! I left some more comments and also I tested the field on the funding table, and I dont see the dropdown options there. I think the query is missing the new field for that one, but I didnt confirm.

@chiaberry i didn't include it in the funding table because it wasn't listed as a bullet point in the issue description but now i'm reminded it is in the title >.< i'll add that!!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: you're right it doesn't display it, it just sets the value. the `findOptionById` helper function returns the option object which includes the subproject name that is displayed in the text field as the existing value when the user opens the dropdown. we no longer need to set the initial value as just the subproject id as we were doing before, so i removed that!

tillyw: i started to worry there might be a case where we only have the ecapris subproject id and not the options array so i put this back in as it was originally! you can disregard all this

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -22,6 +22,27 @@ import {
...
+      if (!option) return "";
+      const id = option?.ecapris_subproject_id ?? "";
+      const name = option?.subproject_name ?? "";
+      const key = `${id}|${name}`;
+      if (cache.has(key)) return cache.get(key);
+      const label = name ? `${id} - ${name}` : id;
+      cache.set(key, label);
+      return label;
+    };
+  }, []);
```
mddilley: @tillyw since this is defined outside of the component, can you remove the memo? Since there are no dependencies this isn't doing anything different than a plain JS function. The caching map could also be removed since I'm not convinced it is optimizing this since the goal of the function is handling primitive strings which isn't expensive.

I think that a more readable version could be something like:
```js
getOptionLabel={(option) => option ? `${option.ecapris_subproject_id} - ${option.subproject_name}` : ""} 
```

I didn't suggest a fallback since there are no eCAPRIS subprojects with null names in eCAPRIS.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -168,7 +193,16 @@ const ProjectSummaryProjectECapris = ({
...
+              // Display subproject name in summary view if available
+              text={
+                eCaprisSubprojectId
+                  ? formatOptionLabel(
+                      findOptionById(options, eCaprisSubprojectId) ?? {
+                        ecapris_subproject_id: eCaprisSubprojectId,
+                      }
+                    )
+                  : ""
+              }
```
mddilley: @tillyw I think that showing the eCAPRIS subproject id alone here is sufficient. The use case here was to be able to identify and typeahead through the eCAPRIS options, and I didn't create any requirements around displaying the names in the project summary because some of them can be quite long.

I think that it makes it less awkward to not have to cobble the data together as well since we don't store the subproject names in the `moped_project` database table. Can you switch back to `text={eCaprisSubprojectId ? eCaprisSubprojectId : ""}`?

mddilley (changes requested):
Looking good, Tilly! I requested two changes before this ships. And thank y'all for troubleshooting the not null constraint curveball. ⚾ 🙏 I like the solution of removing the constraint and then handling in a follow-up task. 💯

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: done!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: done!

tillyw:
thanks for your feedback @mddilley ! made these updates and this is ready for rereview!

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -47,9 +59,9 @@ const ProjectSummaryProjectECapris = ({
   const userEmail = user?.idToken?.payload?.email;
 
   const initialValue = eCaprisSubprojectId
-    ? {
+    ? (findOptionById(options, eCaprisSubprojectId) ?? {
```
mddilley, line 62: @tillyw i noticed that, since `initialValue` depends on the async options, it sometimes includes stale data which causes the `subproject_name` to show as `undefined`

Image<br />

I think that you could:
1. Set the initial value of `selectedValue` as `null` (as well as when state is reset in `handleFieldClose` and in the catch block of `handleFieldSave`)
2. Call `setSelectedValue` in the function passed to `onClickEdit` in the `ProjectSummaryLabel` component and use the same value calculated here on line 62 for `initialValue`

this would make it more event-based too (the subproject name in the initial selection isn't needed until editing starts so it can be called then to make sure the latest option data is used). 

icing on the cake would be adding `loading` to the early return in the function passed to `onClickEdit` to block users from editing until the options are loaded

mddilley (commented):
Thanks for updating, Tilly! Could you merge the latest from `main` into this branch? I had to do that to get the local stack running to test. I also noticed a small bug that stems from a pattern that i started in the eCAPRIS input component. Let me know if you have any questions - i'm happy to walk through it. 🙏

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
tillyw: @mddilley thank you for catching that and for the suggested fixes! i pushed an update :)

mddilley (approved):
Tested this many different ways, and this looks great! Thanks for working through this one, and I'm excited to get this improvement into the hands of users! 🚢 🙌

mddilley:
also here is the follow-up issue that I forgot to link earlier in the PR https://github.com/cityofaustin/atd-data-tech/issues/28070. thanks everyone again for coming up with the solution. 🙏

moped-database/migrations/default/1776300831851_alter_table_public_ecapris_subproject_funding_add_column_subproject_name/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."ecapris_subproject_funding" add column "subproject_name" text
+ null;
```
johnclary, line 2: I was going to suggest indexing this column, but i see we're not actually searching on this column via graphql query so this is not needed.

moped-editor/src/views/projects/projectView/ProjectSummary/ProjectSummaryProjectECapris.js
```diff
@@ -111,7 +123,9 @@ const ProjectSummaryProjectECapris = ({
               value={selectedValue}
               sx={fieldSelectItem}
               options={options}
```
johnclary, line 125: I noticed that this autocomplete isn't afraid to show you hundreds of options in the dropdown menu. Some other autocompletes use the `filterOptions` function from `autocompleteHelpers.js` to cap the number of options displayed. Something to consider here since this list can be quite long.

johnclary (approved):
Looks great! really nice to have the project name when picking an ID.

Just want to note that it might be worth revisiting the caching behavior of the `COMBINED_FUNDING_QUERY`, since it's pulling quite a lot of data whenever the funding tab focuses. One idea would be to split out the lookup values from the project-specific data so that those are fetched once per user session.

chiaberry (approved):
🚢 

Would be nice to add the `filterOptions` John mentioned, to help with the name load in the dropdown, like in the subproject drop down

tillyw:
thank you for the feedback! i will add the `filterOptions` helper in the morning!

mateoclarke (commented):
Step 3 in the testing steps is tripping me up:
> go to a project summary page and note that the eCapris Subproject ID field now shows the eCapris Subproject ID followed by a dash and the subproject name

I do not see the "ID - subproject" name as described.

Maybe the requirements changed and your testing steps are out of date?

I'm not sure what I'm doing wrong. I've run migrations, metadata updates, and I ran the ecapris-funding ETL. On the summary page & funding table, I see the eCapris subproject ID but no name. When I click the ID, I see the dropdown which DOES have the `ID - name` format described. If I select a new value from the dropdown and click check to confirm, the name disappears and I only see the ID number again. 

This isn't the expected behavior, is it? Am I missing a configuration step?

Image

Image

mateoclarke (approved):
Please ignore my previous comment. Got clarified in backlog refinement 

👍 🚢

mddilley:
thanks Tilly! going to merge this one, and then we can handle #1833 after today's release. 🚀

PR #1820, opened 2026-04-13
[Collector] Attach files to eCAPRIS funding

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/27387

This PR implements file attachments for eCAPRIS funding records. This branch will be the collector for shipping attachments for the entire funding table.

## Testing
**URL to test:** 

Local

**Steps to test:**

👋 **Before testing, note that this branch only covers attaching files to eCAPRIS records. The next issue will extend attachments to the Moped-entered funding records so this PR will not ship alone.**

**I'll gather feedback from y'all and MUG, and I'll move the larger changes into the next issue that covers attaching files to Moped funding records (non-eCAPRIS).**

1. Create a new project and then select an eCAPRIS subproject id like `10553.009`
2. See that funding rows appear in the table and that each row now shows a paperclip icon
3. See that the funding table now has a "Files" column
4. Click one of the paperclips to open the attachment dialog and see that the file upload dialog appears
5. See that the bottom part of the dialog shows "Attached files" with none attached yet
6. Now, toggle the "Link to file" switch on and enter a name like "Important funding document", type, description, and a drive path like `G:\ATD\Arterial Management\Programs\PHBs\PHB Studies & Ped Counts\Studies\Huntland Dr, 300 Blk (St Francis)`. Save.
7. Notice that the dialog closes and there is now a condensed version of the path showing in the **Files** column of the row that you tested on. See that you can click the copy button to copy the path. Hover copy button to see "Copy network location" text.
8. Now, go to the Files tab and see the file that you just created with the path not condensed and also copyable. Notice that you can edit it. Edit your file name to "Unimportant funding document".
9. Go back to the **Funding** tab and see that the name updated here too.
10. Now, click the attachment button on the same row. See that you can see your attachment in the bottom part of the dialog with an unlink icon next to it. We'll unlink in a few steps.
11. Add two more attachments:
      - One with file name of "Plans 4/11/2026" and then a link of `https://www.austinmobility.io/`
      - Another with file name "Plans 4/13/2026" and then a link of `https://atd.knack.com/amd`
12. Notice that both attachments show the file names in the **Files** column but show the url on hover. Also notice that our friendly url to convert Data Tracker links to `AMD Data Tracker` is not applied.
13. Hover the names in the **Files** column and see that you can see the full url or path if you need.
[...]

PR #1824, opened 2026-04-20
Attach new files to Moped funding records 

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/27483

## Testing
**URL to test:** 
test locally, use prod data

**Steps to test:**

This PR builds off of the work in https://github.com/cityofaustin/atd-moped/pull/1820

1. Open a project (or create a new one)
2. Navigate to the funding table. Select an eCAPRIS subproject id like `10553.009`
3. Like the previous PR, you see funding rows, paperclip icons and Files column. 
4. Now either add another funding row (not imported from eCAPRIS) or edit an existing row so the record is not synced from eCAPRIS.

5. Click the paperclip to open the attachment dialog. 
6. Toggle the link to file switch and enter a name, type, description and a drive path or link. 

7. You can see the file in the files column of this table, also go to the Files tab and see it there as well. 

8. use your db client to confirm that this record is in the correct table.

```sql
select * from files_project_funding fpf 
```

9. Click the paperclip to open the attached files dialog again. Now unlink the file. 
The file is still on the Files tab but is no longer on the funding table. 

---
#### Ship list
- [ ] Code reviewed 
   - Added items to launch checklist in https://github.com/cityofaustin/atd-moped/pull/1820

### DIFF
--- a/moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
+++ b/moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
@@ -1,7 +1,7 @@
 -- Create join table to link Moped projects, moped project funding rows, and Moped project files
 CREATE TABLE public.files_project_funding (
     id SERIAL PRIMARY KEY,
-    project_id int4 NOT NULL REFERENCES public.moped_project(project_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    entity_id int4 NOT NULL REFERENCES public.moped_proj_funding(proj_funding_id) ON DELETE RESTRICT ON UPDATE CASCADE,
     file_id int4 NOT NULL REFERENCES public.moped_project_files(project_file_id) ON DELETE RESTRICT ON UPDATE CASCADE,
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
@@ -9,8 +9,8 @@ CREATE TABLE public.files_project_funding (
     updated_by_user_id int4 NOT NULL REFERENCES moped_users (user_id) ON DELETE RESTRICT ON UPDATE CASCADE
 );
 
-COMMENT ON TABLE public.files_project_funding IS 'Join table to link Moped projects, Moped funding rows, and Moped project files to create funding file attachments.';
-COMMENT ON COLUMN public.files_project_funding.project_id IS 'References the Moped project to which the file attachment belongs.';
+COMMENT ON TABLE public.files_project_funding IS 'Join table to link Moped funding rows and Moped project files to create funding file attachments.';
+COMMENT ON COLUMN public.files_project_funding.entity_id IS 'References the Moped project funding record to which the file attachment belongs.';
 COMMENT ON COLUMN public.files_project_funding.file_id IS 'References the file that is attached to the moped funding row.';
 COMMENT ON COLUMN public.files_project_funding.created_at IS 'Timestamp for when the record was created.';
 COMMENT ON COLUMN public.files_project_funding.updated_at IS 'Timestamp for when the record was last updated.';
--- a/moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
+++ b/moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
@@ -1,0 +1,18 @@
+-- Create join table to link Moped projects, moped project funding rows, and Moped project files
+CREATE TABLE public.files_project_funding (
+    id SERIAL PRIMARY KEY,
+    project_id int4 NOT NULL REFERENCES public.moped_project(project_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    file_id int4 NOT NULL REFERENCES public.moped_project_files(project_file_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    created_by_user_id int4 NOT NULL REFERENCES moped_users (user_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    updated_by_user_id int4 NOT NULL REFERENCES moped_users (user_id) ON DELETE RESTRICT ON UPDATE CASCADE
+);
+
+COMMENT ON TABLE public.files_project_funding IS 'Join table to link Moped projects, Moped funding rows, and Moped project files to create funding file attachments.';
+COMMENT ON COLUMN public.files_project_funding.project_id IS 'References the Moped project to which the file attachment belongs.';
+COMMENT ON COLUMN public.files_project_funding.file_id IS 'References the file that is attached to the moped funding row.';
+COMMENT ON COLUMN public.files_project_funding.created_at IS 'Timestamp for when the record was created.';
+COMMENT ON COLUMN public.files_project_funding.updated_at IS 'Timestamp for when the record was last updated.';
+COMMENT ON COLUMN public.files_project_funding.created_by_user_id IS 'References the user who created the file attachment record.';
+COMMENT ON COLUMN public.files_project_funding.updated_by_user_id IS 'References the user who last updated the file attachment record.';
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
@@ -36,7 +36,10 @@ const ProjectFundingFilesAttachmentDialog = ({
   dataLookups,
   rows,
 }) => {
-  const fundingRecord = rows.find((row) => row.id === fileAttachmentId); // should i memoize this
+  const fundingRecord = useMemo(
+    () => rows.find((row) => row.id === fileAttachmentId),
+    [rows, fileAttachmentId]
+  );
   const isSyncedFromECapris = fundingRecord.is_synced_from_ecapris;
   const [addFundingFileAttachment] = useMutation(
     isSyncedFromECapris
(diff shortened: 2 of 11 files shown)


moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
```diff
@@ -320,9 +320,12 @@ export const useColumns = ({
         editable: false,
         sortable: false,
         renderCell: ({ row }) => {
+          const filesType = row.is_synced_from_ecapris
+            ? "ecapris_funding_files"
+            : "moped_funding_files";
           return (
             <Stack direction="column" spacing={0.5}>
-              {row.ecapris_funding_files.map((file_record, index) => {
+              {row?.[filesType].map((file_record, index) => {
```
mddilley, line 331: @chiaberry there is a small bug when creating the first Moped funding row in a new project where this line tries to map on `undefined` and breaks.

I made a mistake in my last PR where I added `ecapris_funding_files` to the object added to new rows in `handleAddRecordClick`. could you remove while you are in here? that keeps that object as a template of only the fields required to insert.

I think removing that and then handling the case when a record isn't saved yet in this block would fix. Maybe an early return with a comment? I'm sure there are other routes but that is what i'm seeing right now.

mddilley (commented):
Looking great, Chia! Cool to see that the branching between the eCAPRIS and Moped records isn't bad at all imo. 😎 I found a small bug, but I was able to patch, test, and see attachments to both records types. 🙌 

I'll come back and wrap the review once you've had a chance to look at the one comment.

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
frankhereford: Thank you Mike, I was just about to report the same thing. Chia, FWIW, I asked cursor to take a hack at fixing it, and it offered up this diff. I have not verified anything about it except that it appears to work, so please take it with a grain of salt. Thanks!

```diff
diff --git a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
index d99d54764..031af8288 100644
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
@@ -216,6 +216,7 @@ const ProjectFundingTable = ({
         isNew: true,
         proj_funding_id: id,
         is_manual: true,
+        moped_funding_files: [],
         ecapris_funding_files: [],
       },
       ...oldRows,
diff --git a/moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js b/moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
index 2ba6e4347..7c65dc4c1 100644
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
@@ -323,9 +323,10 @@ export const useColumns = ({
           const filesType = row.is_synced_from_ecapris
             ? "ecapris_funding_files"
             : "moped_funding_files";
+          const files = row?.[filesType] ?? [];
           return (
             <Stack direction="column" spacing={0.5}>
-              {row?.[filesType].map((file_record, index) => {
+              {files.map((file_record, index) => {
                 const file = file_record.moped_project_file;
 
                 if (!file) return null;
```

frankhereford (commented):
When I get to step 4 in the testing instructions, "Now either add another funding row (not imported from eCAPRIS) or edit an existing row so the record is not synced from eCAPRIS.", when I submit that form, I get the following trapped error:

`field 'moped_funding_files' not found in type: 'moped_proj_funding_insert_input'`

I think the field may be missing from the `gridRecord` assignment on line 110 of the ProjectFunding `helpers.js` file.

I think this is a blocker on testing, and I will come back to this PR after dev sync today when I can ask about what you think about these findings. 

Thanks for this PR Chia!! I feel rusty reviewing moped, so if I'm just out in left field, lmk and i'll do whatever I am missing. Thanks!

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
chiaberry: Thanks for testing and catching this bug, I pushed a fix where I return early if there are no files or array of files. 

@frankhereford thanks for including the code suggestion, Mike asked me to remove the `ecapris_funding_files` from the new row object, so I am not going to add the `moped_funding_files` to the object. I'll test a few more times, but I think this fixes the bug

mddilley (approved):
I tested this every which way that I could, and I was able to attach and detach files to both eCAPRIS and Moped funding records. ✨  Thanks and this also helped me catch a small bug that I introduced into the local stack activity log that I'll PR shortly. 🚢 🙌  🚢

mddilley:
@chiaberry I was reviewing issues today, and I realized that I missed scoping the detail of clearing Moped funding file attachments when the file itself is soft-deleted. So, I created https://github.com/cityofaustin/atd-data-tech/issues/28174 to capture in future work.

Charlie-Henry:
I got the same error as @frankhereford 
Screenshot 2026-05-01 at 11 23 16 AM

chiaberry:
@Charlie-Henry I tried running through the test steps again and couldnt replicate the error. did it happen on a new empty project, or was it on an existing project?

moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
```diff
@@ -0,0 +1,18 @@
+-- Create join table to link Moped projects, moped project funding rows, and Moped project files
+CREATE TABLE public.files_project_funding (
+    id SERIAL PRIMARY KEY,
+    entity_id int4 NOT NULL REFERENCES public.moped_proj_funding(proj_funding_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    file_id int4 NOT NULL REFERENCES public.moped_project_files(project_file_id) ON DELETE RESTRICT ON UPDATE CASCADE,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
johnclary, line 7: i can't remember how the `updated_at` timestamp is managed. is this missing a trigger?

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
```diff
@@ -320,9 +320,15 @@ export const useColumns = ({
         editable: false,
         sortable: false,
         renderCell: ({ row }) => {
+          const filesType = row.is_synced_from_ecapris
+            ? "ecapris_funding_files"
+            : "moped_funding_files";
+          if (!row?.[filesType]) {
+            return;
+          }
```
johnclary, line 328: This awkward accessor problem sorta hints to me that it could be nice to combine the two files arrays upstream (client side, or in a view?) so that you can simply map over them here. 

Depending on where this functionality of arbitrary file-entity attachments is heading, it might be worth it now to make a `project_files_view` or whatever that organizes this data more cleanly for the app to consume it.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
johnclary, line 1: Couple of impressions now that I'm seeing this modal for a second time:

1. The modal header has slightly different padding and is missing the close button (`X`) vis a vis the funding record edit modal.
2. I think it could be good to move the submit/cancel buttons to be immediately below the attach file form, to make it clear the attached files are not part of the form/action.
3. I wonder if moving the attached files to the top of the modal would make this slightly more intuitive 🤔 

Image
Image

johnclary (approved):
I tested this heavily and it works beautifully! Nice!

moped-database/migrations/default/1776441267959_funding-file-attach/up.sql
mddilley: i left this out of scope since we were only creating and hard-deleting records, but since I'm rolling the soft deletes into https://github.com/cityofaustin/atd-data-tech/issues/28174 - i'll add this there too. that way we will be able to see when the soft-delete update happened on the attachment tables. 🙏

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
mddilley: this is confirming a feeling that i've had that we could benefit from a reusable dialog component. this one is baked into the file upload dialog currently, and i felt friction when writing code in here that could be reduced by pulling out that dialog wrapper (along with keeping our dialogs looking similar across the board).

I'm going to aim to implement these suggestions in https://github.com/cityofaustin/atd-data-tech/issues/27484

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
```diff
@@ -34,28 +36,53 @@ const ProjectFundingFilesAttachmentDialog = ({
   dataLookups,
   rows,
 }) => {
+  const fundingRecord = useMemo(
+    () => rows.find((row) => row.id === fileAttachmentId),
+    [rows, fileAttachmentId]
+  );
+  const isSyncedFromECapris = fundingRecord.is_synced_from_ecapris;
```
mateoclarke: wondering here if fundingRecord could ever return null (maybe if the `fileAttachmentId` is stale) and could throw an error on line 43. Maybe add a guard like:
```js
const isSyncedFromECapris = fundingRecord?.is_synced_from_ecapris ?? false;
```

mateoclarke (approved):
Tested locally and looks good. 

_I was having a similar error that @Charlie-Henry and @frankhereford called out, but running `hasura metadata apply` seemed to resolve it for me._

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingFilesAttachmentDialog.js
chiaberry: adding! thank you

chiaberry:
Going to merge down to the collector branch, thank you all for your reviews and testing 🫡

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
mddilley: Mentioned to Chia last Friday and forgot to follow up here. I think this is a great idea, and we are going to explore this once getting closer to delivering this whole thing. 🙏

PR #1826, opened 2026-04-22
Prevent out of range error in dollar fields

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/27326

<details>
<summary>My rationale when trying to squeeze this in, before getting more feedback</summary>
I picked up this snackoo yesterday thinking I could potentially squeeze into the release today, but I am not sure and think it might need some UX feedback. 

The bug is that numbers larger than 2,147,183,647 cant be saved in the db because they are too big and we get a `DataError: integer out of range`. I thought about changing the column type in the db, but the issue says to handel it on the front end. 

We strip decimal places and letters from the DollarAmountIntegerField without any feedback, but reducing huge numbers to their upper limit without any indication to the user why felt weird to me. So I instead am checking the field on the funding table and then indicating an error on the field if the number is too big. 

And for the work activities table I updated the yup resolver. But I am not sure if this is the behavior we want. At least it prevents the mutation from crashing. 
</details>

## Testing
**URL to test:** 

https://deploy-preview-1826--atd-moped-main.netlify.app/moped/projects/2207?tab=funding

**Steps to test:**

Add a new funding record. Try to input a giant number as the amount like 10,000,000,000 and see that you cant save the record. 

Do the same for the work activities table amount. 

---
#### Ship list
- [x] Code reviewed 
   - Removed note from script

### DIFF
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -65,7 +65,8 @@ const LookupAutocompleteComponent = ({
           id,
           field: field.fieldName,
           value: field.setFieldValue(newValue),
-          debounceMs: 200
+          // multiple fields being updated and potentially validated, debounce helps prevent race conditions
+          debounceMs: 100
         });
       });
     }
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -52,23 +52,21 @@ const LookupAutocompleteComponent = ({
     }
   }, [open, refetch]);
 
-  const handleChange = (event, newValue) => {
-    apiRef.current.setEditCellValue({
+  const handleChange = async (event, newValue) => {
+    await apiRef.current.setEditCellValue({
       id,
       field,
       value: newValue,
     });
 
     if (dependentFieldsArray && dependentFieldsArray.length > 0) {
-      dependentFieldsArray.forEach((field) => {
-        apiRef.current.setEditCellValue({
+      for (const field of dependentFieldsArray) {
+        await apiRef.current.setEditCellValue({
           id,
           field: field.fieldName,
           value: field.setFieldValue(newValue),
-          // multiple fields being updated and potentially validated, debounce helps prevent race conditions
-          debounceMs: 100
         });
-      });
+      }
     }
   };
 
--- a/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
+++ b/moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
@@ -65,6 +65,7 @@ const LookupAutocompleteComponent = ({
           id,
           field: field.fieldName,
           value: field.setFieldValue(newValue),
+          debounceMs: 200
         });
       });
     }
--- a/moped-editor/src/utils/numberFormatters.js
+++ b/moped-editor/src/utils/numberFormatters.js
@@ -24,3 +24,5 @@ export const removeDecimalsAndTrailingNumbers = (number) =>
  * @return {String} Number string with only 0-9 integer characters
  */
 export const removeNonIntegers = (number) => number.replace(/[^0-9]/g, "");
+
+export const isAmountOutOfRange = number => number >= 2_147_483_647;
--- a/moped-editor/src/utils/numberFormatters.js
+++ b/moped-editor/src/utils/numberFormatters.js
@@ -25,4 +25,5 @@ export const removeDecimalsAndTrailingNumbers = (number) =>
  */
 export const removeNonIntegers = (number) => number.replace(/[^0-9]/g, "");
 
+// check if number exceeds the maximum number that can be saved in the db
 export const isAmountOutOfRange = (number) => number >= 2_147_483_647;
--- a/moped-editor/src/utils/numberFormatters.js
+++ b/moped-editor/src/utils/numberFormatters.js
@@ -25,4 +25,4 @@ export const removeDecimalsAndTrailingNumbers = (number) =>
  */
 export const removeNonIntegers = (number) => number.replace(/[^0-9]/g, "");
 
-export const isAmountOutOfRange = number => number >= 2_147_483_647;
+export const isAmountOutOfRange = (number) => number >= 2_147_483_647;
--- a/moped-editor/src/utils/numberFormatters.js
+++ b/moped-editor/src/utils/numberFormatters.js
@@ -26,6 +26,7 @@ export const removeDecimalsAndTrailingNumbers = (number) =>
 export const removeNonIntegers = (number) => number.replace(/[^0-9]/g, "");
 
 export const INT_4_MAX = 2_147_483_647;
+export const outOfRangeErrorMessage = "Amount is out of range";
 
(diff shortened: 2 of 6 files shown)


moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.js
```diff
@@ -220,7 +220,13 @@ const ProjectWorkActivitiesForm = ({
               control={control}
               onChangeHandler={amountOnChangeHandler}
               size="small"
+              error={formErrors.contract_amount}
             />
+            {formErrors?.contract_amount && (
+              <FormHelperText>
+                {formErrors.contract_amount.message}
+              </FormHelperText>
```
johnclary, line 229: I always forget what the gotcha is, but this help text is gray instead of red

Image

moped-editor/src/utils/numberFormatters.js
```diff
@@ -24,3 +24,6 @@ export const removeDecimalsAndTrailingNumbers = (number) =>
  * @return {String} Number string with only 0-9 integer characters
  */
 export const removeNonIntegers = (number) => number.replace(/[^0-9]/g, "");
+
+// check if number exceeds the maximum number that can be saved in the db
+export const isAmountOutOfRange = (number) => number >= 2_147_483_647;
```
johnclary: taking your word for it that `params.props.value` (below) is always a number. maybe add `Number(number)` if in doubt?

johnclary (approved):
Tested + works great!

roseeichelmann (approved):
Works as expected :)

I did notice something weird (not related to this PR)  that you can save a really large number in a new funding record, but then when you edit it, the amount might be more than the eCAPRIS appropriated amount so you cant save your edits until you you go under that amount.

it seems off that you can put in a number when creating a new record but you cant save that number in edit mode. video below

https://github.com/user-attachments/assets/4880d9d4-4dbd-4f30-b07b-49ff7f152d4b

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
```diff
@@ -205,6 +206,10 @@ const useColumns = ({
         field: "funding_amount",
         width: 100,
         editable: true,
+        preProcessEditCellProps: (params) => ({
+          ...params.props,
+          error: isAmountOutOfRange(params.props.value),
```
mddilley: @chiaberry could you try updating this to return two keys in the object: `error` and `errorMessage` when the validation fails? That way the `DollarAmountIntegerField` input can take a new prop for the message and won't need the hard-coded "out of range" string and will open the door to future validation message we could need later?

Could you also add to the docstrings in `DollarAmountIntegerField` to reflect the change? 🙏

moped-editor/src/utils/numberFormatters.js
mddilley: @chiaberry looks like the type going into this function is either string or null so I think returning `false` if null and then also using the Number constructor like John suggested works. could you update?

mddilley: also i see this number in a couple of places so could you update to a constant called something like `INT_4_MAX`? could help us connect this to a database requirement in the future.

moped-editor/src/views/projects/projectView/ProjectWorkActivity/utils/form.js
```diff
@@ -36,7 +36,10 @@ export const activityValidationSchema = yup.object().shape({
     .string()
     .max(500, "Must be less than 500 characters")
     .nullable(),
-  contract_amount: yup.number().nullable(),
+  contract_amount: yup
+    .number()
+    .max(2_147_483_647, "Amount is out of range") // max number an int can be in the db
```
mddilley: @chiaberry this helper text is also used in a couple of places. could you convert into a constant that uses the string you used here? That would help us keep these consistent in the places where dollar amounts are used.

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.js
mddilley: I think this just needs the `error` prop added to make it read since it will always display an error.

mddilley (changes requested):
I requested a few small code changes, but I also noticed the amount is not automatically populating when selecting an FDU in a manual record anymore. Could you check that out? 🙏

moped-editor/src/views/projects/projectView/ProjectFunding/DollarAmountIntegerField.js
```diff
@@ -43,14 +52,16 @@ const DollarAmountIntegerField = ({ id, value, field, hasFocus }) => {
   return (
     <TextField
       variant="standard"
-      style={{ width: "80px" }}
+      style={{ minWidth: "80px" }}
```
chiaberry, line 55: changed to minWidth so the field can expand if the column is wider

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.js
chiaberry: yes! added to the form control to make the whole thing red.

moped-editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.js
```diff
@@ -203,6 +203,7 @@ const ProjectWorkActivitiesForm = ({
               name="work_assignment_id"
               control={control}
               size="small"
+              error={!!formErrors?.work_assignment_id}
```
chiaberry, line 206: this one and the description one below are tag along fixes, I noticed that the inputs aren't highlighting red with an error for these fields.

chiaberry:
@mddilley this is ready for re-review. I was stumped for a while on the amount not populating when selecting an FDU, and I eventually came across this https://github.com/mui/mui-x/issues/5089#issuecomment-1145441241 comment about inputs behaving erratically. And I think it is the preprocessing validation was conflicting with the setEditCellValue function calls, because the validation happens everytime any cell changes value.

moped-editor/src/components/DataGridPro/LookupAutocompleteComponent.js
```diff
@@ -65,6 +65,8 @@ const LookupAutocompleteComponent = ({
           id,
           field: field.fieldName,
           value: field.setFieldValue(newValue),
+          // multiple fields being updated and potentially validated, debounce helps prevent race conditions
+          debounceMs: 100
```
mddilley: @chiaberry i was reading about this setting and saw that [`setEditCellValue` returns a promise.](https://v7.mui.com/x/api/data-grid/grid-api/#grid-api-prop-setEditCellValue) What do you think about converting `handleChange` to be async so that you could await the `setEditCellValue` calls and guarantee the order? and then we wouldn't have to worry about tuning the value of debounceMs

I tested everything else, and I'll approve once you have a chance to try this out.

mddilley (approved):
Awesome! Thanks for updating, and this was a nice review of some tricky JS things. 🙏 I tested on the funding table and also the team and milestone tables that have dependent fields. All works perfectly! 🙌 🚢

mateoclarke (approved):
Looks good! 🚢

PR #1840, opened 2026-05-06
Profile the `project_list_view` and rewrite for speed increase

### LINKED ISSUE #28199: Look for performance improvements in the project list view
We owe a look at improving the performance of the project list view since it is the first page load of the app and home base for our users. Our last look was a couple years ago, and the robots have improved.

**In Scope**
- Scan the `project_list_view` for possible improvements or suggested indexes that we may be missing - using the robots is encouraged! 🤖  If there are a lot of suggestions or any blockers, we can spin off into separate issues.
   - Acceptance criteria is that the new view version produces the same input as the existing definition and can be verified with a SQL query as a test step

frankhereford:
## Associated issues

This PR hopes to close https://github.com/cityofaustin/atd-data-tech/issues/28199.

## Testing

Local testing on this one.

* Spin up your local stack with production data
* Browse the project listing page. Working? Snappier?
* Compare your listings with those in production - do they match exactly?
* Are you comfortable with the occasional reordering of items in the `project_inspector` and `project_tags` columns? 
  * They have never had an explicit order, and they still don't, so reordering is possible, and does occasionally occur. 

Hard mode:
* Use the [query timing tool](https://github.com/cityofaustin/dts-unified-database/pull/1) to observe the new view operating in about 50% the execution time of the old one.
* Use the [query result set comparator](https://github.com/cityofaustin/dts-unified-database/pull/2) to compare the query driving the new view vs. the old one.
  * You should see the undetermined ordering in the two columns listed above as well as `project_features`. 

---
#### Ship list
- [x] Code reviewed 


moped-database/views/project_list_view.sql
```diff
@@ -201,81 +233,49 @@ SELECT
...
-    ) AS project_tags,
+    project_roles_lookup.project_inspector,
+    project_roles_lookup.project_designer,
+    project_tags_lookup.project_tags,
     concat(added_by_user.first_name, ' ', added_by_user.last_name) AS added_by,
     mpcs.components,
     districts.project_council_districts,
-    districts.project_council_districts_string,
     districts.project_and_child_project_council_districts,
+    districts.project_council_districts_string,
```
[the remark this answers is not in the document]
frankhereford: We consume these through `graphql-engine` middleware, this is a non issue for us.

moped-database/migrations/default/1778169083589_speed_up_project_list_view/up.sql
```diff
@@ -0,0 +1,574 @@
...
+            AND combined_project_notes_view.ecapris_subproject_id = mp.ecapris_subproject_id
+        )
+    )
+    AND combined_project_notes_view.is_status_update = true
+    ORDER BY combined_project_notes_view.created_at DESC
+    LIMIT 1
+) proj_status_update ON true
+WHERE mp.is_deleted = false;
+
+-- Most recent migration: moped-database/migrations/default/1778009019613_council_dist_reporting/up.sql
```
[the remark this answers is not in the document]
frankhereford: Woops, this line just needs to come out. it's from the view bot.

moped-database/views/project_list_view.sql
```diff
@@ -1,4 +1,4 @@
--- Most recent migration: moped-database/migrations/default/1778009019613_council_dist_reporting/up.sql
+-- Most recent migration: moped-database/migrations/default/1778169083589_speed_up_project_list_view/up.sql
 
 CREATE OR REPLACE VIEW project_list_view AS WITH project_person_list_lookup AS (
```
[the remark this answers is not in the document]
frankhereford: This is just the nature of the view bot, we're good here.

moped-database/views/project_list_view.sql
```diff
@@ -76,87 +75,68 @@ moped_proj_components_subtypes AS (
...
         jsonb_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null) AS project_council_districts,
-        array_to_string(array_agg(DISTINCT project_districts.council_district_id) FILTER (WHERE project_districts.council_district_id IS NOT null), ', '::text) AS project_council_districts_string,
         jsonb_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null) AS project_and_child_project_council_districts,
-        array_to_string(array_agg(DISTINCT project_and_children_districts.council_district_id) FILTER (WHERE project_and_children_districts.council_district_id IS NOT null), ', '::text) AS project_and_child_project_council_districts_string
-    FROM parent_child_project_map projects
-    LEFT JOIN project_council_district_map project_and_children_districts ON projects.self_and_children_project_ids = project_and_children_districts.project_id
+        string_agg(DISTINCT project_districts.council_district_id::text, ', '::text ORDER BY (project_districts.council_district_id::text)) AS project_council_districts_string,
```
mddilley: @frankhereford I noticed that the ordering by the text value is making the 10 sort further up than expected like `1, 10, 2, 3, 4, 5, 9` on project 129. Could you update here and in the other string field so it sorts by the id value and we end up with `1, 2, 3, 4, 5, 9, 10`?

Thanks for adding the order! 🙏 I'll create a follow-up issue to add order to the other fields like we talked about in dev sync laster Friday.

moped-database/views/project_list_view.sql
```diff
@@ -76,87 +75,68 @@ moped_proj_components_subtypes AS (
...
-    )
-
+phase_date_inputs AS (
     SELECT
-        min_dates.project_id,
-        min(min_dates.min_date) AS min_phase_date
-    FROM min_dates
-    GROUP BY min_dates.project_id
+        phases.project_id,
+        min(phases.phase_start) FILTER (WHERE phases.phase_start IS NOT null AND phases.is_phase_start_confirmed = true) AS min_confirmed_phase_start,
```
mddilley, line 117: This is pretty cool to achieve the same result of the previous subquery in the select here. TIL about using `FILTER` this way.

moped-database/views/project_list_view.sql
```diff
@@ -166,8 +146,60 @@ project_component_work_types AS (
...
+    WHERE mpc.is_deleted = false AND mpcwt.is_deleted = false
     GROUP BY mpc.project_id
+),
+
+project_partners_lookup AS (
+    SELECT
+        mpp2.project_id,
+        string_agg(DISTINCT me2.entity_name, ', '::text) AS project_partners
+    FROM moped_proj_partners mpp2
+    LEFT JOIN moped_entity me2 ON mpp2.entity_id = me2.entity_id
```
mddilley: @frankhereford Not a huge deal but could you remove the `2` from the end of the table aliases in this CTE? Pretty sure that is leftover from when they could collide with other tables in the `GROUP BY` that is now gone so it would be nice to clean this up alongside the larger change. 🧹 ✨

moped-database/views/project_list_view.sql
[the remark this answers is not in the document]
mddilley: @frankhereford This might affect the Power BI dataflow that reads this view. We do a nightly refresh, and I think that the dataflow may rely on the order of the columns remaining the same. 🤔 

@Charlie-Henry does this sound like changing the order of the columns in the view could break the dataflow? We could always rebuild during the release party like we do when adding a column if needed.

moped-database/views/project_list_view.sql
```diff
@@ -166,8 +146,60 @@ project_component_work_types AS (
...
+construction_start_dates AS (
+    SELECT
+        phases.project_id,
+        min(phases.phase_start) AS construction_start_date
+    FROM moped_proj_phases phases
+    WHERE phases.phase_id = 9 AND phases.is_deleted = false
+    GROUP BY phases.project_id
+),
+
+project_roles_lookup AS (
```
mddilley, line 183: It is so nice to have this zoomed-out pass on the view so we can see that we were querying the same table twice for similar information on staff roles and that those queries could be consolidated into one CTE. 🙌

mddilley (changes requested):
This is sooo much faster! I used your profiling tools, ran some separate queries to compare the output of new vs. old, and had some chats with Claude about the output of the new and old with `EXPLAIN ANALYZE`. All signs point to a big jump in performance. 🙌 And, I think that removing the GROUP BY clause will make updating this view easier in the future as we can focus on CTEs first like the pattern in this diff.

I requested a two small changes but then this is 🚢 🏎️ 🙏

moped-database/views/project_list_view.sql
chiaberry: I checked the project in the UI and in the db, and I am seeing the expected order, with 10 at the end of the list. Is this comment out of date, or does the order change sometimes?

chiaberry (approved):
This does feel faster and its been interesting to see how the view works without needing the group by. 

I don't see the council districts out of order, but I might have been looking in the wrong place. Approving assuming that you are going to square that up with Mike.

moped-database/views/project_list_view.sql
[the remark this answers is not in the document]
frankhereford: Mike, thank you, yes, the order is restored. It's interesting that it thinks it was the autofix, I changed it in my editor and pushed it up like normal, but that is neither here nor there. 

Thank you for thinking about the other ways data is consumed out of Moped!

moped-database/views/project_list_view.sql
frankhereford: I added an explicit order here in https://github.com/cityofaustin/atd-moped/pull/1840/commits/86f684bc8b681fd5ce9d36fb2925aefa82bcee80. Thank you for catching this. It was casting and then sorting instead of the other way around.

moped-database/views/project_list_view.sql
frankhereford: Mike, thanks for asking about this. You're so totally right, both of the number "suffixes" on these table alias can come out of this CTE, and you can see that I made that change in the `up.sql` file. Interestingly, the view bot reinserts a `_1` on the `moped_entities` table alias when it exports it back out of the database after the migration has run, which isn't perfect, but I'm not sure exactly what to do further here. Does this work for you?

mddilley: Oh! that is fine if it is the bot, and thanks for helping me see that! I'll resolve this one.

mddilley (approved):
I retested in my db client and checked it out in the app again too. All looks great! And definitely faster. Thanks again! 🏎️ 🙌

tillyw (approved):
tested locally and seems to be very responsive and a definite improvement while retaining existing data and functionality! no concerns with the column values mentioned potentially being reordered :)

moped-database/views/project_list_view.sql
```diff
@@ -284,5 +285,4 @@ FROM combined_project_notes_view
...
 LIMIT 1) proj_status_update ON true
-WHERE mp.is_deleted = false
-GROUP BY mp.project_id, mp.project_name, mp.project_description, ppll.project_team_members, mp.ecapris_subproject_id, mp.date_added, mp.is_deleted, me.entity_name, mel.entity_name, mp.updated_at, mp.interim_project_id, mp.parent_project_id, mp.knack_project_id, current_phase.phase_name, current_phase.phase_key, current_phase.phase_name_simple, mpcs.components, fsl.funding_source_name, fsl.funding_program_names, fsl.funding_source_and_program_names, added_by_user.first_name, added_by_user.last_name, mpps.name, cpl.children_project_ids, proj_status_update.project_note, proj_status_update.date_created, proj_status_update.author, work_activities.workgroup_contractors, work_activities.contract_numbers, work_activities.task_order_names, work_activities.task_order_names_short, work_activities.task_orders, districts.project_council_districts, districts.project_council_districts_string, districts.project_and_child_project_council_districts, districts.project_and_child_project_council_districts_string, mepd.min_phase_date, mcpd.min_phase_date, pcwt.component_work_type_names;
```
johnclary, line 288: 🎉 So great to see this go—nice work!

PR #1849, opened 2026-05-13
display call to action when user deselects all columns

tillyw:
## Associated issues
closes https://github.com/cityofaustin/atd-data-tech/issues/28058

## Testing
**URL to test:** 

local

**Steps to test:**
- go to the projects list view and deselect all columns
- you should see a simple alert notifying the user they should use the column visibility menu to show columns
- select a column and note that the alert disappears
- how does it look? should the alert be styled any differently?

Screenshot 2026-06-11 at 7 29 59 PM

---
#### Ship list
- [x] Code reviewed 


tillyw:
the column headers disappear when you deselect all columns, so the column visibility menu looks a bit out of place in this state, but it all makes sense when you select a column. should we make any changes to the alignment or spacing or should we just leave it as is since this is an edge case and not a view to be used?

moped-editor/src/views/projects/projectsListView/ProjectsListViewTable.js
```diff
@@ -254,6 +255,7 @@ const ProjectsListViewTable = () => {
                 }}
                 slots={{
                   toolbar: ProjectListToolbar,
+                  noColumnsOverlay: ColumnVisibilityAlert,
```
tillyw, line 258: yay that this prop works now in datagrid pro v8!

roseeichelmann:
image

Im getting two alert banners when testing on the deploy preview :thinking:

johnclary (commented):
This looks good for me locally, but in the deploy preview I'm seeing two amber banners:

Image

Does it make sense to go ahead and add this message to all DataGrid tables in the app?'

Edit: +1 what Rose said

moped-editor/src/styles/dataGridProStylesOverrides.js
```diff
@@ -22,6 +22,9 @@ const dataGridProStyleOverrides = {
     justifyContent: "flex-start",
     borderBottom: "none",
   },
+  "& .MuiDataGrid-overlayWrapperInner": {
+    padding: "16px",
+  },
```
mddilley: @tillyw could you move this padding to the sx prop of the `Box` in `ColumnVisibilityAlert`? I think that could help isolate the space around the alert there and avoid this override affecting other overlay slots (loading, no rows, etc.) down the line.

mddilley (changes requested):
I tested this locally, and it looks great! I requested one small change, ~and I'm going to look into why Netlify didn't create an updated preview.~ 🔍 **Edit:** Nvm Chia helped me realize this started as a branch from `main` and is now branching from a different branch so the preview went stale after that.

mddilley:
@johnclary I think that showing this in all tables would be great too. I'm going to create another issue to bring this up into the `MopedDataGrid` component along with the local storage saving logic since it is tied into each table right now. Also, I tried adding to editable tables, and it isn't working as expected (leaves an empty row with edit/delete buttons) so more dev time is needed on that as well. 🤔  edit: here is the issue https://github.com/cityofaustin/atd-data-tech/issues/28970

chiaberry (approved):
I tested locally and it looks good. 

I was surprised there was a deploy preview at all, since this is a branch off of a branch, but maybe it originally was pointing to main, so thats why the preview initially built? And any updates after that werent picked up?

roseeichelmann:
So should we be testing locally instead of w the netlify preview? @mddilley

mddilley:
@roseeichelmann yes, please review by running locally since this one is no longer branching from the `main` branch and won't generate deploy previews. Thanks!!

@tillyw could you update the test instructions? 🙏

tillyw:
ack i'm sorry for the confusion, y'all >.< i didn't think about the fact that the deploy preview would become stale after i rebased the pr. i updated the test instructions!

moped-editor/src/styles/dataGridProStylesOverrides.js
tillyw: done!

tillyw:
> @johnclary I think that showing this in all tables would be great too. I'm going to create another issue to bring this up into the `MopedDataGrid` component along with the local storage saving logic since it is tied into each table right now. Also, I tried adding to editable tables, and it isn't working as expected (leaves an empty row with edit/delete buttons) so more dev time is needed on that as well. 🤔 edit: here is the issue [cityofaustin/atd-data-tech#28970](https://github.com/cityofaustin/atd-data-tech/issues/28970)

@mddilley thanks for creating this issue!

mddilley (approved):
I tested this in the project list, and it looks much improved! Thank you for doing the work to unblock this feature and then updating in here! 🙏 🚢

tillyw:
thank you!

PR #1851, opened 2026-05-14
Update table dialog forms to disable save button on error

### LINKED ISSUE #27103: Update table dialog forms to disable save button on error
We currently keep the save button enabled when there is an error in the phase and funding dialog forms. There could be others too. As suggested by Frank [here](https://github.com/cityofaustin/atd-moped/pull/1779#pullrequestreview-3855248448), we could update to make it clearer that the error is blocking save.

Would be good to do a sweep of error highlighting too. We added it to the **Amount** field in the funding dialog and have it in other places.

Examples:
- SaveUserViewForm disables save button on error state
- ProjectNameForm does not
- DefineProjectForm does not
- ComponentForm disables save button on error state
- MoveComponentForm disables save button on error state
   - Also needs error highlighting on project input
- OverrideFundingForm does not
- ProjectPhaseForm does not
- ProjectSummaryProjectDescription does not
- ProjectWorkActivitiesForm does not

**In Scope**
- Update the forms above that do not disable the save button when there are errors to disable until errors are fixed
- Add error highlighting to MoveComponentForm project input on error

tillyw:
## Associated issues
closes https://github.com/cityofaustin/atd-data-tech/issues/27103

## Testing
**URL to test:** 

https://deploy-preview-1851--moped-main-and-prs.netlify.app/

**Steps to test:**

Check the following forms to confirm they disable the save/submit button on error:

- Project summary
  - ProjectNameForm
  - ProjectSummaryProjectDescription
- Map
  - ComponentForm
- Funding
  - OverrideFundingForm (edit a row with sync from eCapris selected)
  - ProjectWorkActivitiesForm
- Timeline
  - ProjectPhaseForm
- New project
  - DefineProjectForm

Finally, check `MoveComponentForm` to confirm it highlights errors on the project input field

Please let me know if you encounter any unexpected behavior!

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -58,6 +58,8 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
     resolver: yupResolver(validationSchema),
   });
 
+  const areFormErrors = Object.keys(errors).length > 0;
+
   const [isSignalProject] = watch(["isSignalProject"]);
 
   return (
@@ -211,7 +213,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
           success={success}
           buttonOptions={{
             type: "submit",
-            disabled: !isDirty || loading,
+            disabled: !isDirty || loading || areFormErrors,
           }}
         />
       </Box>
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -45,7 +45,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
     handleSubmit,
     control,
     watch,
-    formState: { errors, isDirty },
+    formState: { errors, isDirty, isValid },
   } = useForm({
     defaultValues: {
       projectName: null,
@@ -57,8 +57,6 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
     mode: "onChange",
     resolver: yupResolver(validationSchema),
   });
-
-  const areFormErrors = Object.keys(errors).length > 0;
 
   const [isSignalProject] = watch(["isSignalProject"]);
 
@@ -213,7 +211,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
           success={success}
           buttonOptions={{
             type: "submit",
-            disabled: !isDirty || loading || areFormErrors,
+            disabled: !isDirty || !isValid || loading,
           }}
         />
       </Box>
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -85,7 +85,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
                   name="projectName"
                   control={control}
                   error={!!errors?.projectName}
-                  helperText={errors?.projectName?.message}
+                  helperText={errors?.projectName?.message || "Required"}
                   InputProps={{
                     disabled: loading,
                   }}
@@ -108,7 +108,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
                       textInputOptions={{
                         variant: "standard",
                         error: !!errors?.signal,
-                        helperText: errors?.signal?.message,
+                        helperText: errors?.signal?.message || "Required",
                         disabled: loading,
                       }}
                     />
--- a/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
+++ b/moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
@@ -188,7 +188,7 @@ const DefineProjectForm = ({ handleSave, loading, success }) => {
                 size="small"
                 control={control}
                 error={!!errors?.description}
-                helperText={errors?.description?.message}
+                helperText={errors?.description?.message || "Required"}
                 InputProps={{
                   disabled: loading,
                 }}
--- a/moped-editor/src/views/projects/projectView/ProjectComponents/MoveComponentForm.js
+++ b/moped-editor/src/views/projects/projectView/ProjectComponents/MoveComponentForm.js
@@ -64,6 +64,7 @@ const MoveComponentForm = ({ onSave, component }) => {
             getOptionLabel={getOptionLabel}
             name="project"
             control={control}
+            error={!!errors?.project}
             autoFocus
             helperText="Required"
           />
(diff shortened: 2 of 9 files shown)


moped-editor/src/views/projects/projectView/ProjectComponents/MoveComponentForm.js
mddilley, line 1: while I was in the map tab, I noticed that the **Component Type** in the new and edit component form (`ComponentForm`) isn't highlighting red when there is an error.

@tillyw could you add the same error highlighting that is on the **Work Type** input in that form to the **Component Type**? 🙏 🙏

mddilley (approved):
Looks great! Thanks for ironing these all out to match up with other forms. 🙌 🚢

moped-editor/src/views/projects/projectView/ProjectComponents/MoveComponentForm.js
johnclary: There's a bit of wonkiness going on with the button disable behavior on the component form. The button is disabled by default (presumably because of the missing **Component type**), but if you add a tag, the button becomes enabled. For consistency, I think the desired behavior is that the submit button is initially _enabled_ and becomes disabled on submit if errors?

Image

moped-editor/src/views/projects/projectView/ProjectPhase/ProjectPhaseForm.js
```diff
@@ -358,7 +361,7 @@ const ProjectPhaseForm = ({
             variant="contained"
             color="primary"
             type="submit"
-            disabled={!isDirty || mutationState.loading}
+            disabled={!isDirty || mutationState.loading || areFormErrors }
```
johnclary: On the start date, it looks like we need to hide the normal helper text when there is an input error. You can see it's rendering stacked messages:

Image

moped-editor/src/views/projects/projectView/ProjectPhase/ProjectPhaseForm.js
```diff
@@ -60,8 +60,11 @@ const ProjectPhaseForm = ({
   } = useForm({
     defaultValues,
     resolver: yupResolver(phaseValidationSchema),
+    mode: "onChange",
   });
 
+  const areFormErrors = Object.keys(formErrors).length > 0;
```
johnclary: This is another form where the submit button is disabled by default—might be good to change the behavior to enabled by default?

Other notes:

* If i fill in the start date only, the submit button becomes enabled regardless of if **Phase** is populated
* **Phase** is not indicated required on this form—clicking the save button reveals that it is
* The **Save and Mark as Current** button needs the same disabled/enabled treatment as the **Save** button

Image

johnclary (commented):
Looking good, though I ran into a few minor issues.

moped-editor/src/views/projects/projectView/ProjectComponents/MoveComponentForm.js
mddilley: @johnclary thanks for flagging! i looked at staging, and our convention in most create and edit forms is to keep the save button disabled on load until the form is valid (or until something has changed on edit forms). 

this helped me catch that React Hook Form already gives us a free form state, `isValid`, to use instead of `areFormErrors`.

beyond that, we are mixing the usage of yup validation and the `required` attribute in this component form. since that will take some untangling beyond the scope of this issue, I'm going to create a spin-off issue for this and capture other inconsistencies. 

this is encouraging me to expand on the [form style docs](https://atd-dts.gitbook.io/moped-documentation/dev-guides/design-system/component-styles#forms) that I started last week while reviewing this so we'll have a point of reference in the future. 🙏

moped-editor/src/views/projects/projectView/ProjectPhase/ProjectPhaseForm.js
mddilley: 🙏 @tillyw could you update the helper text render on the `phase_start` so that we only render one instead of two component with the content being conditional? The `formState` errors should render first so that the non-error helper text never shows as red like in John's screenshot.

moped-editor/src/views/projects/projectView/ProjectPhase/ProjectPhaseForm.js
mddilley: @tillyw could you add the helper text "Required" to show when there is no error on the **Phase** input with the error message taking precedence? And also update the "Save and Mark as Current" button to disable until valid as well?

mddilley (changes requested):
Updating my review since using React Hook Form's [isValid form state](https://react-hook-form.com/docs/useform/formstate) is the baked in signal for "is this form ok to submit". I know that `areFormErrors` was a pattern we adopted somewhere along the way, but I think it would be amazing to replace with RHF's free feedback. Could you verify and then update? thank you!

There are a couple other small changes that I requested in response to PR feedback.

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
frankhereford, line 1: I find that if I toggle 'Use signal asset' to true, I can't get the 'Create' button to turn active even after I get the form into a valid state.

Image

frankhereford (changes requested):
Thanks for this work Tilly, I think it's adding a lot of polish. I found one spot where the 'Create' button isn't turning active when I think it should, but other than that, this all tests out great. Thanks again!

tillyw:
> Updating my review since using React Hook Form's [isValid form state](https://react-hook-form.com/docs/useform/formstate) is the baked in signal for "is this form ok to submit". I know that `areFormErrors` was a pattern we adopted somewhere along the way, but I think it would be amazing to replace with RHF's free feedback. Could you verify and then update? thank you!
> 
> There are a couple other small changes that I requested in response to PR feedback.

Thanks for catching this in the docs! I went ahead and updated all instances to use the `isValid` form state and removed `areFormErrors`.

moped-editor/src/views/projects/projectView/ProjectPhase/ProjectPhaseForm.js
tillyw: fixed! only one helper shows at a time now with error taking priority and the standard helper not appearing in red

moped-editor/src/views/projects/newProjectView/DefineProjectForm.js
tillyw: thanks for catching this, @frankhereford ! The 'description' field is required even if the user has selected a signal asset, but the form was not displaying the helper text. fixed it so it displays in both cases!

tillyw:
thanks y'all for your feedback! i believe i have everything operating consistently now. required fields include helper text that turns red (and is replaced with a more detailed error message when available) when an invalid state is triggered, default helper text is overridden by error messages, and save buttons are disabled by default until the form reaches a valid state.

mddilley (approved):
Looks so good! Thanks for adding the required helper text to the required fields where they were missing and helping iron out next steps on forms. 🙏 🚢

chiaberry (approved):
went through and tested all of the updates, and the error text looks good and the save buttons are disabled if there are errors or no changes in the form. 🚢

frankhereford (approved):
🚢🚢🚢

mddilley:
thanks Tilly! just merged this one to start on QA testing. 🥼

PR #1852, opened 2026-05-18
Clean up Moped funding file attachments when deleting a file and implement soft delete of connections

### LINKED ISSUE #28174: Clean up Moped funding file attachments when deleting a file and implement soft delete of connections
When we soft-delete a file, we clean up eCAPRIS funding file attachments. We should do the same for Moped funding file attachments since we can't rely on the database "on delete" actions that only fire on hard deletes.

We also are currently deleting funding-file connections. We should update to soft deletes to adhere to the pattern in other Moped tables but also to be able to track connections to files that happened in the past (who created, who deleted, and when). 

**In Scope**
- Update both `files_ecapris_funding` and `files_project_funding` to include `is_deleted` column that is not nullable defaults to `false`
   - Add triggers to update `updated_by` on soft-delete 
- Update ecapris and funding file connection delete mutations to soft-delete when detached in the UI
- Update file delete mutation to soft-delete Moped funding file attachments and eCAPRIS funding file attachments too
- Filter out soft-deleted attachments to show attachments in the table

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/28174

This PR makes updates so:
- file attachments are now soft-deleted
- deleting a file also soft-deletes any attachments related to that file

## Testing
**URL to test:** 

Local because there is a migration

**Steps to test:**
1. Start up your local stack with production data
2. Create a new project or go to an existing one
3. Go to the **Funding** tab and choosing an eCAPRIS subproject id like `10553.005` so that you will have an eCAPRIS funding row in your table
4. Manually add a row with a source and amount only then save it so you will have a Moped funding row in your table
5. Now, for each funding row:
   - Click the attach button and make sure you end up in the "New" tab of the dialog
   - Turn on "Link to file" switch
   - Then, fill out all fields and a valid url in "Link" input
6. You should now have two funding rows with one attachment on each
7. Confirm that you have two files in the **Files** tab
8. Check your local database to see that you see the connection rows in the `files_ecapris_funding` table and `files_project_funding` table with `is_deleted = false` and that the `created_at` and `updated_at` values are the same
9. Now, go back to the **Files** tab and delete both files
10. Go to the **Funding** tab and see that there are no attachment listed on either of your funding rows
11. Go back to the database tables to see that the two attachments in the `files_ecapris_funding` and `files_project_funding` are now marked as `is_deleted = true` and that the `updated_at` value has updated for when the soft delete happened
12. Now, go to the **Files** tab and add a new file
13. Go back to the **Funding** table, and, for each row in the table:
    - Click the attach button and then choose the "Existing" tab of the dialog
    - Select the file that you just created as the existing file to attach
    - Detach the file and then attach the existing file again to test adding/removing/adding the connection

---
#### Ship list
- [x] Code reviewed 


moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
```diff
@@ -224,7 +224,7 @@ export const useColumns = ({
           <LookupAutocompleteComponent
             {...props}
             name={"ecapris_funding"}
-            options={dataFduOptions?.ecapris_subproject_funding}
+            options={dataLookups?.ecapris_subproject_funding}
```
mddilley, line 228: this needs to be updated because I made a mistake when resolving changes after merging the latest from `main` into the collector branch

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
mddilley, line 1: another merge conflict resolution mistake that i made. i missed these unused import when resolving conflicts because i didn't realize that [Vite doesn't complain about them in the dev server console](https://austininnovation.slack.com/archives/CHZE6BC6L/p1778797145786059)

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
chiaberry: ah I saw this update in the other PR and wondered. good that you caught this

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
chiaberry: I am going to miss those complaints

chiaberry (approved):
🚢 !!

I tested the down migration, then applied the up migration again and then tested adding and removing a file from the funding table and got momentarily thrown off when my previous test files reappeared in the funding table but not the files table -- but thats expected with the down/up migration, it reset the is_deleted to be false.

moped-editor/src/views/projects/projectView/ProjectFunding/ProjectFundingTable.js
mddilley: same! it seems that there are Vite plugins that could help us get this back but might be too noisy until we do https://github.com/cityofaustin/atd-data-tech/issues/28277

https://github.com/fi3ework/vite-plugin-checker

tillyw (approved):
tested and confirmed the rows in the `moped_ecapris_funding` and `moped_project_funding` tables appeared with the correct `is_deleted` status and timestamps! deleted them in the UI and confirmed they were set to `is_deleted = TRUE` and the `updated_at` timestamps were updated! confirmed the down migration as well!

mddilley:
y'all - i mentioned in dev sync that I caught bugs while working on https://github.com/cityofaustin/atd-data-tech/issues/27911. I've made some changes to this PR:
- added conflict handling for inserts into `files_project_funding` and `files_ecapris_funding` to handle when an attachment is added, removed, and then added again on the same file and funding row
- added a new unique constraint to `files_project_funding` so that we can support the conflict handling and also handle uniqueness in the DB in addition to the UI
- updated the check used to show a "File already attached" tooltip which had a regression

@chiaberry @tillyw i re-requested review since I added test steps 12 and 13

chiaberry (approved):
ran through the test steps again, added, deleted and then re-added an existing file and it all works smoothly!

roseeichelmann (approved):
all the test steps worked as described, this looks great!

johnclary (approved):
I created and deleted a lot of files. Looks great and works great 🚀 

Just a passing observation that it could be nice to have a badge in the **Existing** tab that shows the number of files attached.

Image

mateoclarke (approved):
Tested and works as described. 🚢 

Also, I liked John's suggestion about displaying the number of existing files. Would be a nice little dynamic touch.

One thing I was kind of pondering with this change to soft deletes is, should we change the permissions metadata yml so the moped-admin and moped-editor roles don't have delete permissions for these records? I doubt it matters, but if the true intent is to disable the ability to do hard deletes, it seems crufty to leave that door open in the metadata where permissions are configured.

mddilley:
@mateoclarke thank you for being so thorough! I just removed the delete permissions 🔒 🙏

tillyw (approved):
ran through all the test steps again, including 12 and 13 where i added, attached, and detached files and noted that everything looked as expected in the ui and in the associated records in the database! i like john's suggestion for a preview badge for existing files as well! thank you for considering all these possible changes and keeping the associated records clean and tidy ✨ this is going to be a huge feature!

PR #1861, opened 2026-06-01
Move file attachment UI out of dialog

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/28525

## Testing
**URL to test:** 

locally, branching off of branches that branch

**Steps to test:**

Like in previous testing, we are going to start out by setting up a project with funding records and files. 

1. Start up your local stack with production data
2. Create a new project or go to an existing one
3. Go to the Funding tab and choosing an eCAPRIS subproject id like 10553.005 so that you will have an eCAPRIS funding row in your table
4. Manually add a row with a source and amount only then save it so you will have a Moped funding row in your table
5. Attach a file to each row, using the link to file option. The detach file section is no longer on the attach modal.
6. Add another file to one of the rows so you have multiple files on one row. 
7. Notice that your files now have the MoreHoriz icon right aligned in their same row. 
8. Click on the icon to open a Menu. The only option is "Detach"
9. Click Detach to launch the delete confirmation modal.
10. Detach the file. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
@@ -96,7 +96,6 @@ const FundingFile = ({
         keepMounted
         open={menuOpen}
         onClose={handleMenuClose}
-        autoFocus={false}
         TransitionComponent={Fade}
         anchorOrigin={{
           vertical: "bottom",
--- a/moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
+++ b/moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
@@ -10,6 +10,7 @@ import {
 } from "@mui/material";
 import MoreHorizIcon from "@mui/icons-material/MoreHoriz";
 import { LinkOff } from "@mui/icons-material";
+import { useMutation } from "@apollo/client";
 import ProjectFileLink from "src/views/projects/projectView/ProjectFiles/ProjectFileLink";
 import DeleteConfirmationModal from "src/views/projects/projectView/DeleteConfirmationModal";
 import {
@@ -17,11 +18,12 @@ import {
   DETACH_FILE_MOPED_FUNDING_ATTACHMENT,
 } from "src/queries/project";
 
-const FundingFile = ({ file, isSyncedFromECapris }) => {
+const FundingFile = ({ file, isSyncedFromECapris, refetch, fileRecordId }) => {
   const [anchorElement, setAnchorElement] = useState(null);
-  const [detachConfirmationFileId, setDetachConfirmationFileId] =
-    useState(null);
+  const [isDeleteConfirmationOpen, setIsDeleteConfirmationOpen] =
+    useState(false);
   const menuOpen = Boolean(anchorElement);
+
 
   const handleMenuOpen = (event) => {
     setAnchorElement(event.currentTarget);
@@ -44,13 +46,13 @@ const FundingFile = ({ file, isSyncedFromECapris }) => {
       },
     })
       .then(() => {
-        setDetachConfirmationFileId(null);
-        onClose();
-        handleSnackbar(true, "File attachment detached", "success");
+        setIsDeleteConfirmationOpen(null);
+        // handleSnackbar(true, "File attachment detached", "success");
       })
       .catch((error) => {
-        setDetachConfirmationFileId(null);
-        handleSnackbar(true, "Error detaching file attachment", "error", error);
+        setIsDeleteConfirmationOpen(null);
+        console.error(error);
+        // handleSnackbar(true, "Error detaching file attachment", "error", error);
       })
       .finally(() => {
         refetch();
@@ -93,7 +95,7 @@ const FundingFile = ({ file, isSyncedFromECapris }) => {
         }}
       >
         <MenuItem
-          onClick={() => console.log("unlink this file")}
+          onClick={() => setIsDeleteConfirmationOpen(true)}
           selected={false}
         >
           <ListItemIcon>
@@ -102,17 +104,15 @@ const FundingFile = ({ file, isSyncedFromECapris }) => {
           <ListItemText primary="Detach" />
         </MenuItem>
       </Menu>
-      {detachConfirmationFileId && (
+      {isDeleteConfirmationOpen && (
         <DeleteConfirmationModal
           type="file attachment"
           actionButtonText="Detach"
           additionalConfirmationText="This will not delete the file, only detach it from this funding record."
           actionButtonIcon={<LinkOff />}
-          submitDelete={() => handleUnlinkFileAttachment(file.id)}
-          isDeleteConfirmationOpen={detachConfirmationFileId === file.id}
-          setIsDeleteConfirmationOpen={(open) =>
-            setDetachConfirmationFileId(open ? file.id : null)
-          }
+          submitDelete={() => handleUnlinkFileAttachment(fileRecordId)}
+          isDeleteConfirmationOpen={isDeleteConfirmationOpen}
+          setIsDeleteConfirmationOpen={setIsDeleteConfirmationOpen}
         />
       )}
     </Box>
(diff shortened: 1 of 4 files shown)


moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
```diff
@@ -0,0 +1,135 @@
...
+          selected={false}
+        >
+          <ListItemIcon>
+            <LinkOff />
+          </ListItemIcon>
+          <ListItemText primary="Detach" />
+        </MenuItem>
+      </Menu>
+      {isDeleteConfirmationOpen && (
+        <DeleteConfirmationModal
```
mddilley, line 120: I noticed that we don't support disabling the delete button when the mutation fires in this component so I created a snackoo 🍫 https://github.com/cityofaustin/atd-data-tech/issues/28756

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
mddilley, line 1: thanks for extracting this component! (and the .jsx extension ⚛️ 🙏 )

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
```diff
@@ -0,0 +1,135 @@
...
+        <MoreHorizIcon />
+      </IconButton>
+      <Menu
+        id="funding-file-action-menu"
+        anchorEl={anchorElement}
+        keepMounted
+        open={menuOpen}
+        onClose={handleMenuClose}
+        autoFocus={false}
+        TransitionComponent={Fade}
```
mddilley, line 99: I saw a warning in my code editor that TransitionComponent will deprecate in v7 so I noted in https://github.com/cityofaustin/atd-data-tech/issues/27446 cc @tillyw

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
mddilley, line 120: @chiaberry I think that having this dialog nested in this Box is possibly taking some space in the DOM when it opens? Could you check this out?

Image

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
```diff
@@ -0,0 +1,135 @@
...
+      <IconButton onClick={handleMenuOpen}>
+        <MoreHorizIcon />
+      </IconButton>
+      <Menu
+        id="funding-file-action-menu"
+        anchorEl={anchorElement}
+        keepMounted
+        open={menuOpen}
+        onClose={handleMenuClose}
+        autoFocus={false}
```
mddilley: Is this autoFocus here to address something specific? I'm just curious because I couldn't spot a difference w/o it and the MUI docs note it could have accessibility implications which I could also not surface so I feel a little lost on it. 🙃

mddilley (commented):
Works so well! Thank you for rearranging this, and I am liking this direction a lot. 🙌 I mentioned you in one small quirk that I think is straightforward to address, and I had one question. Thank you Chia!!!!

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
chiaberry: I noticed this too, since i was taking my time between opening the modal and detaching the file, the token refresh made me wish i had some visual feedback that requests were being completed

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
chiaberry: it is not, I copied the Menu from our delete project Menu and kept the same props so it would behave the same. I am happy to remove. (fwiw I also couldnt tell what TransitionComponent was doing, I tried swapping Grow in there for Fade and I honestly couldnt see a difference)

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
johnclary: I tested removing this prop, and I can't tell a difference, or possibly I'm seeing that excluding this prob makes the menu appear _more_ smoothly? Seems removable.

johnclary (approved):
Looks great and works very great! Nice!!
🚢

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
tillyw: good to know for the mui upgrade, thank you!

moped-editor/src/views/projects/projectView/ProjectFunding/helpers.js
```diff
@@ -353,23 +355,23 @@ export const useColumns = ({
...
+              spacing={0.5}
+              divider={<Divider sx={{ my: 0.5 }} />}
+            >
+              {row?.[filesType].map((file_record) => {
                 const file = file_record.moped_project_file;
-
                 if (!file) return null;
-
                 return (
-                  <React.Fragment key={file.project_file_id}>
```
tillyw, line 363: 🧹

tillyw (approved):
i like this workflow! and the multiple files on one row looks very clean. having a separate FundingFile component is also very helpful. thank you!

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
mddilley: i couldn't replicate this when i tried again just now

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
```diff
@@ -0,0 +1,134 @@
...
+  );
+
+  const handleUnlinkFileAttachment = (id) => {
+    detachFundingFileAttachment({
+      variables: {
+        id,
+      },
+    })
+      .then(() => {
+        setIsDeleteConfirmationOpen(null);
```
mateoclarke: Should this be set to `false` here and on line 67? That's the initial state given on line 38-39 using `setState`. Not sure it effects anything functionally but consistency seems good.

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
```diff
@@ -0,0 +1,134 @@
...
+        fileUrl={file.file_url}
+        fileName={file.file_name}
+        condensed
+        showNetworkPathStyles={false}
+      />
+      <IconButton onClick={handleMenuOpen}>
+        <MoreHorizIcon />
+      </IconButton>
+      <Menu
+        id="funding-file-action-menu"
```
mateoclarke: ID attributes are supposed to be unique in HTML, but if you have more than one instance of a funding file, there will be duplicates in the DOM. Probably doesn't matter but I could see something buggy happening. Since you have access to `fileRecordId` as a prop, maybe just tack that on the end for uniqueness? 
```
id=`funding-file-action-menu-${fileRecordId}`
```

mateoclarke (approved):
Works great, just as described! 🚢

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
chiaberry: good call, yes. it used to hold an id and i didn't update this to be consistent.

moped-editor/src/views/projects/projectView/ProjectFunding/FundingFile.jsx
chiaberry: 🙏🏼

mddilley (approved):
Thank you for updating! Retested and I can't reproduce what I was seeing earlier. Everything looks ✨ ✨ ✨ 🚢

PR #1862, opened 2026-06-01
Copy attachments from eCAPRIS funding rows to Moped funding rows when overriding

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/27911

This PR handles attachments so that they move with funding rows through the cycle of syncing, overriding, and deleting overrides so that they are not unintentionally lost by users.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Start up your local stack from a production snapshot
2. Create a new project and then set the eCAPRIS subproject ID to `10553.037` so that funding and status sync will turn on and you'll have one FDU in your funding table
3. Go the **Files** tab and create one file that is a url. This will give you an existing file to attach in the next step.
4. Go back to the **Funding** tab and click the paperclip icon on your one funding row.
5. Create a new file attachments in the **New** tab that is a link
6. Click the paperclip icon again and go to the **Existing** tab. Attach the file that you created in step 3
7. You should now have two file attachments on your one funding row.
8. Click the pencil icon, turn on the override switch and enter an amount lower than appropriated, then save
9. Notice that your row is still in the table but it now has an asterisk next to your override amount
10. More importantly, see that the two file attachments have moved from being associated to the eCAPRIS record synced earlier and are now associated with your override record
11. Now, click the trash can icon to undo the override and see the asterisk is now gone.
12. Notice that your file attachments are still there.
13. Last, turn eCAPRIS sync off and back on to see that attachments persist unless a user detaches them
14. Try other combinations of creating manual rows, attaching, and deleting to try to break it

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/moped-editor/src/queries/funding.js
+++ b/moped-editor/src/queries/funding.js
@@ -169,8 +169,10 @@ export const DELETE_PROJECT_FUNDING = gql`
   }
 `;
 
+/* Delete funding record and also transfer file attachments back to
+synced eCAPRIS record that is restored. Executes in one transaction. */
 export const DELETE_PROJECT_FUNDING_AND_REATTACH = gql`
-  mutation DeleteProjectFunding(
+  mutation DeleteProjectFundingAndReattach(
     $proj_funding_id: Int!
     $attachmentObjects: [files_ecapris_funding_insert_input!]!
   ) {
@@ -198,8 +200,6 @@ export const DELETE_PROJECT_FUNDING_AND_REATTACH = gql`
   }
 `;
 
-/* Add funding records and optionally soft-delete eCAPRIS file attachments and then
-insert Moped file attachments on override funding record. */
 export const ADD_PROJECT_FUNDING = gql`
   mutation AddProjectFunding(
     $fundingObjects: [moped_proj_funding_insert_input!]!
@@ -212,10 +212,11 @@ export const ADD_PROJECT_FUNDING = gql`
   }
 `;
 
-/* Add funding records and optionally soft-delete eCAPRIS file attachments and then
-insert Moped file attachments on override funding record. */
+/* Add funding record and also transfer synced eCAPRIS row attachments to
+the new Moped funding record inserted as an override of eCAPRIS. Executes in 
+one transaction. */
 export const ADD_PROJECT_FUNDING_AND_REATTACH = gql`
-  mutation AddProjectFunding(
+  mutation AddProjectFundingAndReattach(
     $fundingObjects: [moped_proj_funding_insert_input!]!
     $entityId: Int!
     $projectId: Int!
--- a/moped-editor/src/queries/funding.js
+++ b/moped-editor/src/queries/funding.js
@@ -153,6 +153,23 @@ export const UPDATE_PROJECT_FUNDING = gql`
 `;
 
 export const DELETE_PROJECT_FUNDING = gql`
+  mutation DeleteProjectFunding($proj_funding_id: Int!) {
+    update_moped_proj_funding(
+      _set: { is_deleted: true }
+      where: { proj_funding_id: { _eq: $proj_funding_id } }
+    ) {
+      affected_rows
+    }
+    update_files_project_funding(
+      where: { entity_id: { _eq: $proj_funding_id } }
+      _set: { is_deleted: true }
+    ) {
+      affected_rows
+    }
+  }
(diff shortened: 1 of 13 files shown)


moped-editor/src/queries/funding.js
```diff
@@ -43,6 +43,9 @@ export const COMBINED_FUNDING_QUERY = gql`
           file_name
         }
       }
+      ecapris_funding {
+        id
+      }
```
mddilley, line 48: This part was a little tricky, but this gives us the primary key of the row in the `ecapris_subproject_funding` table so that we can move attachments from `files_project_funding` back to `files_ecapris_funding` if an override is deleted.

Also why the new object relationship needed to be added to `combined_project_funding_view`.

moped-editor/src/queries/funding.js
```diff
@@ -157,19 +160,81 @@ export const DELETE_PROJECT_FUNDING = gql`
     ) {
       affected_rows
     }
+    update_files_project_funding(
+      where: { entity_id: { _eq: $proj_funding_id } }
+      _set: { is_deleted: true }
+    ) {
+      affected_rows
+    }
```
mddilley, line 168: This was outside the scope of the issue, but this makes sure that we remove file attachments when a funding row is deleted. I noticed some `is_deleted = false` attachments lingering when testing this PR, and I missed this case in scoping.

roseeichelmann:
Something im confused about is that if i add a manual funding record with an FDU, it doesnt get the eCAPRIS badge, but when i edit it i have to toggle the override eCAPRIS amount button in order to change the amount. so im confused about whether or not this is an eCAPRIS record (im probably missing some context about what an FDU is)

moped-editor/src/queries/funding.js
roseeichelmann: what if someone changes the ecapris id of a project, should associated attachments be removed then?

roseeichelmann (approved):
followed all the test steps and the code, its working as expected though some of it is a little confusing to me, but i feel like that may just come from my ignorance about how all of this ties together.

maybe there could be a tooltip on trash can hover that emphasizes how its not really deleting the funding record but reverting to the originally synced ecapris funding record?

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 

I ended up testing this ~5 times because the first time i tested the override, the attachments did not transfer. but after refreshing my page the test worked every time. i think i was looking at a stale version of the app after testing a different branch.

this looks great and I'm glad you found a clear path for dealing with the tricky mechanics of this. I am curious what you think about what should happen to file attachments to a non-overridden funding row when the project's eCapris project ID changes.

chiaberry (approved):
I tried different combinations of attaching and overriding and I couldnt break it. 

I am glad you caught the bug where we werent removing file attachments when rows were deleted, good to fix that before this all ships!

mateoclarke (approved):
I can tell how tricky this is, thinking through the edge cases of overrides and multiple file attachments but it all worked as described in the testing instructions. 🚢

moped-editor/src/queries/funding.js
mddilley: that is a good question! I ran this by Nathan today, and he thought it was reasonable for attachments to persist if the "Sync from eCAPRIS" switch is turned off (intentional or accidental) or if the id changes. i'm imagining a project with several synced FDUs with file attachments that could be lost with a single click and trying to avoid that.

Overall, I'm hoping to give users the ability to recover and detach if needed. The attachments stay associated with the original eCAPRIS funding row, so if sync is toggled back on or the ID is corrected, the connections reappear.

mddilley:
@roseeichelmann thanks for the feedback, and i know the various states can get confusing since there are a lot of combinations of funding info moving into this table. There is a [funding user guide](https://atd-dts.gitbook.io/moped-documentation/user-guides/add-funding-to-a-project) for users that you could review.

i love the idea of a tooltip that expands on the revert behavior when a project still has syncing enabled so i created https://github.com/cityofaustin/atd-data-tech/issues/28771

tillyw (approved):
tested and all the cases checked out!! thanks for accounting for all the different cases of overrides and reversions, it works seamlessly and behaves as i would hope it would as a user in cases where i might need to revert a change without having to redo all my previous work. thank you!

PR #1871, opened 2026-06-11
Address lint errors in Vite-updated project

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/28277

This PR performs auto and manual fixes to ESLint errors in our project. Some errors were ignored because they overlap with future upgrades (https://github.com/cityofaustin/atd-data-tech/issues/24261) and some were ignored in the past. I also created some spinoff issues for refactors that I thought were too much to bundle in here:
- https://github.com/cityofaustin/atd-data-tech/issues/28905
- https://github.com/cityofaustin/atd-data-tech/issues/28901
- https://github.com/cityofaustin/atd-data-tech/issues/28934

## Testing
**URL to test:** 

Local

**Steps to test:**

**No exhaustive code review is needed on this one. I commented on important changes, and I'm open to discussion on ignored errors. I didn't want to get bogged down in complex refactors but I could be overestimating effort on some that were ignored.**

1. First, from the `/editor/` folder, run: `npm run lint` and you should see 5 warnings
2. Next, you can search the editor code for `eslint-disable` to see that there are 18 errors ignored
3. To test refactored code:
   - Open up a project with components and check that the map in the summary tab and the map tab zooms in to the project's components when it first loads
   -  In the project list, update the hidden columns, leave the list to view a project page, and then go back to the list to make sure your hidden columns are saved
   - Visit https://localhost:3000/moped/projects/930?tab=map&project_component_id=1230 and make sure the component list item is focused and not collapsed on the list and that the map is zoomed to the matching line geometry
   - To test that the map still shows related components (parent, child, sibling):
      - Create three projects and add components. Then, make one the parent of the other two.
      - Make sure that you can see:
         - the child sibling components in green on the parent project maps
         - the sibling and parent components on the child components

---
#### Ship list
- [x] Code reviewed 


editor/src/auth/user.js
mddilley, line 1: this is one of many examples of JS code that is exported to be used in many files was pulled out into a new file away from exported JSX components. These were caught by our `eslint-plugin-react-refresh` plugin which flags code that break breaks [hot module replacement](https://vite.dev/guide/features#hot-module-replacement).

```
ERROR(ESLint)  Fast refresh only works when a file only exports components. Use a new file to share constants or functions between components. (react-refresh/only-export-components)
```

editor/vite.config.js
```diff
@@ -14,7 +15,23 @@ export default defineConfig(() => {
...
+          dev: {
+            logLevel: ["error"],
+          },
+          overlay: {
+            initialIsOpen: true,
+            position: "tl", // top-left
+          },
+        },
+      }),
+    ],
```
mddilley, line 35: this is the config for the [plugin](https://github.com/fi3ework/vite-plugin-checker) that runs a separate worker to display lint errors in the command line and rendered in the local app.

this was something we had for free with CRA so I'm curious about opinions to keep or not. It is helpful to me and very similar to what we see in Next.js too.

Image

Image

editor/src/utils/localStorageHelpers.js
```diff
@@ -31,7 +31,7 @@ const getPreviousHiddenColumns = (defaultColumnSettings, storageKey) => {
     // By iterating the defaults, we also remove outdated columns no longer in config.
     currentHiddenColumnSettings = Object.keys(defaultColumnSettings).reduce(
       (acc, columnName) => {
-        if (previousHiddenColumnSettings.hasOwnProperty(columnName)) {
+        if (Object.hasOwn(previousHiddenColumnSettings, columnName)) {
```
mddilley, line 34: This was one of a couple fixes for https://eslint.org/docs/latest/rules/no-prototype-builtins

editor/src/utils/localStorageHelpers.js
```diff
@@ -51,18 +51,12 @@ export const useHiddenColumnsSettings = ({
...
-  useEffect(() => {
-    const initialHiddenColumnSettings = getPreviousHiddenColumns(
-      defaultHiddenColumnSettings,
-      storageKey
-    );
-    setHiddenColumns(initialHiddenColumnSettings);
-  }, [defaultHiddenColumnSettings, storageKey]);
+  const [hiddenColumns, setHiddenColumns] = useState(() =>
+    getPreviousHiddenColumns(defaultHiddenColumnSettings, storageKey)
+  );
```
mddilley, line 59: This is a refactor where we were initializing state as an empty object and then immediately using `useEffect` to set the initial hidden columns from local storage.

this change uses the [initializer function pattern](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state) so we don't need the `useEffect`, and the initial value is set on the first render.

editor/src/views/dashboard/DashboardView.jsx
```diff
@@ -240,19 +242,16 @@ const DashboardView = () => {
...
-      }
+  const rows = useMemo(() => {
+    if (!data) return [];
+    if (TABS[activeTab].label === "My projects") {
+      return data.moped_proj_personnel.map((row) => row.project);
+    }
+    if (TABS[activeTab].label === "Following") {
+      return data.moped_user_followed_projects.map((row) => row.project);
     }
+    return [];
```
mddilley, line 254: since this table doesn't have any edit/delete actions, we can derive the rows from the fetched data and also fallback to an empty array with `useMemo`

editor/src/views/dev/LookupsView/index.jsx
```diff
@@ -143,7 +142,8 @@ const LookupsView = () => {
               ))}
             </Grid2>
           </Paper>
-
+          {/* Ignoring because refs is a stable map of { current } objects created in useMemo above */}
+          {/* eslint-disable-next-line react-hooks/refs */}
```
mddilley, line 146: This one is a judgement call, and it doesn't feel like the right place to put effort into at the moment. We are calling `createRef` in a function component when [it is recommended for use in class components](https://react.dev/reference/react/createRef) which caused the eslint error. 

All that said, this code has held up for a long time, and the refactor isn't super clear since we can't map a list of refs using a hook like `useRef`.

editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
```diff
@@ -19,7 +19,8 @@ export const advancedSearchIsOrParamName = "isOr";
 const makeAdvancedSearchWhereFilters = (filters) =>
   Object.keys(filters)
     .map((filter) => {
-      let { field, value, operator } = filters[filter];
+      const { field, operator } = filters[filter];
+      let { value } = filters[filter];
```
mddilley, line 23: ESLint was complaining about a mixture of reassigned and non-reassigned variables both using `let`.

editor/src/views/projects/projectsListView/useProjectListViewQuery/useCurrentData.js
```diff
@@ -8,14 +8,12 @@ import { useState, useMemo } from "react";
...
-      return data;
-    } else {
-      return previousData;
     }
-  }, [data, setPreviousData, previousData]);
+  }, [data]);
 
-  return currentData;
+  return data ?? previousData;
 };
```
mddilley, line 19: Refactored because we were setting state inside `useMemo` which can be unpredictable - `Error: Calling setState from useMemo may trigger an infinite loop`

Ignored the `react-hooks/set-state-in-effect` rule since we are synchronizing with an external data source. We are caching results to display in the UI from the previous data fetch until the next page of data populates the project list.

editor/src/views/projects/projectView/ProjectComponents/utils/map.js
```diff
@@ -144,10 +144,13 @@ export const useZoomToExistingComponents = (
   featureCollection,
   refreshOnComponentsUpdate = false
 ) => {
-  const [hasMapZoomedInitially, setHasMapZoomedInitially] = useState(false);
+  const hasMapZoomedInitially = useRef(false);
```
mddilley, line 147: This and the refactor of `useComponentLinksParams` revolve around `Error: Calling setState synchronously within an effect can trigger cascading renders`. 

Both of these can rely on `useRef` since updating the value doesn't need to trigger a re-render and both are one-time operations:
- map only needs to initialize zoom once
- component links only need to be read from the url once to scroll to the matching component

editor/src/views/projects/projectView/ProjectComponents/utils/useCreateComponent.js
```diff
@@ -39,7 +39,7 @@ const createReducer = (state, action) => {
         draftComponent: action.payload,
         isCreatingComponent: true,
       };
-    case "remove_draft_component_feature":
+    case "remove_draft_component_feature": {
```
mddilley, line 42: this was a TIL for me that cases do not create their own scope but you can make them do it with these curly braces. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#lexical_scoping

editor/src/views/projects/projectView/ProjectComponents/utils/useProjectComponents.js
```diff
@@ -38,21 +38,27 @@ export const useProjectComponents = (data) => {
...
-  const parentComponents = useMemo(() => {
-    if (!data?.parentProjectComponents) return [];
-
-    return data.parentProjectComponents;
-  }, [data]);
-
-  const siblingComponents = useMemo(() => {
-    if (!data?.siblingProjects) return [];
+  const allRelatedComponents = useMemo(() => {
+    if (!data) return [];
```
mddilley, line 42: This one was a little gnarly, and it had to be refactored to address https://react.dev/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization.

The core of the refactor is that all of the `useMemo`s only depend on `data` instead of building up dependencies.

editor/src/auth/user.js
chiaberry: ah, so the UserProvider component used to be in user.jsx and now its moved into its own file?

editor/vite.config.js
chiaberry: this seems like it will be helpful, i vote keep

editor/src/views/dev/LookupsView/index.jsx
chiaberry: i agree with skipping this one

editor/src/views/dev/LookupsView/ProjectTagsTable.jsx
```diff
@@ -136,6 +136,7 @@ const ProjectTagsTable = ({ canEdit, handleSnackbar, onScrollToTop }) => {
   const isEditMode = getIsEditMode(rowModesModel);
 
   useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- rows state is also used for optimistic updates after mutations; refactor in issue #28901
```
chiaberry, line 141: ✨ including the issue number is nice

editor/src/auth/user.js
mddilley: yep! we exported helpers and vars used in more than one place out of files that export JSX components which breaks the fast refresh. so the component exports and non-component exports are split into separate files now.

editor/src/views/projects/projectView/ProjectComponents/utils/useCreateComponent.js
chiaberry: TIL too!

chiaberry (approved):
I read the comments you left on files and glanced through a chunk of your other updates and tested the app -- a lot of things being fixed, and some nice TILs to boot. 

🚢

editor/src/views/projects/projectView/ProjectComponents/index.jsx
```diff
@@ -160,6 +162,7 @@ export default function MapView({
         (component) => component.project_component_id === clickedComponentId
       );
 
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- syncs clickedComponent with latest project data after refetch
     setClickedComponent(updatedClickedComponent);
```
roseeichelmann, line 166: how did you decide when to ignore vs refactor this error? im running across a lot of set state in effect errors and so far have been trying to refactor all of them, but are there cases where its valid to leave it?

editor/src/views/projects/projectView/ProjectComponents/utils/map.js
roseeichelmann: im learning a lot from seeing how you handled these diff cases of this error!

editor/src/utils/localStorageHelpers.js
johnclary: TIL 🙏

editor/src/views/projects/projectsListView/useProjectListViewQuery/useAdvancedSearch.js
johnclary: Cool that it caught this 👍

editor/src/views/projects/projectView/ProjectComponents/utils/map.js
johnclary: yeah it's great to see the use case for refs in these scenarios

editor/src/views/projects/projectView/ProjectComponents/utils/useCreateComponent.js
johnclary: 💡

editor/src/views/projects/projectView/ProjectComponents/index.jsx
mddilley: It was definitely a judgement call! I felt like, if the feature is working as expected and the refactor seemed to touch more than a few components working together, then it was too big for this PR. Not the most definitive answer, but I mostly didn't want to break things or create a bunch of testing for y'all in this PR.

johnclary: I know this isn't really helpful, but for me the first test is "can i think of another way to create this effect/sync that doesn't use `useEffect` in this way?". Or simply asking, why can't we memoize/compute this value instead?

Other questions to ask:
- Is the dependency that triggers the effect is distinct from the state being set?
- Is the state being set a cached/synced copy of external data?

Tbh, this one is showing some code smell—`makeClickedComponentUpdates` is calling `setClickedComponent(clickedComponent)`, and then this hook fires and makes the same update. But I'd need to spend a lot more time with this code to have a strong opinion about. This some of the most complicated state management in the app and I don't know that it's worth flagging for later refactor versus letting it bubble up if/when it becomes an issue.

editor/vite.config.js
johnclary: yeah, this is neat! glad to see it!

johnclary (approved):
Wow, so much clean up here along with some TILs. Hooray for being down to just 5 errors plus a few backlog issues 🚀 

I love how this is also clearing the way for easier typescript implementation, once the time comes. Really nice tech debt payoff here!!

🚢 🚢 🚢 🚢 🚢

roseeichelmann (approved):
super helpful to see how you handled all these errors especially as im working through the same thing with VZ

app run great locally, followed all test steps. this looks ready to ship!

editor/src/views/projects/projectView/ProjectComponents/index.jsx
mddilley: it is helpful and i agree! i was planning on letting this one slide since it is flawed but working, but i made an issue after all since it will be easier to dive into on its own. i just pushed a change to point to the issue in the code. thank y'all for the discussion on this! https://github.com/cityofaustin/atd-data-tech/issues/29002

tillyw (approved):
tested locally and all the test steps checked out! 5 lint errors, 18 exceptions, maps and menu items appeared as expected. thanks Mike!

mddilley:
thank y'all for the reviews. 🙏  going to merge this one now 🚢

PR #1886, opened 2026-06-29
Enhance local execution of moped AGOL ETL to use varlock

frankhereford:
## Associated issues

This is a proof of concept as part of https://github.com/cityofaustin/atd-data-tech/issues/28702.

## Testing

⚠️ When you start the ETL as the last step in the testing process, please use the commands with the `-n` or `--dry-run` flag to prevent the ETL from writing to the AGOL dataset. 
  * Commands that start like this need to have that `-n` or `--dry-run` flag.

```
npx varlock run -- docker compose run arcgis
```

* Spin up your local database and `graphql-engine` instances,
* Follow the [README's](https://github.com/cityofaustin/atd-moped/blob/frank/arcgis-etl-varlock/moped-etl/arcgis/README.md) instructions on how to use.  
  * If you have any questions about how to install 1Password's CLI or turn on support for it from the desktop application, please let me know.

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager checked if any columns in views used by the Power BI dataflow were added, deleted, or renamed


moped-etl/arcgis/.env.schema
```diff
@@ -0,0 +1,10 @@
+# This env file uses @env-spec - see https://varlock.dev/env-spec for more info
+
+# @plugin(@varlock/1password-plugin)
+# @initOp(allowAppAuth=true)
+# @defaultRequired=false @defaultSensitive=false
+
+AGOL_USERNAME=op("op://ocp5yzivvhbyr226xvbo33eoqi/goljxwns3yu5totrop2wo3vhtu/production/username")
+AGOL_PASSWORD=op("op://ocp5yzivvhbyr226xvbo33eoqi/goljxwns3yu5totrop2wo3vhtu/production/password")
```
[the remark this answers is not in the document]
frankhereford: I think these IDs are more explicit and easier to handle since they are readily available from the desktop app and don't have spaces in them.

moped-etl/arcgis/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "moped-agol-etl",
+  "version": "1.0.0",
+  "description": "Python script integration pushing Moped data to ESRI's ArcGIS Online (AGOL) platform",
+  "private": true,
+  "scripts": {
+    "test": "echo \"No tests\""
+  }
```
johnclary: @frankhereford heads-up that there is missing comma here which is rendering this JSON invalid.

moped-etl/arcgis/.env.schema
johnclary: I guess this does raise the question of if we want to go with vault names instead of IDs (I forgot that's an option). Vault names seem preferable to me because:

* From a devx perspective, if i see the plain-text vault name I know where a secret is coming from. I'm not sure how I would reverse lookup a vault by its ID, if that were something I needed to do.
* The optics are better than putting vault UUIDs in git, for whatever that's worth (avoiding questions from CISO?)

moped-etl/arcgis/.env.schema
```diff
@@ -0,0 +1,10 @@
+# This env file uses @env-spec - see https://varlock.dev/env-spec for more info
+
+# @plugin(@varlock/1password-plugin)
+# @initOp(allowAppAuth=true)
+# @defaultRequired=false @defaultSensitive=false
```
johnclary: I am curious about both of these settings: they look like things we want to set to `true`, especially sensitivity, right?

moped-etl/arcgis/.env.schema
```diff
@@ -0,0 +1,10 @@
...
+
+AGOL_USERNAME=op("op://ocp5yzivvhbyr226xvbo33eoqi/goljxwns3yu5totrop2wo3vhtu/production/username")
+AGOL_PASSWORD=op("op://ocp5yzivvhbyr226xvbo33eoqi/goljxwns3yu5totrop2wo3vhtu/production/password")
+HASURA_ENDPOINT=http://host.docker.internal:8082/v1/graphql
+HASURA_ADMIN_SECRET=hasurapassword
```
johnclary, line 12: Should we go ahead and drop these in 1password as well?

moped-etl/arcgis/.env.schema
johnclary, line 1: It might be worth proving out the pattern for [env-specific files](https://varlock.dev/guides/environments/#loading-environment-specific-env-files) in this PR. 

I was feeling pretty on the fence about Varlock vs simply using the 1password CLI until reading through their feature set around this. You can set up multiple env files—committed to code—which override the base schema based on your `APP_ENV` 👀

johnclary (commented):
Very cool to see this in action, Frank! I like varlock, and think it'd be good to go a bit further down the road to prove out how we would handle env-specific variables for local/staging/prod.

And this is just an aside for Mike, but now that the AGOL dataset is relatively stable, maybe it's time to have the GIS folks create a staging dataset we can use for ETL testing?

moped-etl/arcgis/package.json
frankhereford: Thank you John! Fixed

moped-etl/arcgis/.env.schema
mddilley: yep i agree. seems like [their definition](https://varlock.dev/guides/secrets/) of "sensitive" matches up with anything that we would store in 1PW - "any value that should not be exposed to the outside world"

moped-etl/arcgis/.env.schema
mddilley: Yeah, it is worth thinking about! iirc we need to use vault ids but then can use plaintext for entry names like we do in the [certbot renewal scripts](https://github.com/cityofaustin/atd-airflow/blob/production/toolbox/certbot/renew_domain_with_certbot.sh#L19). i think that would be a good option here (if possible) since we're in the habit of doing the same in Airflow DAGs.

moped-etl/arcgis/.env.schema
mddilley: oo yeah good catch. 🙏  we have them ready and waiting in the `Moped Hasura Admin` entry which can be dropped right in here

moped-etl/arcgis/.env.schema
mddilley: this is very nice. feels very familiar to our patterns in Airflow and the way Next.js and Vite handle envs too so it does feel like the little bit of overhead could be worth it if we have a clear reference to what secrets are used/needed per environment. 

this ETL is a perfect example of where that could be helpful (if we had a staging dataset) since operating on the prod dataset is a just few keystrokes away. having a safe path committed and default would be rad. 💯

mddilley (commented):
I had this up and running in no time at all. I also think it would be cool to explore the multi-environment path, and I'll look into setting up a staging dataset for this ETL tomorrow. Could be one of those "feed two birds with one scone" situations. 🥐 🐦 🐦 

I'll also add this to refinement next week so we can all talk about next steps. Thanks, Frank!

moped-etl/arcgis/.env.schema
frankhereford: > I'm not sure how I would reverse lookup a vault by its ID, if that were something I needed to do.

It's not obvious at all, but you can actually search for these in 1PW and it will find the specific entry or vault. 

That said, I think y'all are right that the human readable ones are easier to deal with and I'm about to try to switch over to using those.

frankhereford: Interesting resolution on using human readable names VS IDs. On this particular entry, we use parentheses in our entry title `(AGOL)` which forces us, on this one particular entry, to use the ID. But generally, yes, we can use named Vault + Entry combo. I did get us switched to using the named vault at least here.

As an aside, you actually get this whole `op://` URL out of the 1PW app by right clicking on an field and selecting "Copy Secret Reference" and it will default to using the named URL path components, but will use the ID if there are breaking characters in it.

moped-etl/arcgis/.env.schema
frankhereford: Thank you for asking. I turned the "required" default to `true` -- we want every secret to be available. I did leave sensitive as false, but I have decorated the actual secret secret as sensitive so it's obscured on the terminal now. 🤭 I think this makes for a cleaner reading file, but I am very open to discussing these defaults more -- if y'all still think they should be default sensitive, I'll conform the file to whatever y'all think is best. 👍

moped-etl/arcgis/.env.schema
frankhereford: I think so yea, I agree. Anything that changes between deployment environments or is sensitive should get pulled into 1PW, I think. This work feels like it'd be a part of the upcoming issue https://github.com/cityofaustin/atd-data-tech/issues/29621.

moped-etl/arcgis/docker-compose.yaml
```diff
@@ -6,5 +6,8 @@ services:
       - .:/app
     entrypoint: python /app/components_to_agol.py
     command: -d
-    env_file:
-      - env_file
+    environment:
+      - AGOL_USERNAME=${AGOL_USERNAME?AGOL_USERNAME is required}
+      - AGOL_PASSWORD=${AGOL_PASSWORD?AGOL_PASSWORD is required}
+      - HASURA_ENDPOINT=${HASURA_ENDPOINT?HASURA_ENDPOINT is required}
+      - HASURA_ADMIN_SECRET=${HASURA_ADMIN_SECRET?HASURA_ADMIN_SECRET is required}
```
mateoclarke, line 13: It seems like `AGOL_COMPONENTS_ENDPOINT` is in the `.env.schema`, but missing here. Without it in the docker-compose file, varlock just prepares a secret that isn't passed into the container, right?

We should add:
`- AGOL_COMPONENTS_ENDPOINT=${AGOL_COMPONENTS_ENDPOINT?AGOL_COMPONENTS_ENDPOINT is required}`

moped-etl/arcgis/.env.schema
```diff
@@ -0,0 +1,12 @@
...
+# @plugin(@varlock/1password-plugin)
+# @initOp(allowAppAuth=true)
+# @defaultRequired=true @defaultSensitive=false
+
+AGOL_USERNAME=op("op://TPW DTS API Accessible Secrets/goljxwns3yu5totrop2wo3vhtu/production/username")
+# @sensitive=true
+AGOL_PASSWORD=op('op://TPW DTS API Accessible Secrets/goljxwns3yu5totrop2wo3vhtu/production/password')
+AGOL_COMPONENTS_ENDPOINT=https://services.arcgis.com/0L95CJ0VTaxqcmED/arcgis/rest/services/Staging_Moped_Project_Components/FeatureServer
+HASURA_ENDPOINT=http://host.docker.internal:8082/v1/graphql
+HASURA_ADMIN_SECRET=hasurapassword
```
mateoclarke, line 12: Should we give`HASURA_ADMIN_SECRET` the `@sensitive` decorator as well? Otherwise it could end up in production logs.

Screenshot 2026-09-09 at 11 59 28 AM

mateoclarke (approved):
Looks good, I was able to install `op` and configure the 1pw cli to load varlock envs with a successful dry-run. I added a few comments, the `AGOL_COMPONENTS_ENDPOINT` being a higher priority. 

I think the dry-run didn't catch it as missing because `AGOL_COMPONENTS_ENDPOINT` is only needed to write/delete using the `get_endpoint` method in [utils.py#33](https://github.com/cityofaustin/atd-moped/blob/frank/arcgis-etl-varlock/moped-etl/arcgis/utils.py#L33-L34), which the dry run intentionally skips.

PR #1892, opened 2026-07-10
Explore dev experience of GraphQL Codegen for type safety

### LINKED ISSUE #29055: Explore dev experience of GraphQL Codegen for type safety
To start planning a JS to TS migration, we should explore end-to-end type safety for our GraphQL operations using GraphQL Codegen that introspects our Hasura GraphQL endpoint. We want to understand what the query/mutation/component conversion process looks like for a developer. This spike should use a query like `SUMMARY_QUERY` as a test since it uses a query with fragment interpolation.

Also, the Hasura guide shows installing many plugins ([guide](https://hasura.io/learn/graphql/typescript-react-apollo/codegen/)) but that is outdated and [GraphQL codegen](https://the-guild.dev/graphql/codegen) now recommends using the `client-preset` package [for better devx and small bundle size](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-react-apollo).

**In scope:**
- Install @graphql-codegen/cli and @graphql-codegen/client-preset and then create a codegen.yml that targets our `/queries/` folder where all of our GraphQL queries and mutations live
   - Run codegen locally against the our local Hasura instance
   - Figure out what a developer imports from the generated file and how they find it
- Convert the data fetching of the component that uses `SUMMARY_QUERY`
- Replace `useQuery(SUMMARY_QUERY)` call with the equivalent generated hook
- Make sure that the `${PROJECT_COMPONENT_FIELDS}` fragment interpolation in `SUMMARY_QUERY` works as expected or note if there is an important change regarding fragments
- Capture a summary of the process in this issue's comments

**Not in scope:**
- Netlify build command changes
- Adding typecheck to CI
- Migrating query files from .js to .graphql format which might be needed later on
- Converting any other files or queries

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29055

This PR implements a local dev workflow that incorporates [GraphQL Codegen](https://the-guild.dev/graphql/codegen) for type safety and a [Vite plugin](https://www.npmjs.com/package/vite-plugin-graphql-codegen) that hooks codegen into the local development process.

Test steps will cover the details, but the high-level proposal is:
1. Start local Hasura stack and Vite development server
2. The Vite development server now has a plugin that immediately runs the codegen on start and continues to generate every time code changes are saved
3. The codegen tool introspects the local Hasura engine through access we give to the local GraphQL endpoint
4. We let the codegen tool know what types to generate by wrapping our queries and mutations with the `graphql()` function which makes the tool "aware" of the query or mutation
5. Just like how we develop migrations locally and then code review them, we would generate codegen diffs and code review them. If the schema changes, we would expect a change in the generated code in a PR so we could see the impact of the schema change.
6. Last, since the generated types are committed in code, we can add `npm run typecheck` to our Netlify build commands. So, if there is a type error in a deploy preview (or other) build then it fails just like what happens with our lint errors. We could also avoid developing our CI further to reach our staging and production GraphQL endpoints to run codegen and then type check which would require sharing more sensitive secrets and thinking about how to let more places reach our endpoints. All type checking would be contained in our build.

## Testing
**URL to test:** 

Local

**Steps to test:**

Looking for feedback on this proposed workflow. No need to review files in `/gql/` folder since they are generated but feel free to take a look. Please poke holes and flag anything issues that you think about when testing through this. Thanks! 🙏 

Also, I converted a few files to .tsx and .ts to get a feel for what converting code would look like in practice. Using `@ts-expect-error` and noting an issue to convert components felt like a good pattern for marking boundaries since we did similar with lint errors. Open to feedback!

### Setting up
1. Start up the local stack from the `moped-database` folder: `./hasura-cluster replicate`
2. Install new dependencies and start up Vite server from `editor` folder: `npm i && npm run dev`

### Testing codegen on save by adding an existing column to an existing query
1. In the `SUMMARY_QUERY` query in `src/queries/project.js`, add `date_added` to the query just below `project_id`
2. On save, the codegen plugin will go to work. Check your Git status and see new changes that include the new column in the generated types.

### Testing type errors will occur if you remove a referenced column from a query
1. In the `SUMMARY_QUERY` query in `src/queries/project.js`, remove `is_deleted` from the query
2. Open `ProjectView.tsx` and you should see a type error on the line where `isProjectDeleted` is defined that `is_deleted` does not exist
3. Add `is_deleted` back to the query

### Testing type errors on columns referenced that are not in a query
1. Open `ProjectView.tsx` and add `const addedBy = data.moped_project[0].added_by;` on line 309. You should see a type error that `added_by` does not exist

### Testing the addition of a new column to the project table
1. Start the Hasura console from the `moped-database` folder: `hasura console`
2. Go to http://localhost:9695/console/data/default/schema/public/tables/moped_project/modify and click "Add a new column"
3. Add a nullable boolean column called `is_cool` with a default of `true`, then save. You'll see in your Git status that a migration was generated
4. Now, go back to `ProjectView.tsx` and add `const isCool = data.moped_project[0].is_cool;`
5. Notice that you see a type error because we haven't added `is_cool` to the query yet
6. Go back to `SUMMARY_QUERY` and add `is_cool` to the query just below `date_added` in the project query
7. Go back to `ProjectView.tsx` and see that the type error is gone

### Testing type errors in mutation payloads
1. Go to `ProjectFollowButton.tsx` and change `Number(projectId)` to `false` on line 61
2. You'll see a type error on `project_id` that a number is expected

### Test the type check
1. From the `editor` folder, run `npm run typecheck` and see that it fails since there are type errors created by the previous test steps

---
#### Ship list
- [x] Code reviewed 


editor/src/components/Page.tsx
mddilley, line 1: no new logic here, only went from jsx to tsx

editor/src/components/QueryContextProvider.ts
mddilley, line 1: same, js to ts

editor/src/components/useFeedbackSnackbar.ts
mddilley, line 1: you guessed it, js to ts

editor/src/gql/fragment-masking.ts
mddilley, line 1: this file creates information that is consumed through the [Gql Codegen's `useFragment`](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#the-usefragment-helper) which we don't currently have a use for. We use GraphQL fragments in one place that I'll note elsewhere.

editor/src/gql/gql.ts
mddilley, line 1: this file exports the `graphql()` function that gives us typed data and variable objects used by `useQuery` and `useMutation`

editor/src/gql/graphql.ts
mddilley, line 1: this file contains all of the generated types

editor/src/queries/components.js
```diff
@@ -166,6 +167,113 @@ export const PROJECT_COMPONENT_FIELDS = gql`
   }
 `;
 
+export const PROJECT_COMPONENT_FIELDS_FRAGMENT = graphql(`
```
mddilley, line 170: This is the one GraphQL fragment that we use in the codebase to capture all of the data needed when we query components.

Wrapping this fragment in `graphql()` makes it accessible by other queries wrapped in `graphql()`. You can see this later in the code where we no longer need to use string interpolation to reuse it. The `SUMMARY_QUERY` can use the `...projectComponentFields` syntax to reuse the fragment defined here.

editor/src/queries/project.js
```diff
@@ -17,8 +17,7 @@ export const ADD_PROJECT = gql`
   }
 `;
 
-export const SUMMARY_QUERY = gql`
-  ${PROJECT_COMPONENT_FIELDS}
```
mddilley, line 21: Here is where the interpolation can be removed. The `projectComponentFields` fragment is referenced later in this query.

editor/src/queries/project.js
```diff
@@ -602,7 +601,7 @@ export const PROJECT_ACTIVITY_LOG = gql`
 `;
 
 export const PROJECT_ACTIVITY_LOG_DETAILS = gql`
-  query getMopedProjectChanges($activityId: uuid!) {
```
mddilley, line 605: the codegen tool flagged a few queries that had duplicate names or were not queried anywhere so the rest of this file is renaming or removing unused queries.

editor/codegen.ts
mddilley, line 1: the codegen config:
- gets our Hasura endpoint and admin secret to reach the local instance
- is scoped to two files that contain the queries and mutations updated in this PR; we would open this up to the whole queries folder if we adopt this workflow
- has a couple of config options that keep Apollo cache working which depends on the __typename field that we see all the time
- formats with prettier using our existing config

editor/eslint.config.mjs
```diff
@@ -19,6 +19,7 @@ export default tseslint.config(
       "coverage",
       "node_modules",
       "**/*.test.{js,jsx,ts,tsx}",
+      "src/gql/**",
```
mddilley, line 22: skip linting the generated types and related code

editor/netlify.toml
mddilley, line 1: here's where we run typecheck before building and deploying

editor/vite.config.js
```diff
@@ -32,6 +33,9 @@ export default defineConfig(() => {
         // Disable during build since we lint in our Netlify build command
         enableBuild: false,
       }),
+      codegen({
+        runOnBuild: false,
+      }),
```
mddilley, line 46: tells the Vite server to run codegen on save and not to run during build. we run a type check instead

editor/src/gql/gql.ts
```diff
@@ -0,0 +1,536 @@
...
+/**
+ * Map of all GraphQL operations in the project.
+ *
+ * This map has several performance disadvantages:
+ * 1. It is not tree-shakeable, so it will include all operations in the project.
+ * 2. It is not minifiable, so the string of a GraphQL query will be multiple times inside the bundle.
+ * 3. It does not support dead code elimination, so it will add unused operations.
+ *
+ * Therefore it is highly recommended to use the babel or swc plugin for production.
+ * Learn more about it here: https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#reducing-bundle-size
```
johnclary, line 14: Not sure how much you explored this as a potential concern, but they do have [a section](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#reducing-bundle-size) for Vite React here that mentions the SWC plugin.

editor/src/queries/components.js
johnclary: Getting type safety magic on a graphql fragment is cool!

editor/src/gql/gql.ts
```diff
@@ -0,0 +1,536 @@
...
+ */
+export function graphql(
+  source: "\n  mutation DeleteMopedComponent($projectComponentId: Int!) {\n    update_moped_proj_components_by_pk(\n      pk_columns: { project_component_id: $projectComponentId }\n      _set: { is_deleted: true }\n    ) {\n      project_component_id\n    }\n    update_moped_proj_components_subcomponents(\n      where: { project_component_id: { _eq: $projectComponentId } }\n      _set: { is_deleted: true }\n    ) {\n      affected_rows\n    }\n  }\n"
+): (typeof documents)["\n  mutation DeleteMopedComponent($projectComponentId: Int!) {\n    update_moped_proj_components_by_pk(\n      pk_columns: { project_component_id: $projectComponentId }\n      _set: { is_deleted: true }\n    ) {\n      project_component_id\n    }\n    update_moped_proj_components_subcomponents(\n      where: { project_component_id: { _eq: $projectComponentId } }\n      _set: { is_deleted: true }\n    ) {\n      affected_rows\n    }\n  }\n"];
+/**
+ * The graphql function is used to parse GraphQL queries into a document that can be used by GraphQL clients.
+ */
+export function graphql(
```
johnclary, line 257: I don't quite follow how all of these repeated exports work but I can tell they definitely are working 😄

editor/src/utils/phases.ts
```diff
@@ -13,24 +14,27 @@ export const defaultLabel = "Unknown";
 /**
  * Font colors
  */
-const primary = theme?.palette?.text?.primary;
-const white = theme?.palette?.background?.paper;
+const primary = theme.palette.text.primary;
+const white = theme.palette.background.paper;
```
johnclary, line 18: The removal of doubt ✨

editor/src/queries/project.js
```diff
@@ -501,24 +500,24 @@ export const ADD_PROJECT_MILESTONE = gql`
   }
 `;
 
-export const PROJECT_FOLLOW = gql`
+export const PROJECT_FOLLOW = graphql(`
   mutation FollowProject($object: moped_user_followed_projects_insert_input!) {
```
johnclary, line 504: I ran a few different tests by adding invalid columns to queries and also invalid types to variables and `where` expressions.

You don't get a red squiggly line in query declaration, but you can see that the query type flips to `unknown`, and type errors immediately pop up in the parts of the code that consume this query: Apollo's `useQuery` and `useMutation` choke on the `unknown` type. Super cool 👍 

Image

editor/src/gql/graphql.ts
johnclary: One minor issue I noticed is that timestamps columns are being typed as `unknown`. Quick googling suggested that we will need to define a custom scalars for these. I'm curious to see how to resolve that because our use of plain `string` types in the VZE is quite lackluster.

  Image

johnclary (approved):
Dang, this was super fun to test out. The graphql typing works like a charm, and, wow, you've got full type safety on the project summary page. Incredible stuff.

I'm pretty sold on the codegen approach. I was impressed by how responsive the type checking was—both when updating Hasura and modifying queries. The `graphql.ts` file is quite massive, but it's also easy enough to read and follow what's going on. Thanks for advancing this approach, Mike! It feels really solid.

🚢 🚢 🚢 🚢

editor/src/queries/project.js
```diff
@@ -966,38 +965,6 @@ export const PROJECT_CLEAR_INTERIM_ID = gql`
   }
 `;
 
-export const PROJECT_CLEAR_NO_CURRENT_PHASE = gql`
```
chiaberry, line 1024: 🧹

editor/src/views/projects/projectView/ProjectFollowButton.tsx
```diff
@@ -1,25 +1,37 @@
...
  * Icon button to follow/unfollow a project
- * @param {Number} projectId - The id of the current project to follow/unfollow
- * @param {Boolean} isFollowing - Whether the user is currently following the project
- * @param {Function} refetch - The refetch function from Apollo
- * @param {Function} handleSnackbar - The function to show the snackbar
- * @return {JSX.Element}
+ * @param projectId - The id of the current project to follow/unfollow
+ * @param isFollowing - Whether the user is currently following the project
+ * @param refetch - The refetch function from Apollo
+ * @param handleSnackbar - The function to show the snackbar
```
chiaberry: these variables are pretty self explanatory, but in the future should we move these descriptions into the type definition, that way the definition is commented and not the component? That way we can hover over props and see our comments when using components

chiaberry (approved):
This was easy to test out and it is going to be so good to have this extra layer of protection while developing. I appreciated when it would warn that something could be undefined and needs optional chaining. 

🚀

editor/src/gql/gql.ts
mddilley: thanks for catching this John! just added since it is _highly_ recommended for prod. that section made it a breeze. 🙏

editor/src/gql/graphql.ts
mddilley: this is a good question! i'm going to take a closer look at this tomorrow. I'm seeing there is a straightforward way to set these in the codegen config to treat them as strings, but I'm wondering how far we can take it 🤔 

Image

```ts
import { type CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  schema: {
    "http://localhost:8082/v1/graphql": {
      headers: {
        "x-hasura-admin-secret": "hasurapassword",
      },
    },
  },
  documents: ["src/queries/project.js", "src/queries/components.js"],
  generates: {
    "./src/gql/": {
      preset: "client",
      config: {
        skipTypeNameForRoot: true,
        nonOptionalTypename: true,
        scalars: {
          timestamptz: "string",
        },
      },
    },
  },
  hooks: { afterAllFileWrite: ["prettier --write"] },
};

export default config;

```

editor/src/views/projects/projectView/ProjectFollowButton.tsx
mddilley: thanks for flagging, Chia! i'm going to update so that I can see it firsthand. 🙏

mddilley: Worked out great and is a TIL 🙏

editor/src/gql/graphql.ts
mddilley: This encouraged me to add to the readme to cover the overall codegen workflow and also touch on how to introduce new scalars as we run into them. I kept timestamptz as string for now.

mateoclarke (approved):
This is super neat. I was worried the plugin would slow down the devex but it's snappy and all makes sense given your trial and error testing steps. Thanks for detailing that and making it very clear how this works seemingly magically behind the scenes. And hooray for blazing the trail with more types! 🎉 🚀 

In the test step for mutation payloads, I think it should be:
"Go to ProjectFollowButton.tsx and change *Number(projectId)* to false on line *61*"
_I did a ⌘F Number(project_id) and it didn't work and I found the code you say to test 10 lines down from 51. Just in case this helps any other reviewers_

mddilley:
thanks for reviewing and catching that typo in the test steps, @mateoclarke! i just fixed for any future reviewers. 🙏

editor/src/gql/graphql.ts
johnclary: Nice—that's a clean fix 🚀

frankhereford (approved):
This is so impressive and responsive. I worked through your testing steps and everything just worked. It's such a level up for the DX bringing typing to the data coming off the graphql service, and you've integrated it into the stack so well. Thank you Mike! Thank you also for the extra time you gave us to get to test this -- it's nice to get to practice with our future dev environment. 🚢🚢🚢

PR #1885, opened 2026-06-26
[Collector] Add database table for work activity file attachments

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29049, Closes https://github.com/cityofaustin/atd-data-tech/issues/29109

This PR sets up the table and permissions to support work activity file attachments in the UI. It also adds some foreign key constraints in our audit fields in a few tables and patches up some traces of our previous admin user in the project components table.

## Testing
**URL to test:** 

Local

**Steps to test:**
1. Spin up the local stack
2. Run the query below on the production replica and the local database and see that we no longer have user id 29 in the update audit field after this migration like we have in our production database. You can also check the staging replica to see that we don't have any rows that match.
3. Check the project list view of your local app to make sure the main feed is not flooded with updates related to the migration to make sure the trigger suppression worked. You can also check a project activity log for one of the projects that did have user 29 in the `updated_by_user_id` column like in https://localhost:3000/moped/projects/827?tab=activity_log to see that no activities were generated in our `moped_activity_log` table since this change isn't relevant to users.
4. Check the `moped_proj_components`, `moped_proj_milestones`, and `moped_proj_notes` tables in your database client to make sure that all `created_by_user_id` and `updated_by_user_id` columns have a foreign key constraint 
5. Optional but you can test the GraphQL mutations below in the Hasura console that you can start by running `hasura console` from the `/moped-database/` folder
6. Run the down migration by running `hasura migrate apply --down 1` from the `moped-database` folder to make sure it works

```sql
-- Find project component rows with former admin account user id
SELECT * FROM moped_proj_components WHERE updated_by_user_id = 29;
```

```gql
# Create work activity attachments between existing work activity and project file
mutation AttachExistingFileToWorkActivity(
  $object: files_project_work_activities_insert_input!
) {
  insert_files_project_work_activities_one(
    object: $object
    on_conflict: {
      constraint: files_project_work_activities_entity_id_file_id_key
      update_columns: [is_deleted]
    }
  ) {
    id
  }
}

# Query variables
{
  "object": {
    "entity_id": 4106,
    "file_id": 1777,
    "created_by_user_id": 1,
    "updated_by_user_id": 1
  }
}

# Soft delete attachment
mutation DetachFileWorkActivityAttachment($id: Int!) {
  update_files_project_work_activities_by_pk(
    pk_columns: { id: $id }
    _set: { is_deleted: true }
  ) {
    id
  }
}

# Query variables
{
  "id": 1
}
```
---
[...]

PR #1898, opened 2026-07-16
Migrate work activity links to be work activity attachments

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/29136

## Testing

First you will test migrating existing work order links and then you will test the UI updates

**Steps to test:**

Spin up your local stack, use production data. Migrations are applied automatically. 

Now you can test migrating existing work order links. 
There is a script in `toolbox/migrate_work_activity_links`, I have included the briefest of readme's in that directory. Create a venv and pip install the requirements (its just requests) and you are ready to start. The script is _not_ idempotent, it goes through all work activity records, finds work order urls and creates file records for those urls. 

From your fav DB GUI, you can see there are no file records with the file type "work order url"
`select * from moped_project_files where file_type = 5`

Anyway, back in the toolbox. Copy the secrets template
```bash
cp secrets_template.py secrets.py
```

There is a dry run flag! `python run.py -n` to do a dry run, `python run.py` to run it on your local db. When I ran it, i got 339 files created, you can double check with 
```sql
select count(*) from moped_project_files where file_type = 5;
```

Great, now to the UI! ✨ 

Visit https://localhost:3000/moped/projects/1715?tab=files and see some new files created as a result of running the script. 

Now go to the funding tab and scroll down to the activities table. The "work order link" column is no longer there. 

Click to edit an existing record (✏️ ), the modal opens and the "work order link" at the bottom is empty. Input a valid url, ex: `https://atd.knack.com/amd#project-requests/view-project-request-details3/`. The link is added as a file attachment. 

Click to add a new record. If the url does not match the signs and markings or data tracker pattern, it will be saved as "work order link". 

---
#### Ship list
- [x] Code reviewed 
   - Tests captured in https://github.com/cityofaustin/atd-moped/pull/1885 and will be added while testing the release


moped-database/views/exploded_component_arcgis_online_view.sql
chiaberry, line 1: I dont know why the view formatter chose to do this now

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.jsx
```diff
@@ -312,7 +322,9 @@ const ProjectWorkActivitiesForm = ({
             color="primary"
             startIcon={<CheckCircle />}
             type="submit"
-            disabled={(!isDirty && !isNewActivity) || mutationState.loading || !isValid}
+            disabled={
+              (!isDirty && !isNewActivity) || mutationState.loading || !isValid
+            }
```
chiaberry, line 326: just linting change

editor/src/components/ExternalLink.jsx
```diff
@@ -1,18 +1,7 @@
...
-const getExternalLinkText = (url) => {
-  switch (true) {
-    case url.toLowerCase().includes("https://atd.knack.com/amd"):
-      return "AMD Data Tracker";
-    case url.toLowerCase().includes("https://atd.knack.com/signs-markings"):
-      return "Signs & Markings";
-    default:
-      return null;
-  }
-};
```
chiaberry, line 15: moved this to different file so it could be exported

moped-database/migrations/default/1784138295424_add_moped_file_type/down.sql
chiaberry, line 1: If we try to delete the work order link file type after records of that type have been added, there is a foreign key constraint error. Getting around that by changing all work order link types to "other". Open to feedback on if this is a good idea

chiaberry:
Got some feedback from Mike, going to change up the PR a bit and will rerequest review when im ready!

editor/src/utils/urls.js
```diff
@@ -38,3 +38,15 @@ export const makeUrlValid = (url) => {
 export const createBugReportLink = (error, userEmail) => {
   return `https://atd.knack.com/dts#new-service-request/?view_249_vars=%7B%22field_398%22%3A%22Bug%20Report%20%E2%80%94%20Something%20is%20not%20working%22%2C%22field_399%22%3A%22Moped%22%2C%22field_400%22%3A%22${error.message}%22%2C%22field_406%22%3A%22${userEmail}%22%7D`;
 };
+
+// Returns link text based on url when no text prop is provided, can be expanded to include more cases as needed
+export const getExternalLinkText = (url) => {
```
johnclary, line 43: It looks like this would hit a TypeError if `url` wasn't a string. Might be good to check that before the switch.

editor/src/views/projects/projectView/ProjectWorkActivity/utils/form.js
```diff
@@ -115,18 +142,26 @@ export const onSubmitActivity = ({
...
   const variables = { object: payload };
 
+  // if there is an id, this is an update mutation, otherwise its an add mutation
   if (id) {
     variables.id = id;
+    variables.fileObject = filePayload;
   } else {
     variables.object.project_id = data.project_id;
+    variables.object["work_activity_files"] = { data: [filePayload] };
   }
```
johnclary, line 167: I am getting this error if I try to create a work activity without a work order URL.

Image

toolbox/migrate_work_activity_links/run.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        for new_file in new_file_records:
+            response = make_hasura_request(
+                query=ADD_FILE_WITH_WORK_ACTIVITY_CONNECTION_MUTATION,
+                variables={"object": new_file},
+                env=env,
+            )["insert_moped_project_files_one"]
+            print(
+                f"Inserted file {response["project_file_id"]} for project {response["project_id"]}"
+            )
```
johnclary, line 111: Would it make sense to update the work activity by nullifying the old work order URL after the file has been created? That would have the effect of making this script idempotent.

johnclary (commented):
The script worked great, and the UI is looking good except for the error I'm getting when trying to submit an activity without a work order link.

editor/src/views/projects/projectView/ProjectWorkActivity/utils/form.js
chiaberry: I have update the mutation and the payload, this should be addressed now. I retested adding a new activity with and without a url and updating an existing activity with and without a url.

editor/src/utils/urls.js
chiaberry: good call, updated!

editor/src/queries/funding.js
```diff
@@ -306,7 +305,7 @@ export const WORK_ACTIVITY_STATUSES_QUERY = gql`
   }
 `;
 
-export const ADD_WORK_ACTIVITIY = gql`
```
mddilley, line 309: 🙏

editor/src/queries/funding.js
```diff
@@ -318,13 +317,19 @@ export const UPDATE_WORK_ACTIVITY = gql`
...
   ) {
     update_moped_proj_work_activity_by_pk(
       pk_columns: { id: $id }
       _set: $object
     ) {
       id
     }
+    insert_moped_project_files(objects: $fileObjects) {
+      returning {
+        api_response
```
mddilley: another files column that I didn't know (or remember) existed! @chiaberry could you update to return `affected_rows` instead? I'm going to add this column to https://github.com/cityofaustin/atd-data-tech/issues/29483 since we've never populated values for it

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.jsx
mddilley, line 1: @chiaberry could you update the order of the columns so that the "Files" column is to the right of "Task Order(s)" so it is in the same place as the removed "Work Order Link" column was?

We received good feedback that this change wouldn't interrupt existing workflows, but, after clicking around in here, i'm concerned that the "Files" column being so far to the right and possibly hidden out of view for some could create so whiplash or confusion about it seeming like a feature is removed.

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.jsx
```diff
@@ -291,10 +291,14 @@ const ProjectWorkActivitiesForm = ({
               name="work_order_url"
               size="small"
             />
-            {formErrors?.work_order_url && (
+            {formErrors?.work_order_url ? (
```
mddilley, line 299: I don't feel strongly about this, but I'm wondering if we should add a divider above the work order link input? I might be overthinking but maybe this would help align this behavior more with the look of the similar phase status updates?

Image

moped-database/migrations/default/1784138295424_add_moped_file_type/down.sql
mddilley, line 1: @chiaberry I've thought more about this down migration, and I think that it would be preferable to be no-op and make this up-only. I don't anticipate us needing to roll this back during the release, but, if we ever did run this sql, then we would lose the ability to distinguish migration-applied types from user-applied types.

This would be consistent with how we have handled lookup tables that support soft deletes - following up by retiring them instead of hard-deleting. We would just have to add soft-deletes to this table to make that happen if we ever needed.

toolbox/migrate_work_activity_links/run.py
mddilley: I'm open to this, but I'm also not concerned about needing to run this more than once. I'm also probably clinging a little too hard to the original user-captured data in the deprecated column even though it will still exist in a slightly different form. 

@chiaberry what do you think?

editor/src/queries/funding.js
mddilley, line 1: @chiaberry I'm commenting here because this PR doesn't touch the mutations that handle this, but, I noticed that, if I delete a work activity that has a file attachment in the UI, the files attachment row is not soft-deleted like it is when doing the same thing in the funding table.

I wanted to see if you were up for adding in this PR. I can also create a follow-up issue to handle separately so just let me know. Sorry for missing this earlier in the last PR. You can see how we did it in `DELETE_PROJECT_FUNDING`

mddilley (changes requested):
Looks great, Chia! I tested this a bunch, and I'm requesting a couple small changes that I mentioned you in and wanted to get your thoughts on a few things. Thanks for talking through this one with me. 🙏

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.jsx
chiaberry: updated!

moped-database/migrations/default/1784138295424_add_moped_file_type/down.sql
chiaberry: I agree, it would mess up distinguishing where the records came from. Updated to be up-only

toolbox/migrate_work_activity_links/run.py
chiaberry: I feel like I'm also clinging to not touching the data in the deprecated column and I don't have a good reason why.

editor/src/queries/funding.js
chiaberry: I will tackle this tomorrow

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityTable.jsx
mddilley: thanks Chia!!

editor/src/queries/funding.js
chiaberry: tackled! https://github.com/cityofaustin/atd-moped/pull/1898/changes#diff-cb1bbc75cec528c8ae79ad886c14d3d51e785617af9159cbcb30b61591e3f6a2R342

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.jsx
chiaberry: I added the divider -- its growing on me!

toolbox/migrate_work_activity_links/run.py
mddilley: Ok then let's treat this like the deprecated columns in the funding table by keeping historical data and not nulling out the values of processed rows. I will retain the logged output and paste in the issue so we'll have a record of what changes were made to what. 🙏 

Thanks Chia!

editor/src/views/projects/projectView/ProjectWorkActivity/ProjectWorkActivityForm.jsx
mddilley: i like it too! let's keep it. thanks!

mddilley (approved):
I read through the latest updates, and I retested adding, editing, attaching, detaching, deleting rows with attachments, deleting files that are attached, and all the other things that I could come up with. All works great! Thanks Chia! 🚢 😎

editor/src/queries/funding.js
roseeichelmann: ✨

roseeichelmann (approved):
i followed all the test steps and i saw 339 files created, testing adding editing deleting attachments etc, this looks great to me!!

tillyw (approved):
i got 340 files! all the ui test steps worked as expected as well. thanks for this enhancement!

mddilley:
thanks Chia! merging this down to test in the collector. 🙏 🚀

PR #1908, opened 2026-08-05
29207 upgrade to react v19

### LINKED ISSUE #29207: Upgrade to React v19
We are currently on React v18 which is one version behind the latest release, and it is also causing us to need one additional dependency for MUI v7 (https://mui.com/material-ui/migration/upgrade-to-v7/#react-18-and-below). We should upgrade.

**In Scope**
- Update to React v19 ([guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide))
- Remove the dependency needed by MUI v7 and poke around the app to make sure the UI is still working as expected

tillyw:
## Associated issues
closes https://github.com/cityofaustin/atd-data-tech/issues/29207

## Testing
**URL to test:** 

https://deploy-preview-1908--moped-main-and-prs.netlify.app/

**Steps to test:**

A few notes before testing:
- `defaultProps` has been removed in React 19. we only used it in two files (`GeocoderControl` and `ComponentsDrawControl`), so I replicated it by destructuring defaults directly in the parameter list
- while troubleshooting the above issue, I updated `react-map-gl` and `@mapbox/mapbox-gl-geocoder` to their current versions. this resulted in only one breaking change in how we import `react-map-gl`, which I addressed in all files. I'm not totally sure if this was necessary, but I figured we could bundle these updates with this PR to save us on future updates.
- `html-react-parser` was 5 versions out-of-date and was preventing us from upgrading to React 19, so I made that upgrade as well. It may be worth exploring whether that library is still needed with the current version of React or if they have implemented some native functionality to replace it.
- `react-helmet` is no longer maintained or supported in React 19, so we dropped it in favor of implementing a custom hook, as we do in the Vision Zero Editor.
- I also updated the associated type libraries `@types/react` and `@types/react-dom`
- we were able to drop the `react-is` override
- we will need to update `react-router-dom` to v8 to address some console and audit warnings, which will be done in a separate pr branched from this one, since `react-router-dom` v8 requires React 19

testing:
- unfortunately, I don't know of a way to test this PR aside from clicking through the app and noting any breaking changes, new console warnings (aside from any related to `react-router-dom`), missing functionality, or differences
- please pay close attention to map-related functionality, particularly when it comes to using the geocoder (search bar) or drawing in or editing component features (as these files required some minor changes)

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager checked if any columns in views used by the Power BI dataflow were added, deleted, or renamed


tillyw:
apologies, this branch went out of sync with main but is now up to date!

editor/src/components/Maps/ComponentsDrawControl.jsx
```diff
@@ -1,42 +1,42 @@
...
+      ({ map }) => {
+        map.on("draw.create", onCreate);
+        map.on("draw.update", onUpdate);
+        map.on("draw.delete", onDelete);
+        map.on("draw.modechange", props.onModeChange);
+        map.on("draw.selectionchange", props.onSelectionChange);
 
-  return null;
-});
+        return new MapboxDraw(props);
```
mddilley, line 23: I'm not 100% this is a concern, but a small change in the new code is that the `props` object here no longer includes onCreate, onUpdate, or onDelete. 

I'm not familiar enough with the draw library internals, but I think we should keep that behavior by returning those functions in the object passed to this draw instance like: `{onCreate, onUpdate, onDelete, ...props}` in case it affects future development.

editor/src/components/Maps/GeocoderControl.jsx
```diff
@@ -118,13 +125,3 @@ export default function GeocoderControl(props) {
   }
   return marker;
 }
-
-const noop = () => {};
-
-GeocoderControl.defaultProps = {
-  marker: true,
```
mddilley, line 125: @tillyw could you add this marker key/value to the `GeocoderControl` props alongside the other functions? Thanks for navigating this code btw. iirc this and the draw control were mostly copypasta from react-map-gl examples 🍝

editor/src/utils/documentTitle.jsx
```diff
@@ -0,0 +1,12 @@
+"use client";
```
mddilley, line 1: @tillyw could you remove this? i know we have it over in the VZE code but we aren't currently doing any server side rendering with React or React Router that we need to opt out of. [here](https://react.dev/reference/rsc/use-client) is some more info since we'll be seeing more and more of this out in the world.

editor/package.json
```diff
@@ -47,30 +47,29 @@
     "formik": "^2.4.5",
     "graphql": "^15.8.0",
     "history": "^5.3.0",
-    "html-react-parser": "^1.4.14",
+    "html-react-parser": "^6.0.0",
```
mddilley, line 50: from v1 to v6, nice! i scanned the breaking changes across versions related to our uses, and nothing stood out to me so this feels safe to me. I don't think that we'll be missing Internet Explorer support. 👋 😄

editor/package.json
```diff
@@ -47,30 +47,29 @@
...
     "lexical": "^0.16.1",
     "lodash.clonedeep": "^4.5.0",
     "lodash.get": "^4.4.2",
     "lodash.isequal": "^4.5.0",
     "mapbox-gl": "^3.0.1",
```
mddilley, line 55: @tillyw it looks like [we need to upgrade mapbox-gl to 3.5.0 or greater](https://visgl.github.io/react-map-gl/docs/upgrade-guide#upgrading-to-v80) to match up with the `react-map-gl/mapbox` imports in this diff. could you update this one while you are in here too? We are on `"mapbox-gl": "^3.28.0",` over in VZE land so that gives me confidence that we won't run into issues.

editor/package.json
```diff
@@ -83,8 +82,5 @@
     "vite": "^8.0.16",
     "vite-plugin-checker": "^0.14.1",
     "vitest": "^4.1.5"
-  },
-  "overrides": {
-    "react-is": "^18.3.1"
```
mddilley, line 88: thank you for remembering to remove this [MUI-related override](https://mui.com/material-ui/getting-started/installation/#react-18-and-below) 🙏🙏🙏

mddilley (changes requested):
I flagged a few small changes to fix before I dive further into testing the app. But, yay for version updates ✅ and less dependencies ✅ - that is the good stuff!

editor/src/views/projects/projectView/ProjectComponents/TheMap.jsx
```diff
@@ -1,7 +1,7 @@
 import { useEffect, useState } from "react";
-import MapGL, { NavigationControl } from "react-map-gl";
+import MapGL, { NavigationControl } from "react-map-gl/mapbox";
 import cloneDeep from "lodash.clonedeep";
-import GeocoderControl from "src/components/Maps/GeocoderControl";
+import GeocoderControl from "../../../../components/Maps/GeocoderControl";
```
chiaberry: could this go back to the absolute import style?

tillyw: done!

tillyw:
thanks, y'all! i made the updates you requested and am rerequesting review!!

mddilley:
thanks @tillyw! I'm merging this down for the release. 🚀🚀🚀

